PR #55, opened 2020-11-20
Unix Domain Socket transport

gabrik:
This PR contains the code to implement the Unix Domain Sockets as transport for Zenoh

Creating a draft to keep track of the work

- [x] Implement Unix Domain Socket Link, LinkManager
- [x] Update locator with `unix/<domain socket path>` 
- [x] Add `unix` feature for this transport
- [x] Add conditional building to build only on UNIX systems
- [x] Add tests

### DIFF
--- a/zenoh-protocol/src/link/locator.rs
+++ b/zenoh-protocol/src/link/locator.rs
@@ -15,7 +15,7 @@
 use async_std::net::SocketAddr;
 #[cfg(any(feature = "tcp", feature = "udp"))]
 use async_std::net::ToSocketAddrs;
-#[cfg(all(feature = "unix", target_family = "unix"))]
+#[cfg(all(feature = "transport_unixsock-stream", target_family = "unix"))]
 use async_std::path::PathBuf;
 #[cfg(any(feature = "tcp", feature = "udp", feat))]
 use async_std::task;
@@ -38,8 +38,8 @@ pub const PORT_SEPARATOR: char = ':';
 pub const STR_TCP: &str = "tcp";
 #[cfg(feature = "udp")]
 pub const STR_UDP: &str = "udp";
-#[cfg(all(feature = "unix", target_family = "unix"))]
-pub const STR_UNIX: &str = "unix";
+#[cfg(all(feature = "transport_unixsock-stream", target_family = "unix"))]
+pub const STR_UNIXSOCK_STREAM: &str = "unixsock-stream";
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub enum LocatorProtocol {
@@ -47,8 +47,8 @@ pub enum LocatorProtocol {
     Tcp,
     #[cfg(feature = "udp")]
     Udp,
-    #[cfg(all(feature = "unix", target_family = "unix"))]
-    Unix,
+    #[cfg(all(feature = "transport_unixsock-stream", target_family = "unix"))]
+    UnixSockStream,
 }
 
 impl fmt::Display for LocatorProtocol {
@@ -58,8 +58,8 @@ impl fmt::Display for LocatorProtocol {
             LocatorProtocol::Tcp => write!(f, "{}", STR_TCP)?,
             #[cfg(feature = "udp")]
             LocatorProtocol::Udp => write!(f, "{}", STR_UDP)?,
-            #[cfg(all(feature = "unix", target_family = "unix"))]
-            LocatorProtocol::Unix => write!(f, "{}", STR_UNIX)?,
+            #[cfg(all(feature = "transport_unixsock-stream", target_family = "unix"))]
+            LocatorProtocol::UnixSockStream => write!(f, "{}", STR_UNIXSOCK_STREAM)?,
         }
         Ok(())
     }
@@ -71,8 +71,8 @@ pub enum Locator {
     Tcp(SocketAddr),
     #[cfg(feature = "udp")]
     Udp(SocketAddr),
-    #[cfg(all(feature = "unix", target_family = "unix"))]
-    Unix(PathBuf),
+    #[cfg(all(feature = "transport_unixsock-stream", target_family = "unix"))]
+    UnixSockStream(PathBuf),
 }
 
 impl FromStr for Locator {
@@ -145,8 +145,8 @@ impl FromStr for Locator {
                 });
                 addr.map(Locator::Udp)
             }
-            #[cfg(all(feature = "unix", target_family = "unix"))]
-            STR_UNIX => {
+            #[cfg(all(feature = "transport_unixsock-stream", target_family = "unix"))]
+            STR_UNIXSOCK_STREAM => {
                 let addr = task::block_on(async {
                     match PathBuf::from(addr).to_str() {
                         Some(path) => Ok(PathBuf::from(path)),
@@ -157,7 +157,7 @@ impl FromStr for Locator {
                         }
                     }
                 });
-                addr.map(Locator::Unix)
+                addr.map(Locator::UnixSockStream)
             }
             _ => {
                 let e = format!("Invalid protocol locator: {}", proto);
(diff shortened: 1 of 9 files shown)


JEnoch:
I'm not sure **"unix**" is a good idea for the Cargo feature name as:
 - it's the same than the "unix" target name
 - on user's perspective, a "unix" feature doesn't sound clear on its purpose.
Why not **"unix-socket"** feature ?

zenoh-protocol/tests/channel.rs
```diff
@@ -305,6 +305,27 @@ fn channel_udp() {
     });
 }
 
+#[cfg(all(feature = "unix", target_family = "unix"))]
+#[test]
+fn channel_unix() {
+    let _ = std::fs::remove_file("/tmp/socket.sock");
+    // Define the locator
+    let locators: Vec<Locator> = vec!["unix//tmp/socket.sock".parse().unwrap()];
+    // Define the reliability and congestgino control
```
Mallets, line 314: Typo in congestion control.

zenoh-protocol/tests/channel.rs
```diff
@@ -305,6 +305,27 @@ fn channel_udp() {
     });
 }
 
+#[cfg(all(feature = "unix", target_family = "unix"))]
+#[test]
+fn channel_unix() {
+    let _ = std::fs::remove_file("/tmp/socket.sock");
+    // Define the locator
+    let locators: Vec<Locator> = vec!["unix//tmp/socket.sock".parse().unwrap()];
```
Mallets, line 313: Although /tmp is available in the totality of the system, I believe that using the current directory for the unix socket path is a better approach:
1. It is guaranteed that the cargo test will have write permissions while executing the test
2. System files are not created outside the test directory

I also suggest to use a more explanatory name for the socket, e.g., zenoh-test-unix-socket.sock

zenoh-protocol/src/link/locator.rs
```diff
@@ -63,19 +71,35 @@ pub enum Locator {
     Tcp(SocketAddr),
     #[cfg(feature = "udp")]
     Udp(SocketAddr),
+    #[cfg(all(feature = "unix", target_family = "unix"))]
+    Unix(PathBuf),
```
Mallets, line 75: I suggest to call the locator UnixDomainSocket and not simply Unix, alternatively we can use Uds.

zenoh-protocol/src/link/manager.rs
```diff
@@ -17,6 +17,8 @@ use async_std::sync::Arc;
 use crate::link::tcp::ManagerTcp;
 #[cfg(feature = "udp")]
 use crate::link::udp::ManagerUdp;
+#[cfg(all(feature = "unix", target_family = "unix"))]
+use crate::link::unix::ManagerUnix;
```
Mallets, line 21: I suggest to call the ManagerUnixSocket instead of ManagerUnix

zenoh-protocol/tests/channel.rs
```diff
@@ -326,3 +347,76 @@ fn channel_tcp_udp() {
...
 }
+
+#[cfg(all(feature = "unix", target_family = "unix"))]
+#[test]
+fn channel_tcp_unix() {
+    let _ = std::fs::remove_file("/tmp/socket_1.sock");
+    // Define the locator
+    let locators: Vec<Locator> = vec![
+        "tcp/127.0.0.1:7449".parse().unwrap(),
+        "unix//tmp/socket_1.sock".parse().unwrap(),
```
Mallets, line 358: Change the file name to something more self-explanatory, e.g., zenoh-test-unix-socket_1.sock

zenoh-protocol/tests/channel.rs
```diff
@@ -326,3 +347,76 @@ fn channel_tcp_udp() {
         close_session(router_manager, client_session, locators).await;
     });
 }
+
+#[cfg(all(feature = "unix", target_family = "unix"))]
+#[test]
+fn channel_tcp_unix() {
+    let _ = std::fs::remove_file("/tmp/socket_1.sock");
```
Mallets, line 354: Although /tmp is available in the totality of the system, I believe that using the current directory for the unix socket path is a better approach:

It is guaranteed that the cargo test will have write permissions while executing the test
System files are not created outside the test directory
I also suggest to use a more explanatory name for the socket, e.g., zenoh-test-unix-socket_1.sock

zenoh-protocol/tests/channel.rs
```diff
@@ -305,6 +305,27 @@ fn channel_udp() {
     });
 }
 
+#[cfg(all(feature = "unix", target_family = "unix"))]
+#[test]
+fn channel_unix() {
+    let _ = std::fs::remove_file("/tmp/socket.sock");
```
Mallets, line 311: Although /tmp is available in the totality of the system, I believe that using the current directory for the unix socket path is a better approach:

It is guaranteed that the cargo test will have write permissions while executing the test
System files are not created outside the test directory
I also suggest to use a more explanatory name for the socket, e.g., zenoh-test-unix-socket.sock

zenoh-protocol/tests/channel.rs
```diff
@@ -326,3 +347,76 @@ fn channel_tcp_udp() {
...
+            }
+        }
+        close_session(router_manager, client_session, locators).await;
+    });
+}
+
+#[cfg(all(feature = "unix", target_family = "unix"))]
+#[test]
+fn channel_udp_unix() {
+    let _ = std::fs::remove_file("/tmp/socket_2.sock");
```
Mallets, line 378: Change the file name to something more self-explanatory, e.g., zenoh-test-unix-socket_2.sock

zenoh-protocol/tests/channel.rs
```diff
@@ -326,3 +347,76 @@ fn channel_tcp_udp() {
...
+    // Define the locator
+    let locators: Vec<Locator> = vec![
+        "udp/127.0.0.1:7449".parse().unwrap(),
+        "unix//tmp/socket_2.sock".parse().unwrap(),
```
Mallets, line 382: Change the file name to something more self-explanatory, e.g., zenoh-test-unix-socket_2.sock

zenoh-protocol/src/link/locator.rs
```diff
@@ -36,13 +38,17 @@ pub const PORT_SEPARATOR: char = ':';
 pub const STR_TCP: &str = "tcp";
 #[cfg(feature = "udp")]
 pub const STR_UDP: &str = "udp";
+#[cfg(all(feature = "unix", target_family = "unix"))]
+pub const STR_UNIX: &str = "unix";
```
Mallets, line 42: My proposal for the locator protocol is to use "uds" from "unix domain socket". Other proposals are also accepted.

zenoh-protocol/src/link/unix.rs
```diff
@@ -0,0 +1,921 @@
...
+                "None".to_string()
+            }
+        }
+    };
+}
+
+/*************************************/
+/*              LINK                 */
+/*************************************/
+pub struct Unix {
```
Mallets, line 94: We should avoid ambiguity on the usage of term Unix here. I suggest either to use:
- UnixDomainSocket
- Uds

Mallets (changes requested):

zenoh-protocol/tests/channel.rs
JEnoch: Another solution is use the [tempfile](https://crates.io/crates/tempfile) that relies on [std::env::temp_dir](https://doc.rust-lang.org/std/env/fn.temp_dir.html) and performs automatic deletion of file.

Mallets: I like the idea of using the [tempfile](https://crates.io/crates/tempfile) crate. This should be added to the dev dependencies.

Mallets:
> I'm not sure **"unix**" is a good idea for the Cargo feature name as:
> 
> * it's the same than the "unix" target name
> * on user's perspective, a "unix" feature doesn't sound clear on its purpose.
>   Why not **"unix-socket"** feature ?

I would be tempted to go even further and have a clearer approach in the whole zenoh. Any feature related to a transport technology should be _trasport\_*_. I.e., existing _tcp_ and _udp_ features should be also renamed to _transport\_tcp_ and _transport\_udp_. What do you think?

Mallets (approved):

PR #82, opened 2021-04-16
Add convenient path/pathexpr/selector creation fn and division operator

jerry73204:
This PR implements the `/` operator on paths and creation functions that could panic. It works as the `TryFrom` alternative to allow users to write more succinct code. The usage looks like this.

```rust
let a = zenoh::path("parent");
let b = a / "child";
```


JEnoch:
Thanks @jerry73204, that's a good idea!

Just few comments:

- Can you please add the rustdoc comments for the {path,pathexpr,selector} creation functions, detailing that they will panic if the syntax is not correct.
- I'm not sure that the Div trait is appropriate for Selector, as if there is a predicate part (after `'?'` the result might differ from what user expects. e.g.:
  ```rust
  let s = zenoh::selector("/a/b?x=1");
  let s2 = s / "c";
  ```
  With your code the resulting Selector is `"/a/b?x=1/c"' which syntactically correct but is probably not what user expects.
  What he probably wants to do is rather:
  ```rust
  let s = zenoh::selector("/a/b?x=1");
  let mut s2 = s.clone();
  s2.path_expr = s2.path_expr / "c";
  // or s2.path_expr /= "c"; if we also implement DivAssign
  ```

jerry73204:
@JEnoch I thought of  `"/a/b?x=1" / "c"` to be `"/a/b/c?x=1"` alternative, but felt a bit uncomfortable for it because `x=1` may not apply for `c`

Alternatively, we can provide explicit  methods to help users to modify the selector. It looks like this.

```rust
use zenoh as z;
let selector = z::selector("/a/b").add_predicate("x=3&y=4").join("c"); // "/a/b/c?x=3&y=4"
```

I will drop the Div on Selector and open another discussion so that we can have an early merge.

jerry73204:
New fixes updated and extra doc comments added.

zenoh/src/selector.rs
```diff
@@ -171,6 +171,10 @@ impl TryFrom<&Query> for Selector {
     }
 }
 
+pub fn selector(selector: impl AsRef<str>) -> Selector {
```
JEnoch, line 174: Please add such a doc:
```rust
/// Creates a [`Selector`] from a string.
///
/// # Panics
/// Panics if the string is not a valid [`Selector`].
```

JEnoch (changes requested):

JEnoch:
> I will drop the Div on Selector and open another discussion so that we can have an early merge.

Sure! I just opened #87 to discuss

zenoh/src/selector.rs
jerry73204: Fixed

JEnoch:
Merged. Thanks @jerry73204 !

PR #92, opened 2021-05-17
Compile on stable

bfrog:
Compile against stable rather than nightly

### DIFF
--- a/plugins/zenoh-rest/examples/zenoh-net/zn_serve_sse.rs
+++ b/plugins/zenoh-rest/examples/zenoh-net/zn_serve_sse.rs
@@ -11,7 +11,6 @@
 // Contributors:
 //   ADLINK zenoh team, <zenoh@adlink-labs.tech>
 //
-#![feature(async_closure)]
 
 use clap::{App, Arg};
 use futures::prelude::*;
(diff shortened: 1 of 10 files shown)


bfrog:
It compiles, and most tests pass, I haven't compared against the normal sync branch to see if the same tests fail yet

Mallets:
Hello @bfrog ! Thanks for submitting the pull request. The CI seems to fail because the `rust-toolchain` file has been removed and it is required by the CI workflow.

Could you just restore the `rust-toolchain` fail and set its content to `stable`?

Mallets:
The CI is failing because of code format. 

Can you please run the following:
- `cargo fmt`
- `cargo clippy`

You shouldn't see any warning with those two commands.
Then `cargo test` should succeed according to your changes.

bfrog:
Clippy reported errors not related to my changes from what I could tell, ran cargo fmt

bfrog:
I did verify that net::session::Session::declare_callback_subscriber fails on both sync and this branch so its not a change I made that causes that particular test failure.

Mallets:
It seems the clippy error you were seeing was solved after you forked the zenoh repo. 
The checks have all successfully passed so I have just merged the PR!
Thank you very much for contributing this, being able to compile zenoh in stable it's certainly a great plus!

PR #94, opened 2021-06-01
Compile with Rust 1.47

esteve:
Before submitting the zenoh bridge to the ROS index, I tested if it could be built by the version of the Rust compiler that Ubuntu ships with (1.47). I'm avoiding rustup so that the ROS buildfarm can build packages cleanly without dependencies that can't be installed via APT.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2018,6 +2018,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "rustversion"
+version = "1.0.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


esteve:
I've updated this PR after the sync branch has been merged, unfortunately flume depends on spinning_top, which uses the newer spin_loop API (https://github.com/rust-osdev/spinning_top/blob/master/src/spinlock.rs#L57).

I could submit a patch to spinning_top so that it uses one API or the other depending on the version of the Rust compiler, or I could package a more recent version of Rust and Cargo in a vendor package for ROS. The latter seems the path of least resistance, so that's probably the one I'll follow.

milyin:
This PR is inactive for more than a year. Closing it

Issue #102, opened 2021-06-17
QueryingSubscriber (to receive past publications)


JEnoch:
Some use cases require a Subscriber to also receive the data that have been published before its creation.

Using the zenoh primitives, this can be done with a "QueryingSubscriber" that would:

- do a query at start-up to retrieve the historical data
- declare the subscription at the same time to receive live publications, but not delivering those to the user for now
- merge the query results and the live data into a same queue, filtering duplicates and ordering by timestamps
- deliver the content of this queue to the user
- finally, go on delivering live publications to the user

The queryable storing the past publications and replying to the query could be implemented in different ways:
 - a zenoh storage deployed in a zenoh router
 - or a "publication cache" colocalized with the publisher (to support peer-to-peer use cases without any router)

JEnoch:
Implemented in **zenoh-net** API.
An implementation in **zenoh** API will come later.

PR #105, opened 2021-06-22
Add a QueryingSubscriber in zenoh::net (#102)

JEnoch:


zenoh-ext/src/net/querying_subscriber.rs
```diff
@@ -0,0 +1,483 @@
...
+        let receiver = QueryingSubscriberReceiver::new(subscriber.receiver().clone());
+
+        let mut query_subscriber = QueryingSubscriber {
+            conf,
+            subscriber,
+            receiver,
+        };
+
+        // start query
+        query_subscriber.query()?;
```
kydos, line 147: Maybe we need to add a timeout for the query. Otherwise the querying subscriber will block if one of the query target is under deadlock or live-lock. This will protect us until we don't allow to support manually asserted liveliness for queryables.

zenoh-ext/src/net/querying_subscriber.rs
```diff
@@ -0,0 +1,483 @@
...
+use std::time::{Duration, Instant};
+use zenoh::net::*;
+use zenoh_util::core::{ZError, ZErrorKind, ZResult};
+use zenoh_util::sync::channel::{RecvError, RecvTimeoutError, TryRecvError};
+use zenoh_util::sync::ZFuture;
+use zenoh_util::{zerror, zwrite};
+
+/// The builder of QueryingSubscriber, allowing to configure it.
+#[derive(Clone)]
+pub struct QueryingSubscriberBuilder<'a> {
```
kydos, line 29: I think we need to add a query time-out parameter (See comment below)

zenoh-ext/src/net/querying_subscriber.rs
```diff
@@ -0,0 +1,483 @@
...
+        self.query_on(
+            &self.conf.query_reskey.clone(),
+            &self.conf.query_predicate.clone(),
+            self.conf.query_target.clone(),
+            self.conf.query_consolidation.clone(),
+        )
+    }
+
+    /// Issue a new query on the specified resource key and predicate.
+    pub fn query_on(
```
kydos, line 175: Shouldn't this operation be async ? It issues a query and waits for it, shouldn't it await?

kydos (changes requested):

zenoh-ext/src/net/querying_subscriber.rs
OlivierHecart: This doesn't wait for the replies. But to be coherent with zenoh::net::Session::query() it could return a ZFuture anyway. (Basically, almost everything return a ZFuture now in zenoh APIs)

zenoh-ext/src/net/querying_subscriber.rs
```diff
@@ -0,0 +1,483 @@
...
+            Ok(())
+        } else {
+            log::error!(
+                "Cannot start query on {}?{} - one is already in progress",
+                reskey,
+                predicate
+            );
+            zerror!(ZErrorKind::Other {
+                descr: "Query already in progress".to_string()
+            })
```
OlivierHecart, line 201: Do we really want to return an error here ? The user has no clue if a query is already pending or not when calling query(). Returning an error when the user did "nothing bad" and have no control on it seems wrong to me.

zenoh-ext/src/net/querying_subscriber.rs
JEnoch: No, this operation doesn't wait for the query. It triggers the query and store the resulting replies receiver in its InnerState. Then, the reception operations (poll_next/recv/try_recv...) are in charge to treat the query replies, adding those to the merge_queue, and then at query completion to sort/filter/forward the merge_queue to the user.

The result is that the wait for query completion is actually made in the reception operations (or not in the case of `poll_next()` and `try_recv()` that return respectively `Poll::Pending` and `TryRecvError::Empty` if the query is not complete)

JEnoch: > This doesn't wait for the replies. But to be coherent with zenoh::net::Session::query() it could return a ZFuture anyway. (Basically, almost everything return a ZFuture now in zenoh APIs)

OK. I'll add this as well as management of a list of queries in progress, instead of returning an error if one is already in progress.

zenoh-ext/src/net/querying_subscriber.rs
JEnoch: Sure. But it would be simpler to rely on timeout implementation in the `zenoh::net::Session::query()`.
#106 created for this.

zenoh-ext/src/net/querying_subscriber.rs
```diff
@@ -0,0 +1,483 @@
...
+}
+
+pub struct QueryingSubscriber<'a> {
+    conf: QueryingSubscriberBuilder<'a>,
+    subscriber: Subscriber<'a>,
+    receiver: QueryingSubscriberReceiver,
+}
+
+impl QueryingSubscriber<'_> {
+    fn new(conf: QueryingSubscriberBuilder<'_>) -> ZResult<QueryingSubscriber<'_>> {
```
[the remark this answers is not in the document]
JEnoch: No need since this new() operation is private. What matters is the `Session::declare_querying_subscriber()` that returns a `QueryingSubscriberBuilder` which implement the `ZFuture` trait.

zenoh-ext/src/net/querying_subscriber.rs
JEnoch: Sure. I'll change the implementation to support several queries at the same time. Thus no error will be returned, but the new query will be added to an inner list of queries in progress.

zenoh-ext/src/net/querying_subscriber.rs
JEnoch: Done in c253b7a.

PR #114, opened 2021-07-07
Define a trait for Plugins and allow them to be dynamically loaded or statically linked

p-avital:
Opening this PR as an RFC. When approved, I'll convert the existing plugins.


p-avital:
@JEnoch @OlivierHecart  :)

Cargo.lock
```diff
@@ -2975,6 +2975,14 @@ dependencies = [
  "zenoh_backend_traits",
 ]
 
+[[package]]
+name = "zenoh-plugin-trait"
+version = "0.1.0-dev"
```
JEnoch, line 2980: Version should be the same as zenoh (0.5.0-dev) since we will release all together.

plugins/zenoh-plugin-trait/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+# which is available at https://www.apache.org/licenses/LICENSE-2.0.
+#
+# SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+#
+# Contributors:
+#   ADLINK zenoh team, <zenoh@adlink-labs.tech>
+#
+[package]
+name = "zenoh-plugin-trait"
+version = "0.1.0-dev"
```
JEnoch, line 16: Version should be the same as zenoh (0.5.0-dev) since we will release all together.

plugins/example-plugin/Cargo.toml
```diff
@@ -14,10 +14,12 @@
...
-authors = ["kydos <angelo@icorsaro.net>",
-           "Julien Enoch <julien@enoch.fr>",
-           "Olivier Hécart <olivier.hecart@adlinktech.com>",
-		   "Luca Cominardi <luca.cominardi@adlinktech.com>"]
+authors = [
+    "kydos <angelo@icorsaro.net>",
+    "Julien Enoch <julien@enoch.fr>",
+    "Olivier Hécart <olivier.hecart@adlinktech.com>",
+    "Luca Cominardi <luca.cominardi@adlinktech.com>",
+]
```
JEnoch, line 22: Feel free to add yourself! 😉

plugins/zenoh-plugin-trait/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+    /// Returns this plugin's [Compatibility].
+    fn compatibility() -> Compatibility;
+
+    /// As Zenoh instanciates plugins, it will append their [Compatibility] to an array.
+    /// This array's current state will be shown to the next plugin.
+    ///
+    /// To signal that your plugin is incompatible with a previously instanciated plugin, return `Err`,
+    /// Otherwise, return `Ok(Self::compatibility())`.
+    ///
+    /// By default, a plugin is non-reentrant to avoir reinstanciation if its dlib is accessible despite it already being statically linked.
```
JEnoch, line 55: "avoir" => "avoid"

plugins/zenoh-plugin-trait/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+        fn new() -> Self {
+            PluginVTablePadding {
+                __padding: [0; PADDING_LENGTH],
+            }
+        }
+    }
+
+    /// For use with dynamically loaded plugins. Its size will not change accross versions, but its internal structure might.
+    ///
+    /// To ensure compatibility, its size and alignment must allow `size_of::<Result<PluginVTable, PluginVTableVersion>>() == 64` (one cache line).
```
JEnoch, line 134: size_of::<Result<**PluginVTable**, PluginVTableVersion>> or size_of::<Result<**PluginVTableInner**, PluginVTableVersion>> ?

plugins/zenoh-plugin-trait/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+        get_expected_args: fn() -> Vec<Arg<'static, 'static>>,
+    }
+
+    /// Automagical padding such that [PluginVTable::init]'s result is the size of a cache line
+    #[repr(C)]
+    struct PluginVTablePadding {
+        __padding: [u8; PADDING_LENGTH],
+    }
+    const PADDING_LENGTH: usize =
+        64 - std::mem::size_of::<Result<PluginVTableInner, PluginVTableVersion>>();
```
JEnoch, line 123: Since the `PluginVTablePadding` is used along `PluginVTableInner` within `PluginVTable`, I don't understand why its length depends on `Result<PluginVTableInner, PluginVTableVersion>` rather than just `PluginVTableInner`. Can you explain ?

plugins/zenoh-plugin-trait/src/lib.rs
p-avital: **PluginVTable**, the goal is for the size of the outer type to be fixed even if we change the size of the inner one (by adding more functions to the vtable for example).

The constant should let us know that the pointers may not point to the same things, but I'm afraid of what would happen if the load_plugin function returns something bigger (or smaller) than what is expected by the caller.

plugins/zenoh-plugin-trait/src/lib.rs
p-avital: It feels a bit less "magic" to set the padding length as "a cache line size minus the size of the type that actually interests us".

Since `Result` should use an 8 bytes marker due to `align_of::<PluginVTable>()` being 8 (and the same as `align_of::<PluginVTableInner>()`), but that might change (for example if it finds a niche to exploit in a few compiler updates), so I'd rather use the `Result` type to ensure that type has consistent size.

plugins/zenoh-plugin-trait/Cargo.toml
p-avital: Good point, although I've removed `zenoh` when I realized it was causing a dependency cycle when I started working on integrating it into `zenoh` itself

plugins/zenoh-plugin-trait/src/lib.rs
Mallets: There is a crate that does that: [https://docs.rs/cache-padded/1.1.1/cache_padded/](https://docs.rs/cache-padded/1.1.1/cache_padded/). I suggest to use this crate that is already widely used.

PR #133, opened 2021-09-10
Configuration refactor

p-avital:
Zenoh configuration is now a statically typed, nested structure, with int-key and str-key map-like accessors, as well as accessors that ensure static predicates are enforced onto set values.

This structure is generated through the `validated_struct` crate's macro, as well as the addition of a `zenoh-macros` crate which will from now on will allow us to define procedural macros (for derives for example) :)

The configuration, when modified through the str-key map-like accessors, can send notifications to subsribers through flume channels to allow any field to be hot-configurable.


p-avital:
* `Locator` is leaked, must export
* `--json` cli option needs work (non-mandatory, better help)
* join config.rs files
* (non-urgent) Find a way to give `Notifier` proper accessors (notify on mutable access)

OlivierHecart:
There is a copy paste error there : https://github.com/p-avital/zenoh/blob/6fb2e0fe319e2a3e19452558575b60f00b472178/zenoh/src/bin/zenohd.rs#L175

p-avital:
CI issue was due to a test that only runs on macs, apparently because only the mac version of gitlab CI can run said test. Might have to go back on that one at some point

p-avital:
Merging the config.rs isn't really feasible, as one is part of zenoh and the other part of zenoh-util :(

Also, in the low priority todo-list: convert examples to using the new config structure instead of the old ways

PR #135, opened 2021-09-14
Implement `futures::stream::FusedStream` for receiver types

phil-opp:
Implementing the [`FusedStream`](https://docs.rs/futures/0.3.17/futures/stream/trait.FusedStream.html) trait makes it possible to use the receiver types together with the [`futures::select!`](https://docs.rs/futures/0.3.17/futures/macro.select.html) macro.


OlivierHecart:
This will indeed ease zenoh usage. Thanks for the contribution @phil-opp ! I'll integrate that right now.

Issue #142, opened 2021-11-02
Feature: Default systemd service for zenoh router


Aposhian:
When running `zenohd` routers in a persistent network topology, it is desirable to configure `zenohd` as a service. It seems like it would be ergonomic to provide a default systemd service with the package install.

For the short term, it could simply use `Type=Simple`, but it would be great if it could use `Type=notify`, so it could notify the system when zenoh is ready.

gabrik:
Hi @Aposhian, indeed having a service file for `zenohd` will simplify a lot of use-cases like yours.
We have created a PR (see #143) that implements a service file with `Type=Simple`, for `Type=Notify` we need more internal discussion as it requires to implement some logic in `zenohd`. 

Please feel free to check the code from the PR and let us know if it solves (or partially) your issue.

PR #143, opened 2021-11-03
Adding systemd service file, updating Debian metadata informations 

gabrik:
This PR contains the service file for the zenoh router, with `Type=simple`, along with some maintainer scripts ( postinst, and postrm).
Logs are managed also by systemd and can be seen using ` journalctl -u zenoh`.
This should address #142.

Ideally, I would like also to update the rmp generation to embed the service file and the postinst/postrm scripts, but I did not find the related metadata in `zenoh/Cargo.toml`


drauschenbach:
There's a lot that goes into Debian APT packaging. My opinion is that use of https://github.com/kornelski/cargo-deb would be the more idiomatic way to support APT packaging. It encapsulates a non-trivial amount of intelligence around multi-architecture target support, and it streamlines configuration.

gabrik:
@drauschenbach that is already the case, we are using `cargo deb`, please look [here](https://github.com/gabrik/zenoh/blob/feat/systemd-service/zenoh/Cargo.toml#L105), it manages the maintainer scripts and the service file.

Ideally the same should be done for rpm packages with `cargo rpm`, but the documentation of this one is not as good as the `cargo deb` one.

drauschenbach:
Nice! Sorry I missed it.

zenoh/.service/zenoh.service
```diff
@@ -0,0 +1,20 @@
+[Unit]
+Description = Eclipse Zenoh Router
+
```
Aposhian, line 3: It seems like it would make sense to add network dependencies to the `[Unit]` section:
```
After=network-online.target
Wants=network-online.target
```

Aposhian: We can also link to the zenoh docs from here:
```
Documentation=https://zenoh.io/docs/getting-started/key-concepts/
```

Aposhian (changes requested):
LGTM. We could just add some more directives to the `[Unit]` section for completeness.

zenoh/.service/zenoh.service
Aposhian, line 3:
```suggestion
Documentation=https://zenoh.io/docs/getting-started/key-concepts/
After=network-online.target
Wants=network-online.target
```

JEnoch (approved):

PR #144, opened 2021-11-09
Configuration now supports plugins

p-avital:


p-avital:
As soon as this PR is merged, I'll make corresponding PRs on all plugins and storage backends. Unless a matching branch exists for these repos, I'll have to make the PRs to their masters

PR #147, opened 2021-11-18
Fix zenohd start issue where non-configured plugins were still loaded

p-avital:


p-avital:
Plugin regressions seem fixed :)

PR #152, opened 2021-12-02
Fix string-like serde visitors to accept any string-like type

p-avital:


zenoh/src/bin/zenohd.rs
```diff
@@ -66,11 +66,11 @@ fn main() {
             ))
             .arg(Arg::from_usage(
                 "-P, --plugin=[PATH_TO_PLUGIN_LIB]... \
-             'A plugin that must be loaded. Repeat this option to load several plugins.'",
+             'A plugin that MUST be loaded. Repeat this option to load several plugins. If loading failed, zenohd will exit. Use --try-plugin if you just want a load attempt to be made.'",
```
JEnoch, line 69: `--try-plugin` doesn't exist

JEnoch (changes requested):

PR #190, opened 2022-01-19
WIP: zenoh-util split

p-avital:
Zenoh Util was simple to split, but a LOT of macros needed to be reimported, hence the huge diff

Gonna work on splitting Zenoh now, but merging this first will avoid potential conflicts for later


p-avital:
To make things clearer, I've extracted into crates most of the top-level modules of zenoh-util. Specifically:
* `zenoh_util::core` -> `zenoh-core`
* `zenoh_util::collections` -> `zenoh-collections`
* `zenoh_util::crypto` -> `zenoh_crypto`
* `zenoh_util::properties` -> `zenoh-cfg-properties`
* `zenoh_util::sync` -> `zenoh_sync`

To avoid breaking changes, I've re-exported all these crates in zenoh-util. The macros are being annoying though.

I've changed most uses of `zenoh_util::<mod>` to `zenoh_<mod>` uses. This should let us drop the dependency on `zenoh_util` in subcrates of `zenoh` itself that only require `zenoh-core` for example, which should help with parallelization of compilation :)

@JEnoch should we merge this before it becomes painful to do so ?

p-avital:
Switching to WIP as I'm gonna check other repos for breakage and do associated PRs if necessary

p-avital:
Going to edit this commit to keep track
* `zenoh-c`: cleared
* `zenoh-python`: https://github.com/eclipse-zenoh/zenoh-python/pull/47

JEnoch:
Some comments:
 - for cleaner repo org, I would have grouped all those utility crates under a `utils/` directory (or any other relevant name)
 - for publication to crates.io, all Cargo.toml must include some fields: https://doc.rust-lang.org/cargo/reference/publishing.html#before-publishing-a-new-crate
 - you should also add the copyright header to files that don't have it
 - I think all the crates versions should be aligned to zenoh version (currently 0.6.0-dev.0), and thus aligned to the git tag for each release (I'll do a script to automatise version change in all Cargo.toml)
 - did you measure the build time gain ? :-)

p-avital:
> * for cleaner repo org, I would have grouped all those utility crates under a `utils/` directory (or any other relevant name)

Agreed
> * for publication to crates.io, all Cargo.toml must include some fields: https://doc.rust-lang.org/cargo/reference/publishing.html#before-publishing-a-new-crate

Right, forgot, that's gonna be a LOT of copy pasting \o/
> * you should also add the copyright header to files that don't have it
> * I think all the crates versions should be aligned to zenoh version (currently 0.6.0-dev.0), and thus aligned to the git tag for each release (I'll do a script to automatise version change in all Cargo.toml)

More copy pasting \o/
> * did you measure the build time gain ? :-)

No measure. It feels slightly faster, but `zenoh` is still so damn slow that it'll mostly make a difference when it's also parallelized. You do get to see parts of zenoh-util get built WAY earlier than they used to :)

p-avital:
Everything should be OK now. This PR can be merged and the others can follow :)

JEnoch (approved):
Looks good to me, as it builds and survives to Clippy!

p-avital:
@JEnoch don't forget to merge this PR's friends, otherwise other masters will be broken (path of some macros got broken :( )

JEnoch:
Yep! In progress... (I re-kicked all CI checks)

PR #195, opened 2022-01-27
Crate splitting

p-avital:
Please pass all tests...


p-avital:
For gains:
* Debug build went from 48s to 43s on my machine.
* Release build went from 2m20s to 1m51s \o/
* Most tests only need their actual dependencies instead of all of `zenoh`, Hurray!
* rust-analyzer is much happier now, with linting being almost instant in the better split crates.
* I cut off some circular dependencies between components. `TransportManager` is now slightly less of a god-object

p-avital:
Just fixed the merge issue on `Cargo.lock` introduced by Olivier adding the `stop-token` dependency

@OlivierHecart @JEnoch @kydos @Mallets everyone happy with merging this PR? :)

JEnoch (approved):
LGTM (and cleaner than it used to be :-) !

PR #200, opened 2022-02-07
Make Selector ownable

JEnoch:
This makes the `Selector` ownable by:
 - changing its `pub value_selector: &'a str` field's type into `Cow<'a, str>`
 - adding a `Selector::to_owned(&self)` operation
 - adding a `From<String> for Selector` impl

### DIFF
--- a/zenoh-ext/src/querying_subscriber.rs
+++ b/zenoh-ext/src/querying_subscriber.rs
@@ -135,7 +135,7 @@ impl<'a, 'b> QueryingSubscriberBuilder<'a, 'b> {
     {
         let selector = query_selector.into();
         self.query_key_expr = selector.key_selector.to_owned();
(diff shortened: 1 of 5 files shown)


p-avital:
As discussed off-issue:
* This adds 16 bytes to `Selector`'s stack size, which is sad (especially since the 8 bytes discriminant could be seen as redundant with the inner `KeyExpr`'s `Cow`'s)
* This is useful, as this allows selectors to become `'static`, which is for example needed by the Python bindings.
We're starting to have quite a few "maybe owned" types, which is natural as we try to avoid making copies as much as possible, yet provide async functionalities which tend to require being able to guarantee `'static` lifetimes. I'm going to start studying how we can write these types more efficiently, both runtime-cost-wise and semantics-wise :)

PR #210, opened 2022-02-21
use ZError in WhatAmI for better parsing enum match

YuanYuYuan:
As we discussed before, this PR lets users handle the enum match of WhatAmI.

### DIFF
--- a/commons/zenoh-protocol-core/src/whatami.rs
+++ b/commons/zenoh-protocol-core/src/whatami.rs
@@ -1,4 +1,5 @@
 use super::{NonZeroZInt, ZInt};
+use zenoh_core::{bail, zresult::ZError};
 
 #[repr(u8)]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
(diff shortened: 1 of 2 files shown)


commons/zenoh-protocol-core/src/whatami.rs
```diff
@@ -9,14 +11,14 @@ pub enum WhatAmI {
...
-    type Err = ();
+    type Err = ZError;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         match s {
             "router" => Ok(WhatAmI::Router),
             "peer" => Ok(WhatAmI::Peer),
             "client" => Ok(WhatAmI::Client),
-            _ => Err(()),
+            _ => bail!("{} is not one of the router, peer, or client.", s),
```
Mallets, line 21: I suggest to change the error message as follows:

bail!("{} is not a valid WhatAmI value. Valid values are: [\"router\", \"peer\", \"client\"]., s)

YuanYuYuan: Okay, let me change the writing. Thanks!

Mallets:
The CI is failing. 

Please run `cargo clippy --all-targets` to verify that `WhatAmI` error is handled properly in all the rest of the code.

YuanYuYuan:
> The CI is failing.
> 
> Please run `cargo clippy --all-targets` to verify that `WhatAmI` error is handled properly in all the rest of the code.

Oops, thanks for your reminder!  I had resolved that issue and rebased this PR to the latest commit

PR #211, opened 2022-02-22
Fix late coming waiters halting on triggered Signal

jerry73204:
The previous Signal impl wakes up waiters by [event_listener::Event](https://docs.rs/event-listener/latest/event_listener/struct.Event.html), which does not take effect if a waiter comes later than the signal is triggered. This patch changes Event to tokio's [Semaphore](https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html) to remedy this issue. I realized this issue when working on my [async-watch-once](https://github.com/jerry73204/async-once-watch).

Tokio's semaphore is the only implementation that supports async/await and adding permits after creations. Unfortunately, I cannot find equivalent impls and introduce tokio to the dependency tree.

This PR provides an additional task to check slow waiters are not blocked by triggered signal.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3582,7 +3582,9 @@ dependencies = [
  "async-std",
  "event-listener",
  "flume",
+ "futures",
  "futures-lite",
+ "tokio",
  "zenoh-core",
 ]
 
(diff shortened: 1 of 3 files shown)


Mallets:
The CI is failing. It seems that the introduced change breaks the expected behaviour in the tests.

jerry73204:
Thanks that CI catches a bug, showing that triggering twice causes panic. It's fixed for now.

PR #214, opened 2022-03-01
Here comes the airplane!

p-avital:
This PR contains the following:
* The CI has been updated to parallelize more:
  * cargo-nextest is now used to parallelize testing
  * rustfmt and clippy checks now run in separate jobs, letting tests begin sooner
  * Through this, CI completion time has gone down from 1h25 to 22 minutes (with ubuntu finishing in 17 minutes).
* Configuration has been organized a bit better before 0.6.0-beta0 hits.
* The example configuration values now make sense. Except for the `plugins` section which is explicitly non-default to show how _our_ plugins behave, the values are now the default ones.
* `Encoding` has been refactored to an enum. It's size is now 5usize, instead of 1u64+4usize. Its capabilities are unchanged, but using it properly should now be more intuitive through pattern matching.

This PR causes a few API breaks, so sister-PRs are necessary for:
- [x] `zenoh-c` : eclipse-zenoh/zenoh-c#45
- [x] `zenoh-python` : eclipse-zenoh/zenoh-python#53
- [x] `zenoh-plugin-dds` : eclipse-zenoh/zenoh-plugin-dds#76
- [x] `zenoh-plugin-webserver` : nothing to do
- [x] `zenoh-backend-filesystem` : eclipse-zenoh/zenoh-backend-filesystem#11
- [x] `zenoh-backend-influxdb` : eclipse-zenoh/zenoh-backend-influxdb#10
- [x] `zenoh-backend-rocksdb` : eclipse-zenoh/zenoh-backend-rocksdb#6


p-avital:
Once this PR is merged, we need to add a commit to each sister PR to update the cargo lock, otherwise it's likely not going to work. Checklist is back :)
- [ ] `zenoh-c`
- [ ] `zenoh-python`
- [ ] `zenoh-plugin-dds`
- [ ] `zenoh-plugin-webserver`
- [ ] `zenoh-backend-filesystem`
- [ ] `zenoh-backend-influxdb`
- [ ] `zenoh-backend-rocksdb`

PR #218, opened 2022-03-09
add support for ZENOH_CFG_FILE env var

p-avital:


Mallets:
I would simply call it `ZENOH_CONFIG_FILE` to be aligned with the CLI parameter.

JEnoch:
Few points worth to discuss:
 - env var name: why not just `ZENOH_CONFIG` (more aligned with `--config` CLI parameter) ?
 - in current PR it only applies to `zenohd`. Why not also having the zenoh apps to use this ENV VAR (same way than CycloneDDS apps do get their config from `CYCLONEDDS_URI`) ? This could be either the behaviour of `Config::default()`, either of a more explicit `Config::from_env()` operation.

Mallets:
After a discussion with @p-avital , we have concluded that having an environment variable for pointing to the configuration is a mistake we should avoid. 
Here the reasons:

1. Precedence of importance. What is more important? The configuration passed by command line? Or the one set by environment variables? Precedence rules should be defined and clearly explained to the users.
2. Two ways to do the same thing. While sometimes it's good to have multiple ways to achieve the same goal in different ways, in case of critical parameters, like the router configuration, it is better to have one single way to do it that is clear and easily reproducible for everyone.
3. Debuggability. In case of troubleshooting a system, it's harder to have the complete view of the system state if it scattered in different places. Environment variables is one of those cases where it's not apparent what's the actual state or content and may lead to unfortunate misconfiguration and misbehaviour, specially if combined with 1.

For the above reasons, we believe that is better to not add such options that would bring little benefits and many drawbacks. 

Opinions?

PR #230, opened 2022-03-28
fix --cfg behaviour being broken with commas

p-avital:


p-avital:
@Mallets flaky test? I didn't touch that :(

JEnoch:
All green after 2nd attempt => flaky test

PR #252, opened 2022-04-14
Modifying on_sample API of storage management plugin

sreeja:
Storage informs back the plugin whether the given sample was used or not.


JEnoch (commented):
LGTM!

PR #266, opened 2022-04-22
New API to get all (key, timestamp) entries in the database

sreeja:


plugins/zenoh-plugin-storage-manager/src/memory_backend/mod.rs
```diff
@@ -227,6 +227,11 @@ impl Storage for MemoryStorage {
 
     async fn on_query(&mut self, query: Query) -> ZResult<()> {
         trace!("on_query for {}", query.key_selector());
+        // Note: uncomment these lines to test get_all_entries API
```
JEnoch, line 230: This commented code should be removed

plugins/zenoh-plugin-storage-manager/src/memory_backend/mod.rs
```diff
@@ -246,6 +251,15 @@ impl Storage for MemoryStorage {
         }
         Ok(())
     }
+
+    async fn get_all_entries(&self) -> ZResult<Vec<(String, Timestamp)>> {
+        let mut result = Vec::new();
```
JEnoch, line 256: It would be more efficient to create the Vec as such (after getting the read lock on map):
```rust
let mut result = Vec::with_capacity(map.len());
```

plugins/zenoh-plugin-storage-manager/src/memory_backend/mod.rs
```diff
@@ -246,6 +251,15 @@ impl Storage for MemoryStorage {
...
+        let map = self.map.read().await;
+        for k in map.keys() {
```
JEnoch, line 258: It would be more efficient to loop on `map.iter()` to get each entry. That would avoid the `map.get()` for each key.

JEnoch (changes requested):

JEnoch (approved):

PR #269, opened 2022-04-28
Use unified async executor in one crate

jerry73204:
This PR moves commonly used runtime functions, `spawn()`, `block_on()` and `sleep()` into a `zenoh-async-rt` crate, and code to use functions from that crate. It ensures the whole repo runs on the unified runtime, and eases the version update of `async-std`. It also paves the way  to async-runtime agnostic feature.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3468,6 +3468,7 @@ dependencies = [
  "uhlc",
  "uuid",
  "vec_map",
+ "zenoh-async-rt",
  "zenoh-buffers",
  "zenoh-cfg-properties",
  "zenoh-collections",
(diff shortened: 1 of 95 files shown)


p-avital:
Thanks @jerry73204, we were planning to do something similar. We'd like to have the executor become an instantiable rather than a singleton, but this at least paves the way.

I'm on holiday until Monday, and would like to check a few things (mainly if this pr doesn't break the fact that we used a dedicated executor to fix a starvation issue we had a couple months ago), but I'm happy this work gets started :)

milyin:
This PR is inactive for a year. Closing it

PR #280, opened 2022-06-01
WIP: Keyexpr refactor and some method renaming

p-avital:


zenoh/src/publication.rs
```diff
@@ -67,7 +69,7 @@ pub type DeleteBuilder<'a> = PutBuilder<'a>;
 /// ```
 #[derive(Debug, Clone)]
 pub struct PutBuilder<'a> {
-    pub(crate) publisher: Publisher<'a>,
+    pub(crate) publisher: PublishBuilder<'a>,
```
Mallets, line 72: Shouldn't be `PublisherBuilder`?

zenoh/src/query.rs
```diff
@@ -143,15 +143,28 @@ pub(crate) struct QueryState {
...
+            selector: match &self.selector {
+                Ok(s) => Ok(s.clone()),
+                Err(e) => Err(zerror!("Cloned Selector Error: {}", e).into()),
+            },
+            target: self.target,
+            consolidation: self.consolidation,
+            local_routing: self.local_routing,
+        }
+    }
+}
```
Mallets, line 167: Similar to a previous comment, why a builder should be clonable?

zenoh/src/session.rs
```diff
@@ -133,42 +138,69 @@ impl SessionState {
     }
 
     #[inline]
-    fn localid_to_expr(&self, id: &ExprId) -> ZResult<String> {
+    fn localid_to_expr<'a>(&'a self, id: &ExprId) -> ZResult<KeyExpr<'a>> {
```
Mallets, line 141: Rename it to `local_wireid_to_keyexpr`?

zenoh/src/session.rs
```diff
@@ -133,42 +138,69 @@ impl SessionState {
...
+                        key_expr.suffix.as_ref(),
+                    ]
+                    .concat(),
+                )
+            })
         }
     }
 
-    pub fn key_expr_to_expr(&self, key_expr: &KeyExpr, local: bool) -> ZResult<String> {
+    pub(crate) fn key_expr_to_expr<'a, 'b>(
```
Mallets, line 199: Rename it to `wireexpr_to_keyexpr`?

zenoh/src/session.rs
```diff
@@ -11,14 +11,17 @@
 // Contributors:
```
Mallets, line 11: I didn't see the renaming of `declare_expr` into `declare_keyexpr`.

zenoh/src/prelude.rs
```diff
@@ -23,1155 +23,1220 @@
...
-    /// if the [`Sample`] was caused by a `delete` operation.
-    Delete = data_kind::DELETE as isize,
-}
+    /// The kind of a [`Sample`].
+    #[derive(Debug, Copy, Clone, PartialEq)]
+    pub enum SampleKind {
+        /// if the [`Sample`] was caused by a `put` operation.
+        Put = data_kind::PUT as isize,
+        /// if the [`Sample`] was caused by a `patch` operation.
+        Patch = data_kind::PATCH as isize,
```
Mallets, line 489: There is no `Patch` operation implemented as of today. Let's remove it from this enum as well from the `data_kind`.

zenoh-ext/src/querying_subscriber.rs
```diff
@@ -18,31 +19,49 @@ use zenoh::query::{QueryConsolidation, QueryTarget};
...
     reliability: Reliability,
     mode: SubMode,
     period: Option<Period>,
-    query_key_expr: KeyExpr<'b>,
-    query_value_selector: String,
+    query_selector: Option<ZResult<Selector<'b>>>,
     query_target: QueryTarget,
     query_consolidation: QueryConsolidation,
 }
+impl Clone for QueryingSubscriberBuilder<'_, '_> {
```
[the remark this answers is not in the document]
p-avital: These are just for consistency with before, where all builders were indeed `Clone`. I'm not against removing that, and maybe re-introducing it later, though.

zenoh/src/key_expr/mod.rs
```diff
@@ -0,0 +1,210 @@
...
+
+pub use zenoh_protocol_core::key_expr::*;
+
+#[derive(Clone)]
+pub(crate) enum KeyExprInner<'a> {
+    Borrowed(&'a keyexpr),
+    Owned(OwnedKeyExpr),
+    Wire {
+        key_expr: OwnedKeyExpr,
+        expr_id: u32,
```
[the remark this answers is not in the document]
p-avital: I'm fine with making `expr_id` a `u64`, we can review whether `u64` makes sense for `WireExpr` later. Hopefully `rustc` will spot the gap with the other `u32` and keep `KeyExpr` at its current 4usize by using the niche :)

zenoh/src/prelude.rs
```diff
@@ -23,1155 +23,1220 @@
...
-                kind: data_info.kind.unwrap_or(data_kind::DEFAULT).into(),
-                timestamp: data_info.timestamp,
-                source_info: data_info.into(),
-            }
-        } else {
+    /// A zenoh sample.
+    #[derive(Clone, Debug)]
+    pub struct Sample {
+        /// The key expression on which this Sample was published.
+        pub key_expr: KeyExpr<'static>,
```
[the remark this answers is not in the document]
p-avital: We've been using `key_expr` consistently as a value name. I'd like to reserve `keyexpr` for the name of the newtype on `str`. If there are inconsistencies where `keyexpr` is used to name a value, these should be the ones to correct IMO :)

zenoh/src/prelude.rs
```diff
@@ -23,1155 +23,1220 @@
...
+    ///  * __fragment__: a list of fields names allowing to return a sub-part of each value.
+    ///    This feature only applies to structured values using a “self-describing” encoding, such as JSON or XML.
+    ///    It allows to select only some fields within the structure. A new structure with only the selected fields
+    ///    will be used in place of the original value.
+    ///
+    /// _**NOTE**_: _the filters and fragments are not yet supported in current zenoh version._
+    pub struct Selector<'a> {
+        /// The part of this selector identifying which keys should be part of the selection.
+        /// I.e. all characters before `?`.
+        pub key_selector: KeyExpr<'a>,
```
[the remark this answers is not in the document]
p-avital: `key_expr`? I feel like `Selector` fields are very poorly named.

Maybe they should be `key_expr` and `arguments` (or `query`, as this is what this segment of a URL is named in REST).

Also, we should have a quick review on `Selector` construction (that leading `?` doesn't quite sit right with me) and either specify it properly like we finally did with `KeyExpr`, or stop acting like it's specified at all.

zenoh/src/session.rs
p-avital: because I borked my grep in the API harmonisation discussion xD

zenoh/src/prelude.rs
p-avital: I'm kind of on the fence on that: if someone wants to implement a queryable that supports path operations, wouldn't it be nice for them to have this?

Just because we don't support it anywhere yet doesn't mean it doesn't make sense for these queries to be able to exist.

But on the other hand, I get that it would be confusing for users that first-party queryables don't support this operation despite it existing, because it is indeed a pointless operation for now.

plugins/example-plugin/src/lib.rs
```diff
@@ -16,14 +16,15 @@
...
+use std::convert::TryFrom;
 use std::sync::{
     atomic::{AtomicBool, Ordering::Relaxed},
     Arc, Mutex,
 };
 use zenoh::net::runtime::Runtime;
 use zenoh::plugins::{Plugin, RunningPluginTrait, ValidationFunction, ZenohPlugin};
 use zenoh::prelude::*;
-use zenoh::utils::key_expr;
+use zenoh::utils::wire_expr;
```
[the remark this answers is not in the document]
p-avital: They should

zenoh/src/prelude.rs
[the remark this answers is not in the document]
Mallets: Ok then :)

zenoh/src/prelude.rs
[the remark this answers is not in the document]
Mallets: Ok. I'm fine with it.

PR #294, opened 2022-07-27
CI: Improve Release workflow wrt. Docker builds

JEnoch:
Prior to this PR the Release workflow was running for +3h, including +2h to build the Docker images.
This was mainly caused by the build for an arm64 image that was relying on Qemu.
Moreover, the build was [failing sometime](https://github.com/eclipse-zenoh/zenoh/actions/runs/2552804647), probably because of lack of memory in the container

This PR changes the way the Docker images are built:
instead of running the Rust compilation within a Docker container, it's run on the Github runner targeting both `x86_64-unknown-linux-musl` and `aarch64-unknown-linux-musl` platforms, for respectively Alpine x86+64 and arm64 images.
Then, the `.github/workflows/Dockerfile` just copies the binaries (`zenohd` + libs) in Docker images.

Note that I also deactivated the build of examples binaries, as it takes a lot of time and leads to far bigger archives, for unlikely usage.


JEnoch:
Build time down from +3h to ~1h  😄

Issue #302, opened 2022-08-16
[Bug] Zenoh does not build on FreeBSD


gabrik:
### Describe the bug

Neither router nor the examples compile in FreeBSD.
It seems related to the dependency `quinn-udp`, supporting the **QUIC** protocol.

Disabling the features does not solve the issue, the library is still compiled, causing the build to fail.

Here is the complete error:
```
$ cargo build -p zenohd --no-default-features --features transport_tcp
....
Compiling quinn-udp v0.1.2
error[E0422]: cannot find struct, variant or union type `in_pktinfo` in crate `libc`
   --> /home/ato/.cargo/registry/src/github.com-1ecc6299db9ec823/quinn-udp-0.1.2/src/unix.rs:456:41
    |
456 |                       let pktinfo = libc::in_pktinfo {
    |                                           ^^^^^^^^^^ help: a struct with a similar name exists: `in6_pktinfo`
    |
   ::: /home/ato/.cargo/registry/src/github.com-1ecc6299db9ec823/libc-0.2.126/src/unix/bsd/freebsdlike/mod.rs:83:1
    |
83  | / s! {
84  | |     pub struct in_addr {
85  | |         pub s_addr: ::in_addr_t,
86  | |     }
...   |
374 | |     }
375 | | }
    | |_- similarly named struct `in6_pktinfo` defined here

error[E0425]: cannot find value `IP_PKTINFO` in crate `libc`
   --> /home/ato/.cargo/registry/src/github.com-1ecc6299db9ec823/quinn-udp-0.1.2/src/unix.rs:463:58
    |
463 |                     encoder.push(libc::IPPROTO_IP, libc::IP_PKTINFO, pktinfo);
    |                                                          ^^^^^^^^^^ help: a constant with a similar name exists: `IPV6_PKTINFO`
    |
   ::: /home/ato/.cargo/registry/src/github.com-1ecc6299db9ec823/libc-0.2.126/src/unix/bsd/freebsdlike/mod.rs:968:1
    |
968 | pub const IPV6_PKTINFO: ::c_int = 46;
    | ------------------------------------- similarly named constant `IPV6_PKTINFO` defined here

error[E0531]: cannot find unit struct, unit variant or constant `IP_PKTINFO` in crate `libc`
   --> /home/ato/.cargo/registry/src/github.com-1ecc6299db9ec823/quinn-udp-0.1.2/src/unix.rs:523:38
    |
[...]

PR #303, opened 2022-08-16
Allow build of zenoh artifacts with specified features

gabrik:
This PR updates the `Cargo.toml` files for `zenohd`, `examples` and `zenoh-ext` crates.
Now all the features are re-exported, thus allowing compilation only of the needed dependencies.

This solves #302, thus now Zenoh can build on FreeBSD.

Now this works.
```
$ cargo build --bin zenohd --example z_pub --example z_sub --no-default-features --features transport_tcp
...
Finished dev [unoptimized + debuginfo] target(s) in 0.16s
```

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -32,28 +32,7 @@ readme = "README.md"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
-auth_pubkey = ["zenoh/auth_pubkey"]
-auth_usrpwd = ["zenoh/auth_usrpwd"]
-complete_n = ["zenoh/complete_n"]
 shared-memory = ["zenoh/shared-memory"]
-stats = ["zenoh/stats"]
-transport_quic = ["zenoh/transport_quic"]
-transport_tcp = ["zenoh/transport_tcp"]
-transport_tls = ["zenoh/transport_tls"]
-transport_udp = ["zenoh/transport_udp"]
-transport_unixsock-stream = ["zenoh/transport_unixsock-stream"]
-transport_serial = ["zenoh/transport_serial"]
-transport_ws = ["zenoh/transport_ws"]
-default = [
-    "auth_pubkey",
-    "auth_usrpwd",
-    "shared-memory",
-    "transport_quic",
-    "transport_tcp",
-    "transport_tls",
-    "transport_udp",
-    "transport_unixsock-stream",
-]
 
 [dependencies]
 zenoh = { path = "../zenoh/" , default-features = false }
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -56,8 +56,8 @@ default = [
 ]
 
 [dependencies]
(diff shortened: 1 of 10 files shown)


Mallets (approved):

Mallets:
The PR seems to work fine. However, it's unclear to me how cargo features are propagated and enabled along the chain of crates. I would like @JEnoch to also have a look at it.

JEnoch:
The issue with the Cargo.tomls prior to this PR was that all packages depending on `zenoh` package (zenohd, examples, backends...) use its default features (including QUIC transport which is problematic on FreeBSD).
And building with `--no-default-features` disable the features only for the selected package(s), not for its dependencies. So there was no way to disable some transport via command-line option.

Changing all dependencies to `zenoh` to use `default-features = false` might look weird, since one would assume zenohd and examples will be built by default without any transport. But since its build within the workspace that contains the `zenoh` package, this one is by default enabling (almost) all the transports, and `zenohd` will rely on this `zenoh` package build with transports enabled.
Still, we can build the zenohd and all examples with TCP and UDP transports only as such:
```
cargo build -p zenohd -p zenoh-examples -p zenoh-ext --no-default-features --features zenoh/transport_tcp,zenoh/transport_udp --examples
```
This disables all features in `zenoh`, but TCP and UDP transports. And the same applies to zenohd and examples.

However this causes 5 warnings building `io/zenoh-transport` that have to be solved:
```log
warning: unused import: `crate::unicast::establishment::authenticator::PeerAuthenticatorId`
  --> io/zenoh-transport/src/unicast/establishment/open/init_ack.rs:14:5
   |
14 | use crate::unicast::establishment::authenticator::PeerAuthenticatorId;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: variable does not need to be mutable
   --> io/zenoh-transport/src/unicast/establishment/authenticator/mod.rs:145:13
    |
145 |         let mut pas = HashSet::new();
    |             ----^^^
    |             |
    |             help: remove this `mut`
    |
    = note: `#[warn(unused_mut)]` on by default

warning: variable does not need to be mutable
   --> io/zenoh-transport/src/unicast/establishment/open/init_ack.rs:114:9
    |
114 |     let mut is_shm = false;
    |         ----^^^^^^
    |         |
    |         help: remove this `mut`

warning: variable does not need to be mutable
   --> io/zenoh-transport/src/unicast/establishment/open/init_ack.rs:117:13
    |
117 |         let mut att = pa
    |             ----^^^
    |             |
    |             help: remove this `mut`

warning: variable does not need to be mutable
   --> io/zenoh-transport/src/unicast/transport.rs:438:35
    |
438 |     pub(crate) fn schedule(&self, mut message: ZenohMessage) -> bool {
    |                                   ----^^^^^^^
    |                                   |
    |                                   help: remove this `mut`

warning: `zenoh-transport` (lib) generated 5 warnings
```

JEnoch (approved):

PR #304, opened 2022-08-16
Simplify consolidation to reality, removing strategy

p-avital:


commons/zenoh-protocol-core/src/lib.rs
```diff
@@ -461,99 +461,12 @@ pub mod queryable {
 
 /// The kind of consolidation.
 #[derive(Debug, Clone, PartialEq, Eq, Copy)]
-#[repr(u8)]
 pub enum ConsolidationMode {
     None,
     Lazy,
     Full,
 }
```
sreeja, line 468: I prefer the words `Monotonic` and `LastValue` (as Pierre mentioned on discord) to `Lazy` and `Full`. It is more explicit and clear to the user.

Mallets: Since the `LastValue` is based on `timestamp`, shouldn't be `LatestValue` to be more explicit about its relationship with ordered time?

p-avital: Agreed, just forgot to actually write it, thaks @sreeja :)

Mallets: Moreover, `Latest` would be simpler and more aligned with `None` and `Monotonic`.

commons/zenoh-protocol-core/src/lib.rs
```diff
@@ -462,9 +462,18 @@ pub mod queryable {
...
+    /// Monotonic consolidation immediately forwards samples, except if one with an equal or more recent timestamp
+    /// has already been sent with the same key.
+    ///
+    /// This optimizes latency while potentially reducing bandwidth.
+    ///
+    /// Note that this doesn't cause re-ordering, but drops the samples for which a more recent timestamp has already
+    /// been observed with the same key.
+    Monotonic,
+    /// Holds back samples to only send the set of samples that had the highest timestamp for their key.
+    LatestValue,
```
Mallets, line 476: Let's simply use `Latest`.

PR #306, opened 2022-08-17
Removed kind from queryables

Mallets:


commons/zenoh-protocol/src/proto/msg.rs
```diff
@@ -1187,7 +1175,7 @@ impl ZenohMessage {
         key: WireExpr<'static>,
         value_selector: String,
         qid: ZInt,
-        target: Option<QueryTAK>,
+        target: Option<QueryTarget>,
```
p-avital, line 1178: Shouldn't QueryTarget be mandatory at this point? Seems odd to me that the protocol would allow not specifying any.

commons/zenoh-protocol/src/proto/msg_writer.rs
```diff
@@ -521,7 +516,7 @@ impl MessageWriter for WBuf {
         zcheck!(self.write_string(&query.value_selector));
         zcheck!(self.write_zint(query.qid));
         if let Some(t) = query.target.as_ref() {
```
p-avital, line 518: if query.target != BestMatching instead? Here, if `query.target == Some(BestMatching)`, we're gonna spend a byte just to send it

p-avital: don't forget to change the `impl Header for Query` to match, and the reader side

p-avital (changes requested):

commons/zenoh-protocol/src/proto/msg.rs
Mallets: Yes, but I wouldn't change it now. In the forthcoming version of the protocol default values will be treated as such and not as options.

commons/zenoh-protocol/src/proto/msg_writer.rs
Mallets: As said in the previous comment, I wouldn't change it now. In the forthcoming version of the protocol default values will be treated as such and not as options.

zenoh/src/net/routing/queries.rs
```diff
@@ -1087,14 +944,13 @@ pub(crate) fn queries_tree_change(
                         WhatAmI::Router => &res.context().router_qabls,
                         _ => &res.context().peer_qabls,
                     };
-                    for ((qabl, kind), qabl_info) in qabls {
+                    for (qabl, qabl_info) in qabls {
```
p-avital, line 947: Should be `if let Some(quabl_info) = quabls.get(tree_id)`

p-avital (changes requested):

p-avital (approved):

PR #309, opened 2022-08-18
Reopening PR on removing queryable kind

Mallets:
It seems that a regression is still in place when interoperating with zenoh-pico that needs to be fixed.


Mallets:
Integration test with zenoh-pico is ok: https://github.com/eclipse-zenoh/zenoh-pico/actions/runs/2884853882
Proceeding in re-merging this PR.

PR #310, opened 2022-08-25
Anti-entropy protocol

sreeja:
This PR looks into aligning multiple key-value stores subscribing to the same key expression. Addressing https://github.com/eclipse-zenoh/roadmap/issues/43


plugins/zenoh-plugin-storage-manager/src/replica/storage.rs
```diff
@@ -0,0 +1,248 @@
...
+            Ok(storage_queryable) => storage_queryable,
+            Err(e) => {
+                error!("Error starting storage {} : {}", self.name, e);
+                return Err(e);
+            }
+        };
+
+        if self.replication.is_some() {
+            let aligner_updates = &self.replication.as_ref().unwrap().aligner_updates;
+            loop {
```
JEnoch, line 97: This loop is blocking the caller. It never returns unless stopped via the channel. But the channel's `tx` is returned to the caller only if loop is stopped...
This loop should be running in a side task.

plugins/zenoh-plugin-storage-manager/src/replica/storage.rs
```diff
@@ -0,0 +1,248 @@
...
+                            Err(e) => {
+                                error!("Storage Message Channel Error: {}", e);
+                                return Err(e.into());
+                            },
+                        };
+                    }
+                );
+            }
+        } else {
+            loop {
```
JEnoch, line 138: Same comment: this loop should be running in a side task.

plugins/zenoh-plugin-storage-manager/src/replica/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+            replica.session.clone(),
+            replica.key_expr.clone(),
+            &replica.name,
+            storage,
+            in_interceptor,
+            out_interceptor,
+            Some(replication),
+        );
+
+        let result = join!(
```
JEnoch, line 147: This `join` will block until all the tasks are done. But there is no mechanism to stop the tasks, except storage_task.
The storage_task must stop the other tasks when ending.
And this operation must not wait for the tasks to end.

JEnoch (changes requested):

EXAMPLE_CONFIG.json5
```diff
@@ -147,6 +147,34 @@
           /// The "memory" volume is always available, and doesn't require any per-storage options, so requesting "memory" by string is always sufficient.
           volume: "memory",
         },
+        demo2: {
+          key_expr: "demo/memory2/**",
+          volume: "memory",
+          /// If multiple storages subscribing to the same key_expr should be synchronized, declare them as replicas.
+          /// In the absence of this configuration, a normal storage is initialized
+          replica_config: {
```
Mallets, line 155: I wonder wether this type of configuration should be done per "storage", or to have a more generic section with common parameters that I expect to be rarely changed per storage-basis.

sreeja: It should ideally be per `key_expr` since that is the basic unit of replication. But some parameters should be system-wide, for example, the `align_prefix`; and I am not sure if we should let the user configure it. Maybe we don't expose it to the user until it is deemed necessary?

sreeja: I removed `align_prefix`, `subintervals`, `hot`, `warm` since it would be problematic if we let the user configure it. The first two is now hard coded into the system, and the last two are computed from other user-provided configuration.

sreeja: Moreover it is "per key_expr" rather than per storage

plugins/zenoh-plugin-storage-manager/src/replica/storage.rs
sreeja: Moved the entire `Replica` and `StorageService` into independent tasks. This solves the problem here. Avoiding creating multiple subtasks spawns since it will be problematic/error-prone to manage the replica state then.

plugins/zenoh-plugin-storage-manager/src/replica/storage.rs
sreeja: The previous solution is applicable here too.

plugins/zenoh-backend-traits/src/config.rs
```diff
@@ -45,10 +46,55 @@ pub struct StorageConfig {
...
+    pub hot: usize,
+    pub warm: usize,
+}
+
+impl Default for ReplicaConfig {
+    fn default() -> Self {
+        Self {
+            // This is the prefix that is used by the alignment protocol
+            // TODO: discuss if this needs to be configurable
+            align_prefix: String::from("@-digest"),
```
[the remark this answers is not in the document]
sreeja: Removed `align_prefix` option from config. Instead, it is built-in to the system as it is a system-wide setting. Easy for users to mess with otherwise.

plugins/zenoh-plugin-storage-manager/src/backends_mgt.rs
```diff
@@ -30,9 +30,10 @@ pub(crate) async fn create_and_start_storage(
 ) -> ZResult<Sender<StorageMessage>> {
     trace!("Create storage {}", &admin_key);
     let key_expr = config.key_expr.clone();
-    let storage = backend.create_storage(config).await?;
+    let storage = backend.create_storage(config.clone()).await?;
```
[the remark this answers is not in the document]
sreeja: Fixed.

plugins/zenoh-plugin-storage-manager/src/replica/mod.rs
sreeja: Using `select` instead of join. Hence when the `storage_task` is killed through a signal, it stops the entire replica as well.

JEnoch (changes requested):

plugins/zenoh-plugin-storage-manager/src/memory_backend/mod.rs
```diff
@@ -138,6 +138,8 @@ impl Storage for MemoryStorage {
 
     async fn on_sample(&mut self, mut sample: Sample) -> ZResult<StorageInsertionResult> {
         trace!("on_sample for {}", sample.key_expr);
+        // NOTE: this will lead to inconsistencies, with different timestamps being generated for same sample
```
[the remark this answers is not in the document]
sreeja: Our current approach fixes this issue by paying for extra data messages.

Assume we have a sample with no timestamp, and two replicas receiving it. The replicas will assign a timestamp locally. When they reconcile with the alignment protocol, the one with the latest timestamp wins, i.e., we will have only one sample left in the system anyway. But there will be some data exchanges to reconcile them. 

Do you think it is better to drop the message with an error, or pay the price of this extra data exchange? I prefer the second option since there is no sample lost.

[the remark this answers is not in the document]
JEnoch: My concern is that with the more publication the storage gets without timestamp, the more traffic between storages is required for alignment. This might lead to system overload and instabilities that might be hard to analyse.

In my opinion it's anyway a bad design to not have publication timestamped at the source when those publications needs to be stored in replicated storages. And better have to forbid it.
So I would prefer we clearly state in doc that for storage replication usage it's required to activate timestamping in Publisher's Sessions. And that we log an error in replicas receiving a non-timestamped sample. When acknowledgement on PUT will be implemented, the replicas will return an error to the Publisher.

@Mallets do you agree ?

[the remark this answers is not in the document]
sreeja: After a discussion with @Mallets and @JEnoch , the conclusion is to force the user to timestamp the samples if they need replication. 
A follow up question would be whether we should force user to timestamp all entries that are meant to be stored. This would be required for our consolidation policies (in the future) to work correctly. 
For now, in this PR, I will enforce it only for replication.

PR #311, opened 2022-08-25
Add single line comment in zenoh-examples

sreeja:


examples/src/lib.rs
```diff
@@ -1 +1 @@
-
+//! Examples on using Zenoh. Check the Readme for usage.
```
JEnoch, line 1: For even more clarity I would reword as such:
```rust
//! Examples on using Zenoh. 
//! See the code in ../examples/ 
//! Check ../README.md for usage.
```

sreeja: I agree.

JEnoch (approved):

PR #312, opened 2022-08-29
Enable PushMode/PullMode switch for generic SubscriberBuilder

YuanYuYuan:
During porting the zenoh-perf to api-changes, I found that the usage of SubscriberBuilder is a little bit counter-intuitive. For example, we can't set the subscriber into `push_mode` since `declare_subscriber` returns a SubscriberBuilder in PushMode by default. And only changes from PushMode to PullMode and converse way are allowed. Maybe let's consider this PR to make the switch be idempotent.

```rust
 let mut subscriber = session
    .declare_subscriber(KEY_EXPR)
    .reliable()
    .push_mode()     // unsupported
    .res()
    .await.unwrap();
```

### DIFF
--- a/zenoh/src/subscriber.rs
+++ b/zenoh/src/subscriber.rs
@@ -444,9 +444,7 @@ impl<'a, 'b, Mode, Handler> SubscriberBuilder<'a, 'b, Mode, Handler> {
         self.local = true;
         self
     }
-}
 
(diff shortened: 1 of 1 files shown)


p-avital (approved):
Might as well, yes. It does offer the opportunity to turn an incoming builder that's still generic over Mode into one with a specific mode.

PR #314, opened 2022-09-06
Dalek the value_selector

p-avital:


commons/zenoh-protocol/src/proto/msg.rs
```diff
@@ -953,7 +953,7 @@ impl Header for Pull {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Query {
     pub key: WireExpr<'static>,
-    pub value_selector: String,
+    pub selector_parameters: String,
```
Mallets, line 956: Shouldn't be called simply `parameters`?

kydos: I agree with Luca, I think it should simply be parameters, with potential parameters being: predicate, projection, arguments, etc. etc.

Mallets (approved):

PR #321, opened 2022-09-14
Improve locality default and optionality

Mallets:


OlivierHecart:
This forces to take the lock on the session on each call to `allowed_destination` and `allowed_origin`. So I prefer the way it was before.

Mallets:
Can you please point out which specific lines are your referring to?

OlivierHecart:
Those lines : 
https://github.com/eclipse-zenoh/zenoh/blob/286629ee643eee298056fa01b2bbc4a318ed2f45/zenoh/src/session.rs#L542
https://github.com/eclipse-zenoh/zenoh/blob/286629ee643eee298056fa01b2bbc4a318ed2f45/zenoh/src/session.rs#L582
https://github.com/eclipse-zenoh/zenoh/blob/286629ee643eee298056fa01b2bbc4a318ed2f45/zenoh/src/session.rs#L616
https://github.com/eclipse-zenoh/zenoh/blob/286629ee643eee298056fa01b2bbc4a318ed2f45/zenoh/src/session.rs#L542
https://github.com/eclipse-zenoh/zenoh/blob/286629ee643eee298056fa01b2bbc4a318ed2f45/zenoh/src/session.rs#L777 https://github.com/eclipse-zenoh/zenoh/blob/286629ee643eee298056fa01b2bbc4a318ed2f45/zenoh/src/session.rs#L1507
https://github.com/eclipse-zenoh/zenoh/blob/286629ee643eee298056fa01b2bbc4a318ed2f45/zenoh/src/session.rs#L1549

Mallets:
Indeed, it would incur in an additional lock. That changes may be reverted then.

PR #330, opened 2022-09-16
Check for KeyExpr intersection of replies to queries

p-avital:
@OlivierHecart can you do the receiver side checking please? I trust you more with stopping a forbidden reply mid-network than myself :)


Mallets:
No longer relevant. Superseded by #334 .

PR #334, opened 2022-09-20
Implement reply filtering based on intersection with query, as well as escape hatch

p-avital:
As discussed here: https://github.com/eclipse-zenoh/roadmap/discussions/55

### DIFF
--- a/zenoh/src/query.rs
+++ b/zenoh/src/query.rs
@@ -140,6 +140,8 @@ pub struct GetBuilder<'a, 'b, Handler> {
     pub(crate) consolidation: QueryConsolidation,
     pub(crate) timeout: Duration,
     pub(crate) handler: Handler,
+    #[cfg(feature = "unstable")]
+    pub(crate) allow_disjoint_replies: bool,
 }
 
 impl<'a, 'b> GetBuilder<'a, 'b, DefaultHandler> {
@@ -172,6 +174,8 @@ impl<'a, 'b> GetBuilder<'a, 'b, DefaultHandler> {
             consolidation,
             timeout,
             handler: _,
+            #[cfg(feature = "unstable")]
+            allow_disjoint_replies,
         } = self;
         GetBuilder {
             session,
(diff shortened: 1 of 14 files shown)


zenoh/src/query.rs
```diff
@@ -279,6 +280,61 @@ impl<'a, 'b, Handler> GetBuilder<'a, 'b, Handler> {
         self.timeout = timeout;
         self
     }
+
+    /// By default, `get` will filter out
```
Mallets, line 284: Will filter out what?

zenoh/src/query.rs
```diff
@@ -279,6 +280,61 @@ impl<'a, 'b, Handler> GetBuilder<'a, 'b, Handler> {
...
+                        let s = s.parameters_mut();
+                        let mut start = index as usize;
+                        let pend = start + _REPLY_KEY_EXPR_ANY_SEL_PARAM.len();
+                        if start != 0 {
+                            start -= 1
+                        }
+                        match s[pend..].find('&') {
+                            Some(end) => std::mem::drop(s.drain(start..end + pend)),
+                            None => s.truncate(pend),
+                        }
```
Mallets, line 317: I believe it's worth adding a test to verify the correctness of the parameters truncation.

zenoh/src/queryable.rs
```diff
@@ -62,13 +64,35 @@ impl Query {
     }
 
     /// Sends a reply to this Query.
+    ///
+    /// By default, queries only accept replies which's key expression intersects with the query's.
```
Mallets, line 68: *whose key expression

zenoh/src/session.rs
```diff
@@ -1272,7 +1272,7 @@ impl Session {
 
     pub(crate) fn query(
         &self,
-        selector: &Selector<'_>,
+        selector: Selector<'_>,
```
Mallets, line 1275: Check if the `Selector` really needs to be passed by value or it can still be passed by reference.

Mallets (changes requested):

Mallets (approved):

PR #338, opened 2022-09-21
Unmatching queries

sreeja:


plugins/zenoh-plugin-storage-manager/src/replica/align_queryable.rs
```diff
@@ -92,21 +92,21 @@ impl AlignQueryable {
                     match value {
                         AlignData::Interval(i, c) => {
                             let sample = Sample::new(
-                                zenoh::key_expr::KeyExpr::from_str(&self.digest_key).unwrap(),
+                                zenoh::key_expr::KeyExpr::from_str(query.key_expr()).unwrap(),
```
Mallets, line 95: If you are getting the `key_expr` directly from the query, why reconstructing it?

Mallets (changes requested):

Mallets (approved):

PR #339, opened 2022-09-21
EXAMPLE_CONFIG.json5 -> DEFAULT_CONFIG.json5

OlivierHecart:


DEFAULT_CONFIG.json5
```diff
@@ -74,115 +74,115 @@
...
       subscribers: [
-        // key/expression
+        // key_expression
       ],
       /// A list of key-expressions for which all included publishers will be aggregated into.
       publishers: [
-        // key/expression
+        // key_expression
       ],
   },
```
Mallets, line 91: Shouldn't this also commented out by default?

p-avital: since the default is an empty array, which is the case here, I'm ok with it not being commented, but having some examples KEs to give a vibe. They do need to be quoted though.

PR #349, opened 2022-09-23
Implement IntoFuture for async API

Mallets:
As announced in the latest [Rust 1.64](https://blog.rust-lang.org/2022/09/22/Rust-1.64.0.html) release, the trait `IntoFuture` has been finally finalised. 

Since Zenoh API adopts a builder pattern, it would be highly beneficial for the async API to implement the `IntoFuture` trait for the sake of API ergonomics.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,35 +14,35 @@
 [workspace]
 
 members = [
-  "commons/zenoh-config",
   "commons/zenoh-buffers",
-  "commons/zenoh-protocol-core",
-  "commons/zenoh-protocol",
   "commons/zenoh-cfg-properties",
   "commons/zenoh-collections",
+  "commons/zenoh-config",
   "commons/zenoh-core",
   "commons/zenoh-crypto",
+  "commons/zenoh-macros",
+  "commons/zenoh-protocol",
+  "commons/zenoh-protocol-core",
   "commons/zenoh-sync",
   "commons/zenoh-util",
-  "commons/zenoh-macros",
+  "examples",
+  "io/zenoh-link",
   "io/zenoh-link-commons",
-  "io/zenoh-links/zenoh-link-udp/",
+  "io/zenoh-links/zenoh-link-quic/",
   "io/zenoh-links/zenoh-link-tcp/",
   "io/zenoh-links/zenoh-link-tls/",
-  "io/zenoh-links/zenoh-link-quic/",
+  "io/zenoh-links/zenoh-link-udp/",
   "io/zenoh-links/zenoh-link-unixsock_stream/",
   "io/zenoh-links/zenoh-link-ws/",
-  "io/zenoh-link",
   "io/zenoh-transport",
+  "plugins/example-plugin",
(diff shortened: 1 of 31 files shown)


Mallets:
This PR will be put on hold for the time being. 
The reason is the following:
- `IntoFuture` trait does not require the future to be `Send`
- In certain cases, especially in the API returning `impl Resolve<T>`, the compiler is not capable of verifying that the returned `Future` is Send`
- This lead to the need of defining a `IntoFutureSend` trait, which explicitly sets the `Send` boundary to the returned `Future`

As a result of that, the user might incur in such case:
```rust
fn foo() {
    let ret = myintofuture.await;
    do_other_stuff.await;
    return ret; // Not Send
}
```
where `ret` is not detected to be `Send`.
To force the `Future` to be `Send`, the user would need to convert the `Future` explicitly via the `IntoFutureSend` trait.
```rust
fn foo() {
    let ret = myintofuture.into_future_send().await;
    do_other_stuff.await;
    return ret; // Send
}
```
However, knowing when the `Future` needs to be explicitly created would be quite obscure for the user.

An example in zenoh code where that was necessary is here: https://github.com/eclipse-zenoh/zenoh/blob/bc34b84e8cde2c55d715ef9a02c4c9f06a0fedac/zenoh/src/session.rs#L777.
If `.into_future_send()` is not called on `Self::init`, this is the quite cryptic error message being returned:
```
error: future cannot be sent between threads safely
   --> zenoh/src/session.rs:770:9
    |
770 |         ResolveFuture(async move {
    |         ^^^^^^^^^^^^^ future created by async block is not `Send`
    |
    = help: within `impl futures::Future<Output = Result<session::Session, Box<dyn StdError + std::marker::Send + std::marker::Sync>>>`, the trait `std::marker::Send` is not implemented for `<impl Wait<session::Session> as std::future::IntoFuture>::IntoFuture`
note: future is not `Send` as this value is used across an await
   --> zenoh/src/session.rs:776:101
    |
776 |                       let session = Self::init(runtime, aggregated_subscribers, aggregated_publishers)
    |  _____________________________________________________________________________________________________-
777 | |                             // .into_future_send()
778 | |                             .await;
    | |                                  ^
    | |                                  |
    | |__________________________________await occurs here, with the value maybe used later
    |                                    has type `<impl Wait<session::Session> as std::future::IntoFuture>::IntoFuture` which is not `Send`
note: the value is later dropped here
   --> zenoh/src/session.rs:778:35
    |
778 | ...                   .await;
    |                             ^
note: required by a bound in `ResolveFuture`
   --> /Users/luca/Repository/eclipse-zenoh/commons/zenoh-core/src/lib.rs:106:30
    |
106 |     F: Future<Output = To> + Send;
    |                              ^^^^ required by this bound in `ResolveFuture`
```

As consequence of all of the above, we believe it's not safe yet to bring the `IntoFuture` implementation into zenoh API.

milyin:
The PR is untouched for about 10 months, closing it

PR #351, opened 2022-09-26
New config: "adminspace.permissions"

JEnoch:
This PR adds a new configuration:
```json5
{
  adminspace: {
    permissions: {
      read: true,
      write: false,
  }
}
```
The equivalent command line option for `zenohd` is `--adminspace-permissions=<[r|w|rw|none]>`.

By default the admin space is now read only, meaning that by default configuration changes at runtime is no longer possible (including loading of plugins and backends libraries).

### DIFF
--- a/commons/zenoh-config/src/lib.rs
+++ b/commons/zenoh-config/src/lib.rs
@@ -292,6 +292,12 @@ validated_struct::validator! {
                 },
             },
         },
+        /// Configuration of the admin space.
+        pub adminspace:
+        AdminSpaceConf {
+            /// Whether the admin space accepts config changes at runtime.
+            readonly: bool,
+        },
         /// A list of directories where plugins may be searched for if no `__path__` was specified for them.
         /// The executable's current directory will be added to the search paths.
         plugins_search_dirs: Vec<String>, // TODO (low-prio): Switch this String to a PathBuf? (applies to other paths in the config as well)
@@ -300,6 +306,12 @@ validated_struct::validator! {
         ///
         /// Please refer to [`PluginsConfig`]'s documentation for further details.
         plugins: PluginsConfig,
+    }
+}
+
+impl Default for AdminSpaceConf {
+    fn default() -> Self {
+        AdminSpaceConf { readonly:true }
     }
 }
 
--- a/commons/zenoh-config/src/lib.rs
+++ b/commons/zenoh-config/src/lib.rs
@@ -293,10 +293,11 @@ validated_struct::validator! {
             },
         },
         /// Configuration of the admin space.
-        pub adminspace:
+        pub adminspace: #[derive(Default)]
         AdminSpaceConf {
-            /// Whether the admin space accepts config changes at runtime.
-            readonly: bool,
+            /// Whether the admin space accepts config changes at runtime (false by default).
+            /// [unstable]: this configuration may change in the future.
+            changes: bool,
         },
         /// A list of directories where plugins may be searched for if no `__path__` was specified for them.
         /// The executable's current directory will be added to the search paths.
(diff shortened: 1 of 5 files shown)


commons/zenoh-config/src/lib.rs
```diff
@@ -292,6 +292,13 @@ validated_struct::validator! {
                 },
             },
         },
+        /// Configuration of the admin space.
+        pub adminspace: #[derive(Default)]
+        AdminSpaceConf {
+            /// Whether the admin space accepts config changes at runtime (false by default).
+            /// [unstable]: this configuration may change in the future.
+            changes: bool,
```
p-avital, line 300: `changes` is a bit of a weird name for this.

`writeable: bool` with the possibility to turn that into a structure that allows more precise filtering of the keys that are writeable sounds better to me

zenoh/src/net/runtime/adminspace.rs
```diff
@@ -307,6 +307,17 @@ impl Primitives for AdminSpace {
             data_info,
         );
 
+        {
+            let conf = self.context.runtime.config.lock();
+            if !*conf.adminspace.changes() {
+                error!(
```
p-avital, line 313: prefer `log::error` to `error` to avoid confusion with more aggressive macros and for easier greps

zenohd/src/main.rs
```diff
@@ -66,7 +66,8 @@ r#"Allows arbitrary configuration changes as column-separated KEY:VALUE pairs, w
   - VALUE must be a valid JSON5 string that can be deserialized to the expected type for the KEY field.
 Examples:
 --cfg='startup/subscribe:["demo/**"]'
---cfg='plugins/storage_manager/storages/demo:{key_expr:"demo/example/**",volume:"memory"}'"#)
+--cfg='plugins/storage_manager/storages/demo:{key_expr:"demo/example/**",volume:"memory"}'"#),
+clap::arg!(--"adminspace-changes" r"By default zenohd doesn't accept any runtime configuration changes via its admin space. This option allows such changes"),
```
p-avital, line 70: `--writeable-adminspace`?

JEnoch:
I think that's a minimum for the incoming release.

Or do we need an intermediate configuration level allowing only creation of Storages ?
But what about Volumes creation? If we allow this, it implies runtime loading of backends. Unless we introduce also a new configuration to specify at startup the backends to be loaded, without Volumes creation.
@Mallets, @OlivierHecart, @p-avital: opinions ?

In the future we will add Authentication + Access Control that will allow more fine-grained configuration of what's allowed or not, depending the credentials.

p-avital (changes requested):
Inner workings seem good (taking the lock on the config to check writeability allows a grace-period pattern where the adminspace is left writeable until it's locked by a config write).

The naming of the config field rings a bit weird to me, as `changes` sounds more like something that happens passively than the permission it is.

Alternative config change: `{permissions: {read: bool, write: bool}}`, where `read` could be introduced in a later release, allowing the adminspace to be completely locked down to avoid the risk of data extraction from it :)

Mallets:
I believe for the time being is enough to have the whole adminspace readonly by default. Volume creation are runtime it would be a nice feature to have but it would require a more mature support for access control.

commons/zenoh-config/src/lib.rs
JEnoch: Agreed for:
```json5
// [unstable]: this configuration may change in the future
adminspace: {
  permissions: {
    read: true,
    write: false,
  }
}
```

And for `zenohd` command line:
`--adminspace-permissions=[r|w|rw|none]` (default to `r`)

p-avital (approved):

Issue #316, opened 2022-09-08
[Bug] Undefined Behavior in `zenoh_config::Notifier`


jhdcs:
### Describe the bug

The function `mutable` found on https://github.com/eclipse-zenoh/zenoh/blob/45b7ad70bddf45dfcdcae80e9c9a7680c70ac289/commons/zenoh-config/src/lib.rs#L500 transmutes a reference to a mutable reference, which is defined in the [Rust language documentation](https://doc.rust-lang.org/reference/behavior-considered-undefined.html) as always being Undefined Behavior (since it breaks LLVM's pointer aliasing rules).

Is it possible to refactor the code to use an `UnsafeCell` instead?

### To reproduce

N/A - found via looking through code

### System info

Zenoh Commit: 45b7ad70bddf45dfcdcae80e9c9a7680c70ac289

milyin:
I suppose there is no need to mutable() at all. 'Notifier' is just an Arc wrapper and cloning of Arc is cheap operation. I think this solution should be enough: https://github.com/eclipse-zenoh/zenoh/pull/350

p-avital:
As explained in #353, the `UnsafeCell` is indeed there in the internals, and there aren't any reason to externally use the illegally granted mutability, which would have possibly caused UB.

`mutable` allowed to skip incrementing and decrementing the internal `Arc`'s strong counter, so as not to invalidate its cache line, but since it should be on the same cache line as the `Mutex`'s spin-prelock, that cache line will be invalidated anyway, and cause the exact same cache misses.

Even if it wasn't, that cost is extremely small.

Potential improvement: implement the traits that required mutable references on `&Notifier<T>` rather than on `Notifier<T>`. Rust's method resolution order should then allow calls without `mutable` or `clone` :)

PR #353, opened 2022-09-27
#316 mutable to clone

milyin:
Fix for https://github.com/eclipse-zenoh/zenoh/issues/316

### DIFF
--- a/commons/zenoh-config/src/lib.rs
+++ b/commons/zenoh-config/src/lib.rs
@@ -576,13 +576,6 @@ impl<T: ValidatedMap> Notifier<T> {
     pub fn lock(&self) -> MutexGuard<T> {
         zlock!(self.inner.inner)
     }
-    /// Since this type is fully interior-mutable, this method can be used to obtain a mutable reference for trait-compatibility with [`validated_struct::ValidatedMap`]
-    /// # Safety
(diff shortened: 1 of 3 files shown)


p-avital:
This doesn't pass tests because the cloned notifier is a temporary, which needs to be stored separately.

This is easily solved by assigning that temporary, but does make the API _slightly_ less comfortable.

While `transmute`ing an immutable reference to a mutable one can cause various UB, these are mostly related to optimisations that may make false assumptions:
- No two mutable references point to the same address, equivalent to `restrict`ed pointers in C++. I don't see someone using `config.mutable()` twice on the same config before using both obtained references in an operation that would behave poorly with restrict.
- The compiler may assume that anything that isn't inside an `UnsafeCell` hasn't been muted since it last looked at it through an immutable reference. However, since the data is behind a `Mutex` which does wrap it into an `UnsafeCell`, that assumption won't be used for the config data which the notifier wraps.

Cloning an `Arc` doesn't add cost here (the Mutex's spinner will invalidate the same cache-line as the counter increment), so it's really an API choice. Might as well be safe for such a small API-cost, rather than toy with unsafe and risk scaring users away. I'll approve once the doctests are fixed :)

p-avital (approved):

p-avital:
Closes #316

PR #361, opened 2022-10-05
Faster mutex macros

p-avital:
The `zlock`, `zread` and `zwrite` macros were written long ago. The stdlib has improved and now does the `try_lock` trick that these macros implement internally.

When benched alone, these macros yielded 10-20% worse performance than simply calling the `lock` method.

On my machine, this modification to the macros yields a 1-2% increase in throughput. This might be noise though, as my machine has a noisy CPU. Requesting @Mallets to try it out on c18 to judge whether this is worth it.

### DIFF
--- a/commons/zenoh-core/src/macros.rs
+++ b/commons/zenoh-core/src/macros.rs
@@ -18,10 +18,7 @@
 #[macro_export]
 macro_rules! zlock {
     ($var:expr) => {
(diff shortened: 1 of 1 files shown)


Mallets:
On c18 testing machine I can observe the same ~1% throughput improvement.

Mallets (approved):

PR #363, opened 2022-10-12
Tries multiple entries on DNS lookup when opening TCP or UDP link

Mallets:
When establishing a TCP or UDP link a DNS query might be performed in case the locator contains a DNS name.
Today only the first IP address returned by the DNS lookup is used to establish a TCP or UDP link. 

This PR handles multiple IP addresses in DNS queries:
- When establishing a TCP or UDP link the IP addresses are tried in order
- The first successful IP address is used
- TCP or UDP link establishment fails iff none of the returned IP addresses can be used

### DIFF
--- a/io/zenoh-links/zenoh-link-udp/src/unicast.rs
+++ b/io/zenoh-links/zenoh-link-udp/src/unicast.rs
@@ -23,8 +23,9 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex, RwLock, Weak};
 use std::time::Duration;
 use zenoh_collections::{RecyclingObject, RecyclingObjectPool};
-use zenoh_core::Result as ZResult;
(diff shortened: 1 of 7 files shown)


io/zenoh-links/zenoh-link-udp/src/unicast.rs
```diff
@@ -347,50 +337,132 @@ impl LinkManagerUnicastTrait for LinkManagerUnicastUdp {
...
-        Ok(locator)
+        bail!(
+            "Can not create a new UDP link bound to {}: {:?}",
+            endpoint,
+            errs
+        )
+    }
+
+    async fn new_listener(&self, mut endpoint: EndPoint) -> ZResult<Locator> {
+        let addrs = get_udp_addrs(&endpoint.locator).await?;
```
OlivierHecart, line 386: Shouldn't multicast addresses be filtered out here as well ?

Mallets: Fixed.

PR #370, opened 2022-10-27
add support for error numbers to ZError, so that bindings may expose them

p-avital:
by default, errors are numbered -128

### DIFF
--- a/commons/zenoh-protocol-core/src/key_expr/borrowed.rs
+++ b/commons/zenoh-protocol-core/src/key_expr/borrowed.rs
@@ -326,21 +326,21 @@ impl<'a> TryFrom<&'a str> for &'a keyexpr {
         let mut in_big_wild = false;
         for chunk in value.split('/') {
             if chunk.is_empty() {
-                bail!("Invalid Key Expr `{}`: empty chunks are forbidden, as well as leading and trailing slashes", value)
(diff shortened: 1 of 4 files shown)


commons/zenoh-protocol-core/src/key_expr/borrowed.rs
```diff
@@ -326,21 +326,21 @@ impl<'a> TryFrom<&'a str> for &'a keyexpr {
         let mut in_big_wild = false;
         for chunk in value.split('/') {
             if chunk.is_empty() {
-                bail!("Invalid Key Expr `{}`: empty chunks are forbidden, as well as leading and trailing slashes", value)
+                bail!((-4) "Invalid Key Expr `{}`: empty chunks are forbidden, as well as leading and trailing slashes", value)
```
Mallets, line 329: The various `errorno` should be defined in an enum to have them all grouped.

Mallets (changes requested):

cguimaraes:
It would be interesting to have a common definition of error codes for Zenoh across the different implementations/bindings.
In Zenoh-Pico, they are being defined here: https://github.com/eclipse-zenoh/zenoh-pico/blob/cf9033bf7349905778b1bc705b8bff13c94c5c5a/include/zenoh-pico/utils/result.h 
Note still, that the current numbering should be revised before exposing them through the user API.

PR #374, opened 2022-11-02
Make alignment compatible for the future

sreeja:
In the future, when we have support for storing wild card updates, multiple keys might have the same timestamp. In order to maintain the uniqueness of the hash function and detect any misaligned entries, we need to hash the key and timestamp together. This PR fixes this issue.

### DIFF
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -71,6 +71,14 @@
 
   /// The routing strategy to use and it's configuration.
   routing: {
+      /// The routing strategy to use in routers and it's configuration.
+      router: {
+          /// When set to true a router will forward data between two peers
+          /// directly connected to it if it detects that those peers are not
+          /// connected to each other.
+          /// The failover brokering only works if gossip discovery is enabled.
+          peers_failover_brokering: true,
+      },
       /// The routing strategy to use in peers and it's configuration.
       peer: {
           /// The routing strategy to use in peers. ("peer_to_peer" or "linkstate").
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 [![CI](https://github.com/eclipse-zenoh/zenoh/workflows/CI/badge.svg)](https://github.com/eclipse-zenoh/zenoh/actions?query=workflow%3A%22CI%22)
 [![Documentation Status](https://readthedocs.org/projects/zenoh-rust/badge/?version=latest)](https://zenoh-rust.readthedocs.io/en/latest/?badge=latest)
 [![Discussion](https://img.shields.io/badge/discussion-on%20github-blue)](https://github.com/eclipse-zenoh/roadmap/discussions)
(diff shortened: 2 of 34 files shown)


plugins/zenoh-plugin-storage-manager/src/replica/align_queryable.rs
```diff
@@ -212,46 +212,36 @@ impl AlignQueryable {
 
 // replying queries
 impl AlignQueryable {
-    async fn get_entry_with_ts(&self, timestamp: Timestamp) -> Option<Sample> {
+    async fn get_entry(&self, logentry: LogEntry) -> Option<Sample> {
```
JEnoch, line 215: Passing `logentry` as a ref (`&LogEntry`) is sufficient and will avoid to clone it at line 163

plugins/zenoh-plugin-storage-manager/src/replica/align_queryable.rs
```diff
@@ -212,46 +212,36 @@ impl AlignQueryable {
...
+        let replies = self.session.get(&logentry.key).res().await.unwrap();
+        if let Ok(reply) = replies.recv_async().await {
+            match reply.sample {
+                Ok(sample) => {
+                    trace!(
+                        "[ALIGN QUERYABLE] Received ('{}': '{}')",
+                        sample.key_expr.as_str(),
+                        sample.value
+                    );
+                    if sample.timestamp.is_some() {
```
JEnoch, line 226: Better to rewrite this as such:
```rust
if let Some(timestamp) = &sample.timestamp {
   match timestamp.cmp(&logentry.timestamp) {
     ...
```
Thus, no need to call `.unwrap()`

plugins/zenoh-plugin-storage-manager/src/replica/snapshotter.rs
```diff
@@ -268,11 +278,17 @@ impl Snapshotter {
             if ts > *last_snapshot_time {
                 remains_volatile.insert(k, ts);
             } else {
-                let deleted = stable.insert(k, ts);
+                let deleted = stable.insert(k.clone(), ts);
                 if deleted.is_some() {
```
JEnoch, line 282: Same comment than above: better to replace `if x.is_some() { ... x.unwrap() ... }` 
with `if let Some(y) = x { ... }`

plugins/zenoh-plugin-storage-manager/src/replica/snapshotter.rs
```diff
@@ -235,12 +242,15 @@ impl Snapshotter {
             drop(log);
         } else {
             let mut log = replica_data.stable_log.write().await;
-            let deleted = (*log).insert(key, ts);
+            let deleted = (*log).insert(key.clone(), ts);
             if deleted.is_some() {
```
JEnoch, line 246: Same comment than above: better to replace `if x.is_some() { ... x.unwrap() ... }`
with `if let Some(y) = x { ... }`

JEnoch (changes requested):

JEnoch (approved):

PR #375, opened 2022-11-07
Make multihop gossip configurable

OlivierHecart:


commons/zenoh-config/src/lib.rs
```diff
@@ -147,6 +147,9 @@ validated_struct::validator! {
             GossipConf {
                 /// Whether gossip scouting is enabled or not.
                 enabled: Option<bool>,
+                /// When true, gossip scouting informations are propagated multiple hops to all nodes in the local network.
+                /// When flase, gossip scouting informations are only propagated to the next hop.
```
Mallets, line 151: Typo: `When flase` -> `When false`

zenoh/src/net/runtime/mod.rs
```diff
@@ -110,6 +110,7 @@ impl Runtime {
             .unwrap_or(false);
 
         let gossip = config.scouting().gossip().enabled().unwrap_or(true);
+        let gossip_multihop = config.scouting().gossip().multihop().unwrap_or(false);
```
Mallets, line 113: A generic comment: `zenoh-config` crate has a [default.rs](https://github.com/eclipse-zenoh/zenoh/blob/master/commons/zenoh-config/src/defaults.rs) file containing all default values. 

Wouldn't be better to use the `Default` implementation for the config instead of manually setting the value to `false`?

commons/zenoh-config/src/lib.rs
p-avital: Don't forget to align `lib.rs` and `DEFAULT_CONFIG.json5`'s documentations

zenoh/src/net/runtime/mod.rs
p-avital: We've had this `Default` debate multiple times, using options helps distinguish whether the value is set by user or left at default behaviour. This may be useful when merging configuration, or when other configuration fields may change the default value for a given field.

p-avital (changes requested):
Apart from the documentation comments, I'm not certain why we'd want multihop gossip (nor do I have any argument against). Can you elaborate on it somewhere so the knowledge on it doesn't get lost? (Or link the somewhere if it already exists).

Otherwise, update the docstrings as Mallets asked and you get my green light :)

commons/zenoh-config/src/lib.rs
OlivierHecart: Done

DEFAULT_CONFIG.json5
```diff
@@ -51,6 +51,9 @@
     gossip: {
       /// Whether gossip scouting is enabled or not
       enabled: true,
+      /// When true, gossip scouting informations are propagated multiple hops to all nodes in the local network.
+      /// When flase, gossip scouting informations are only propagated to the next hop.
```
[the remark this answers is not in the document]
OlivierHecart: Done

zenoh/src/net/runtime/mod.rs
OlivierHecart: I think the point of @Mallets is not to remove the `Option` but to centralise the hard coded defaults into the default.rs file instead of spreader around in the code. A agree with @Mallets but let's deal with this in another PR.

zenoh/src/net/routing/network.rs
```diff
@@ -436,17 +440,19 @@ impl Network {
...
-                                idx,
-                                Details {
-                                    zid: true,
-                                    locators: self.gossip,
-                                    links: false,
-                                },
-                            )],
-                            |link| link.zid != zid,
-                        );
+                        if self.gossip_multihop || self.links.values().any(|link| link.zid == zid) {
```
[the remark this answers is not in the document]
OlivierHecart: Done

Mallets (approved):

PR #379, opened 2022-11-14
Mutual authentication

DariusIMP:
# Features

* Upgraded the async-rustls library, now using  rustls 0.20
* Adapted the zenoh-link-tls package to work with this new version of async-rustls
* Added two way authentication if configured that way.

# How to test the 2 way authentication
Let's create 2 certificates, one for the "server" and one for the "client". For this, it is explained in the zenoh documentation regarding TLS (https://zenoh.io/docs/manual/tls/) how to create certificates using MiniCA.

You can store the certificates as follows, under a certificates folder on your zenoh folder. It should look like this:

```
certificates
├── client
│   ├── localhost
│   │   ├── cert.pem
│   │   └── key.pem
│   ├── minica-key.pem
│   └── minica.pem
└── server
    ├── localhost
    │   ├── cert.pem
    │   └── key.pem
    ├── minica-key.pem
    └── minica.pem
```
From your zenoh folder, after having build this branch,

1- run the router with: `RUST_LOG=debug cargo run --bin=zenohd -- -c configs/router.json5` 

2- run a publisher with: `RUST_LOG=debug ./target/debug/examples/z_pub -c configs/client.json5`

3- run a subscriber with: `RUST_LOG=debug ./target/debug/examples/z_sub -c configs/client.json5`

where `configs/client.json5` is
```
{
  /// The node's mode (router, peer or client)
  mode: "client",
  connect: {
    endpoints: [ "tls/localhost:7447" ]
  },
  transport: {
    link: {
      tls: {
        root_ca_certificate: "./certificates/client/minica.pem",
        client_auth: true,
        client_private_key: "./certificates/server/localhost/key.pem",
        client_certificate: "./certificates/server/localhost/cert.pem",
      },
    },
  },
}
```

and `configs/router.json5` is:

```
{
  /// The node's mode (router, peer or client)
  mode: "router",
  listen: {
    endpoints: [ "tls/localhost:7447" ]
  },
  transport: {
    link: {
      tls: {
        client_auth: true,
        root_ca_certificate: "./certificates/server/minica.pem",
        server_private_key: "./certificates/client/localhost/key.pem",
        server_certificate: "./certificates/client/localhost/cert.pem",
      },
    },
  },
}
```

(see the configs are inside a folder `configs` under your `/zenoh` folder).

### DIFF
--- a/io/zenoh-transport/tests/unicast_transport.rs
+++ b/io/zenoh-transport/tests/unicast_transport.rs
@@ -735,7 +735,9 @@ fn transport_unicast_quic_only() {
 // in the expected error messages from the tests below.
 //
 // See: https://docs.rs/rustls/latest/src/rustls/msgs/enums.rs.html#128
+#[cfg(all(feature = "transport_tls", target_family = "unix"))]
 const RUSTLS_HANDSHAKE_FAILURE_ALERT_DESCRIPTION: &str = "HandshakeFailure";
+#[cfg(all(feature = "transport_tls", target_family = "unix"))]
 const RUSTLS_CERTIFICATE_REQUIRED_ALERT_DESCRIPTION: &str = "CertificateRequired";
 
 #[cfg(all(feature = "transport_tls", target_family = "unix"))]
@@ -976,6 +978,7 @@ fn transport_unicast_tls_two_way_auth_wrong_certs_fail() {
 // certificate brought in by the server. Similarly the server's certificate authority will validate
 // the key and certificate brought in by the client.
 //
+#[cfg(all(feature = "transport_tls", target_family = "unix"))]
 const CLIENT_KEY: &str = "-----BEGIN RSA PRIVATE KEY-----
 MIIEpAIBAAKCAQEArNthaKa7u/T7X5LTykqctYnmFmZcx9zUL1R7qXC/uWJWlKk6
 3xvQbUs2IDeIxL3yC6djJulbrqw+XvuclCM4nMFmUB8hcidesoTRb2agapompNOH
@@ -1004,6 +1007,7 @@ Fz66VkcYPMu05ggFIzrsJqPK6LUCb+h8sYnN4464+cJXhIrPvRJ3Pu30NoVBB0yz
 AbQCDGFMp3XCC+FLajMdQQhuXfUfSGjbidhQI87hmMF+gZ+WkNc8+A==
 -----END RSA PRIVATE KEY-----";
 
+#[cfg(all(feature = "transport_tls", target_family = "unix"))]
 const CLIENT_CERT: &str = "-----BEGIN CERTIFICATE-----
 MIIDLDCCAhSgAwIBAgIIdTp0cmbVlKswDQYJKoZIhvcNAQELBQAwIDEeMBwGA1UE
 AxMVbWluaWNhIHJvb3QgY2EgN2M1NWNjMB4XDTIyMTEwOTE1MzkwNVoXDTI0MTIw
@@ -1024,6 +1028,7 @@ FnK0IPRdnOl5v3j/62DDBllnJER5aahQcbNx9WszP2ZZb/SNzzQghVJ8yWBrbAJn
 SRCU86jw504Zx5q/SbuXJsPXJbiFF7eclvKEumdF3XmJeMRGPg2ysQ/nfco0nBz8
 -----END CERTIFICATE-----";
 
+#[cfg(all(feature = "transport_tls", target_family = "unix"))]
 const CLIENT_CA: &str = "-----BEGIN CERTIFICATE-----
 MIIDSzCCAjOgAwIBAgIIC3MWFI+HOvowDQYJKoZIhvcNAQELBQAwIDEeMBwGA1UE
 AxMVbWluaWNhIHJvb3QgY2EgMGI3MzE2MCAXDTIyMTAyNDEzMDIwNloYDzIxMjIx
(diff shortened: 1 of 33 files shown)


Mallets:
I reckon the benefit of using the latest version (i.e. pointing to git) of `async-rustls` for security reasons.

However, to be able accept the PR, we need a specific `async-rustls` dependency (i.e. published on [crates.io](https://crates.io/)) in the `Cargo.toml` to be able in turn to publish zenoh on [crates.io](https://crates.io/).
Unfortunately [crates.io](https://crates.io/) does not allow packages to be published with [git dependencies](
https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories).

I've opened an issue on `async-rustls` to ask for more information about their intended plan: https://github.com/smol-rs/async-rustls/issues/10

io/zenoh-transport/tests/unicast_transport.rs
```diff
@@ -839,5 +872,398 @@ tOzot3pwe+3SJtpk90xAQrABEO0Zh2unrC8i83ySfg==
...
+        &MSG_SIZE_ALL,
+    ));
+    assert!(result.is_err());
+    let err = result.unwrap_err();
+    assert!(err
+        .to_string()
+        .contains(RUSTLS_HANDSHAKE_FAILURE_ALERT_DESCRIPTION));
+}
+
+//*************************************/
```
Mallets, line 1095: Since you are adding this sections, shouldn't we remove the in-line certificates from the `tls` and `quic` tests and only refers to these ones?

io/zenoh-transport/tests/unicast_transport.rs
```diff
@@ -839,5 +872,398 @@ tOzot3pwe+3SJtpk90xAQrABEO0Zh2unrC8i83ySfg==
...
+    assert!(result.is_err());
+    let err = result.unwrap_err();
+    assert!(err
+        .to_string()
+        .contains(RUSTLS_CERTIFICATE_REQUIRED_ALERT_DESCRIPTION));
+}
+
+#[cfg(all(feature = "transport_tls", target_family = "unix"))]
+#[test]
+fn transport_unicast_tls_two_way_auth_wrong_certs_fail() {
```
Mallets, line 1023: Let's simply call the test `transport_unicast_tls_only_mutual_fail`?

io/zenoh-transport/tests/unicast_transport.rs
```diff
@@ -839,5 +872,398 @@ tOzot3pwe+3SJtpk90xAQrABEO0Zh2unrC8i83ySfg==
...
+        &MSG_SIZE_ALL,
+    ));
+    assert!(result.is_err());
+    let err = result.unwrap_err();
+    assert!(err
+        .to_string()
+        .contains(RUSTLS_CERTIFICATE_REQUIRED_ALERT_DESCRIPTION));
+}
+
+#[cfg(all(feature = "transport_tls", target_family = "unix"))]
```
Mallets, line 1021: I've just spotted this (also in the old `tls` test): should we try to remove the `target_family` to also target `windows` environments?

Mallets (changes requested):

io/zenoh-transport/tests/unicast_transport.rs
DariusIMP: What about `transport_unicast_tls_only_mutual_wrong_certs_failure`? The convention I use to follow for these tests is to have a first section of the name referring to a family of tests (`transport_unicast_tls_only_mutual`) and the last part of the name the particular test case we want to check with its desired result (`wrong_certs_failure`), I think the wrong_certs is important in the name.

io/zenoh-transport/tests/unicast_transport.rs
```diff
@@ -149,20 +149,21 @@ impl TransportPeerEventHandler for SCClient {
 }
 
 async fn open_transport(
-    endpoints: &[EndPoint],
-) -> (
+    client_endpoints: &[EndPoint],
+    server_endpoint: Option<&EndPoint>,
+) -> ZResult<(
```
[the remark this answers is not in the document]
DariusIMP: Yes, I wanted to be able to catch the error result and verify the message for the mTLS tests. I found a better way now I believe.

io/zenoh-transport/tests/unicast_transport.rs
Mallets: Ok, let's keep it then :)

io/zenoh-transport/tests/unicast_transport.rs
Mallets: Your proposed naming convention is fine to me. Thanks for the explanation.

Ricardicus:
Thank you for this fix!

Just for future reference, I would like to point out that this set up for testing 2 way authentication only works for the localhost case.

I used this on two containers with their own hostnames and needed another configuration. The fix is valid and works well for this use case as well. For future reference, I'll just specify another layout. 

Say we have the hostnames "server" and "client" as above. Then the minica workflow for generating the keys would be like this:

````BASH
# Getting minica
$ git clone https://github.com/jsha/minica.git
$ cd minica
$ go build
# create CA and server certificates
$ ./minica --domains server,client
$ ./minica --domains client,server
find . 
...
./minica-key.pem
./server
./server/cert.pem
./server/key.pem
./minica.pem
./client
./client/cert.pem
./client/key.pem
...
````

The configuration files

For the server:

```JSON5
{
  mode: "router",
  listen: {
    endpoints: [ "tls/server:7447" ]
  },
  transport: {
    link: {
      tls: {
        client_auth: true,
        root_ca_certificate: "./minica/minica.pem",
	server_private_key: "./minica/client/key.pem",
	server_certificate: "./minica/client/cert.pem",
      },
    },
  },
}
````

For the client:

```JSON5
{
  mode: "client",
  connect: {
    endpoints: [ "tls/server:7447" ]
  },
  transport: {
    link: {
      tls: {
        root_ca_certificate: "./minica/minica.pem",
        client_auth: true,
        client_private_key: "./server/key.pem",
        client_certificate: "./server/cert.pem",
      },
    },
  },
}
```

PR #384, opened 2022-11-24
Config defaults

OlivierHecart:
The first commit gathers config defaults as modules and constants in `zenoh_config::defaults`.

The second commit may be more controversial. It redefines `listen.endpoints` and `connect.endpoints` as optional `ModeDependentValues`. This is more accurate and allows users to explicitly set an empty list of listen endpoints and instruct Zenoh not to open any listening port. But while remaining backward compatible concerning the config file format, it slightly changes the way to access an update the conf programmatically. Thus strictly speaking it may be considered an API change. 
Note that `connect.endpoints` did not need to be retyped to a `ModeDependentValue`, but it makes things more homogeneous to have both `listen.endpoints` and `connect.endpoints` having the same type and be accessed the same way.

### DIFF
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -12,6 +12,7 @@
 
   /// Which endpoints to connect to. E.g. tcp/localhost:7447.
   /// By configuring the endpoints, it is possible to tell zenoh which router/peer to connect to at startup.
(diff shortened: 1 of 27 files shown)


fuzzypixelz:
Please change your pull request's base branch to main (new default branch). And rebase your branch against `main` as it is missing a status check necessary to merge this pull request but which is only available on `main`.

PR #385, opened 2022-11-25
Connectivity events

OlivierHecart:


Mallets (changes requested):
Before merging I suggest to add a test (e.g. in `zenoh/tests/events.rs`) to verify the feature.

Mallets (approved):

PR #387, opened 2022-12-01
Whatamimatcher fix

OlivierHecart:


commons/zenoh-protocol-core/src/whatami.rs
```diff
@@ -245,7 +245,7 @@ where
 impl BitOr<WhatAmI> for WhatAmIMatcher {
     type Output = Self;
     fn bitor(self, rhs: WhatAmI) -> Self::Output {
-        self | rhs as u8
+        self | (rhs as u8 + 128)
```
p-avital, line 248: unnecessary because `self` should already have its highest bit set

commons/zenoh-protocol-core/src/whatami.rs
```diff
@@ -259,12 +259,12 @@ impl BitOr for WhatAmIMatcher {
 impl BitOr for WhatAmI {
     type Output = WhatAmIMatcher;
     fn bitor(self, rhs: Self) -> Self::Output {
-        WhatAmIMatcher(unsafe { NonZeroU8::new_unchecked(self as u8 | rhs as u8) })
+        WhatAmIMatcher(unsafe { NonZeroU8::new_unchecked((self as u8 | rhs as u8) + 128) })
```
p-avital, line 262: I'd prefer `|` to `+`, as there's no risk of accidental carries ruining our day here

commons/zenoh-protocol-core/src/whatami.rs
```diff
@@ -259,12 +259,12 @@ impl BitOr for WhatAmIMatcher {
...
     }
 }
 
 impl From<WhatAmI> for WhatAmIMatcher {
     fn from(w: WhatAmI) -> Self {
-        WhatAmIMatcher(unsafe { NonZeroU8::new_unchecked(w as u8) })
+        WhatAmIMatcher(unsafe { NonZeroU8::new_unchecked(w as u8 + 128) })
```
p-avital, line 268: Same, use `|` instead of `+`

p-avital (changes requested):

commons/zenoh-protocol-core/src/whatami.rs
Mallets: To make things more explicit that we are actually setting the bit at position 7, the following could be used:
```rust
w as u8 | (1 << 7)
```

p-avital (approved):

PR #388, opened 2022-12-01
Complete default to false

Mallets:
This PR sets to false the default `complete` value for all queryables.


zenoh/src/sample.rs
```diff
@@ -33,6 +33,7 @@ pub enum Locality {
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub(crate) enum Locality {
     SessionLocal,
+    Remote,
```
sreeja, line 36: Is this an intended change?

Mallets: Yes, otherwise a library importing zenoh `feature = "unstable)` won't compile.

PR #400, opened 2022-12-13
Link establishment logging

OlivierHecart:
Here is what it does running z_pub/z_sub examples with different protocol versions: 

```bash
$ z_sub --no-multicast-scouting -l tcp/127.0.0.1:7447
Opening session...
Declaring Subscriber on 'demo/example/**'...
Enter 'q' to quit...
[2022-12-13T09:56:08Z ERROR zenoh_transport::unicast::establishment::accept] Rejecting InitSyn on tcp/127.0.0.1:7447 => tcp/127.0.0.1:63931 because of unsupported Zenoh version from peer: DA8E2B12900D40A0AF00DAE3D8022576 at io/zenoh-transport/src/unicast/establishment/accept/init_syn.rs:84.
```

```bash
$ z_pub --no-multicast-scouting -e tcp/127.0.0.1:7447
Opening session...
[2022-12-13T09:56:08Z ERROR zenoh_transport::unicast::establishment::open::init_ack] Received a close message (reason INVALID) in response to an InitSyn on: tcp/127.0.0.1:63931 => tcp/127.0.0.1:7447 at io/zenoh-transport/src/unicast/establishment/open/init_ack.rs:68.
Declaring Publisher on 'demo/example/zenoh-rs-pub'...
```

### DIFF
--- a/io/zenoh-transport/src/unicast/manager.rs
+++ b/io/zenoh-transport/src/unicast/manager.rs
@@ -565,19 +565,13 @@ impl TransportManager {
                 peer_id,
             };
 
-            let res = super::establishment::accept::accept_link(&link, &c_manager, &mut auth_link)
-                .timeout(c_manager.config.unicast.accept_timeout)
-                .await;
-            match res {
-                Ok(res) => {
-                    if let Err(e) = res {
-                        log::debug!("{}", e);
-                    }
-                }
-                Err(e) => {
-                    log::debug!("{}", e);
-                    let _ = link.close().await;
-                }
+            if let Err(e) =
+                super::establishment::accept::accept_link(&link, &c_manager, &mut auth_link)
+                    .timeout(c_manager.config.unicast.accept_timeout)
+                    .await
+            {
+                log::debug!("{}", e);
+                let _ = link.close().await;
             }
             let mut guard = zasynclock!(c_manager.state.unicast.incoming);
             *guard -= 1;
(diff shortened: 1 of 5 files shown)


io/zenoh-transport/src/unicast/manager.rs
```diff
@@ -565,19 +565,13 @@ impl TransportManager {
...
-                Ok(res) => {
-                    if let Err(e) = res {
-                        log::debug!("{}", e);
-                    }
-                }
-                Err(e) => {
-                    log::debug!("{}", e);
-                    let _ = link.close().await;
-                }
+            if let Err(e) =
```
Mallets, line 568: This check shouldn't be changed. `timeout()` returns a `Result`, if `Ok` it contains the `Result` of `accept_link`. In practice it's a `Result<Result<_, _>, _>`. So, the proposed change only catches the `Error` generated by the `timeout`.

Mallets (changes requested):

io/zenoh-transport/src/unicast/manager.rs
OlivierHecart: Yes but nothing was done except logging in case `Ok(Err(_))`. So as we don't want to log this case any more (logging occurs in `accept_link`) there is no point checking this case any more.

Mallets (approved):

Issue #410, opened 2022-12-21
[Bug] Router panic answering admin queries when gossip scouting disabled


OlivierHecart:
### Describe the bug

When a Zenoh router with 
  - `peer_to_peer` mode enabled (default)
  - `gossip` scouting disabled

answers admin queries matching `@/router/<router_id>/linkstate/peers`, it logs a panic message:
```
thread 'async-std/runtime' panicked at 'called `Option::unwrap()` on a `None` value', /Users/olivier/workspaces/eclipse-zenoh/zenoh/zenoh/src/net/runtime/adminspace.rs:614:10
```
The query times out and may miss some replies. 

### To reproduce

1. Start `zenohd` with `gossip` scouting disabled:
```bash
$ zenohd --cfg=scouting/gossip/enabled:false
```
2. Start `z_get` example on `@/**`:
```bash
$ z_get -s @/**
```

### System info

Zenoh v0.7.0-rc

PR #411, opened 2022-12-21
Fix router panic in adminspace (#410)

OlivierHecart:


zenoh/src/net/runtime/adminspace.rs
```diff
@@ -577,46 +578,29 @@ pub async fn router_data(
     )
 }
 
-pub async fn linkstate_routers_data(
+pub async fn linkstate_data(
```
p-avital, line 581: this predates the PR, but this function is marked async despite awaiting nothing (and blocking on a mutex, so it can only be run by the executor, but will park one of its threads :( )

OlivierHecart: Good point.
This function is used as an admin space `Handler` which is an async function called in an async context anyway (task). This could be changed but is not part of this PR anyway.

PR #412, opened 2022-12-27
Refactoring replica module

sreeja:
This PR fixes a bug in the alignment protocol that keeps on identifying misalignment: 
- When a replica starts with an empty log, it creates a digest with 0 subinterval, which changes the digest checksums even though it has no impact semantically. When a replica starts with some data, the digest will not have this 0 subinterval, leading the replica to detect misalignment. 

This PR also refactors the replica module, includes more tests, and improves logging.

### DIFF
--- a/plugins/zenoh-plugin-storage-manager/Cargo.toml
+++ b/plugins/zenoh-plugin-storage-manager/Cargo.toml
@@ -37,6 +37,7 @@ name = "zplugin_storage_manager"
 crate-type = ["cdylib", "rlib"]
 
 [dependencies]
+async-global-executor = "2.3.1"
 async-std = "=1.12.0"
 async-trait = "0.1.59"
 clap = "3.2.23"
--- a/plugins/zenoh-plugin-storage-manager/Cargo.toml
+++ b/plugins/zenoh-plugin-storage-manager/Cargo.toml
@@ -37,7 +37,6 @@ name = "zplugin_storage_manager"
 crate-type = ["cdylib", "rlib"]
 
 [dependencies]
-async-global-executor = "2.3.1"
 async-std = "=1.12.0"
 async-trait = "0.1.59"
 clap = "3.2.23"
@@ -59,6 +58,9 @@ zenoh-core = { version = "0.7.0-rc", path = "../../commons/zenoh-core/" }
 zenoh-plugin-trait = { version = "0.7.0-rc", path = "../zenoh-plugin-trait", default-features = false }
 zenoh-util = { version = "0.7.0-rc", path = "../../commons/zenoh-util" }
 zenoh_backend_traits = { version = "0.7.0-rc", path = "../zenoh-backend-traits/" }
+
+[dev-dependencies]
+async-global-executor = "2.3.1"
 
 [build-dependencies]
 rustc_version = "0.4.0"
--- a/plugins/zenoh-plugin-storage-manager/src/replica/aligner.rs
+++ b/plugins/zenoh-plugin-storage-manager/src/replica/aligner.rs
@@ -16,7 +16,7 @@ use super::{Digest, EraType, LogEntry, Snapshotter};
 use super::{CONTENTS, ERA, INTERVALS, SUBINTERVALS};
 use async_std::sync::{Arc, RwLock};
 use flume::{Receiver, Sender};
-use log::{error, trace};
+use log::{error, debug, trace};
 use std::collections::{HashMap, HashSet};
 use std::str;
 use zenoh::key_expr::{KeyExpr, OwnedKeyExpr};
@@ -69,7 +69,7 @@ impl Aligner {
                 continue;
             } else {
                 // process this digest
-                trace!(
+                debug!(
                     "[ALIGNER]Processing digest: {:?} from {}",
                     incoming_digest,
                     from
@@ -103,7 +103,7 @@ impl Aligner {
 
             for (key, (ts, value)) in missing_data {
                 let sample = Sample::new(key, value).with_timestamp(ts);
-                trace!("[ALIGNER] Adding sample {:?} to storage", sample);
+                debug!("[ALIGNER] Adding sample {:?} to storage", sample);
                 match self.tx_sample.send_async(sample).await {
                     Ok(()) => continue,
                     Err(e) => error!("[ALIGNER] Error adding sample to storage: {}", e),
@@ -317,7 +317,7 @@ impl Aligner {
             .join(&from)
             .unwrap()
             .with_parameters(&properties);
-        trace!("[ALIGNER]Sending Query '{}'...", selector);
+        trace!("[ALIGNER] Sending Query '{}'...", selector);
         let mut return_val = Vec::new();
         let replies = self
             .session
--- a/plugins/zenoh-plugin-storage-manager/src/replica/aligner.rs
+++ b/plugins/zenoh-plugin-storage-manager/src/replica/aligner.rs
@@ -56,10 +56,16 @@ impl Aligner {
     pub async fn start(&self) {
         while let Ok((from, incoming_digest)) = self.rx_digest.recv_async().await {
             if self.in_processed(incoming_digest.checksum).await {
-                trace!("[ALIGNER]Skipping already processed digest: {}", incoming_digest.checksum);
+                trace!(
+                    "[ALIGNER]Skipping already processed digest: {}",
+                    incoming_digest.checksum
+                );
                 continue;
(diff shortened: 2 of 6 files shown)


plugins/zenoh-plugin-storage-manager/src/replica/aligner.rs
```diff
@@ -134,267 +140,208 @@ impl Aligner {
...
             }
             // get subintervals diff
-            return this.get_full_content_diff(other_content);
+            let result = this.get_full_content_diff(other_content);
+            return (result, no_err);
+        }
+        (Vec::new(), true)
+    }
+
+    async fn perform_query(&self, from: String, properties: String) -> (Vec<Sample>, bool) {
```
JEnoch, line 313: `from` should be a `&str` since `perform_query()` is always called passing a '<&str>.to_string()`. This will avoid String allocation.

Better: could it be a `&keyexpr` or a `&OwnedKeyExpr` ?
If yes, the `selector` below can be computed without validity/canonical check as such:
```rust
let selector = (KeyExpr::from(&self.digest_key) / from).with_parameters(&properties);
```

plugins/zenoh-plugin-storage-manager/src/replica/digest.rs
```diff
@@ -290,50 +293,59 @@ impl Digest {
...
-            let checksum = Digest::get_subinterval_checksum(
-                &subintervals
-                    .get(&sub)
-                    .unwrap()
-                    .content
-                    .clone()
-                    .into_iter()
-                    .collect::<Vec<LogEntry>>(),
-            );
+            let content = &subintervals.get(&sub).unwrap().content;
```
JEnoch, line 296: Are you sure this `unwrap()` cannot panic ?
Also: you could directly call `get_mut()` rather than `get()`. It would avoid the 2nd `get_mut()` below.

plugins/zenoh-plugin-storage-manager/src/replica/digest.rs
```diff
@@ -290,50 +293,59 @@ impl Digest {
...
-                &intervals
-                    .get(&int)
-                    .unwrap()
-                    .content
-                    .clone()
-                    .into_iter()
-                    .collect::<Vec<u64>>(),
-                &subintervals,
-            );
+            let content = &intervals.get(&int).unwrap().content;
```
JEnoch, line 310: Same comment than above.

plugins/zenoh-plugin-storage-manager/src/replica/digest.rs
```diff
@@ -290,50 +293,59 @@ impl Digest {
...
-                &eras
-                    .get(&era)
-                    .unwrap()
-                    .content
-                    .iter()
-                    .copied()
-                    .collect::<Vec<u64>>(),
-                &intervals,
-            );
+            let content = &eras.get(&era).unwrap().content;
```
JEnoch, line 331: Same comment than above.

plugins/zenoh-plugin-storage-manager/src/replica/digest.rs
```diff
@@ -418,11 +436,11 @@ impl Digest {
                 .subintervals
                 .entry(subinterval)
                 .and_modify(|e| {
-                    e.content.insert(ts.clone());
+                    e.content.insert(log_entry.clone());
                 })
                 .or_insert(SubInterval {
                     checksum: 0,
-                    content: [ts].iter().cloned().collect(),
+                    content: [log_entry].iter().cloned().collect(),
```
JEnoch, line 443: Can be replaced by `content: [log_entry].into()`
(and the same with `Interval`s' content in following lines)

plugins/zenoh-plugin-storage-manager/src/replica/digest.rs
```diff
@@ -683,21 +701,18 @@ impl Digest {
...
-            if other.contains_key(&era) && other.get(&era).unwrap().checksum != 0 {
-                if self.eras.contains_key(&era) {
-                    if self.eras.get(&era).unwrap().checksum != other.get(&era).unwrap().checksum {
-                        result.insert(era);
-                    }
-                } else {
-                    result.insert(era);
+    // check of the other era has more content
+    pub fn era_has_diff(&self, era: &EraType, other: &HashMap<EraType, Interval>) -> bool {
+        if other.contains_key(era) {
```
JEnoch, line 706: This code could be more clear as such:
```rust
match (other.get(era), self.eras.get(era)) {
    (Some(other_interval), Some(my_interval)) => other_interval != my_interval,
    (Some(_), None) => true,
    _ => false
}
```


BTW, if `self` contains the `era` and `other` doesn't, isn't it a diff ?

JEnoch (changes requested):

plugins/zenoh-plugin-storage-manager/Cargo.toml
```diff
@@ -37,6 +37,7 @@ name = "zplugin_storage_manager"
 crate-type = ["cdylib", "rlib"]
 
 [dependencies]
+async-global-executor = "2.3.1"
```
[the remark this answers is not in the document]
sreeja: This dependency is for tests. Specifically, the following code inside tests requires this dependency:
```
async_std::task::block_on(async {
    zenoh_core::zasync_executor_init!();
});
```

plugins/zenoh-plugin-storage-manager/src/replica/digest.rs
sreeja: For the `unwrap()`, yes, they cannot panic. This is because `subintervals_to_update` will only contain the keys that are already present in `subintervals`.
I will use `get_mut()` instead of `get()`.

plugins/zenoh-plugin-storage-manager/Cargo.toml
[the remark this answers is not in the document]
Mallets: If it's only used in tests, then I suggest to move it to the dev block:
```
[dev-dependencies]
async-global-executor = "2.3.1"
```

plugins/zenoh-plugin-storage-manager/src/replica/digest.rs
sreeja: > if `self` contains the era and `other` doesn't, isn't it a diff ?

It is a diff, but as far as this replica is concerned (who has the `self`), it is not missing anything, it is upto the replica that sent the `other` digest to process it.

plugins/zenoh-plugin-storage-manager/src/replica/aligner.rs
sreeja: `from` indicates the replica that this replica needs to talk to in order to fetch the missing updates. We can convert the type from `&str` into a `&keyexpr` or a `&OwnedKeyExpr`, but I thought it might not be intuitive.  I agree though that it would reduce canonical check inside the function that is frequently used. 
Should I proceed with making it a `&OwnedKeyExpr`?

plugins/zenoh-plugin-storage-manager/src/replica/aligner.rs
```diff
@@ -116,285 +121,226 @@ impl Aligner {
...
+                        error!(
+                            "[ALIGNER] Received error for query on selector {} :{}",
+                            selector, err
+                        );
+                        no_err = false;
+                    }
+                }
+            }
+        } else {
+            error!("[ALIGNER] Query failed on selector `{}`", selector);
```
JEnoch, line 342: Shouldn't `no_err` be set to `false` if the query fails ?
Also, it would help debugging to log the cause of failure. Use a `match` to log the error message.

plugins/zenoh-plugin-storage-manager/src/replica/aligner.rs
JEnoch: Let's do it in a separate PR after this one, so we better track/explain it.

plugins/zenoh-plugin-storage-manager/src/replica/aligner.rs
sreeja: > Shouldn't no_err be set to false if the query fails ?
Thank you for catching this. It is in fact an error.

> Also, it would help debugging to log the cause of failure. Use a match to log the error message.
Agreed

plugins/zenoh-plugin-storage-manager/src/replica/aligner.rs
sreeja: Agreed

JEnoch (approved):

PR #419, opened 2023-01-09
`no_std` support

davidedellagiustina:
Main changes:
- `zenoh-collections` and `zenoh-buffers` tested and working on STM32F7 board (minor fixes);
- in `zenoh-collections`/`zenoh-buffers` added `defmt` feature (opt-in) useful for debugging in `no_std`;
- [WIP] `zenoh-protocol` should be also ok in `no_std`, apart from the few dependencies that it still has in `zenoh-core`. `zenoh-codec` depends on `zenoh-protocol`;
- [WIP] `uhlc` dependency now relies on the GitHub version instead of the `crates.io` one, in order to have `no_std` support;
- ~[WIP] `defmt` feature for `zenoh-protocol`/`zenoh-codec` is almost done, but needs a couple of additions dependent on `defmt` feature in `uhlc` ([PR waiting](https://github.com/atolab/uhlc-rs/pull/6))~ `defmt::Format` is implemented also for structures in `zenoh-protocol`/`zenoh-codec`.

### DIFF
--- a/commons/zenoh-buffers/Cargo.toml
+++ b/commons/zenoh-buffers/Cargo.toml
@@ -31,8 +31,6 @@ description = "Internal crate for zenoh."
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 [features]
-default = ["std"]
-std = ["zenoh-collections/std"]
 test = ["rand"]
 defmt = ["dep:defmt", "zenoh-collections/defmt"]
 
--- a/commons/zenoh-buffers/Cargo.toml
+++ b/commons/zenoh-buffers/Cargo.toml
@@ -31,8 +31,12 @@ description = "Internal crate for zenoh."
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 [features]
-test = [ "rand" ]
+default = ["std"]
+std = ["zenoh-collections/std"]
+test = ["rand"]
+defmt = ["dep:defmt", "zenoh-collections/defmt"]
 
 [dependencies]
+defmt = { version = "0.3.2", features = ["alloc"], optional = true }
 rand = { version = "0.8.5", optional = true }
 zenoh-collections = { version = "0.7.0-rc", path = "../zenoh-collections/", default-features = false }
--- a/commons/zenoh-buffers/src/bbuf.rs
+++ b/commons/zenoh-buffers/src/bbuf.rs
@@ -11,8 +11,6 @@
 // Contributors:
 //   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
 //
-extern crate alloc;
-
 use crate::{
     reader::HasReader,
     vec,
@@ -22,6 +20,7 @@ use alloc::boxed::Box;
 use core::num::NonZeroUsize;
 
 #[derive(Clone, Debug, PartialEq, Eq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub struct BBuf {
     buffer: Box<[u8]>,
     len: usize,
@@ -131,6 +130,7 @@ impl<'a> HasReader for &'a BBuf {
 #[cfg(feature = "test")]
 impl BBuf {
     pub fn rand(len: usize) -> Self {
+        #[allow(unused_imports)]
         use alloc::vec::Vec;
         use rand::Rng;
 
--- a/commons/zenoh-buffers/src/bbuf.rs
+++ b/commons/zenoh-buffers/src/bbuf.rs
@@ -130,7 +130,6 @@ impl<'a> HasReader for &'a BBuf {
 #[cfg(feature = "test")]
 impl BBuf {
     pub fn rand(len: usize) -> Self {
-        #[allow(unused_imports)]
         use alloc::vec::Vec;
         use rand::Rng;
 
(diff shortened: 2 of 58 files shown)


commons/zenoh-buffers/Cargo.toml
```diff
@@ -31,8 +31,12 @@ description = "Internal crate for zenoh."
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 [features]
-test = [ "rand" ]
+default = ["std"]
+std = ["zenoh-collections/std"]
```
Mallets, line 35: zenoh-buffers shouldn't depend on `std` from `zenoh-collections`.

commons/zenoh-buffers/src/bbuf.rs
```diff
@@ -131,6 +130,7 @@ impl<'a> HasReader for &'a BBuf {
 #[cfg(feature = "test")]
 impl BBuf {
     pub fn rand(len: usize) -> Self {
+        #[allow(unused_imports)]
```
Mallets, line 133: Double check if this is really needed.

commons/zenoh-buffers/src/lib.rs
```diff
@@ -13,7 +13,7 @@
 //
 
 //! Provide differnt buffer implementations used for serialization and deserialization.
-#![no_std]
+#![cfg_attr(not(feature = "std"), no_std)]
```
Mallets, line 16: zenoh-buffers should be always `no_std`?

commons/zenoh-codec/Cargo.toml
```diff
@@ -29,20 +29,34 @@ description = "Internal crate for zenoh."
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
-shared-memory = ["zenoh-shm", "zenoh-protocol/shared-memory"]
+default = ["std"]
+std = [
+    "uhlc/std",
+    "zenoh-buffers/std",
```
Mallets, line 35: zenoh-codec shouldn't depend on `std` from `zenoh-buffers`.

commons/zenoh-codec/src/core/zbuf.rs
```diff
@@ -103,7 +103,7 @@ where
                     reader.read_zslices(len, |s| zbuf.push_zslice(s))?;
                 }
                 super::zslice::kind::SHM_INFO => {
-                    let bytes: Vec<u8> = self.codec.read(&mut *reader)?;
+                    let bytes: alloc::vec::Vec<u8> = self.codec.read(&mut *reader)?;
```
Mallets, line 106: I would use `use alloc::vec::Vec` at the beginning of the file.

commons/zenoh-protocol/src/core/endpoint.rs
```diff
@@ -310,6 +345,13 @@ impl fmt::Display for MetadataMut<'_> {
     }
 }
 
+#[cfg(feature = "defmt")]
```
Mallets, line 348: Can we use `derive(defmt::Format)`?

Mallets (changes requested):

commons/zenoh-codec/src/core/zbuf.rs
davidedellagiustina: This piece of code is compiled only when `#[cfg(feature = "shared-memory")]`, hence putting `use alloc::vec::Vec` at the beginning would result in a `unused import` warning whenever the `shared-memory` feature is not enabled.
Should I move the import to the top anyways, and add a feature gate to it?

commons/zenoh-buffers/src/bbuf.rs
davidedellagiustina: It wasn't anymore. Removed in last commit.

commons/zenoh-protocol/src/core/endpoint.rs
davidedellagiustina: I think `derive` would derive an implementation based on the fields of a struct, instead of "copying" the behaviour of `Display` (which is what we want here). One solution to avoid duplicating code would be using [this](https://docs.rs/defmt/latest/defmt/struct.Display2Format.html), but there will still be the problem of quite a lot boilerplate code. I'll continue investigating if there is a better solution.

davidedellagiustina:
With the last commit, `defmt::Format` is now implemented for all the types that also implement at least one among `fmt::Display` and `fmt::Debug` (relies on latest changes merged into `uhlc`).

commons/zenoh-codec/src/core/zbuf.rs
davidedellagiustina: Rust seems to complain:
```bash
warning: the item `Vec` is imported redundantly
   --> commons/zenoh-codec/src/core/zbuf.rs:96:13
    |
96  |         use alloc::vec::Vec;
    |             ^^^^^^^^^^^^^^^
    |
   ::: /home/vagrant/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/prelude/mod.rs:119:13
    |
119 |     pub use super::v1::*;
    |             ------------ the item `Vec` is already defined here
    |
    = note: `#[warn(unused_imports)]` on by default
```

commons/zenoh-protocol/src/core/endpoint.rs
davidedellagiustina: The solution I mentioned previously may no be suitable either. From `defmt`'s docs:

immagine

Mallets (approved):

PR #424, opened 2023-01-12
`no_std`: `zenoh-protocol` and `zenoh-codec`

davidedellagiustina:
**In this PR:**
- fixed problem with `default-features` in workspace `Cargo.toml`;
- `zenoh-protocol` and `zenoh-codec` are made competely `no_std`:
  - `zenoh-result` was split from `zenoh-core` (`zenoh-result` supports `no_std`, while `zenoh-core` depends on stuff in `std`);
  - own `IError` in place of `std::core::Error` (until `core::error::Error` is made stable, see its [tracking issue](https://github.com/rust-lang/rust/issues/103765)) $-$ the `derive_tryfrom` macro was returning `uhlc::SizeError`s, which are now converted into a new struct `zenoh_protocol::SizeError` in order to implement `zenoh_result::IError`;
  - `AtomicZInt` (`= AtomicU64`) is not portable to `no_std` target, hence it has been moved out of `zenoh_protocol` to the only place in which it is used (namely `zenoh::session`).

### DIFF
--- a/commons/zenoh-core/src/lib.rs
+++ b/commons/zenoh-core/src/lib.rs
@@ -22,8 +22,8 @@ pub mod macros;
 pub use macros::*;
 use std::future::{Future, Ready};
 pub use zenoh_macros::*;
-pub use zenoh_result::*;
 pub use zenoh_result::ZResult as Result;
+pub use zenoh_result::*;
 
 pub trait Resolvable {
     type To: Sized + Send;
--- a/commons/zenoh-core/src/lib.rs
+++ b/commons/zenoh-core/src/lib.rs
@@ -22,6 +22,7 @@ pub mod macros;
 pub use macros::*;
 use std::future::{Future, Ready};
 pub use zenoh_macros::*;
+pub use zenoh_result::*;
 
 pub trait Resolvable {
     type To: Sized + Send;
--- a/commons/zenoh-core/src/lib.rs
+++ b/commons/zenoh-core/src/lib.rs
@@ -22,9 +22,6 @@ pub mod macros;
 pub use macros::*;
 use std::future::{Future, Ready};
 pub use zenoh_macros::*;
-pub mod zresult;
-pub use zresult::Error;
-pub use zresult::ZResult as Result;
 
 pub trait Resolvable {
     type To: Sized + Send;
--- a/commons/zenoh-core/src/lib.rs
+++ b/commons/zenoh-core/src/lib.rs
@@ -23,6 +23,7 @@ pub use macros::*;
 use std::future::{Future, Ready};
 pub use zenoh_macros::*;
 pub use zenoh_result::*;
+pub use zenoh_result::ZResult as Result;
 
 pub trait Resolvable {
     type To: Sized + Send;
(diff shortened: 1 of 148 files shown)


commons/zenoh-core/src/lib.rs
```diff
@@ -22,9 +22,6 @@ pub mod macros;
 pub use macros::*;
 use std::future::{Future, Ready};
 pub use zenoh_macros::*;
-pub mod zresult;
```
p-avital, line 25: A reexport of zenoh-result might be helpful here, to avoid a breaking change. It could also have saved this PR a lot of lines that change the imports. @Mallets ?

commons/zenoh-core/src/macros.rs
```diff
@@ -115,14 +115,14 @@ macro_rules! zasyncrecv {
...
-        if !$op {
-            return false;
-        }
-    };
-}
+// #[macro_export]
+// macro_rules! zcheck {
+//     ($op:expr) => {
+//         if !$op {
+//             return false;
```
p-avital: commented code might as well get deleted

p-avital (changes requested):
LGTM, I'd like the reexport in core to be there to avoid a breaking change for users that may have used ZResult/zenoh_core::Error in their signatures :)

commons/zenoh-core/src/macros.rs
davidedellagiustina: I forgot about that piece of code, but I re-found it literally 5 minutes ago and removed in the last commit already 😃

commons/zenoh-core/src/lib.rs
davidedellagiustina: Muy only "complain" is that this would require an explicit dependency from `zenoh_core` to `zenoh_result`. Anyway, I'm adding it so everybody is happy :)

p-avital (approved):

OlivierHecart (approved):

p-avital:
Rerunning the jobs now that the uhlc PR is merged

davidedellagiustina:
@p-avital @JEnoch should be releasing a new version of uhlc to crates.io right now, I'm waiting for that to update our `Cargo.toml` with the updated source.

davidedellagiustina:
Done + fixed re-export in `zenoh-core`: `ZResult` was previously exported with the alias `Result`, so I re-added that in order to avoid breaking API changes, as discussed.

davidedellagiustina:
**Question:** should we add a CI job that checks the _`no_std`-ness_ of the crates `zenoh-collections`/`zenoh-result`/`zenoh-buffers`/`zenoh-protocol`/`zenoh-codec`? Perhaps a [`cargo nono check --no-default-features`](https://github.com/hobofan/cargo-nono) would suffice (maybe with [`cargo no-dev-deps`](https://lib.rs/crates/cargo-no-dev-deps) on `zenoh-protocol`/`zenoh-codec`)?

davidedellagiustina:
As discussed in the meeting we had this morning, I added a CI job that performs some checks for the stuff that should be `no_std`. Considered what `cargo-nono` does, I think there is no need to run the checks on different target operating systems.
Note: this is my first time using GitHub actions/workflows, so I hope I did nothing wrong 🤞🏻

Mallets (approved):

Mallets:
LGTM

@davidedellagiustina before merging I would you ask you to double check that all other sister repos in eclipse Zenoh are compatible with these changes.

PR #427, opened 2023-01-23
Fix Ping/Pong examples

bachp:
The code was expecting a config file argument but it wasn't declared in clap.

Signed-off-by: Pascal Bach <pascal.bach@nextrem.ch>


Mallets:
Thanks @bachp for the PR and the fix! It is now merged.

PR #430, opened 2023-02-02
Liveliness

OlivierHecart:
There are still details discussed there https://github.com/eclipse-zenoh/roadmap/discussions/72 . But Angelo wants this to be merged ASAP so that the bridge can leverage it ASAP. 
Details can be changed after if needed but I believe the base is there and won't change much.


zenoh/src/net/routing/queries.rs
```diff
@@ -1461,6 +1461,45 @@ fn compute_final_route(
...
 
+#[inline]
+fn compute_local_replies(
+    tables: &Tables,
+    prefix: &Arc<Resource>,
+    suffix: &str,
+    face: &Arc<FaceState>,
+) -> Vec<(WireExpr<'static>, ZBuf)> {
+    let mut result = vec![];
+    if face.whatami == WhatAmI::Client {
```
Mallets, line 1472: It's not clear to me why this computation would apply only to clients.

zenoh/src/session.rs
```diff
@@ -102,6 +106,8 @@ impl SessionState {
             publications: Vec::new(),
             subscribers: HashMap::new(),
             queryables: HashMap::new(),
+            #[cfg(feature = "unstable")]
```
Mallets, line 109: Here we are changing the layout of the session struct. Did you verify the impact on the bindings and plugins? E.g. zenoh-c and zenoh-python?

zenoh/src/session.rs
```diff
@@ -948,7 +987,15 @@ impl Session {
             callback,
         });
 
-        let declared_sub = (origin != Locality::SessionLocal)
+        #[cfg(not(feature = "unstable"))]
+        let declared_sub = origin != Locality::SessionLocal;
```
Mallets, line 991: Locality should be an unstable feature. So both cases should be under the "unstable" feature gate.

zenoh/src/session.rs
```diff
@@ -1044,7 +1091,15 @@ impl Session {
                 res.subscribers.retain(|sub| sub.id != sub_state.id);
             }
 
-            if sub_state.origin != Locality::SessionLocal {
+            #[cfg(not(feature = "unstable"))]
```
Mallets, line 1094: Locality should be an unstable feature. So both cases should be under the "unstable" feature gate.

Mallets (changes requested):

examples/Cargo.toml
```diff
@@ -114,6 +114,10 @@ path = "examples/z_get.rs"
 name = "z_forward"
 path = "examples/z_forward.rs"
 
+[[example]]
+name = "z_liveliness"
+path = "examples/z_liveliness.rs"
```
Mallets, line 119: You should add `required-features = ["unstable"]`.

zenoh/src/net/routing/queries.rs
OlivierHecart: Added comment

examples/Cargo.toml
OlivierHecart: Done

zenoh/src/session.rs
OlivierHecart: Locality is public when unstable feature is enabled and private otherwise.

zenoh/src/session.rs
OlivierHecart: Locality is public when unstable feature is enabled and private otherwise.

zenoh/src/session.rs
p-avital: This doesn't actually change the Session's layout, since `SessionState` is Arc'd in it. So no impact on zenoh-c and zenoh-python apart from new features to expose :)

OlivierHecart:
Merged with #441

PR #434, opened 2023-02-14
introduce custom cowstr to guarantee its layout

p-avital:


commons/zenoh-protocol/src/core/cowstr.rs
```diff
@@ -0,0 +1,105 @@
+use core::fmt::{Debug, Display, Formatter};
```
Mallets, line 1: Should the `cowstr` part of `zenoh-collections` instead of `zenoh-protocol`?

commons/zenoh-protocol/src/core/cowstr.rs
```diff
@@ -0,0 +1,105 @@
...
+use core::num::NonZeroUsize;
+
+enum CowStrInner<'a> {
+    Borrowed(&'a str),
+    Owned { s: Box<str>, capacity: NonZeroUsize },
```
Mallets, line 6: If I got correctly, an empty string can be represented by a `Borrowed("")`, right?

commons/zenoh-protocol/src/core/encoding.rs
```diff
@@ -135,20 +139,18 @@ impl Encoding {
         if suffix.as_ref().is_empty() {
             Some(Encoding::Exact(prefix))
         } else {
-            Some(Encoding::WithSuffix(prefix, suffix.into()))
+            Some(Encoding::WithSuffix(prefix, suffix.into().into()))
```
Mallets, line 142: The double `into()` without annotation is a bit weird to me. What are the conversions steps?

commons/zenoh-protocol/src/core/encoding.rs
```diff
@@ -135,20 +139,18 @@ impl Encoding {
...
     where
-        IntoCowStr: Into<Cow<'static, str>>,
+        IntoCowStr: Into<Cow<'static, str>> + AsRef<str>,
     {
         match self {
-            Encoding::Exact(e) => Encoding::WithSuffix(e, suffix.into()),
-            Encoding::WithSuffix(e, s) => {
-                Encoding::WithSuffix(e, Cow::Owned(format!("{}{}", s, suffix.into())))
-            }
+            Encoding::Exact(e) => Encoding::WithSuffix(e, suffix.into().into()),
```
Mallets, line 152: Double `into()` also here.

commons/zenoh-protocol/src/core/cowstr.rs
p-avital: Correct

commons/zenoh-protocol/src/core/encoding.rs
p-avital: Cow<str> is used as intermediate to maintain the previous API contract and legibility

commons/zenoh-protocol/src/core/cowstr.rs
p-avital: Meh, I also hesitated, not quite a collection either, we can move it if you want

Mallets: You are right, I'm also hesitating on that. But I see it more generic than just the protocol...

commons/zenoh-protocol/src/core/encoding.rs
Mallets: Then I just wonder how the compiler manages to correctly infer that... I wasn't expecting this to work :)

p-avital: It's thanks to the function's bounds: the only thing it knows about `suffix` here is that it can be `into()`ed into a `Cow<str>`, so that's the only thing the first `into()` can do. The second into knows the slot expects a `CowStr`, and boom, magic :D

Mallets: That's right, I have overlooked the function bounds for a moment. It LGTM.

PR #435, opened 2023-02-16
impl Send+Sync for KeyExprTreeNode

p-avital:


commons/zenoh-util/src/keyexpr_tree/box_tree.rs
```diff
@@ -303,6 +303,15 @@ pub struct KeyExprTreeNode<Weight, Wildness: IWildness, Children: IChildrenProvi
...
 }
 
+unsafe impl<Weight: Send, Wildness: IWildness + Send, Children: IChildrenProvider<Box<Self>> + Send>
+    Send for KeyExprTreeNode<Weight, Wildness, Children>
+{
+}
+unsafe impl<Weight: Sync, Wildness: IWildness + Sync, Children: IChildrenProvider<Box<Self>> + Sync>
+    Sync for KeyExprTreeNode<Weight, Wildness, Children>
+{
+}
```
Mallets, line 313: Something I've seen quite a lot in the core Rust is to always ask for `Send + Sync` boundaries when implementing unsafe `Send` and `Sync`. I wonder if it is the case also here.
E.g.: https://marabos.nl/atomics/building-arc.html

p-avital: This isn't an Arc, but more akin to a Box, which only requires Send for Send and Sync for Sync, so we should be good :)

Mallets (approved):

PR #437, opened 2023-02-20
Introducing keformat!

p-avital:
To help users with processing Key Expressions, here come KeFormats.

`examples/examples/z_formats` shows how zenoh-rs users are expected to use the `keformat` macro:
- `keformat!($spec: lit, $module: ident)` will declare `mod $module` which contain all of the formatting utilities wrapped in helpers that provide accessors based on the `$spec`:
  - `Format` is a ZST representing this exact format.
  - `formatter() -> Formatter` constructs a formatter with setters for each field of the spec.
  - `parser<'s>(&'s keyexpr) -> ZResult<Parsed<'s>>` parses the keyexpr into an annotated parsing result, with getters for each field of the spec.
- `keformat!($formatter: expr, $($field: ident [ = $value: expr])*)` works on formatters constructed by the previously mentioned `formatter()` function. It sets each `$field` in the formatter, returning a `ZResult<&mut Formatter>` that will error if `$value` doesn't fit `$field`'s spec. If no `$value` is provided, the macro will assume a `$field` variable exists and use it as `$value`. The macro will detect attempt to set fields that don't exist.

`KeFormat`s can also be constructed at runtime, and the formatters and parse-results may be used respectively like read-only and write-only maps. APIs for other languages may be built on to of these.

TODOs:
- [x] More documentation
- [ ] C and Python APIs


examples/examples/z_formats.rs
```diff
@@ -0,0 +1,16 @@
+use zenoh::{keformat, prelude::keyexpr};
+
+keformat!("user_id/${user_id:*}/file/${file:**}", format);
+
+fn main() {
+    let mut formatter = format::formatter();
+    let file = "hi/there";
+    let ke = keformat!(formatter, user_id = 42, file)
+        .unwrap()
```
gabrik, line 9: Maybe I'm dumb, but why this can fail twice?
I imagine that the first one is when the `formatter` used does not match the parameters (or vice-versa).
But why the `build()` can fail?

gabrik (approved):

examples/examples/z_formats.rs
p-avital: This can fail if some of the fields haven't received values yet, but expected non-empty values

p-avital:
This PR introduces a very weird bug with rust-analyzer: https://github.com/rust-lang/rust-analyzer/issues/14190

I propose we refrain on merging it to avoid users having a subpar experience with the macros and dismissing the feature until that issue is fixed.

Can someone on the team with a Mac let me know if they're also affected (proc_macro build fails in RA only, I've confirmed it on Linux and Windows).

Mallets:
I've just cloned the repo and on my Mac (Intel i5 - Ventura 13.1) I don't see any problem.
- `rustc 1.67.1`
- `rust-analyzer 0.3.1418`

examples/examples/z_formats.rs
```diff
@@ -0,0 +1,16 @@
+use zenoh::{keformat, prelude::keyexpr};
```
Mallets, line 1: Header missing.

examples/examples/z_formats.rs
```diff
@@ -0,0 +1,16 @@
...
+
+fn main() {
+    let mut formatter: format::Formatter = format::formatter();
+    let file = "hi/there";
+    let ke = keformat!(formatter, user_id = 42, file)
+        .unwrap()
+        .build()
+        .unwrap();
+    println!("{formatter:?} => {ke}");
+    let parsed = format::parse(&ke).unwrap();
```
Mallets, line 13: Why there is the need to parse the `ke` if it has just been built at line 8?

Mallets: Can I use it to retrieve a `KeyExpr` e.g. from a `Sample` in a subscriber callback, parse it, and access the field? 
I believe so, but I think it would be better to be more explicit in the example.

p-avital: I changed the example to parse an arbitrary KE with a different format.

I avoided making subscribers/queryables to avoid distracting from the formatting aspects.

Mallets (approved):

Mallets:
Approving this pull request. 
As next step it would be nice to create a `zenoh-keyexpr` crate where all type definitions, macros, utils (e.g. `ketree`), etc. are grouped. I fill that too many things are pushed into `zenoh-protocol` now.  The issue #445 has been created to keep track of this.

PR #439, opened 2023-02-22
Fix CI in `no_std`

davidedellagiustina:
The "easy" checks I previously introduced in the CI to allow compiling in `no_std` have already proven their limits (`zenoh-protocol` is already broken).
In this PR:
- fixed `zenoh-protocol`;
- improved CI checks with separate crate (that also defines a dummy `no_std` `main()`).


commons/zenoh-codec/src/bin/no_std_ci_check.rs
```diff
@@ -0,0 +1,28 @@
...
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+//
+// Contributors:
+//   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
+//
+
+#![no_std]
+
+#[panic_handler]
+fn dummy_panic_handler(_: &core::panic::PanicInfo) -> ! {
```
p-avital, line 18: This doesn't pass `cargo check --all-features` because std defines the panic handler. Add `#[cfg(not(feature="std"))]` to fix it :)

p-avital (changes requested):

ci/nostd-check/rust-toolchain.toml
```diff
@@ -0,0 +1,2 @@
+[toolchain]
+channel = "nightly"
```
Mallets, line 2: Why `nightly` is needed here?

p-avital: By reading through, for the `#[global_allocator]` use in the test, but I'm also confused as to why override it

davidedellagiustina: I couldn't find any simple allocator that works in Rust stable. The alternative is define our own dummy one that does completely nothing (it just needs to compile, it doesn't need to actually run). By the way, as I did it the nightly toolchain is applied only to the `nostd-check` crate.

davidedellagiustina: @p-avital Just saw your message. The target `x86_64-unknown-none` does not have a default allocator, hence Rust complains if none is defined (since Zenoh uses the `alloc` crate)

PR #444, opened 2023-03-13
Resolves #443

### LINKED ISSUE #443: Allow whitelisting of supported protocols in config
### Describe the feature

It would nice to offer to the users the possibility of whitelisting which protocols should be used in a Zenoh session.
For example, in the config:

```
  /// Configure internal transport parameters
  transport: {
    // An optional whitelist of protocols to be used for accepting and opening sessions.
    protocols: ["tcp", "udp"]
  }
```

Mallets:


OlivierHecart (commented):
Starting a router or a peer with no defined listeners and tcp transport black listed leads to the following error: 
`Unable to open listener tcp/[::]:0: Unauthorized protocol: tcp. Whitelisted protocols are: ...`
This is a bad user experience. I believe Zenoh should not try to open a tcp listener if not explicitly configured and tcp is disabled.

PR #452, opened 2023-03-27
Fix #451

### LINKED ISSUE #451: [Bug] Default endpoint listener not working after #444
### Describe the bug

#444 introduced a bug when no endpoints are defined via config:
- default locator is not added to the list of candidates endpoints to listen on when the protocol is supported
- as a result, two peers are not able to scout each other because none of those is listening 

### To reproduce

1. Start a peer subscriber
2. Start a peer publisher
3. The peers will not connect 

### System info

Any

Mallets:


yellowhatter:
I checked, this fixes the issue

PR #456, opened 2023-03-28
Tables locking

OlivierHecart:


Mallets (commented):
The proposed changes allow to reduce the time a write lock is held on the tables upon reception of new declarations. However, the proposed changes seem to introduce a performance regression on the machine used to mesure Zenoh's performance (see https://zenoh.io/blog/2023-03-21-zenoh-vs-mqtt-kafka-dds/).

From `4.35M msg/s` -> `4.05M msg/s` with a data payload of 8 bytes (i.e., `300K msg/s` went missing). I believe this should be further investigated before merging.

Mallets:
I've been investigating and rerunning the tests, it seems a fresh compilation provides better results. I still experience a small drop about `20K msg/s` but I believe they are negligible compared against the benefits brought by the new locking strategy.

Mallets (approved):

PR #457, opened 2023-03-29
Make `zenoh-util` (`ketree`) `no_std`

davidedellagiustina:
`zenoh-util` crate already has a `std` feature that can be disabled, but dependencies were a mess and the standard library was still required even after disabling the feature. This PR addresses the issue.


davidedellagiustina:
Better addressed by #458, closing.

PR #458, opened 2023-03-29
move keyexpr concerns to its own, no_std, crate

p-avital:


commons/zenoh-keyexpr/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+description = "Internal crate for zenoh."
+
+[features]
+default = ["std"]
+std = ["zenoh-result/std", "rand?/std", "rand?/std_rng"]
+test = ["rand"]
+
+[dependencies]
+keyed-set = { workspace = true }
+rand = { workspace = true, features = ["alloc", "getrandom"], optional = true }
```
Mallets, line 33: Should `rand` be part of `[dependencies]` as well or only of `[dev-dependencies]`?

commons/zenoh-keyexpr/src/key_expr/tests.rs
```diff
@@ -151,7 +151,7 @@ fn inclusions() {
 #[test]
 fn fuzz() {
     const FUZZ_ROUNDS: usize = 100_000;
-    let rng = rand::thread_rng();
+    let rng = rand::rngs::OsRng::default();
```
Mallets, line 154: I'm curious about this change. Why replacing `thread_rng()` with `OsRng::default()`?

davidedellagiustina: `thread_rng()` is available only in `std`. I think what we preciously did in similar situations is add a `test` feature that enables `std` in cascade, and have `rand` methods be enabled only when `test` is enabled.

commons/zenoh-keyexpr/Cargo.toml
p-avital: It's a dependency if we want the KeFuzzer to be available to other crates

commons/zenoh-keyexpr/src/key_expr/tests.rs
p-avital: like davide explained, except I rolled it back and missed this instance because it got lost in all the file moves ^^"

commons/zenoh-keyexpr/Cargo.toml
Mallets: Right, it optionally included it the `test` features is enabled for `zenoh-keyexpr`. Fine with that.

commons/zenoh-keyexpr/src/key_expr/tests.rs
Mallets: Not a big deal for me. Up to you wether to roll it back to `thread_rng`.

Mallets (approved):

Issue #461, opened 2023-03-31
Zenoh-Ext Group management should provide an operation to get the eventual leader


kydos:
### Describe the feature

zenoh_ext provides a group management abstraction but at the present stage does not provide a way of agreeing on a leader. As we have an eventual group view, we could use this to elect an eventual leader. This would be a useful extension for the group management users that require to differentiate one of the members as the leader.

PR #462, opened 2023-03-31
Added eventual leader on group management (see #461)

kydos:
Added eventual leader to the zenoh-ext group management.


kydos:
The email update on eclipse has been taken into account yet. I'll re-submit the PR using the previous email which should be OK w.r.t. ECA.

PR #466, opened 2023-04-04
Compression

DariusIMP:
Compression PR.

This PR contains the logic to compress batches hop to hop for particular use cases.

In order to enable this feature, it's necessary to build Zenoh with the `transport_compression` feature.


commons/zenoh-compression/src/lib.rs
```diff
@@ -0,0 +1,89 @@
+//
```
Mallets, line 1: I believe this codec should go directly under io/zenoh-transport/src/common

io/zenoh-transport/src/unicast/link.rs
```diff
@@ -177,13 +193,33 @@ async fn tx_task(
...
         match pipeline.pull().timeout(keep_alive).await {
             Ok(res) => match res {
                 Some((batch, priority)) => {
                     // Send the buffer on the link
-                    let bytes = batch.as_bytes();
+                    #[allow(unused_mut)]
+                    let mut bytes = batch.as_bytes();
+
+                    #[cfg(feature = "transport_compression")]
+                    let compression: Vec<u8>;
```
Mallets, line 210: I believe this is a non-valid Rust statement since the vector is non-initialized.

io/zenoh-transport/src/unicast/link.rs
```diff
@@ -391,3 +470,63 @@ async fn rx_task(
...
+        if link.is_streamed() {
+            let compression_size_u16: u16 = compression_size.try_into().map_err(|e| {
+                zerror!(
+                    "Compression error: unable to convert compression size into u16: {}",
+                    e
+                )
+            })?;
+            // including is compressed byte
+            let batch_size_le = (compression_size_u16 + 1).to_le_bytes();
+            compressed_batch.append(&mut batch_size_le.to_vec());
```
Mallets, line 497: This allocates everytime on the heap. I suggest to use slice copy instead of `push` and `append`.

io/zenoh-transport/src/unicast/link.rs
```diff
@@ -391,3 +470,63 @@ async fn rx_task(
...
+        }
+        buff = compressed_batch;
+    } else {
+        // Add compression byte as false
+        // Increment the batch size number from the 2 initial bytes
+        buff = bytes.to_vec();
+        buff.insert(2, false as u8);
+        let (batch_size, batch_payload) = buff.split_at_mut(2);
+        let batch_size = u16::from_le_bytes(batch_size.try_into().unwrap()) + 1;
+        buff = [batch_size.to_le_bytes().to_vec(), batch_payload.to_vec()].concat();
```
Mallets, line 512: This allocates everytime on the heap. I suggest to use slice copy instead of to_vec() and concat().

io/zenoh-transport/src/unicast/link.rs
```diff
@@ -391,3 +470,63 @@ async fn rx_task(
...
+            compressed_batch.append(&mut buff);
+        } else {
+            compressed_batch.push(true as u8);
+            compressed_batch.append(&mut buff);
+        }
+        buff = compressed_batch;
+    } else {
+        // Add compression byte as false
+        // Increment the batch size number from the 2 initial bytes
+        buff = bytes.to_vec();
```
Mallets, line 508: This allocates on the heap. It should be avoided.

io/zenoh-transport/src/unicast/link.rs
```diff
@@ -391,3 +470,63 @@ async fn rx_task(
...
+        let (batch_size, batch_payload) = buff.split_at_mut(2);
+        let batch_size = u16::from_le_bytes(batch_size.try_into().unwrap()) + 1;
+        buff = [batch_size.to_le_bytes().to_vec(), batch_payload.to_vec()].concat();
+    }
+    Ok(buff)
+}
+
+#[cfg(feature = "transport_compression")]
+fn rx_decompress(
+    zenoh_compress: &ZenohCompress,
```
Mallets, line 519: ZenohCompress is a ZST, there is no need to pass it as argument.

io/zenoh-transport/src/unicast/link.rs
```diff
@@ -391,3 +470,63 @@ async fn rx_task(
...
+    compression: &[u8],
+    aux_buffer: &mut [u8],
+) -> ZResult<usize> {
+    Ok(zenoh_compress
+        .read((&compression, aux_buffer))
+        .map_err(|e| {
+            zerror!(
+                "Decompression error {:?}. Unable to decompress batch. {:?}",
+                e,
+                compression.to_vec()
```
Mallets, line 529: There is no need to call to_vec() here. Moreover, if you want to print out the batch, I'd rather use log::trace! rather then putting in as string in the error message.

io/zenoh-transport/src/unicast/link.rs
```diff
@@ -391,3 +470,63 @@ async fn rx_task(
...
+    compression_enabled: bool,
+    bytes: &[u8],
+    compression_aux_buff: &mut Box<[u8]>,
+    link: &LinkUnicast,
+) -> ZResult<Vec<u8>> {
+    let mut buff = vec![];
+    if compression_enabled {
+        let zenoh_compress = ZenohCompress::default();
+        let compression_size = zenoh_compress
+            .write(&mut buff.writer(), (&bytes, compression_aux_buff))
```
Mallets, line 485: It's not clear to me why the compression_aux_buff is passed as argument and not its writer?

io/zenoh-transport/src/unicast/link.rs
```diff
@@ -391,3 +470,63 @@ async fn rx_task(
...
+            .map_err(|e| zerror!("Compression error: {:?}", e))?;
+        let mut compressed_batch: Vec<u8> = vec![];
```
Mallets, line 487: Again, here is another allocation, all of this should be handled by the compression_aux_buff.

io/zenoh-transport/src/unicast/link.rs
```diff
@@ -391,3 +470,63 @@ async fn rx_task(
...
+            let compression_size_u16: u16 = compression_size.try_into().map_err(|e| {
+                zerror!(
+                    "Compression error: unable to convert compression size into u16: {}",
+                    e
+                )
+            })?;
+            // including is compressed byte
+            let batch_size_le = (compression_size_u16 + 1).to_le_bytes();
+            compressed_batch.append(&mut batch_size_le.to_vec());
+            compressed_batch.push(true as u8);
```
Mallets, line 498: Even though `true` is garantueed to be `1` in Rust, I suggest to write `1_u8` to be simmetric with the read part.

io/zenoh-transport/src/unicast/link.rs
```diff
@@ -82,6 +92,10 @@ impl TransportLinkUnicast {
                 queue_size: self.transport.config.manager.config.queue_size,
                 backoff: self.transport.config.manager.config.queue_backoff,
             };
+
+            #[cfg(feature = "transport_compression")]
+            let compression_enabled = self.transport.config.manager.config.compression_enabled;
```
Mallets, line 97: This could be removed and the configuration parameter passed directly at line `113`.

DariusIMP: The problem when doing that is that the function from the executor would take ownership of "self" (because of the async move), causing trouble later when self is required.

io/zenoh-transport/src/unicast/link.rs
Mallets: Good to know, it's the first time I see it indeed :)

DEFAULT_CONFIG.json5
```diff
@@ -201,6 +201,17 @@
         /// Path to the TLS client public certificate
         client_certificate: null,
       },
+
+      /// **Experimental** compression feature.
+      /// Will compress the batches hop to hop (as opposed to end to end). May cause errors when
+      /// the batches's complexity is too high, causing the resulting compression to be bigger in
```
Mallets, line 207: This is no longer true, isn't?

DEFAULT_CONFIG.json5
```diff
@@ -201,6 +201,17 @@
...
+      /// size than the MTU. The feature "transport_compression" needs to be enabled to run this.
```
Mallets, line 208: The features to be enabled are "unstable" and "transport_compression"

commons/zenoh-config/src/lib.rs
```diff
@@ -283,6 +283,18 @@ validated_struct::validator! {
                     client_private_key: Option<String>,
                     client_certificate: Option<String>,
                 },
+                pub compression: #[derive(Default)]
+                /// **Experimental** compression feature.
+                /// Will compress the batches hop to hop (as opposed to end to end). May cause errors when
+                /// the batches's complexity is too high, causing the resulting compression to be bigger in
+                /// size than the MTU.
+                /// You must use zenoh's feature "transport_compression" to enable this.
+                Compression {
```
Mallets, line 292: For consistency with the rest of the config, I'd call the datatype `CompressionConf`.

DEFAULT_CONFIG.json5
DariusIMP: No, yes it is, it's an edge case but if the complexity of the batch is too high it may happen the end result is bigger than the MTU

DariusIMP: Ok, I had missed the point, yea if the compression result is bigger than the original batch size, then we can send the batch non compressed. However in that case we increment the batch size by 1 byte, and although that's an edge case it may happen that we exceed the maximum size for a batch by 1 byte.

io/zenoh-transport/src/unicast/link.rs
```diff
@@ -33,6 +37,27 @@ use zenoh_protocol::transport::TransportMessage;
...
+const COMPRESSION_ENABLED: u8 = 1_u8;
+
+#[cfg(all(feature = "unstable", feature = "transport_compression"))]
+const COMPRESSION_DISABLED: u8 = 0_u8;
+
+#[cfg(all(feature = "unstable", feature = "transport_compression"))]
+const BATCH_PAYLOAD_START_INDEX: usize = 1;
+
+#[cfg(all(feature = "unstable", feature = "transport_compression"))]
+const MAX_BATCH_SIZE: usize = u16::MAX as usize;
```
DariusIMP, line 59: I'm commiting a mistake here by setting this constant to u16::max? Since I believe the mtu is something to be set in the configuration and that is also negotiated between edges right?

examples/examples/z_pub_thr.rs
```diff
@@ -20,12 +20,19 @@ use zenoh::publication::CongestionControl;
 fn main() {
     // initiate logging
     env_logger::init();
-    let (config, size, prio, print, number) = parse_args();
+    let (config, size, prio, print, number, high_entropy) = parse_args();
```
Mallets, line 23: The `z_pub_thr` example shouldn't be modified. Please revert this change.

io/zenoh-transport/src/unicast/link.rs
Mallets: The maximum batch size is `u16::MAX`. I.e., the maximum amount of useful bytes to be sent excluding the header.

Mallets (changes requested):
Update the docstring before merging.

Mallets (approved):

PR #468, opened 2023-04-05
fix #422

### LINKED ISSUE #422: canonize function doens't pass tests [Bug]
1. Canonization of expression `hello/foo$*$*/bar` makes `hello/foo$*/barar` - the delimiter is not set after shrinking `$*$*` secuence. This is fixed - see pull request below
2. Accordingly to the document  https://github.com/eclipse-zenoh/roadmap/blob/main/rfcs/ALL/Key%20Expressions.md the expression `hello/$*/bye` should be canonized to `hello/*/bye`. This is not the case. The test (which currently fails) for this case is also in the pull request below
3. There is a bunch of canonization tests in zenoh-pico - https://github.com/eclipse-zenoh/zenoh-pico/blob/master/tests/z_keyexpr_test.c. Probably we need to have the same coverage in zenoh

p-avital:


p-avital:
Can't wait for the election to finish so I can assign @milyin, he's very good at finding bugs :D

PR #492, opened 2023-05-01
name the threads spawned to service the TX queue

codebot:
Currently the TX queue threads are unnamed. This is a tiny PR to name them, so it's easier to understand the CPU usage. I'm not attached to the name scheme at all, feel free to edit it :slightly_smiling_face:


Mallets:
I'm fine with the naming scheme. I've just pushed a quick fix to the PR. Let the CI run now :)

PR #499, opened 2023-05-15
quinn 0.10.0 now supports connecting to peers named with IP addresses

YuanYuYuan:
As quinn [0.10.0](https://github.com/quinn-rs/quinn/releases/tag/0.10.0) has been released (the rustls 0.21.0 has been used), we no longer need to parse the DNS manually. (The parsing is now handled by [rustls](https://github.com/rustls/rustls/blob/main/rustls/src/client/client_conn.rs#L363-L374))

- Some simple QUIC pub & sub  tests with domain names and IP addresses are checked.
- I  only upgraded the related dependencies in `Cargo.toml` and left the `Cargo.lock` untouched.


io/zenoh-links/zenoh-link-quic/src/unicast.rs
```diff
@@ -222,11 +222,10 @@ impl LinkManagerUnicastQuic {
 impl LinkManagerUnicastTrait for LinkManagerUnicastQuic {
     async fn new_link(&self, endpoint: EndPoint) -> ZResult<LinkUnicast> {
         let epaddr = endpoint.address();
+        let host = epaddr.as_str().split(':').next().unwrap();
```
Mallets, line 225: Return an `Err` instead of `unrwap()`.

Mallets (changes requested):
Pleas address the comments before merging.

io/zenoh-links/zenoh-link-quic/src/unicast.rs
YuanYuYuan: Okay. Done.

Mallets (approved):

PR #507, opened 2023-05-30
Fixing hypothetic race in zenoh-links

yellowhatter:
I discovered some repeated code that I don't like: the concurrent listening task may (theoretically) finish earlier than the spawning task, leaving `listener` unremoved and thus leaving obsolete data in `self.listeners`. Obviously, 99.9(9)% of time that does not happen, but in some corner cases \ platforms that theoretically could.


Mallets:
Please contribute the PR to master.

Mallets (approved):

PR #509, opened 2023-06-07
Zenoh shm link layer and tests

yellowhatter:


io/zenoh-links/zenoh-link-shm/src/unicast.rs
```diff
@@ -0,0 +1,439 @@
...
+}
+#[async_trait]
+impl LinkUnicastTrait for UnicastPipe {
+    async fn close(&self) -> ZResult<()> {
+        log::trace!("Closing SHM Pipe link: {}", self);
+        Ok(())
+    }
+
+    async fn write(&self, buffer: &[u8]) -> ZResult<usize> {
+        //todo: I completely don't like the idea of locking reader\writer each time, but LinkUnicastTrait requires self.r and self.w to be Sync
```
Mallets, line 250: By construction in the rest of zenoh logic there is only one task writing at the time. So, an `UnsafeCell` could also do the trick.

io/zenoh-links/zenoh-link-shm/src/unicast.rs
```diff
@@ -0,0 +1,439 @@
...
+        self.w
+            .write()
+            .await
+            .write_all(buffer)
+            .await
+    }
+
+    async fn read(&self, buffer: &mut [u8]) -> ZResult<usize> {
+        //todo: I completely don't like the idea of locking reader\writer each time, but LinkUnicastTrait requires self.r and self.w to be Sync
```
Mallets, line 268: By construction in the rest of zenoh logic there is only one task reading at the time. So, an `UnsafeCell` could also do the trick.

io/zenoh-links/zenoh-link-shm/src/unicast.rs
```diff
@@ -0,0 +1,439 @@
...
+        self.r
+            .write()
+            .await
+            .read(buffer)
+            .await
+    }
+
+    async fn read_exact(&self, buffer: &mut [u8]) -> ZResult<()> {
+        //todo: I completely don't like the idea of locking reader\writer each time, but LinkUnicastTrait requires self.r and self.w to be Sync
```
Mallets, line 277: By construction in the rest of zenoh logic there is only one task reading at the time. So, an `UnsafeCell` could also do the trick.

io/zenoh-links/zenoh-link-shm/src/unicast.rs
```diff
@@ -0,0 +1,439 @@
...
+    w: async_std::sync::RwLock<PipeW>,
+    local: Locator,
+    remote: Locator,
+}
+impl Drop for UnicastPipe {
+    fn drop(&mut self) {}
+}
+#[async_trait]
+impl LinkUnicastTrait for UnicastPipe {
+    async fn close(&self) -> ZResult<()> {
```
Mallets, line 244: Close could happen concurrently. In this particular case is a no-op, but in case something needs to be done on the link, a mutex should be used here.

io/zenoh-transport/tests/unicast_authenticator.rs
```diff
@@ -808,6 +808,18 @@ fn authenticator_udp() {
...
 
+#[cfg(feature = "transport_shm")]
+#[test]
+fn authenticator_shm() {
+    let _ = env_logger::try_init();
+    task::block_on(async {
+        zasync_executor_init!();
+    });
+
+    let endpoint: EndPoint = "shm//tmp/authenticator_shm_test".to_string().parse().unwrap();
```
Mallets, line 819: `to_string()` can be removed.

io/zenoh-transport/tests/unicast_concurrent.rs
```diff
@@ -417,3 +417,37 @@ fn transport_ws_concurrent() {
...
+#[cfg(feature = "transport_shm")]
+#[test]
+fn transport_shm_concurrent() {
+    let _ = env_logger::try_init();
+    task::block_on(async {
+        zasync_executor_init!();
+    });
+
+    let endpoint01: Vec<EndPoint> = vec![
+        "shm//tmp/transport_shm_concurrent".to_string().parse().unwrap(),
```
Mallets, line 430: `to_string()` can be removed.

io/zenoh-links/zenoh-link-shm/src/unicast.rs
yellowhatter: In this case, why not relaxing Sync requirement in `LinkUnicastTrait` ?

Mallets: It can be investigated. This trait was defined at the very beginning (it was one of the first pieces of zenoh). So, the whole link trait could be revised. But that's I believe is out-of-scope of this PR.

commons/zenoh-config/src/lib.rs
```diff
@@ -283,6 +283,10 @@ validated_struct::validator! {
                     client_private_key: Option<String>,
                     client_certificate: Option<String>,
                 },
+                pub shared_memory: #[derive(Default)]
```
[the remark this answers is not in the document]
yellowhatter: This is a good reason for discussion. There are link config and transport config for shared memory - these are different layers with different logic and different config options. Moreover, the config options for shm-link would differ depending on shm link implementation (pipes, ringbuffer etc). I'm not sure about mixing transport and link configurations together....

io/zenoh-links/zenoh-link-shm/src/unicast.rs
yellowhatter: https://zettascale.atlassian.net/browse/ZEN-90

.github/workflows/ci.yml
```diff
@@ -68,6 +68,13 @@ jobs:
           command: clippy
           args: --all-targets --features shared-memory -- -D warnings
 
+      - name: Clippy transport_shm
+        if: ${{ matrix.os == 'ubuntu-latest' || matrix.os == 'macOS-latest' }}
+        uses: actions-rs/cargo@v1
+        with:
+          command: clippy
+          args: --all-targets --features shared-memory --features transport_shm -- -D warnings
```
Mallets, line 76: I would just add `--features transport_shm` to the previous test. I would avoid duplicating the clippy check.

.github/workflows/ci.yml
```diff
@@ -93,6 +100,7 @@ jobs:
         run: cargo install cargo-nextest --locked
 
       - name: Run tests
+        if: ${{ matrix.os == 'windows-latest' }}
```
Mallets, line 103: Why restricting the full tests only on `windows`?

.github/workflows/ci.yml
```diff
@@ -101,11 +109,12 @@ jobs:
...
 
-      - name: Run tests shared-memory
+      - name: Run tests with SHM
+        if: ${{ matrix.os == 'ubuntu-latest' }}
         uses: actions-rs/cargo@v1
         with:
-          command: test
-          args: --features "transport_tcp,shared-memory" -p zenoh-transport --test unicast_shm
+          command: nextest
+          args: run -F shared-memory -F transport_shm
```
Mallets, line 117: This would rerun the full test suite, which will take way too long. The reason of having only `unicast_shm` test is to reduce testing time.

.github/workflows/ci.yml
```diff
@@ -101,11 +109,12 @@ jobs:
           CARGO_REGISTRIES_CRATES_IO_PROTOCOL: sparse
           ASYNC_STD_THREAD_COUNT: 4
 
-      - name: Run tests shared-memory
+      - name: Run tests with SHM
+        if: ${{ matrix.os == 'ubuntu-latest' }}
```
Mallets, line 113: I'm not fond of having filtering at CI level. We can better tune it in the code via conditional compilation.

.github/workflows/release.yml
```diff
@@ -111,7 +111,6 @@ jobs:
 
   tests:
     name: Tests
-    if: ${{ !(github.event.inputs.runtests == 'false') }}
```
Mallets, line 114: Why removing this check?

commons/zenoh-config/src/lib.rs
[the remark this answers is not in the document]
Mallets: I'm still not convinced about the need of this configuration.

commons/zenoh-core/src/macros.rs
```diff
@@ -64,6 +64,20 @@ macro_rules! zasyncread {
     };
 }
 
+// This macro performs an async read with upgrade to write option on RwLock<T>
+// For performance reasons, it first performs a try_read() and,
+// if it fails, it falls back on read().await
+#[macro_export]
+macro_rules! zasyncread_upgradable {
```
Mallets, line 71: Oh cool, I didn't realize that upgradeable reads have been added into async rwlock. :)

commons/zenoh-protocol/src/network/oam.rs
```diff
@@ -21,6 +21,12 @@ pub mod flag {
     pub const Z: u8 = 1 << 7; // 0x80 Extensions    if Z==1 then an extension will follow
 }
 
+pub mod id {
```
Mallets, line 24: For style consistency, just `use super::OamId` and then replace the full path expansion`crate::transport::oam::OamId` with `OamId`.

examples/Cargo.toml
```diff
@@ -28,6 +28,19 @@ readme = "README.md"
 [features]
 shared-memory = ["zenoh/shared-memory"]
 unstable = ["zenoh/unstable"]
+transport_shm = ["zenoh/transport_shm"]
+default = [
+    "shared-memory",
+]
+
+# Unfortunately, the feature "transport_shm" is always
```
Mallets, line 36: Here there is some commented directives. Are they relevant? Do they hide a misbehaviour/bug that needs to be further investigated?

examples/examples/z_ping.rs
```diff
@@ -123,6 +124,11 @@ fn parse_args() -> (Config, Duration, usize, usize) {
     if args.is_present("no-multicast-scouting") {
         config.scouting.multicast.set_enabled(Some(false)).unwrap();
     }
+    #[cfg(feature = "shared-memory")]
+    if args.is_present("enable-shm") {
+        config.transport.shared_memory.set_enabled(true).unwrap();
+    }
```
Mallets, line 130: Add a check on `#[cfg(not(feature = "shared-memory"))]` and `if args.is_present("enable-shm")` saying that SHM can not be enabled because Zenoh has been compiled without `shared-memory` feature.

examples/examples/z_pong.rs
```diff
@@ -83,6 +84,10 @@ fn parse_args() -> Config {
     if args.is_present("no-multicast-scouting") {
         config.scouting.multicast.set_enabled(Some(false)).unwrap();
     }
+    #[cfg(feature = "shared-memory")]
+    if args.is_present("enable-shm") {
+        config.transport.shared_memory.set_enabled(true).unwrap();
+    }
```
Mallets, line 90: Add a check on `#[cfg(not(feature = "shared-memory"))]` and `if args.is_present("enable-shm")` saying that SHM can not be enabled because Zenoh has been compiled without `shared-memory` feature.

examples/examples/z_pub_thr.rs
```diff
@@ -106,9 +107,14 @@ fn parse_args() -> (Config, usize, Priority, bool, usize) {
...
         config.listen.endpoints = values.map(|v| v.parse().unwrap()).collect();
     }
+
     if args.is_present("no-multicast-scouting") {
         config.scouting.multicast.set_enabled(Some(false)).unwrap();
     }
+    #[cfg(feature = "shared-memory")]
+    if args.is_present("enable-shm") {
+        config.transport.shared_memory.set_enabled(true).unwrap();
+    }
```
Mallets, line 117: Add a check on `#[cfg(not(feature = "shared-memory"))]` and `if args.is_present("enable-shm")` saying that SHM can not be enabled because Zenoh has been compiled without `shared-memory` feature.

examples/examples/z_sub_thr.rs
```diff
@@ -141,6 +142,10 @@ fn parse_args() -> (Config, usize, usize) {
     if args.is_present("no-multicast-scouting") {
         config.scouting.multicast.set_enabled(Some(false)).unwrap();
     }
+    #[cfg(feature = "shared-memory")]
+    if args.is_present("enable-shm") {
+        config.transport.shared_memory.set_enabled(true).unwrap();
+    }
```
Mallets, line 148: Add a check on `#[cfg(not(feature = "shared-memory"))]` and `if args.is_present("enable-shm")` saying that SHM can not be enabled because Zenoh has been compiled without `shared-memory` feature.

io/zenoh-links/zenoh-link-shm/Cargo.toml
```diff
@@ -0,0 +1,51 @@
...
+async-io = ">= 1.13.0"
+log = { workspace = true }
+rand = { workspace = true }
+socket2 = { workspace = true }
+zenoh-buffers = { workspace = true }
+zenoh-collections = { workspace = true }
+zenoh-core = { workspace = true }
+zenoh-config = { workspace = true }
+zenoh-link-commons = { workspace = true }
+zenoh-protocol = { workspace = true }
```
Mallets, line 42: I don't think `zenoh-protocol` should be a dependency of `zenoh-link-shm`.

io/zenoh-links/zenoh-link-shm/src/lib.rs
```diff
@@ -0,0 +1,30 @@
...
+//!
+//! [Click here for Zenoh's documentation](../zenoh/index.html)
+
+// this error is designed to explicitly forbid transport_shm feature for non-unix platforms, but our
+// clippy with --all-targets flag triggers this on windows and I don't really know how to bypass this
+// I think it would be easier to make windows implementation for transport_shm rather than fighting with this :)
+// #[cfg(all(feature = "transport_shm", not(target_family = "unix")))]
+// compile_error!("Feature transport_shm: platform not supported yet!");
+
+#[cfg(all(feature = "transport_shm", unix))]
```
Mallets, line 27: Why having the `transport_shm` conditional feature here? Is not `target_family = "unix"` enough?

io/zenoh-links/zenoh-link-shm/src/unix/unicast.rs
```diff
@@ -0,0 +1,573 @@
+use crate::config;
```
Mallets, line 1: This `use` slipped before the header.

io/zenoh-links/zenoh-link-shm/src/unix/unicast.rs
```diff
@@ -0,0 +1,573 @@
...
+};
+use zenoh_protocol::core::{EndPoint, Locator};
+use zenoh_result::{bail, ZResult};
+
+use super::SHM_ACCESS_MASK;
+
+const LINUX_PIPE_MAX_MTU: u16 = 65_535;
+const LINUX_PIPE_DEDICATE_TRIES: usize = 100;
+
+static PIPE_INVITATION: &[u8] = &[0xDE, 0xAD, 0xBE, 0xEF];
```
Mallets, line 45: Poor beef :(

io/zenoh-links/zenoh-link-shm/src/unix/unicast.rs
```diff
@@ -0,0 +1,573 @@
...
+            (true, false) => {
+                // drop all the data from the pipe in case if it already exists
+                let mut buf: [u8; 1] = [0; 1];
+                while let Ok(val) = open_result.as_ref().unwrap().read(&mut buf) {
+                    if val == 0 {
+                        break;
+                    }
+                }
+            }
+            _ => {}
```
Mallets, line 156: What happens for the `(false, false)` case? Should the cleanup done as well?

io/zenoh-links/zenoh-link-shm/src/unix/unicast.rs
```diff
@@ -0,0 +1,573 @@
...
+        let open_result = Self::open_unique_pipe_for_read(path_r);
+        match (open_result.is_ok(), r_was_created.is_ok()) {
+            (false, true) => {
+                // clean-up in case of failure
+                let _ = remove_file(path_r).await;
+            }
+            (true, false) => {
+                // drop all the data from the pipe in case if it already exists
+                let mut buf: [u8; 1] = [0; 1];
+                while let Ok(val) = open_result.as_ref().unwrap().read(&mut buf) {
```
Mallets, line 150: Remove the `unwrap()` by replacing the `match { (bool, bool) .. }` with `match { (Ok(val), Err(_)) .. }`. I.e. matching directly on `Result` variants`.

io/zenoh-links/zenoh-link-shm/src/unix/unicast.rs
```diff
@@ -0,0 +1,573 @@
...
+struct UnicastPipe {
+    r: UnsafeCell<PipeR>,
+    w: UnsafeCell<PipeW>,
+    local: Locator,
+    remote: Locator,
+}
+
+impl UnicastPipe {
+    #[allow(clippy::mut_from_ref)]
+    fn get_r_mut(&self) -> &mut PipeR {
```
Mallets, line 422: Add `SAFETY` comment on what are the assumptions that are held in order for the `unsafe` block to actually be safe.

io/zenoh-links/zenoh-link-shm/src/unix/unicast.rs
```diff
@@ -0,0 +1,573 @@
...
+
+    async fn new_listener(&self, endpoint: EndPoint) -> ZResult<Locator> {
+        let listener = UnicastPipeListener::listen(endpoint.clone(), self.manager.clone()).await?;
+        let locator = listener.uplink_locator.clone();
+        self.listeners.write().unwrap().insert(endpoint, listener);
+        Ok(locator)
+    }
+
+    async fn del_listener(&self, endpoint: &EndPoint) -> ZResult<()> {
+        let removed = self.listeners.write().unwrap().remove(endpoint);
```
Mallets, line 537: The `unwrap()` might go away if the `RwLock` is from `async` instead of `std`. In any case, for style consistency, you might want to use `zwrite!` or `zasyncwrite!` macros.

io/zenoh-links/zenoh-link-shm/src/unix/unicast.rs
```diff
@@ -0,0 +1,573 @@
...
+            Some(val) => {
+                val.stop_listening().await;
+                Ok(())
+            }
+            None => bail!("No listener found for endpoint {}", endpoint),
+        }
+    }
+
+    fn get_listeners(&self) -> Vec<EndPoint> {
+        self.listeners.read().unwrap().keys().cloned().collect()
```
Mallets, line 548: The `unwrap()` might go away if the `RwLock` is from `async` instead of `std`. In any case, for style consistency, you might want to use `zread!` or `zasyncread!` macros.

io/zenoh-links/zenoh-link-shm/src/unix/unicast.rs
```diff
@@ -0,0 +1,573 @@
...
+    }
+
+    fn get_locators(&self) -> Vec<Locator> {
+        self.listeners
+            .read()
+            .unwrap()
```
Mallets, line 554: The `unwrap()` might go away if the `RwLock` is from `async` instead of `std`. In any case, for style consistency, you might want to use `zread!` or `zasyncread!` macros.

io/zenoh-transport/Cargo.toml
```diff
@@ -77,5 +78,5 @@ zenoh-util = { workspace = true }
 env_logger = { workspace = true }
 panic-message = { workspace = true }
 zenoh-protocol = { workspace = true, features = ["test"] }
-zenoh-transport = { workspace = true, features = ["test"] }
+zenoh-transport = { workspace = true, features = ["test", "transport_multilink"] }
```
Mallets, line 81: What's the purpose of `"transport_multilink"` feature here?

Mallets (changes requested):

.github/workflows/ci.yml
yellowhatter: Really, why...

.github/workflows/ci.yml
yellowhatter: The main reason why I used CI filtering is that transport_shm feature is not allowed for windows. But I think it would be easy to add windows support for transport_shm feature in the future (using windows named pipes that really seems to be an easy task).

.github/workflows/release.yml
yellowhatter: Hmm, that's a mistake

examples/Cargo.toml
yellowhatter: This seems to be a Cargo bug ;)
Here I tried to automatically enable `transport_shm` feature in examples only for Unix and disable for other platforms, but when uncommenting these lines I get zenoh built with `transport_shm` for windows too...

I left this commented block to highlight that there is a good mechanism for making platform-specific features that unfortunately doesn't work, and I plan to create a minimum example and submit an issue for Cargo some day.

io/zenoh-links/zenoh-link-shm/Cargo.toml
yellowhatter: Locator and EndPoint are defined in zenoh-protocol

yellowhatter: But I removed other unnecessary stuff here, thank you

io/zenoh-links/zenoh-link-shm/src/lib.rs
yellowhatter: Because of a damn Clippy!
I tried just to manage zenoh-link-shm as a `transport_shm` feature-triggered dependency, but `clippy --all-targets -- -D warnings` used in our CI unconditionally checks zenoh-link-shm and fails on windows...

io/zenoh-links/zenoh-link-shm/src/unix/unicast.rs
yellowhatter: No, for `(false, false)` we'll return ZResult containing error and this is that what is intended

io/zenoh-links/zenoh-link-shm/src/unix/unicast.rs
yellowhatter: Yes, very good comment, I will get rid of unwrap through value decomposition as you recommended during the meetup

io/zenoh-links/zenoh-link-shm/src/unix/unicast.rs
```diff
@@ -0,0 +1,573 @@
...
+        f.debug_struct("Shm")
+            .field("src", &self.local)
+            .field("dst", &self.remote)
+            .finish()
+    }
+}
+
+pub struct LinkManagerUnicastPipe {
+    manager: Arc<NewLinkChannelSender>,
+    listeners: std::sync::RwLock<HashMap<EndPoint, UnicastPipeListener>>,
```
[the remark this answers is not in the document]
yellowhatter: Sync RwLock should be more performant, and for the cases of low contention is acceptable inside async calls - that was the reason why I used it here. On the other hand, code around the `listeners` is not a hot path, so it is nice to get rid of unwrap() by means of using async implementation. I will migrate to async RwLock

io/zenoh-transport/Cargo.toml
yellowhatter: So, briefly, the tests for zenoh-transport do not compile without  `transport_multilink` feature!

When we do `cargo test`, `transport_multilink` feature is enabled by default in zenoh, so the tests compile.

But when we compile tests for separate test package, for example as we do on the CI, the compilation fails:
`cargo nextest run -F shared-memory -F transport_shm -p zenoh-transport`.

I think the proper way would be adding a support of `transport_multilink == false` for `zenoh-transport` tests, but this change actually adds conditional compilation to test scenarios and require testing for both modes. A treat this as a part of another workpackage (test refactoring and feature testing)

.github/workflows/ci.yml
Mallets: Shouldn't be `if: ${{ matrix.os == 'ubuntu-latest' || matrix.os == 'macOS-latest' }}`? Including also `macOS-latest` as target?

commons/zenoh-codec/src/network/oam.rs
```diff
@@ -21,8 +21,9 @@ use zenoh_protocol::{
     common::{iext, imsg, ZExtBody},
     network::{
         id,
-        oam::{ext, flag, Oam, OamId},
+        oam::{ext, flag, Oam},
     },
+    transport::oam::OamId,
```
Mallets, line 26: This change is simply wrong. Revert.

commons/zenoh-protocol/src/network/oam.rs
```diff
@@ -12,15 +12,22 @@
 //   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
 //
 use crate::common::ZExtBody;
-
-pub type OamId = u16;
+use crate::transport::oam::OamId;
```
Mallets, line 15: This is wrong. We are in the network layer. Do NOT import things from the transport layer. Revert.

examples/examples/z_ping.rs
```diff
@@ -12,6 +12,8 @@
 //   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
 //
 use clap::{App, Arg};
+#[cfg(not(feature = "shared-memory"))]
```
Mallets, line 15: I don't think logging is needed. Just print and exit.

examples/examples/z_ping.rs
```diff
@@ -123,6 +126,13 @@ fn parse_args() -> (Config, Duration, usize, usize) {
     if args.is_present("no-multicast-scouting") {
         config.scouting.multicast.set_enabled(Some(false)).unwrap();
     }
+    if args.is_present("enable-shm") {
+        #[cfg(feature = "shared-memory")]
+        config.transport.shared_memory.set_enabled(true).unwrap();
+        #[cfg(not(feature = "shared-memory"))]
+        warn!("enable-shm argument: SHM cannot be enabled, because Zenoh is compiled without shared-memory feature!");
```
Mallets, line 133: I don't think logging is needed. Just print and exit. Let the application fail.

examples/examples/z_pong.rs
```diff
@@ -14,6 +14,8 @@ use std::io::{stdin, Read};
 //   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
 //
 use clap::{App, Arg};
+#[cfg(not(feature = "shared-memory"))]
```
Mallets, line 17: Similar comment as `z_pong`.

examples/examples/z_pong.rs
```diff
@@ -83,6 +86,12 @@ fn parse_args() -> Config {
     if args.is_present("no-multicast-scouting") {
         config.scouting.multicast.set_enabled(Some(false)).unwrap();
     }
+    if args.is_present("enable-shm") {
+        #[cfg(feature = "shared-memory")]
+        config.transport.shared_memory.set_enabled(true).unwrap();
+        #[cfg(not(feature = "shared-memory"))]
```
Mallets, line 92: Similar comment as `z_pong`.

examples/examples/z_pub_thr.rs
```diff
@@ -12,6 +12,8 @@
 //   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
 //
 use clap::{App, Arg};
+#[cfg(not(feature = "shared-memory"))]
```
Mallets, line 15: Similar comment as `z_pong`.

examples/examples/z_pub_thr.rs
```diff
@@ -106,9 +109,16 @@ fn parse_args() -> (Config, usize, Priority, bool, usize) {
...
     }
+
     if args.is_present("no-multicast-scouting") {
         config.scouting.multicast.set_enabled(Some(false)).unwrap();
     }
+    if args.is_present("enable-shm") {
+        #[cfg(feature = "shared-memory")]
+        config.transport.shared_memory.set_enabled(true).unwrap();
+        #[cfg(not(feature = "shared-memory"))]
+        warn!("enable-shm argument: SHM cannot be enabled, because Zenoh is compiled without shared-memory feature!");
```
Mallets, line 120: Similar comment as `z_pong`.

examples/examples/z_sub_thr.rs
```diff
@@ -12,6 +12,8 @@
 //   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
 //
 use clap::{App, Arg};
+#[cfg(not(feature = "shared-memory"))]
```
Mallets, line 15: Similar comment as `z_pong`.

examples/examples/z_sub_thr.rs
```diff
@@ -141,6 +144,12 @@ fn parse_args() -> (Config, usize, usize) {
     if args.is_present("no-multicast-scouting") {
         config.scouting.multicast.set_enabled(Some(false)).unwrap();
     }
+    if args.is_present("enable-shm") {
+        #[cfg(feature = "shared-memory")]
+        config.transport.shared_memory.set_enabled(true).unwrap();
+        #[cfg(not(feature = "shared-memory"))]
```
Mallets, line 150: Similar comment as `z_pong`.

io/zenoh-links/zenoh-link-shm/src/unix/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+
+        let c = config.transport().link().shared_memory();
+        let shm_access_mask_;
+        if let Some(shm_access_mask) = c.shm_access_mask() {
+            shm_access_mask_ = shm_access_mask.to_string();
+            properties.push((config::SHM_ACCESS_MASK, &shm_access_mask_));
+        }
+
+        let mut s = String::new();
+        Parameters::extend(properties.drain(..), &mut s);
```
Mallets, line 60: Since access mask is passed as `EndPoint` parameter, I believe the changes in the config are not necessary. Access mask can be configured via the `EndPoint` like: `shm//tmp/myfile.shm#mask=777`.

io/zenoh-transport/tests/transport_whitelist.rs
```diff
@@ -137,3 +137,20 @@ fn transport_whitelist_tcp() {
     // Run
     task::block_on(run(&endpoints));
 }
+
+#[cfg(feature = "transport_shm")]
```
Mallets, line 141: I believe there is no need to run this test every time on the CI (it will make increase its duration).
Mark the test as follows:
```
#[test]
#[ignore]
```
It can be then run with `cargo test -- --ignored`.

io/zenoh-transport/tests/unicast_authenticator.rs
```diff
@@ -606,6 +606,18 @@ fn authenticator_udp() {
     task::block_on(run(&endpoint));
 }
 
+#[cfg(feature = "transport_shm")]
+#[test]
```
Mallets, line 610: I believe there is no need to run this test every time on the CI (it will make increase its duration).
Mark the test as follows:
```
#[test]
#[ignore]
```
It can be then run with `cargo test -- --ignored`.

io/zenoh-transport/tests/unicast_authenticator.rs
```diff
@@ -606,6 +606,18 @@ fn authenticator_udp() {
...
+    task::block_on(async {
+        zasync_executor_init!();
+    });
+
+    let endpoint: EndPoint = "shm//tmp/authenticator_shm_test".parse().unwrap();
+    task::block_on(run(&endpoint));
+}
+
 #[cfg(feature = "transport_ws")]
 #[test]
```
Mallets, line 622: Same applies here (which slipped from some old code):.

I believe there is no need to run this test every time on the CI (it will make increase its duration).
Mark the test as follows:
```
#[test]
#[ignore]
```
It can be then run with `cargo test -- --ignored`.

io/zenoh-transport/tests/unicast_concurrent.rs
```diff
@@ -412,3 +412,37 @@ fn transport_ws_concurrent() {
         transport_concurrent(endpoint01, endpoint02).await;
     });
 }
+
+#[cfg(feature = "transport_shm")]
+#[test]
```
Mallets, line 417: I believe there is no need to run this test every time on the CI (it will make increase its duration).
Mark the test as follows:
```
#[test]
#[ignore]
```
It can be then run with `cargo test -- --ignored`.

Please apply it as well to `transport_ws_concurrent`, which slipped in from the past.

io/zenoh-transport/tests/unicast_conduits.rs
```diff
@@ -346,6 +346,19 @@ fn conduits_tcp_only() {
     task::block_on(run(&endpoints));
 }
 
+#[cfg(feature = "transport_shm")]
+#[test]
```
Mallets, line 350: I believe there is no need to run this test every time on the CI (it will make increase its duration).
Mark the test as follows:
```
#[test]
#[ignore]
```
It can be then run with `cargo test -- --ignored`.

io/zenoh-transport/tests/unicast_defragmentation.rs
```diff
@@ -204,3 +204,42 @@ fn transport_unicast_defragmentation_ws_only() {
         }
     });
 }
+
+#[cfg(feature = "transport_shm")]
+#[test]
```
Mallets, line 209: I believe there is no need to run this test every time on the CI (it will make increase its duration).
Mark the test as follows:
```
#[test]
#[ignore]
```
It can be then run with `cargo test -- --ignored`.

Please apply the changes to `transport_ws` as well.

io/zenoh-transport/tests/unicast_intermittent.rs
```diff
@@ -408,3 +408,15 @@ fn transport_ws_intermittent() {
     let endpoint: EndPoint = format!("ws/127.0.0.1:{}", 12010).parse().unwrap();
     task::block_on(transport_intermittent(&endpoint));
 }
+
+#[cfg(feature = "transport_shm")]
+#[test]
```
Mallets, line 413: I believe there is no need to run this test every time on the CI (it will make increase its duration).
Mark the test as follows:
```
#[test]
#[ignore]
```
It can be then run with `cargo test -- --ignored`.
Please apply the changes to `transport_ws` as well.

io/zenoh-transport/tests/unicast_multilink.rs
```diff
@@ -517,6 +517,18 @@ mod tests {
         task::block_on(multilink_transport(&endpoint));
     }
 
+    #[cfg(feature = "transport_shm")]
+    #[test]
```
Mallets, line 521: I believe there is no need to run this test every time on the CI (it will make increase its duration).
Mark the test as follows:
```
#[test]
#[ignore]
```
It can be then run with `cargo test -- --ignored`.
Please apply the changes to `transport_ws` and `transport_unisock-stream` as well.

io/zenoh-transport/tests/unicast_openclose.rs
```diff
@@ -470,6 +470,18 @@ fn openclose_ws_only() {
     task::block_on(openclose_transport(&endpoint));
 }
 
+#[cfg(feature = "transport_shm")]
+#[test]
```
Mallets, line 474: I believe there is no need to run this test every time on the CI (it will make increase its duration).
Mark the test as follows:
```
#[test]
#[ignore]
```
It can be then run with `cargo test -- --ignored`.
Please apply the changes to `transport_ws` and `transport_unixsock-stream` as well.

io/zenoh-transport/tests/unicast_simultaneous.rs
```diff
@@ -329,6 +329,32 @@ mod tests {
         });
     }
 
+    #[cfg(feature = "transport_shm")]
+    #[test]
```
Mallets, line 333: I believe there is no need to run this test every time on the CI (it will make increase its duration).
Mark the test as follows:
```
#[test]
#[ignore]
```
It can be then run with `cargo test -- --ignored`.
Please apply the changes to `transport_ws` and `transport_unixsock-stream` as well.

io/zenoh-transport/tests/unicast_transport.rs
```diff
@@ -375,7 +404,14 @@ async fn open_transport(
     // Open transport -> This should be accepted
     for e in client_endpoints.iter() {
         println!("Opening transport with {}", e);
-        let _ = ztimeout!(client_manager.open_transport(e.clone())).unwrap();
+        //let _ = ztimeout!(client_manager.open_transport(e.clone())).unwrap();
```
Mallets, line 407: Don't remove the `ztimeout!`.

io/zenoh-transport/tests/unicast_transport.rs
```diff
@@ -375,7 +404,14 @@ async fn open_transport(
...
+        let e = client_manager.open_transport(e.clone()).await;
+        match e {
```
Mallets, line 409: In tests is fine to `unwrap()`. We early catch the problem and we do not continue in case of failure.
In this particular case, if the `open_transport` fails, all the following checks will fail as well. Better be explicit with an `unwrap()` which also return the line number where the error occured.

io/zenoh-transport/tests/unicast_transport.rs
```diff
@@ -441,14 +477,10 @@ async fn test_transport(
...
     );
-    let cctrl = match channel.reliability {
-        Reliability::Reliable => CongestionControl::Block,
-        Reliability::BestEffort => CongestionControl::Drop,
-    };
     // Create the message to send
     let message: NetworkMessage = Push {
         wire_expr: "test".into(),
-        ext_qos: QoSType::new(channel.priority, cctrl, false),
+        ext_qos: QoSType::new(channel.priority, CongestionControl::Block, false),
```
Mallets, line 483: Revert this change.

io/zenoh-transport/tests/unicast_transport.rs
```diff
@@ -441,14 +477,10 @@ async fn test_transport(
         "Sending {} messages... {:?} {}",
         MSG_COUNT, channel, msg_size
     );
-    let cctrl = match channel.reliability {
```
Mallets, line 444: Revert this change.

io/zenoh-transport/tests/unicast_transport.rs
```diff
@@ -576,6 +653,39 @@ fn transport_unicast_tcp_only() {
...
+#[test]
+fn transport_unicast_tcp_only_with_shm() {
+    let _ = env_logger::try_init();
+    task::block_on(async {
+        zasync_executor_init!();
+    });
+
+    // Define the locators
+    let endpoints: Vec<EndPoint> = vec![
+        format!("tcp/127.0.0.1:{}", 16000).parse().unwrap(),
```
Mallets, line 666: Colliding ports with other tcp tests. Please use `16090` and `16091`.

io/zenoh-transport/tests/unicast_transport.rs
```diff
@@ -535,19 +573,58 @@ async fn run_single(
...
+                *ms,
+                #[cfg(feature = "shared-memory")]
+                shm_transport,
+            )
+            .await;
         }
     }
 }
 
+async fn run(
```
Mallets, line 598: Let's explicitly call it `run_with_net` in the same way we call `run_with_shm`. This also reflects the actual code organization under `src`.

io/zenoh-transport/tests/unicast_transport.rs
```diff
@@ -604,6 +714,39 @@ fn transport_unicast_udp_only() {
...
+#[test]
+fn transport_unicast_udp_only_with_shm() {
+    let _ = env_logger::try_init();
+    task::block_on(async {
+        zasync_executor_init!();
+    });
+
+    // Define the locator
+    let endpoints: Vec<EndPoint> = vec![
+        format!("udp/127.0.0.1:{}", 16010).parse().unwrap(),
```
Mallets, line 727: Coliding UDP ports. Please use `16100` and `16101`.

io/zenoh-transport/tests/unicast_transport.rs
```diff
@@ -669,6 +850,112 @@ fn transport_unicast_ws_only() {
     task::block_on(run(&endpoints, &endpoints, &channel, &MSG_SIZE_ALL));
 }
 
+#[cfg(all(feature = "transport_ws", feature = "shared-memory",))]
+#[test]
```
Mallets, line 854: Instead of duplicating every test, wouldn't be easier to execute `run_with_net` and `run_with_shm` in already existing tests? In this way you also do not need to care about port number collisions.

Mallets (changes requested):

Mallets (changes requested):
Comments on CI and tests.

examples/Cargo.toml
```diff
@@ -28,6 +28,19 @@ readme = "README.md"
 [features]
 shared-memory = ["zenoh/shared-memory"]
 unstable = ["zenoh/unstable"]
+transport_shm = ["zenoh/transport_shm"]
+default = [
+    "shared-memory",
```
Mallets, line 33: `shared-memory` can not be default yet. Remove it.

io/zenoh-transport/src/unicast/manager.rs
```diff
@@ -406,12 +422,53 @@ impl TransportManager {
...
                 };
-                let a_t = Arc::new(TransportUnicastInner::make(self.clone(), stc)?);
+
+                // select and create transport implementation depending on the cfg
+                let a_t: Result<Arc<dyn TransportUnicastInnerTrait>, (Error, Option<u8>)> = {
+                    #[cfg(feature = "shared-memory")]
+                    {
+                        match stc.is_shm {
+                            false => {
+                                log::info!("Will use NET transport!");
```
Mallets, line 432: I don't think there is the need to log here. Information about the transport being used can is displayed at debug level upon successful establishment.

Moreover `log::info` is highly discouraged in the transport layer.

io/zenoh-transport/src/unicast/manager.rs
```diff
@@ -406,12 +422,53 @@ impl TransportManager {
...
+                                Ok(ShmTransportUnicastInner::make(self.clone(), stc, link)
+                                    .map_err(|e| (e, Some(close::reason::INVALID)))
+                                    .map(|v| Arc::new(v) as Arc<dyn TransportUnicastInnerTrait>)?)
+                            }
+                        }
+                    }
+                    #[cfg(not(feature = "shared-memory"))]
+                    {
+                        log::info!("Will use the only NET transport!");
+                        let t: Arc<dyn TransportUnicastInnerTrait> =
```
Mallets, line 455: There is some code duplication introduced by the conditional compilation. Need to investigate a patter on how to remove that.

commons/zenoh-codec/src/network/oam.rs
yellowhatter: `For style consistency, just use super::OamId and then replace the full path expansioncrate::transport::oam::OamId with OamId.`
I just misunderstood you here, sorry ;)
As far as we have `OamId` shred between layers, maybe it is good to move `Qam` definition to separate module and share it between transport and network layer modules?

io/zenoh-links/zenoh-link-shm/src/unix/mod.rs
yellowhatter: I propose to keep this configuration parameter as default global access mask - for convenience and more flexibility. For example, automatic SHM links (links created as an optimization during establish procedure) should generate `EndPoint` automatically and there is no way to tune it's access mask rather than through parameters.

io/zenoh-transport/tests/unicast_transport.rs
yellowhatter: That was just the debug code slipped here, sorry

io/zenoh-transport/tests/unicast_transport.rs
yellowhatter: That was just the debug code slipped here, sorry

commons/zenoh-codec/src/network/oam.rs
Mallets: There are 3 protocol levels: `transport`, `network`, `zenoh`. All of those three have an `Oam` message that is semantically different from one and the other and they may evolve in different directions. Let's keep them separate and let's avoid overlap.

io/zenoh-links/zenoh-link-shm/src/unix/mod.rs
Mallets: Right. Let's keep it. Both options should be possible though. As a general rule, the configuration passed in the `EndPoint` configuration `shm//tmp/myfile.shm#mask=777` takes over the more generic configuration.

io/zenoh-transport/src/unicast/shm/transport.rs
```diff
@@ -0,0 +1,329 @@
...
+use zenoh_protocol::core::{WhatAmI, ZenohId};
+use zenoh_protocol::network::NetworkMessage;
+use zenoh_protocol::transport::{Close, TransportSn};
+use zenoh_result::{bail, zerror, ZResult};
+
+/*************************************/
+/*             TRANSPORT             */
+/*************************************/
+#[derive(Clone)]
+pub(crate) struct ShmTransportUnicastInner {
```
Mallets, line 46: For naming consistency I'd call it `TransportUnicastShm` and `TransportUnicastNet`, both implement `TransportUnicastTrait`. Opinion?

io/zenoh-transport/src/unicast/shm/transport.rs
```diff
@@ -0,0 +1,329 @@
...
+
+        #[cfg(not(feature = "transport_shm"))]
+        bail!(
+            "Can not add Link {} with peer {}: link already exists and only unique link is supported!",
+            link,
+            self.config.zid,
+        );
+
+        #[cfg(feature = "transport_shm")]
+        async_std::task::block_on(async {
```
Mallets, line 235: If required, why not changing the `add_link` and `del_link` internal signature to `async`?

io/zenoh-transport/tests/unicast_transport.rs
yellowhatter: Sometimes I get congestion, and with `CongestionControl::Drop` test fails here:
`client_transport.schedule(message.clone()).unwrap(); (line 499 in this branch)`

The original code sets `CongestionControl::Drop` for unreliable transports, BUT the congestion policy is applied at `pipeline.rs` input, which causes `Transport::schedule` to return `false` in case of congestion......
As far as we need to test the transport's message scheduling, we couldn't ignore `Transport::schedule` return value.

io/zenoh-transport/tests/unicast_transport.rs
yellowhatter: see my comment above

commons/zenoh-codec/src/network/oam.rs
yellowhatter: Ok, I've got it! Thought that Oam is intended to be the same for all levels (some kind of cross-layer extendable functionality)

io/zenoh-links/zenoh-link-shm/src/unix/mod.rs
yellowhatter: Yes, currently the mask in EndPoint has higher priority than generic mask

io/zenoh-transport/Cargo.toml
Mallets: Ok, this is then linked to the changes introduced in the CI. We are introducing the additional feature in `[dev-dependencies]` for test purposes. 

If I run the command provided by you (the same as in the CI) I get:
```
$ cargo nextest run -F shared-memory -F transport_shm -p zenoh-transport
error: Found argument '-F' which wasn't expected, or isn't valid in this context

	If you tried to supply `-F` as a value rather than a flag, use `-- -F`

USAGE:
    cargo nextest run [OPTIONS] [FILTERS]...

For more information try --help
```

Then, by following cargo's suggestion no tests are run at all:

```
$ cargo nextest run -- -F shared-memory -F transport_shm -p zenoh-transport
    Blocking waiting for file lock on build directory
   Compiling zenoh-protocol v0.10.0-dev (/Users/luca/Repository/eclipse-zenoh/commons/zenoh-protocol)
   Compiling zenoh-core v0.10.0-dev (/Users/luca/Repository/eclipse-zenoh/commons/zenoh-core)
   Compiling zenoh-sync v0.10.0-dev (/Users/luca/Repository/eclipse-zenoh/commons/zenoh-sync)
   Compiling zenoh-util v0.10.0-dev (/Users/luca/Repository/eclipse-zenoh/commons/zenoh-util)
   Compiling zenoh-codec v0.10.0-dev (/Users/luca/Repository/eclipse-zenoh/commons/zenoh-codec)
   Compiling zenoh-link-commons v0.10.0-dev (/Users/luca/Repository/eclipse-zenoh/io/zenoh-link-commons)
   Compiling zenoh-config v0.10.0-dev (/Users/luca/Repository/eclipse-zenoh/commons/zenoh-config)
   Compiling zenoh-plugin-trait v0.10.0-dev (/Users/luca/Repository/eclipse-zenoh/plugins/zenoh-plugin-trait)
   Compiling zenoh-link-ws v0.10.0-dev (/Users/luca/Repository/eclipse-zenoh/io/zenoh-links/zenoh-link-ws)
   Compiling zenoh-link-tcp v0.10.0-dev (/Users/luca/Repository/eclipse-zenoh/io/zenoh-links/zenoh-link-tcp)
   Compiling zenoh-link-udp v0.10.0-dev (/Users/luca/Repository/eclipse-zenoh/io/zenoh-links/zenoh-link-udp)
   Compiling zenoh-link-unixsock_stream v0.10.0-dev (/Users/luca/Repository/eclipse-zenoh/io/zenoh-links/zenoh-link-unixsock_stream)
   Compiling zenoh-link-quic v0.10.0-dev (/Users/luca/Repository/eclipse-zenoh/io/zenoh-links/zenoh-link-quic)
   Compiling zenoh-link-tls v0.10.0-dev (/Users/luca/Repository/eclipse-zenoh/io/zenoh-links/zenoh-link-tls)
   Compiling zenoh-link-shm v0.10.0-dev (/Users/luca/Repository/eclipse-zenoh/io/zenoh-links/zenoh-link-shm)
   Compiling zenoh-link-serial v0.10.0-dev (/Users/luca/Repository/eclipse-zenoh/io/zenoh-links/zenoh-link-serial)
   Compiling zenoh-link v0.10.0-dev (/Users/luca/Repository/eclipse-zenoh/io/zenoh-link)
   Compiling zenoh-transport v0.10.0-dev (/Users/luca/Repository/eclipse-zenoh/io/zenoh-transport)
   Compiling zenoh v0.10.0-dev (/Users/luca/Repository/eclipse-zenoh/zenoh)
   Compiling zenoh-ext v0.10.0-dev (/Users/luca/Repository/eclipse-zenoh/zenoh-ext)
   Compiling zenoh_backend_traits v0.10.0-dev (/Users/luca/Repository/eclipse-zenoh/plugins/zenoh-backend-traits)
   Compiling zenoh-examples v0.10.0-dev (/Users/luca/Repository/eclipse-zenoh/examples)
   Compiling zenoh-plugin-storage-manager v0.10.0-dev (/Users/luca/Repository/eclipse-zenoh/plugins/zenoh-plugin-storage-manager)
   Compiling zenoh-plugin-rest v0.10.0-dev (/Users/luca/Repository/eclipse-zenoh/plugins/zenoh-plugin-rest)
   Compiling zenoh-plugin-example v0.10.0-dev (/Users/luca/Repository/eclipse-zenoh/plugins/example-plugin)
   Compiling zenohd v0.10.0-dev (/Users/luca/Repository/eclipse-zenoh/zenohd)
    Finished test [unoptimized + debuginfo] target(s) in 1m 59s
    Starting 0 tests across 54 binaries (199 skipped)
     Summary [   0.002s] 0 tests run: 0 passed, 199 skipped
```

There is something off here...

io/zenoh-transport/tests/unicast_transport.rs
Mallets: see my reply above

io/zenoh-transport/Cargo.toml
yellowhatter: I've just copy and pasted the line that works for me:
`cargo nextest run -F shared-memory -F transport_shm -p zenoh-transport`
It is the same as you tried. Maybe the problem in nextest version? I use recent 0.9.57

yellowhatter: `dmitriy@dmitriy-TITAN:~/zenoh$ cargo nextest run -F shared-memory -F transport_shm -p zenoh-transport
warning: constant `MSG_SIZE_ALL` is never used
   --> io/zenoh-transport/tests/unicast_transport.rs:223:7
    |
223 | const MSG_SIZE_ALL: [usize; 2] = [1_024, 131_072];
    |       ^^^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default

warning: `zenoh-transport` (test "unicast_transport") generated 1 warning
    Finished test [unoptimized + debuginfo] target(s) in 0.11s
    Starting 12 tests across 13 binaries (13 skipped)
        PASS [   0.004s] zenoh-transport common::seq_num::tests::sn_gap
        PASS [   0.004s] zenoh-transport common::batch::tests::serialization_batch
        PASS [   0.004s] zenoh-transport common::seq_num::tests::sn_precedence
        PASS [   0.004s] zenoh-transport common::seq_num::tests::sn_generation
        PASS [   0.004s] zenoh-transport common::seq_num::tests::sn_set
        PASS [   0.007s] zenoh-transport common::pipeline::tests::tx_pipeline_blocking
        PASS [   0.087s] zenoh-transport unicast::establishment::cookie::tests::codec_cookie
        PASS [   0.162s] zenoh-transport common::pipeline::tests::tx_pipeline_flow
        PASS [   2.101s] zenoh-transport::endpoints endpoint_shm
        PASS [   5.344s] zenoh-transport::unicast_shm tests::transport_shm_shm
        PASS [   6.427s] zenoh-transport::unicast_transport transport_unicast_shm_only
        PASS [   8.378s] zenoh-transport::unicast_transport transport_unicast_shm_only_with_shm
------------
     Summary [   8.381s] 12 tests run: 12 passed, 13 skipped
dmitriy@dmitriy-TITAN:~/zenoh$ 
`

yellowhatter:
```
dmitriy@dmitriy-TITAN:~/zenoh$ cargo nextest run -F shared-memory -F transport_shm -p zenoh-transport
warning: constant `MSG_SIZE_ALL` is never used
   --> io/zenoh-transport/tests/unicast_transport.rs:223:7
    |
223 | const MSG_SIZE_ALL: [usize; 2] = [1_024, 131_072];
    |       ^^^^^^^^^^^^
    |
    = note: `#[warn(dead_code)]` on by default

warning: `zenoh-transport` (test "unicast_transport") generated 1 warning
    Finished test [unoptimized + debuginfo] target(s) in 0.11s
    Starting 12 tests across 13 binaries (13 skipped)
        PASS [   0.004s] zenoh-transport common::seq_num::tests::sn_gap
        PASS [   0.004s] zenoh-transport common::batch::tests::serialization_batch
        PASS [   0.004s] zenoh-transport common::seq_num::tests::sn_precedence
        PASS [   0.004s] zenoh-transport common::seq_num::tests::sn_generation
        PASS [   0.004s] zenoh-transport common::seq_num::tests::sn_set
        PASS [   0.007s] zenoh-transport common::pipeline::tests::tx_pipeline_blocking
        PASS [   0.087s] zenoh-transport unicast::establishment::cookie::tests::codec_cookie
        PASS [   0.162s] zenoh-transport common::pipeline::tests::tx_pipeline_flow
        PASS [   2.101s] zenoh-transport::endpoints endpoint_shm
        PASS [   5.344s] zenoh-transport::unicast_shm tests::transport_shm_shm
        PASS [   6.427s] zenoh-transport::unicast_transport transport_unicast_shm_only
        PASS [   8.378s] zenoh-transport::unicast_transport transport_unicast_shm_only_with_shm
------------
     Summary [   8.381s] 12 tests run: 12 passed, 13 skipped
dmitriy@dmitriy-TITAN:~/zenoh$ 

```

Mallets: Right, I was still with `nextest 0.9.8`. I've updated it to `0.9.57` and I get the same result as you.

io/zenoh-transport/src/unicast/shm/rx.rs
```diff
@@ -0,0 +1,146 @@
+use super::oam_extensions::unpack_oam_close;
```
Mallets, line 1: It slipped before the header.

Mallets (approved):

Issue #516, opened 2023-06-30
[Bug] Errors in plugins leave zenohd in a stalled state


OlivierHecart:
### Describe the bug

When errors occur in plugins, zenohd does not exit but does not work any more and seems to be in a stalled state.

### To reproduce

1. Run `zenohd` with rest plugin:
```bash
./zenohd 
```
2. Run a second `zenohd` with the rest plugin on different zenoh port but same http port:
```bash
./zenohd -l tcp/[::]:7448
```

The second zenohd reports an error running the rest plugin but keeps running and is not reachable by clients on port 7448.

### System info

- Platform: macOS M2
- Zenoh version 0.7.2-rc

Mallets:
@milyin is this issue still relevant or #583 solved it?

wyfo:
It no longer seems to be the case, I've just started 2 routers with the same REST HTTP port.
The second one expectedly fails to start the REST plugin, but I can then start a pub and a sub (with multicast scouting disabled) that are able to communicate through the second router.

PR #517, opened 2023-07-03
Allow plugins to fail at startup, and Zenohd to react to that failure

p-avital:
Helps mitigate, but doesn't _solve_ #516 

Bonus: fix clippy lints related to compression


zenohd/src/main.rs
```diff
@@ -1,3 +1,5 @@
+use std::collections::HashSet;
```
Mallets, line 1: It slipped before the header.

Mallets (approved):

PR #521, opened 2023-07-24
Support more Value types.

evshary:
integer: i64, i32, i16, i8, isize
unsigned integer: u64, u32, u16, u8, usize
floating point: f64, f32

### DIFF
--- a/zenoh/src/value.rs
+++ b/zenoh/src/value.rs
@@ -275,6 +275,294 @@ impl TryFrom<Value> for i64 {
     }
 }
 
+// i32 conversion
+impl From<i32> for Value {
+    fn from(i: i32) -> Self {
+        Value {
+            payload: ZBuf::from(Vec::<u8>::from(i.to_string())),
+            encoding: KnownEncoding::AppInteger.into(),
+        }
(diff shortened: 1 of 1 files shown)


zenoh/src/value.rs
```diff
@@ -307,6 +595,38 @@ impl TryFrom<Value> for f64 {
     }
 }
 
+// f32 conversion
+impl From<f32> for Value {
+    fn from(f: f32) -> Self {
+        Value {
+            payload: ZBuf::from(f.to_string().as_bytes().to_vec()),
```
Mallets, line 602: Please construct the payload with ` ZBuf::from(Vec::<u8>::from(f.to_string()))` as in the other cases. I saw that also `f64` does it in this way. Do you mind updating it?

Mallets (changes requested):

zenoh/src/value.rs
evshary: Sure, I thought it was designed intentionally.
I'll update f32 and f64 just like the others.

Mallets (approved):

PR #526, opened 2023-08-02
Multicast support

Mallets:


io/zenoh-transport/src/multicast/establishment.rs
```diff
@@ -11,75 +11,81 @@
...
-    _manager: &TransportManager,
-    _link: LinkMulticast,
+    manager: &TransportManager,
+    link: LinkMulticast,
 ) -> ZResult<TransportMulticast> {
-    todo!();
-    // // Create and configure the multicast transport
-    // let mut prng = zasynclock!(manager.prng);
+    // Create and configure the multicast transport
+    let mut prng = zasynclock!(manager.prng);
```
p-avital, line 34: This lock is held until end of function, shouldn't it be released at line 54?

io/zenoh-transport/src/multicast/establishment.rs
```diff
@@ -11,75 +11,81 @@
...
-    //     zlock!(manager.state.multicast.transports).remove(locator);
-    //     let _ = ti.stop_rx();
-    //     e
-    // })?;
+    // Add the link
+    ti.start_tx()?;
 
-    // Ok(transport)
+    // Store the active transport
+    zasynclock!(manager.state.multicast.transports).insert(locator.clone(), ti.clone());
```
p-avital, line 70: What if there's alerady a transport at this locator? (insert would return `Some(previous_ti)`)

io/zenoh-transport/src/multicast/manager.rs
```diff
@@ -170,9 +168,9 @@ impl TransportManager {
     pub async fn close_multicast(&self) {
         log::trace!("TransportManagerMulticast::clear())");
 
-        zlock!(self.state.multicast.protocols).clear();
+        zasynclock!(self.state.multicast.protocols).clear();
 
-        let mut tm_guard = zlock!(self.state.multicast.transports)
+        let mut tm_guard = zasynclock!(self.state.multicast.transports)
```
p-avital, line 173: alternatively, you could swap an empty vec with the transports, and then iterate over transports. Probably overkill since we're closing through, but I also think it'd be easier to understand.

io/zenoh-transport/src/multicast/rx.rs
```diff
@@ -58,197 +68,249 @@ impl TransportMulticastInner {
...
+                "Ingoring Join on {} from peer: {}. Unsupported version: {}. Expected: {}.",
+                locator,
+                join.zid,
+                join.version,
+                self.manager.config.version,
+            );
+            return Ok(());
+        }
+
+        if join.resolution > self.manager.config.resolution {
```
p-avital, line 125: Pico doesn't support resolutions different from its own, wouldn't this mean that a Pico with resolution < manager.config.resoluction would be able to join, but couldn't be joined by this?

io/zenoh-transport/src/unicast/manager.rs
```diff
@@ -473,28 +479,31 @@ impl TransportManager {
         super::establishment::open::open_link(&link, self).await
     }
 
-    pub async fn get_transport_unicast(&self, peer: &ZenohId) -> Option<TransportUnicast> {
-        zasynclock!(self.state.unicast.transports)
-            .get(peer)
-            .map(|t| t.into())
+    pub fn get_transport_unicast(&self, peer: &ZenohId) -> Option<TransportUnicast> {
```
p-avital, line 482: why turn it into a sync function if the inner code is still async? if it's for caller's benefit, shouldn't caller be the one doing task::block_on?

p-avital (changes requested):
A few minor details, and a few semantic check hints

io/zenoh-transport/src/multicast/establishment.rs
Mallets: Good catch.

io/zenoh-transport/src/multicast/manager.rs
Mallets: I've replaced with a single `drain()` iterator. I believe now it should be more compact and easier to read while keeping the lock as before.

io/zenoh-transport/src/multicast/rx.rs
Mallets: I believe if the resolution is smaller it should be able to join. The other way around is not true. So, I believe zenoh-pico should support the case when a smaller resolution is advertised.

io/zenoh-transport/src/unicast/manager.rs
Mallets: To simply avoid changes in the routing code. There are few changes going on there and I'd like to minimise the impact external to the transport.

io/zenoh-transport/src/multicast/manager.rs
p-avital: My point was that you could steal the Vec from the lock (steal 3 usizes), and then do the iteration after the lock is released

Mallets: Yes, but I'd rather guarantee there are no concurrency issues when closing multicast transports and someone, for any reason, still tries to open new multicast transports. In normal situations that would never happen but keeping the lock would keep us safe.

io/zenoh-transport/src/multicast/rx.rs
p-avital: Yeah, but then the sn would need to respect that resolution, which means that every peer on the multicast network must be aware of the minimal resolution, including if it changes as a new peer joins the network. Sounds general's paradoxey to me :/

Mallets: You are right, while on the reading side it would pose no problem, it does on the writing side. So you are correct. In unicast we can do that check and react accordingly with sn resolution, but in multicast is cumbersome. Let's stick to strict equality check for now.

io/zenoh-transport/src/unicast/manager.rs
Mallets: I've moved the functions to `async` and in the caller a `block_on` has been introduced.

p-avital (approved):

PR #528, opened 2023-08-07
Use .readthedocs.yaml for building ReadTheDocs's content. 

evshary:
RTD requests the migration of configuration v2:
https://blog.readthedocs.com/migrate-configuration-v2/


p-avital:
@Mallets have we even used ReadTheDocs with Zenoh ever since 0.5? If not, we could just drop it entirely, since `docs.rs` is our canon doc for zenoh-rs

Mallets:
> @Mallets have we even used ReadTheDocs with Zenoh ever since 0.5? If not, we could just drop it entirely, since `docs.rs` is our canon doc for zenoh-rs

We can drop `ReadTheDocs` entirely and use only `docs.rs`.

p-avital:
@evshary since you've already made this PR on the subject: do you mind repurposing it to remove the RTD stuff from the repo? :)

evshary:
Sure, then I'll remove all `zenoh/docs` folder directly.

p-avital (approved):

PR #534, opened 2023-08-15
external file plugin configuration, json schema support

milyin:
- plugin settings can contain optional string fields `__config__` and `__schema__`.
- if `__config__` is specified, it's content is merged into zenohd plugin config. Settings in `__config__` overrides values in zenohd config
- if `__schema__` is specified, the additional verification of `__config__` file is performed accordingly to the json schema. `__schema__` doesn't affect plugin settings in zenohd config


Mallets (approved):

Mallets:
LGTM code wise. I'd like @gabrik to validate it.

Mallets:
A different PR could come to support the same pattern within the [storage manager plugin](https://github.com/eclipse-zenoh/zenoh/tree/master/plugins/zenoh-plugin-storage-manager). The storage manager loads various storage backends, which in turn would also need to provide a `__config__` and `__schema__`.

milyin:
it was decided not to add `__schema__` option to cofing. Replacement pr is https://github.com/eclipse-zenoh/zenoh/pull/536

PR #537, opened 2023-08-20
automatic generation of json schema for plugin config

milyin:
The schema is generated on build stage by `build.rs` based on rust config structure. When structure is changed, schema is regenerated and should be committed into git together with changes in Rust code.


gabrik:
LGTM, modulo the failing CI, I think there should be some feature-gate for no_std.

plugins/zenoh-plugin-rest/config_schema.json5
```diff
@@ -0,0 +1,26 @@
...
+      ]
+    },
+    "__required__": {
+      "type": [
+        "boolean",
+        "null"
+      ]
+    },
+    "http_port": {
+      "type": "string"
```
p-avital, line 22: The generated schema here is incomplete: a number would also be accepted. Not sure it that's an issue for our purpose though

plugins/zenoh-plugin-rest/config_schema.json5
```diff
@@ -0,0 +1,26 @@
...
+    "__path__": {
+      "type": [
+        "string",
+        "null"
+      ]
+    },
+    "__required__": {
+      "type": [
+        "boolean",
+        "null"
```
p-avital, line 18: I would have expected `{"type": "boolean", "required": false}`. Is this equivalent, or is the `_required_` key now mandatory, but nullable?

plugins/zenoh-plugin-rest/config_schema.json5
milyin: This goes directly from definition of port field in zenoh-plugin-rest/config.rs. If we want to correct it, we should fix this type
```
pub struct Config {
    #[serde(deserialize_with = "deserialize_http_port")]
    pub http_port: String,
    __path__: Option<String>,
    __required__: Option<bool>,
}
```

plugins/zenoh-plugin-rest/config_schema.json5
milyin: I need to check json schema standard, but at least example file without "__required__" field passes the check.
I've added additional verification of sample config file against generated schema. Seems that serde is much less strict than formal schema specification. Sometimes it seems completely wrong: schema requires `volumes` and `storages` to be vectors, but our config sample shows them as objects:
```Rust
pub struct PluginConfig {
    pub name: String,
    pub required: bool,
    pub backend_search_dirs: Option<Vec<String>>,
    pub volumes: Vec<VolumeConfig>,
    pub storages: Vec<StorageConfig>,
...
```

gabrik (approved):

PR #538, opened 2023-08-21
Extend the sleeping time to address the timeout issue.

YuanYuYuan:


Mallets:
I believe it's not a good idea to modify the test to ignore the timeout. In the way the test has been built (module some error in the test itself), the timeout should never happen. If the timeout ultimately happens, it means something weird is going on...

YuanYuYuan:
Well, just to be too optimistic. It seemed that an extended sleep time can't resolve the issue completely.

PR #542, opened 2023-08-25
Change rust-toolchain to rust-toolchain.toml file

JEnoch:
The `rust-toolchain` file format has been deprecated in favour of a `rust-toolchain.toml` file using TOML syntax.
See https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file

The deprecated format is causing troubles to dependabot, which was displaying such errors:

> Dependabot can't parse your rust-toolchain
> Dependabot failed to update your dependencies because there was an error parsing the rust-toolchain found at /rust-toolchain.
> 
> Dependabot encountered the following error:
>
> `only rust-toolchain files formatted as TOML are supported, the non-TOML format was deprecated by Rust`

This PR changes to the new TOML format, keeping Rust 1.71.0 toolchain version.


JEnoch:
As a side effect, the `actions-rs/toolchain@v1` used in workflows had to be replaced because it doesn't support `rust-toolchain.toml` files. Instead, the `rustup show` command is used, as it automatically installed read the file and install the appropriate toolchain.

gabrik (approved):

PR #556, opened 2023-09-21
Prometheus

OlivierHecart:


zenoh/src/net/runtime/adminspace.rs
```diff
@@ -72,6 +72,10 @@ impl AdminSpace {
 
         let mut handlers: HashMap<_, Handler> = HashMap::new();
         handlers.insert(root_key.clone(), Arc::new(router_data));
+        handlers.insert(
+            format!("@/router/{zid_str}/metrics").try_into().unwrap(),
```
Mallets, line 76: Should we call it `@/router/{zid_str}/stats` instead of `@/router/{zid_str}/metrics` for consistency with the rest?

zenoh/src/net/runtime/adminspace.rs
```diff
@@ -471,6 +489,40 @@ fn router_data(context: &AdminContext, query: Query) {
     }
 }
 
+fn router_metrics(context: &AdminContext, query: Query) {
+    let reply_key: OwnedKeyExpr = format!("@/router/{}/metrics", context.zid_str)
```
Mallets, line 493: Should we call it @/router/{zid_str}/stats instead of @/router/{zid_str}/metrics for consistency with the rest?

zenoh/src/net/runtime/adminspace.rs
gabrik: `metrics` is the industry standard.

zenoh/src/net/runtime/adminspace.rs
gabrik: See above.

Mallets (approved):

PR #558, opened 2023-10-02
Add interface bound tcp link

YuanYuYuan:
- socket2 -> async-std tcp stream/listener is synchronous only.
- socket2 has different APIs on different platforms. 

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4718,6 +4718,7 @@ dependencies = [
  "async-std",
  "async-trait",
  "log",
+ "socket2 0.5.4",
  "zenoh-core",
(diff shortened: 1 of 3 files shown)


fuzzypixelz:
@YuanYuYuan Please change your pull request's base branch to main (new default branch). And rebase your branch against `main` as it is missing a status check necessary to merge this pull request but which is only available on `main`.

Mallets:
@YuanYuYuan @sashacmc I believe we can close this PR since it has been addressed by #691 and #742 ?

YuanYuYuan:
Yes. Let's close this.

PR #561, opened 2023-10-04
Rest get body

OlivierHecart:


plugins/zenoh-plugin-rest/src/lib.rs
```diff
@@ -397,14 +398,15 @@ async fn query(req: Request<(Arc<Session>, String)>) -> tide::Result<Response> {
...
-        match req
-            .state()
-            .0
-            .get(&selector)
-            .consolidation(consolidation)
-            .res()
-            .await
-        {
+        let mut query = req.state().0.get(&selector).consolidation(consolidation);
+        if !body.is_empty() {
```
Mallets, line 402: `body` is not used before line `402`. However, line `377` does `unwrap_or_default()` and then we check `if !body.is_empty()`. Would it be possible to merge the two check with an `if let Some(body) = req.body_bytes().await { .. }`?

OlivierHecart: The issue is that `body_bytes()` needs a temporary mutable borrow of `req` while `url` and derived (`key_expr`, `selector`) need an immutable borrow of `req` until used in the query. So options were either:
1. Copy selector to owned.
2. Read the body bytes before taking the immutable borrow for `url`
I opted for option 2 (in normal operation, we'll read the bytes anyway).

Mallets: It sounds reasonable, thanks!

Mallets (approved):

PR #565, opened 2023-10-10
Matching status

OlivierHecart:


Mallets (approved):

zenoh/src/publication.rs
```diff
@@ -198,6 +202,35 @@ use std::pin::Pin;
 use std::task::{Context, Poll};
 use zenoh_result::Error;
 
+#[zenoh_macros::unstable]
+#[derive(Clone)]
+pub enum PublisherRef<'a, 'b: 'a> {
```
p-avital, line 207: Are the 2 lifetimes really necessary? Everything passes a single lifetime from then on

OlivierHecart: Probably not. Removed them.

PR #568, opened 2023-10-12
Fix missing batch length check

Mallets:
When reading from a streamed link, the advertised batch length could exceed the allocated buffer.
This PR introduces a validation check instead of directly subslice the receiving buffer to read from the link.


milyin (commented):
looks good for me

Issue #515, opened 2023-06-27
[Bug] systemd error 217/user on fresh install


eliasdc:
### Describe the bug

Zenohd systemd fails to launch from a fresh install of `zenohd`.

post installation step on debian fails to recognize `systemctl` with the test `command -v systemctl` in `zenoh/zenohd/.deb/postinst`, which then skips adding the zenohd user which is used to launch the service.

Bug introduced in tagged version [0.7.2-dev] and from commit 6beb99c

### To reproduce

1. apt purge zenoh*
2. apt install zenohd
3. systemctl start zenohd.service
4. systemctl status zenohd.service

### System info

- Platform: [Ubuntu 20.04 64-bit, Ubuntu 22.04 64-bit]
- Zenohd version [0.7.2-rc]
- Bug introduced in [0.7.2-dev]

Mallets:
@evshary was this issue solved or is it still open?

evshary:
> @evshary was this issue solved or is it still open?

This should be solved after that bugfix.

Mallets:
Closing the issue then, if the problem still persist let's reopen it.

PR #576, opened 2023-10-30
Bugfix: sh (dash) dose not support `&> /dev/null`.

evshary:
Since sh (link to dash in Ubuntu) does not support `&> /dev/null`, the if-condition will always fail and won't create zenohd users.
I believe this is why systemd fails to run zenohd while bootup.

Is there any way to build the deb file locally? I don't know how to create the deb file easily for testing.

### DIFF
--- a/zenohd/.deb/postinst
+++ b/zenohd/.deb/postinst
@@ -20,12 +20,12 @@ set -e
 
 case "$1" in
     configure)
-        if ! command -v systemctl &> /dev/null
+        if ! command -v systemctl > /dev/null 2>&1
         then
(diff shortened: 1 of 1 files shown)


gabrik:
Hi @evshary 
> 
> Is there any way to build the deb file locally? I don't know how to create the deb file easily for testing.

I think `cargo deb -p zenohd` should do the trick.

evshary:
@gabrik Thanks! I didn't know the trick before!
After some tests, I think the PR fixes the problem.
I just found that it's related to the issue https://github.com/eclipse-zenoh/zenoh/issues/515

OlivierHecart (approved):

PR #580, opened 2023-11-06
Inconsistent duplicated timestamp in Publication

juanjole:
Hi, I've realized that when a Publication is performed to both local and remote destinations, the corresponding timestamps are different since they are generated for each block. I fixed it with a global timestamp for both scenarios as in the Put operation to have the same timestamp. Regards.


JEnoch (approved):

JEnoch:
Thanks for spotting and fixing this @juanjole !

PR #582, opened 2023-11-07
Issue a diagnostic whenever `Publication` is not used

### LINKED ISSUE #581: must_use for put
### Describe the feature

I think it's too easy to do this: 
```rust
publisher.put(value);
```
instead of this 
```rust
publisher.put(value).res().await;
```
So I suggest we add `#[must_use]` to `put` [here](https://github.com/eclipse-zenoh/zenoh/blob/b2959f0f1278ef71b92654338d7796a99ce59722/zenoh/src/publication.rs#L306) so the former creates a compiler warning.

fuzzypixelz:
Resolves #581.

I opted to put `#[must_use]` on `Publication` itself instead of `Publisher::put`. This way we cover `Publisher::delete` and `Publisher::write` too, and any method that returns `Publication`.

### DIFF
--- a/zenoh/src/publication.rs
+++ b/zenoh/src/publication.rs
@@ -404,6 +404,7 @@ impl Drop for Publisher<'_> {
 
 /// A [`Resolvable`] returned by [`Publisher::put()`](Publisher::put),
 /// [`Publisher::delete()`](Publisher::delete) and [`Publisher::write()`](Publisher::write).
+#[must_use]
 pub struct Publication<'a> {
     publisher: &'a Publisher<'a>,
     value: Value,


zenoh/src/publication.rs
```diff
@@ -404,6 +404,7 @@ impl Drop for Publisher<'_> {
 
 /// A [`Resolvable`] returned by [`Publisher::put()`](Publisher::put),
 /// [`Publisher::delete()`](Publisher::delete) and [`Publisher::write()`](Publisher::write).
+#[must_use]
```
p-avital, line 407: Could you use this ```
#[must_use = "Resolvables do nothing unless you resolve them using the `res` method from either `SyncResolve` or `AsyncResolve`"]
``` specifically, as this is the justification we use in the other resolvables

p-avital (changes requested):
Could you apply this to all types that impl `SyncResolve`, you should be able to find them all using Rust Analyzer's reference search.

p-avital (approved):

PR #583, opened 2023-11-09
plugin API rework, plugin status support

milyin:
- internal change: hash of validator closures replaced to single trait - to simplify `RunningPluginTrait`
- plugin compatibility support improved. Now it compares
  - rust version
  - structures used versions
  - build features affecting structures
- PluginManager rework
  - added ability to keep information of plugins not loaded yet
  - added ability to request plugin status
  - pluign manager API made easier to use. Separate traits for declared, loaded and started plugins allows to avoid repetitive error checks
  - Used same plugin manager in both zenohd and storage magager plugin

Corresponging PRs:
https://github.com/eclipse-zenoh/zenoh-backend-influxdb/pull/54
https://github.com/eclipse-zenoh/zenoh-backend-rocksdb/pull/40
https://github.com/eclipse-zenoh/zenoh-backend-filesystem/pull/46
https://github.com/eclipse-zenoh/zenoh-plugin-webserver/pull/32
https://github.com/eclipse-zenoh/zenoh-plugin-mqtt/pull/32
https://github.com/eclipse-zenoh/zenoh-plugin-dds/pull/168
https://github.com/eclipse-zenoh/zenoh-plugin-ros1/pull/30
https://github.com/eclipse-zenoh/zenoh-plugin-ros2dds/pull/51


fuzzypixelz:
@milyin Please change your pull request's base branch to main (new default branch).

plugin_test_config.json5
Mallets, line 1: I believe this file is not needed here. It may be part of the CI though.

plugins/zenoh-backend-example/config.json5
Mallets, line 1: What's the purpose of an empty config file?

plugins/zenoh-backend-example/src/lib.rs
Mallets, line 1: Header is missing.

plugins/zenoh-plugin-storage-manager/src/lib.rs
```diff
@@ -307,20 +262,37 @@ impl From<StorageRuntimeInner> for StorageRuntime {
...
-            log::info!("old: {:?}", &old);
-            log::info!("new: {:?}", &new);
-            let diffs = ConfigDiff::diffs(old, new);
-            log::info!("diff: {:?}", &diffs);
-            { zlock!(runtime).update(diffs) }?;
-            Ok(None)
-        })
+        let old = PluginConfig::try_from((&name, old))?;
+        let new = PluginConfig::try_from((&name, new))?;
+        log::info!("old: {:?}", &old);
```
Mallets, line 290: Is the log intended to stay there or used during development for testing purposes?
If it is meant to stay then I would provide a richer log message. As well, I'd move to the log to debug level.

plugins/zenoh-plugin-trait/src/compatibility.rs
```diff
@@ -0,0 +1,258 @@
...
+            writeln!(f, "Plugin version: {}", plugin_version)?;
+        }
+        Ok(())
+    }
+}
+
+#[repr(C)]
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct RustVersion {
+    major: u64,
```
Mallets, line 206: Not a blocker for this PR but `u64` seems a lot for the version.

plugins/zenoh-plugin-trait/src/manager.rs
```diff
@@ -0,0 +1,300 @@
...
+        self.plugins.push(PluginRecord::new(loader));
+        Ok(self.plugins.last_mut().unwrap())
+    }
+
+    fn get_plugin_index(&self, name: &str) -> Option<usize> {
+        self.plugins.iter().position(|p| p.name() == name)
+    }
+
+    /// Lists all plugins
+    pub fn declared_plugins(
```
Mallets, line 178: Not a blocker, but I'd rather stick to [Rust naming convention](https://rust-lang.github.io/api-guidelines/naming.html#methods-on-collections-that-produce-iterators-follow-iter-iter_mut-into_iter-c-iter) for methods returning iterators.

plugins/zenoh-plugin-trait/src/plugin.rs
```diff
@@ -0,0 +1,240 @@
...
+    }
+    pub fn messages(&self) -> &[Cow<'static, str>] {
+        &self.messages
+    }
+}
+
+pub trait PluginConditionSetter {
+    fn add_error(self, report: &mut PluginReport) -> Self;
+    fn add_warning(self, report: &mut PluginReport) -> Self;
+    fn add_message(self, report: &mut PluginReport) -> Self;
```
Mallets, line 218: As log levels we have "info, warn, err ,debug, trace". Do `add_error`, `add_warning`, and `add_message` refer to those levels? If so, would `add_message` better be named as `add_info`?

plugins/zenoh-backend-example/src/lib.rs
```diff
@@ -0,0 +1,135 @@
+use std::{
+    collections::{hash_map::Entry, HashMap},
+    sync::Arc,
+};
+
+use async_std::sync::RwLock;
+//
```
Mallets, line 7: Please restore the right order of header and `use`.

Mallets (changes requested):
Overall it looks good to me. I've left few minor comments that should be addressed before merging.

plugins/zenoh-plugin-trait/src/plugin.rs
milyin: No, this is related to plugin's error report level:
```
pub enum PluginReportLevel {
    #[default]
    Normal,
    Warning,
    Error,
}
```
I'll think about giving names which shows this relation more clear.

Mallets: Ok, maybe `Normal` could be renamed to `Info` then? What do you think?

plugin_test_config.json5
milyin: Moved to https://github.com/ZettaScaleLabs/zenoh-tests/pull/33

plugins/zenoh-backend-example/config.json5
milyin: yes, it can be safely removed. Done.

plugins/zenoh-backend-example/src/lib.rs
milyin: fixed

plugins/zenoh-backend-example/src/lib.rs
milyin: fixed

plugins/zenoh-plugin-trait/src/compatibility.rs
milyin: fixed, seems like u32 for each field should be enough

plugins/zenoh-plugin-trait/src/manager.rs
milyin: Done

plugins/zenoh-plugin-trait/src/plugin.rs
milyin: Agree. This enum was "PluginStatus" before, the "Normal" remained from this. Now "Info" is ok here.

plugins/zenoh-plugin-storage-manager/src/lib.rs
milyin: This came from existing code without changes. But I agree, this is better to put into debug level. Fixed.

zenoh/src/net/routing/face.rs
```diff
@@ -31,6 +31,7 @@ pub struct FaceState {
     pub(super) id: usize,
     pub(super) zid: ZenohId,
     pub(super) whatami: WhatAmI,
+    #[allow(dead_code)]
```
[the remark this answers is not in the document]
milyin: This is a side-effect of hiding fileds in RuntimeState, which were public before (https://github.com/eclipse-zenoh/zenoh/pull/583/files#diff-d1c4d04e44c0a6c9db07b8efc60893ed69cdcdb7d8c970192a034f94280df339R51-R61) This caused large leak of private structures into public API. So the unused function `FaceState::is_local()` was incorrectly public before.
Seems that this field and function can be removed, but better to ask @OlivierHecart first. So for now they just marked as dead.

milyin:
> Overall it looks good to me. I've left few minor comments that should be addressed before merging.

Done with all the comments.

Mallets (approved):

PR #590, opened 2023-11-15
Add protocol extensions for user attachment

Mallets:
This PR adds an `attachment` extension to `Put`, `Del`, `Query`, and `Reply` Zenoh messages.
This is a necessary step to allow the user to attach any arbitrary metadata to `put`, `del`, `query`, and `reply` operations.

An example of some on-going effort of such API in zenoh-c can be found here: https://github.com/eclipse-zenoh/zenoh-c/pull/190

Although I believe this PR will not evolve any further, I'll keep it in Draft for the time being since a validation with a full attachment API is desirable before merging.

### DIFF
--- a/commons/zenoh-buffers/src/zbuf.rs
+++ b/commons/zenoh-buffers/src/zbuf.rs
@@ -89,7 +89,7 @@ impl PartialEq for ZBuf {
                 (None, _) | (_, None) => return false,
                 (Some(l), Some(r)) => {
                     let cmp_len = l.len().min(r.len());
-                    // SAFETY: cmp_len is the minimum lenght between l and r slices.
+                    // SAFETY: cmp_len is the minimum length between l and r slices.
                     let lhs = crate::unsafe_slice!(l, ..cmp_len);
                     let rhs = crate::unsafe_slice!(r, ..cmp_len);
                     if lhs != rhs {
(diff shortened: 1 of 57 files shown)


examples/examples/z_pub_thr.rs
```diff
@@ -11,24 +11,31 @@
...
+
+    let mut prio = Priority::default();
+    if let Some(p) = args.priority {
+        prio = p.try_into().unwrap();
+    }
+
+    let payload_size = args.payload_size;
 
-    let data: Value = (0usize..size)
+    let data: Value = (0usize..dbg!(payload_size))
```
Mallets, line 33: Why `dbg!`?

p-avital: forgot to remove it, done now

Mallets:
This PR introduces a throughput regression of `~200K msg/s`. However, #610 improved throughput of `~200K msg/s`. So, we are even. Nevertheless, @p-avital I think it's worth investigating why this PR introduces such a throughput drop.

Approving this PR so as to unblock https://github.com/eclipse-zenoh/zenoh-c/pull/203 and https://github.com/eclipse-zenoh/zenoh-cpp/pull/86.

p-avital (approved):

p-avital (approved):

PR #592, opened 2023-11-16
Sifting plugin private fields on debug logs.

DariusIMP:
Sifting private fields on the plugin configs that may be exposed on the logs.

### DIFF
--- a/commons/zenoh-config/src/lib.rs
+++ b/commons/zenoh-config/src/lib.rs
@@ -623,9 +623,15 @@ impl Config {
 
 impl std::fmt::Display for Config {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        let mut json = serde_json::to_value(self).unwrap();
-        sift_privates(&mut json);
-        write!(f, "{json}")
+        serde_json::to_value(self)
+            .map(|mut json| {
+                sift_privates(&mut json);
+                write!(f, "{json}")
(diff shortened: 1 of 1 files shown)


commons/zenoh-config/src/lib.rs
```diff
@@ -1032,7 +1032,9 @@ impl<'a> serde::Deserialize<'a> for PluginsConfig {
 }
 impl std::fmt::Debug for PluginsConfig {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{:?}", &self.values)
+        let mut json = serde_json::to_value(self).unwrap();
```
Mallets, line 1035: Why `unwrap()`? `fmt` returns a `Result`, proper error conversion should be made.

Mallets (changes requested):

commons/zenoh-config/src/lib.rs
DariusIMP: I replicated the logic from: https://github.com/eclipse-zenoh/zenoh/blob/a67f9192d885d2af0d76ed64182944051d8130b7/commons/zenoh-config/src/lib.rs#L624-L630 

Will have to change that there as well then

Mallets: Yes please.

Mallets (approved):

PR #596, opened 2023-11-20
POC: SHM watchdog

yellowhatter:


fuzzypixelz:
@yellowhatter Please change your pull request's base branch to main (new default branch). And rebase your branch against `main` as it is missing a status check necessary to merge this pull request but which is only available on `main`.

commons/zenoh-shm/src/posix_shm/segment.rs
```diff
@@ -0,0 +1,87 @@
...
+use rand::Rng;
+use shared_memory::{Shmem, ShmemConf, ShmemError};
+use zenoh_result::{bail, zerror, ZResult};
+
+const SEGMENT_DEDICATE_TRIES: usize = 100;
+
+/// Segment of shared memory identified by an ID
+pub struct Segment<ID> {
+    pub shmem: Shmem,
+    pub id: ID,
```
Mallets, line 26: Since `ID` is a const generic I believe there is no need to add `ID` on the stack.
One alternative is to have the `ID` as a const like done in the [protocol extensions](https://github.com/eclipse-zenoh/zenoh/blob/0943243da1a9eb14644897b13934be25610c4a94/commons/zenoh-protocol/src/common/extension.rs#L114).

PR #597, opened 2023-11-21
Fix bug causing duplicates in querying subscriber

OlivierHecart:


Mallets (approved):

zenoh-ext/src/querying_subscriber.rs
```diff
@@ -913,3 +901,34 @@ where
...
+    TryIntoSample,
+>(
+    fetch: Fetch,
+    handler: RepliesHandler,
+) -> ZResult<()>
+where
+    TryIntoSample: TryInto<Sample>,
+    <TryIntoSample as TryInto<Sample>>::Error: Into<zenoh_core::Error>,
+{
+    log::debug!("Fetch");
```
Mallets, line 925: I would make the debug message more exhaustive.

Mallets:
LGTM. Just make the debug log more informative.

zenoh-ext/src/querying_subscriber.rs
OlivierHecart: Done

PR #599, opened 2023-11-23
FetchingSubscriber::fetch don't need mutable reference

milyin:
Method `FetchingSubscriber::fetch` works ok with shared reference, `&mut` is not needed and causes problems for zenoh-c wrapper for this method (makes zenoh-c require owned reference instead of loaned for this)


milyin:
Tested compilation with dds and ros2dds plugins

Mallets (approved):

PR #602, opened 2023-11-28
Set example and plugin-example crate to non-publish to reduce release build time 

marvin-hansen:
USER STATEMENT

As a Rust developer 

I want 
to get reasonable build times when adding a new dependency 

But
Currently Zenoh just blows release build times through the roof. 

Therefore
I identified two optional crates "examples" and "plugin-example" that really slow down the release build and so I have marked them as non-publish to prevent them getting shipped and slowing down the build process.

Comment:

I actually don't have hard numbers, but it's painfully obvious that the entire release build of my project hangs for minutes just compiling zenoh code examples that shouldn't be shipped in the first place. Please accept this PR to get the already awful release build time at least a bit shorter by excluding the code examples that really are meant to reside only in the GH repo.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -24,6 +24,7 @@ categories = { workspace = true }
 description = "Internal crate for zenoh."
 readme = "README.md"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+publish = false
 
 [features]
 shared-memory = ["zenoh/shared-memory"]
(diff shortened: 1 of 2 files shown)


marvin-hansen:
I've just signed the ECA. Not sure why it's not showing up.

marvin-hansen:
@Mallets Ok, ECA check is done. Thanks for the clarification.

Mallets (approved):

PR #604, opened 2023-11-28
featue leak fixed, test added

milyin:
there was a leak of zenoh's default features through zenoh-ext. Fixed it, added test to zenohd to control this in future


Mallets:
This PR looks good to me, however I believe it may break our plugins and bindings in cascade without additional PRs.
E.g., [zenoh-c](https://github.com/eclipse-zenoh/zenoh-c/pull/204/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542) needs to explicitly add `default-features = false ` when including zenoh and add `default = ["zenoh/default"]` for the crate's default features.

milyin:
> This PR looks good to me, however I believe it may break our plugins and bindings in cascade without additional PRs. E.g., [zenoh-c](https://github.com/eclipse-zenoh/zenoh-c/pull/204/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542) needs to explicitly add `default-features = false ` when including zenoh and add `default = ["zenoh/default"]` for the crate's default features.

There is no break for dependent crates/projects. zenoh-c is not required to add "default-features = false", it just *can* add this parameter now and be sure that no unnecessary features leak out through another crate.
I.e. it's still possible to include zenoh with or without default features, as before. The only difference that now including zenoh with `default-features = false` really works

Mallets (approved):

PR #606, opened 2023-11-30
Add homebrew lib search path for Apple Silicon systems

fuzzypixelz:
Fixes https://github.com/eclipse-zenoh/homebrew-zenoh/issues/2.

This adds `/opt/homebrew/lib` to the default list of plugin search paths. This is because macOS systems running Apple Silicon no longer use `/usr/local/lib`. Of course, there is always an argument against hard-coding values like this, and there has been [much](https://github.com/Homebrew/brew/issues/13481) [discussion](https://github.com/orgs/Homebrew/discussions/1600) about this.

At the end of the day, little changes the fact that Apple has made it so `/opt/homebrew/lib` is the new `/usr/local/lib` on macOS. Ruby's [ffi](https://github.com/ffi/ffi/blob/82e20928319e499b2e60679205733cb5f088c363/lib/ffi/library.rb#L129) for instance also hard-codes the search path.


fuzzypixelz:
There is also the option of adding `../lib` which is more (too?) general.

PR #609, opened 2023-12-04
Fix type of `__path__` in `Config`

fuzzypixelz:
[Context](https://github.com/eclipse-zenoh/zenoh-plugin-webserver/pull/25#issuecomment-1824448225).

### DIFF
--- a/plugins/zenoh-plugin-rest/src/config.rs
+++ b/plugins/zenoh-plugin-rest/src/config.rs
@@ -23,7 +23,8 @@ const DEFAULT_HTTP_INTERFACE: &str = "[::]";
 pub struct Config {
     #[serde(deserialize_with = "deserialize_http_port")]
     pub http_port: String,
-    __path__: Option<String>,
+    #[serde(default, deserialize_with = "deserialize_path")]
+    __path__: Option<Vec<String>>,
     __required__: Option<bool>,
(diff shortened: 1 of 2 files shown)


plugins/zenoh-plugin-rest/src/config.rs
```diff
@@ -76,3 +77,145 @@ impl<'de> Visitor<'de> for HttpPortVisitor {
...
+        E: de::Error,
+    {
+        Ok(vec![v.into()])
+    }
+
+    fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+    where
+        A: de::SeqAccess<'de>,
+    {
+        let mut v = if let Some(l) = seq.size_hint() {
```
Mallets, line 132: Here you could use [map_or_else](https://doc.rust-lang.org/std/option/enum.Option.html#method.map_or_else) instead of `if { ... } else { ... }`.

Mallets (changes requested):
Just a minor comment about using a more rusty idiom.

plugins/zenoh-plugin-rest/src/config.rs
fuzzypixelz: Done!

Mallets (approved):

PR #610, opened 2023-12-05
Refactor transport trait

yellowhatter:


yellowhatter:
Refactoring summary:
- fixed bug with open_ack race condition
- improved transport initialization code in unicast manager: more readable with clear on-failure cleanup (close link, close transport)
- configuring transport object can never be accessed until it's configuration finished  
- some small fixes and optimizations here and there

yellowhatter:
The only uncertain thing here why I still keep this as a draft is TX executor usage in transports. I need to figure out why it doesn't work in a proper way (that's really strange). Everything other is OK and can be reviewed.

io/zenoh-transport/src/unicast/universal/link.rs
```diff
@@ -38,58 +38,66 @@ use zenoh_sync::{RecyclingObject, RecyclingObjectPool, Signal};
...
-            // The pipeline
-            let (producer, consumer) = TransmissionPipeline::make(config, priority_tx);
-            self.pipeline = Some(producer);
-
             // Spawn the TX task
-            let c_link = self.link.clone();
+            let mut tx = self.link.tx();
             let c_transport = self.transport.clone();
-            let handle = executor.spawn(async move {
+            let handle = async_std::task::spawn(async move {
```
Mallets, line 96: You need to spawn the tx thread in the `TransportExecutor`, not in the `async_std` global one.

io/zenoh-transport/src/unicast/universal/link.rs
```diff
@@ -38,58 +38,66 @@ use zenoh_sync::{RecyclingObject, RecyclingObjectPool, Signal};
...
     // The underlying link
-    pub(super) link: TransportLinkUnicast,
+    pub(super) link: Arc<TransportLinkUnicast>,
     // The transmission pipeline
     pub(super) pipeline: Option<TransmissionPipelineProducer>,
     // The transport this link is associated to
     transport: TransportUnicastUniversal,
     // The signals to stop TX/RX tasks
-    handle_tx: Option<Arc<async_executor::Task<()>>>,
+    handle_tx: Option<Arc<JoinHandle<()>>>, //Option<Arc<async_executor::Task<()>>>,
```
Mallets, line 47: You need to use the `TransportExecutor` to spawn `tx` tasks, not the `async_std` global one.

io/zenoh-transport/src/unicast/universal/link.rs
yellowhatter: That is a problem that I mentioned - the task spawned in the executor seems not to be executed. Sounds strange, but that's it, I'm working on how to understand why that happens

io/zenoh-transport/src/unicast/link.rs
```diff
@@ -38,14 +38,22 @@ pub(crate) struct TransportLinkUnicastConfig {
     pub(crate) is_compression: bool,
 }
 
-#[derive(Clone, PartialEq, Eq)]
+#[derive(PartialEq, Eq)]
 pub(crate) struct TransportLinkUnicast {
-    pub(crate) link: LinkUnicast,
+    pub(crate) link: Arc<LinkUnicast>,
```
Mallets, line 43: Why un `Arc`? [`LinkUnicast`](https://github.com/eclipse-zenoh/zenoh/blob/eb60dd09b6c67ea1d1d0044ff6531d2e6821d4bf/io/zenoh-link-commons/src/unicast.rs#L39) already contains an `Arc`. This introduce an additional indirection that I don't think is needed...

io/zenoh-transport/src/unicast/link.rs
```diff
@@ -273,3 +287,61 @@ impl fmt::Debug for TransportLinkUnicastRx {
             .finish()
     }
 }
+
+pub(crate) struct EstablishAck {
+    link: TransportLinkUnicastTx,
+    ack: Option<TransportMessage>,
+}
```
Mallets, line 294: It's unclear the need of the `EstablishAck` and of the `ack` functions.

yellowhatter: `EstablishAck` stores optional open_ack. The `add_link` return value obliges user to unwrap the `EstablishedTransportLinkUnicast` object in an `add_link` method implementation and return `EstablishAck` in case of success, so that the ack could not be forgotten by the developer. The `TransportManager` takes care of using `EstablishAck` object further to send acknowledge in the right moment. If failure happens anywhere, the `EstablishedTransportLinkUnicast` is unwrapped with `fail()` method that drops `EstablishAck`....

yellowhatter: The input and return arguments around transport initialization methods in the `TransportManager` are designed to put some logical constrains to make initialization procedure more safe: not to forget to send the ack, not to forget to close everything already open in case of failure etc....

Mallets: Ok. Then I'll be more explicit in the naming and in the type. E.g.:
`open_ack: Option<OpenAck>`

io/zenoh-transport/src/unicast/link.rs
Mallets: `LinkUnicast` is not `Clone` anymore (it has never been `Copy`). Still it holds an `Arc` internally and here we are introducing an additional indirection that is traversed every time `send` is called. That means, every time we send a message, we pay the cost of the additional indirection... In any case, we are in the situation of having an `Arc<Arc<..>>`...

yellowhatter: Oh, I mistyped

yellowhatter: The only way to solve this issue is to get back to `Clone` in `LinkUnicast`, I think I don't have another options. Yes, you are right that we're gonna pay the cost on `send` that is not good

yellowhatter:
This is ready!
BTW, I noticed this: #622

io/zenoh-transport/src/common/seq_num.rs
```diff
@@ -121,6 +121,37 @@ impl SeqNum {
...
+    /// the length for the interval that separates them.
+    ///
+    /// ## Case 2: sna > snb
+    ///
+    /// In this case *sna* precedes *snb* iff (sna - snb) > semi_int.
+    ///
+    /// # Arguments
+    ///
+    /// * `value` -  The sequence number which should be checked for precedence relation.
+    pub(crate) fn roll(&mut self, value: TransportSn) -> ZResult<bool> {
```
Mallets, line 143: What's the purpose of this new function?

io/zenoh-transport/src/unicast/universal/rx.rs
```diff
@@ -197,7 +197,7 @@ impl TransportUnicastUniversal {
                 .decode()
                 .map_err(|_| zerror!("{}: decoding error", link))?;
 
-            log::trace!("Received: {:?}", msg);
+            // log::trace!("Received: {:?}", msg);
```
Mallets, line 200: This trace is useful for debugging.

yellowhatter: I tried to keep it away from the hotpath....

Mallets (approved):

Mallets (approved):

PR #613, opened 2023-12-05
Attachment api

p-avital:


zenoh/src/publication.rs
```diff
@@ -117,6 +121,22 @@ impl PutBuilder<'_, '_> {
...
         self
     }
+
+    #[zenoh_macros::unstable]
+    pub fn attachments(&self) -> Option<&Attachments> {
+        self.attachments.as_ref()
+    }
+
+    #[zenoh_macros::unstable]
+    pub fn attachments_mut(&mut self) -> &mut Option<Attachments> {
```
Mallets, line 131: `attachments` returns `Option<&Attachments>`, `attachments_mut` should return `Option<&mut Attachments>`.

zenoh/src/query.rs
```diff
@@ -294,6 +306,22 @@ impl<'a, 'b, Handler> GetBuilder<'a, 'b, Handler> {
         self
     }
 
+    #[zenoh_macros::unstable]
+    pub fn attachments(&self) -> Option<&Attachments> {
+        self.attachments.as_ref()
+    }
+
+    #[zenoh_macros::unstable]
+    pub fn attachments_mut(&mut self) -> &mut Option<Attachments> {
```
Mallets, line 315: It should return `Option<&mut Attachments>`.

zenoh/src/queryable.rs
```diff
@@ -150,6 +160,39 @@ pub struct ReplyBuilder<'a> {
...
+    #[zenoh_macros::unstable]
+    pub fn attachments(&self) -> Option<&Attachments> {
+        match &self.result {
+            Ok(sample) => sample.attachments.as_ref(),
+            Err(_) => None,
+        }
+    }
+
+    #[zenoh_macros::unstable]
+    pub fn attachments_mut(&mut self) -> Option<&mut Option<Attachments>> {
```
Mallets, line 173: Why double option?

zenoh/tests/attachments.rs
```diff
@@ -0,0 +1,112 @@
...
+        .res()
+        .unwrap();
+    for i in 0..10 {
+        let mut backer = [(
+            [0; std::mem::size_of::<usize>()],
+            [0; std::mem::size_of::<usize>()],
+        ); 10];
+        for (j, backer) in backer.iter_mut().enumerate() {
+            *backer = ((i * 10 + j).to_le_bytes(), (i * 10 + j).to_be_bytes())
+        }
```
Mallets, line 93: I understand that the `Iterator` pattern being used here allows to avoid some extra copy. However, API-wise is very verbose and not-so elegant. Before going any further, I'd like to see a benchmark (e.g. pub/sub throughput) using the `assert.insert(...)` and the `Iterator` approach. I.e.: I'd like to understand the tradeoff between API usability and performance impact.

commons/zenoh-buffers/src/zbuf.rs
```diff
@@ -55,6 +55,77 @@ impl ZBuf {
...
+        let start_slice = &mut self.slices.as_mut()[start_slice_idx];
+        start_slice.end = start_slice.start + start_idx_in_start_slice;
+        let drain_start = start_slice_idx + (start_slice.start < start_slice.end) as usize;
+        let end_slice = &mut self.slices.as_mut()[end_slice_idx];
+        end_slice.start += end_idx_in_end_slice;
+        let drain_end = end_slice_idx + (end_slice.start >= end_slice.end) as usize;
+        self.slices.drain(drain_start..drain_end);
+    }
+    fn insert(&mut self, mut at: usize, slice: &[u8]) {
+        if slice.is_empty() {
```
sashacmc, line 104: It would be nice to check if `at` less than buffer size

zenoh/src/sample.rs
```diff
@@ -98,6 +98,193 @@ impl From<Option<DataInfo>> for SourceInfo {
...
+        pub fn is_empty(&self) -> bool {
+            self.len() == 0
+        }
+        pub fn len(&self) -> usize {
+            Zenoh080.read(&mut self.inner.reader()).unwrap_or(0)
+        }
+        pub fn iter(&self) -> AttachmentsIterator {
+            self.into_iter()
+        }
+        fn _get(&self, key: &[u8]) -> Option<ZSlice> {
```
sashacmc, line 148: Why we need `_get`, if it used only in the `get`? (the same for the insert)
Is it some rust syntax sugar?

sashacmc (commented):
It would be nice to add usage examples, preferably containing both approaches through an iterator and through a map.

sashacmc:
More thoughts.
Based on the name of the branch and for general reasons (we have en attachment which has map of key value pairs), I thought it would be correct to name the feature as an `attachment` not  `attachments` and in my [PR](https://github.com/eclipse-zenoh/zenoh-c/pull/203/files#diff-00776be53d5f126d657ea68a085286124f2255c73d98df956676e65861abde16R29) I already did it.
But there still some inconsistency between  `attachment` not  `attachments`.
I think we need to agree on a common name. 
@Mallets

zenoh/src/publication.rs
p-avital: With the mutable accessor, I wanted to give the user the opportunity to straight up remove the attachments, so a reference to the whole option is needed

zenoh/src/query.rs
p-avital: same as above

p-avital:
> More thoughts. Based on the name of the branch and for general reasons (we have en attachment which has map of key value pairs), I thought it would be correct to name the feature as an `attachment` not `attachments` and in my [PR](https://github.com/eclipse-zenoh/zenoh-c/pull/203/files#diff-00776be53d5f126d657ea68a085286124f2255c73d98df956676e65861abde16R29) I already did it. But there still some inconsistency between `attachment` not `attachments`. I think we need to agree on a common name. @Mallets

@Mallets agrees with you, I named the type `Attachments` because I view it as a collection of attachments, each attachment being named (so essentially, I call each value in the key-value pairs an attachment). I'm fine with the plural form being eliminated for consistency.

sashacmc:
> @Mallets agrees with you, I named the type `Attachments` because I view it as a collection of attachments, each attachment being named (so essentially, I call each value in the key-value pairs an attachment). I'm fine with the plural form being eliminated for consistency.

but what confuses me is that in rust we will have `with_attachments`/`attachments` methods but in the same time for the same functionality in C/C++ we will have `as_attachment` / `set_attachment` / `attachment` etc.

p-avital:
> but what confuses me is that in rust we will have `with_attachments`/`attachments` methods but in the same time for the same functionality in C/C++ we will have `as_attachment` / `set_attachment` / `attachment` etc.

In Rust, it's common practice to use `with_X(self, x) -> Self` in builder patterns; `set_X(&mut self, x: X)-> Option<X>` being used for setters. Adding a setter could be done, but this role is mostly fulfilled by `attachments_mut` providing mutable access.

examples/examples/z_pub_thr.rs
```diff
@@ -40,10 +53,28 @@ fn main() {
     let mut count: usize = 0;
     let mut start = std::time::Instant::now();
     loop {
-        publisher.put(data.clone()).res().unwrap();
+        let attachments = (args.attachments_number != 0).then(|| {
```
Mallets, line 56: I suggest to remove the `attachments` from the `z_pub_thr` and to have it only on the `z_pub/z_sub`.

Mallets: `z_sub` example should also show how to use the attachment.

zenoh/src/publication.rs
```diff
@@ -117,6 +121,22 @@ impl PutBuilder<'_, '_> {
         self.kind = kind;
         self
     }
+
+    #[zenoh_macros::unstable]
+    pub fn attachment(&self) -> Option<&Attachment> {
```
Mallets, line 126: Looking at this more closely. What's the interest of `attachment` and `attachment_mut` in the `PutBuilder`? Shouldn't `with_attachment` be enough? It's a builder, so I would expect to be additive for the attachment...

zenoh/src/query.rs
```diff
@@ -294,6 +306,22 @@ impl<'a, 'b, Handler> GetBuilder<'a, 'b, Handler> {
         self
     }
 
+    #[zenoh_macros::unstable]
+    pub fn attachment(&self) -> Option<&Attachment> {
```
Mallets, line 310: Similar comment as per other builders. What's the interest of attachment and attachment_mut in the PutBuilder? Shouldn't with_attachment be enough? It's a builder, so I would expect to be additive for the attachment...

zenoh/src/queryable.rs
```diff
@@ -150,6 +160,36 @@ pub struct ReplyBuilder<'a> {
     result: Result<Sample, Value>,
 }
 
+impl<'a> ReplyBuilder<'a> {
+    #[zenoh_macros::unstable]
+    pub fn attachment(&self) -> Option<&Attachment> {
```
Mallets, line 165: Similar comment as in the other builders... What's the interest of attachment and attachment_mut in the PutBuilder? Shouldn't with_attachment be enough? It's a builder, so I would expect to be additive for the attachment...

examples/examples/z_pub_thr.rs
sashacmc: @Mallets in [zenoh-c](https://github.com/eclipse-zenoh/zenoh-c/pull/203/files) I decided to move attachments to separate sample to don't overload main simple z_pub/z_sub samples.
Maybe we should do the same here too?

Mallets: I'm fine to move it to separate examples.

zenoh/src/publication.rs
p-avital: Agreed for builders

zenoh/src/queryable.rs
p-avital: Agreed, and this spot is especially painful

Mallets (approved):

PR #619, opened 2023-12-11
Fix latency regression

Mallets:
This PR fix a latency regression introduced by #585 .
It removes an additional syscall that was introduced that caused `3us` additional latency.


Mallets:
@yellowhatter could you please review this PR?

yellowhatter (approved):
All good!

Issue #586, opened 2023-11-13
[Bug] TransportExecutor should be used in multicast transport


YuanYuYuan:
### Describe the bug

As titled.

### To reproduce

Skip

### System info

Skip

yellowhatter:
@YuanYuYuan fixed in #626

PR #626, opened 2023-12-15
Use TX executor for multicast TX

yellowhatter:
for #586


yellowhatter:
@Mallets please review

Mallets (approved):

PR #635, opened 2023-12-26
Fix cli examples

Jarsop:
Some mistakes are present in CLI examples, this PR attempts to fix it.

BTW thanks for your awesome work ! :crown:

### DIFF
--- a/plugins/zenoh-plugin-rest/examples/z_serve_sse.rs
+++ b/plugins/zenoh-plugin-rest/examples/z_serve_sse.rs
@@ -94,7 +94,7 @@ fn parse_args() -> Config {
         .arg(arg!(-l --listen [ENDPOINT]...  "Endpoints to listen on."))
         .arg(arg!(-c --config [FILE] "A configuration file."))
         .arg(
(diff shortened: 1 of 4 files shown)


kydos (approved):
Looks good, thanks for your contribution!

fuzzypixelz:
Hi! Good catch :)

You seems to have missed a few incorrect downcasts to `&String` in `ArgMatches::get_many` and `ArgMatches::get_one` calls. You can try something like `rg 'get_.*::<&String>' .` to get all of them.

kydos:
Thanks for you PR @Jarsop.

Jarsop:
> Hi! Good catch :)
> 
> You seems to have missed a few incorrect downcasts to `&String` in `ArgMatches::get_many` and `ArgMatches::get_one` calls. You can try something like `rg 'get_.*::<&String>' .` to get all of them.

@fuzzypixelz yes I've seen other downcasts but as it's optional arguments, when you request `get_*::<&String>()` it converts to `Option<&String>` which is valid. I ran all the examples to see where there was this downcast problem and normally everything is fine but maybe I missed something :)

PR #639, opened 2024-01-03
Reduce status checks

fuzzypixelz:
In GitHub repository settings, the "Require status checks to pass before merging" branch protection rule ensures that commits are only merged from branches where specific status checks have passed. These checks are specified manually as a list of workflow job names.

To properly use the branch protection rule, we need this additional 'all encompassing' status check. Without this, GitHub won't wait for status checks when auto-merging pull requests.

Further context and a more feature-full implementation is available at: https://github.com/re-actors/alls-green. Although we don't (currently) need an 'allow skips' or 'allow failures' option so the value of adding yet another dependency is questionable.

Needless to say, this change needs to be propagated to all other `eclipse-zenoh/zenoh-*` repositories. We could add this logic in an action under `.github/actions/reduce-status-checks.yml` and use that from the other repositories. However, all we would factor out is the `jq` command (and maybe the comment/documentation). This makes for a good argument for using the existing re-actors/alls-green action.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,6 +58,15 @@ jobs:
         if: ${{ matrix.os == 'ubuntu-latest' || matrix.os == 'macOS-latest' }}
         run: cargo clippy --all-targets --all-features -- --deny warnings
 
+      - name: Install nightly Rust toolchain
+        # Generic no_std target architecture is x86_64-unknown-none
+        run: rustup target add x86_64-unknown-none
+
+      - name: Perform no_std checks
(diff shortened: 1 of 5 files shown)


fuzzypixelz:
When running tests, the `cargo install nextest` step take 5-8mins, which is nearly half the time is takes for the tests themselves to run. I switched to using [pre-built binaries](https://nexte.st/book/pre-built-binaries.html#using-nextest-in-github-actions).

fuzzypixelz:
I removed the nightly requirement for `nostd-check` as it's no longer necessary. Moreover, now the `nostd-check` binary runs on Ubuntu/Windows/macOS instead of just Ubuntu as part of the `check` job.

p-avital (approved):

PR #646, opened 2024-01-12
Sync lockfiles

fuzzypixelz:
This is an improvement upon [sync-lockfile-and-toolchain](https://github.com/ZettaScaleLabs/zenoh-sync-lockfile-and-toolchain/blob/main/README.md) which implies adding workflows to all dependent repositories to sync their lockfiles with eclipse-zenoh/zenoh. This instead makes eclipse-zenoh/zenoh itself push lockfile changes on its dependants in the form of auto-merged pull requests.


fuzzypixelz:
Once this is done, I'll add another workflow to sync the toolchain too. But instead of running nightly, it would only be triggered when `./rust-toolchain.toml` changes. But first we need to make sure all repositories have the same toolchain filename (i.e. not `./rust-toolchain` without the TOML extension).

JEnoch (approved):

PR #653, opened 2024-01-19
Add a CallbackPair handler implementing Drop and IntoCallbackReceiverPair

JEnoch:
This struct holds 2 callbacks functions:
  - `callback`: the typical callback function. `context` will be passed as its last argument.
  - `drop`: a callback called when this handler is dropped.

It can be passed to any subscriber or getter function accepting a callback.
E.g.:
 - [`GetBuilder::with(Handler)`](https://github.com/eclipse-zenoh/zenoh/blob/b669489bc814a758741f09c671ecc3a0683697a0/zenoh/src/query.rs#L236)
 - [`SubscriberBuilder::with(Handler)`](https://github.com/eclipse-zenoh/zenoh/blob/b669489bc814a758741f09c671ecc3a0683697a0/zenoh/src/subscriber.rs#L426)

### DIFF
--- a/zenoh/src/handlers.rs
+++ b/zenoh/src/handlers.rs
@@ -86,3 +86,41 @@ pub fn locked<T>(fnmut: impl FnMut(T)) -> impl Fn(T) {
     let lock = std::sync::Mutex::new(fnmut);
     move |x| zlock!(lock)(x)
 }
+
+/// A handler containing 2 callback functions:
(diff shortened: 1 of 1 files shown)


p-avital (approved):

p-avital:
@Mallets requesting your review in case you'd like to change a field name or two from the proposed ones :)

Mallets (approved):

PR #657, opened 2024-01-23
Align the name of zenoh-plugin-example and set the used unstable feature correctly

YuanYuYuan:
Users can't follow the example plugin or build zenoh-plugin-example by simply running

```bash
cargo build -p zenoh-plugin-example
```

Since the "unstable" feature is used in this example but not specified. Besides, let's align the crate name and the folder name as proposed in this PR. I've checked with the blog and no influence would occur.


YuanYuYuan:
@Mallets can you help review it?

PR #662, opened 2024-01-25
Fix matching_status_* test failures

YuanYuYuan:
Since we don't avoid the multicast port usage while testing. Sometimes the CI test on the main branch would fail due to the conflict of

```bash
cargo nextest run --exclude zenoh-examples --exclude zenoh-plugin-example --workspace -E 'test(=zenoh_session_multicast) | test(zenoh_matching_status)'
```

The same error happens frequently on the tokio branch. That's why we found it. And it's better to have `Result` to verbose the failure.

### DIFF
--- a/zenoh/tests/matching.rs
+++ b/zenoh/tests/matching.rs
@@ -29,7 +29,7 @@ macro_rules! ztimeout {
 }
 
 #[cfg(feature = "unstable")]
-async fn create_session_pair(locator: &str) -> Result<(Session, Session)> {
+async fn create_session_pair(locator: &str) -> (Session, Session) {
     let config1 = {
         let mut config = zenoh::config::peer();
         config.scouting.multicast.set_enabled(Some(false)).unwrap();
@@ -39,11 +39,11 @@ async fn create_session_pair(locator: &str) -> Result<(Session, Session)> {
             .unwrap();
         config
     };
-    let config2 = zenoh::config::client([Locator::from_str(locator)?]);
+    let config2 = zenoh::config::client([Locator::from_str(locator).unwrap()]);
 
-    let session1 = ztimeout!(zenoh::open(config1).res_async())?;
-    let session2 = ztimeout!(zenoh::open(config2).res_async())?;
-    Ok((session1, session2))
+    let session1 = ztimeout!(zenoh::open(config1).res_async()).unwrap();
+    let session2 = ztimeout!(zenoh::open(config2).res_async()).unwrap();
+    (session1, session2)
 }
 
 #[cfg(feature = "unstable")]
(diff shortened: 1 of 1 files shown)


zenoh/tests/matching.rs
```diff
@@ -26,200 +28,212 @@ macro_rules! ztimeout {
...
-
-        let session2 = ztimeout!(zenoh::open(config::peer()).res_async()).unwrap();
+        let (session1, session2) = create_session_pair("tcp/127.0.0.1:18001").await?;
 
         let publisher1 = ztimeout!(session1
             .declare_publisher("zenoh_matching_status_any_test")
             .allowed_destination(Locality::Any)
-            .res_async())
-        .unwrap();
+            .res_async())?;
```
Mallets, line 62: In case of tests it's better to use `unwrap()` because the error message will tell exactly at which line the test has panicked. If `?` is used, it's then harder to identify which is the line that fails.

zenoh/tests/matching.rs
```diff
@@ -26,200 +28,212 @@ macro_rules! ztimeout {
...
+
+    let session1 = ztimeout!(zenoh::open(config1).res_async())?;
+    let session2 = ztimeout!(zenoh::open(config2).res_async())?;
+    Ok((session1, session2))
+}
+
 #[cfg(feature = "unstable")]
 #[test]
-fn zenoh_matching_status_any() {
+fn zenoh_matching_status_any() -> Result<()> {
```
Mallets, line 51: Based on my comment below on `unwrap()`, there is no need to return a `Result`.

zenoh/tests/matching.rs
```diff
@@ -26,200 +28,212 @@ macro_rules! ztimeout {
...
+        config.scouting.multicast.set_enabled(Some(false)).unwrap();
+        config
+            .listen
+            .set_endpoints(vec![locator.clone().parse().unwrap()])
+            .unwrap();
+        config
+    };
+    let config2 = zenoh::config::client([Locator::from_str(locator)?]);
+
+    let session1 = ztimeout!(zenoh::open(config1).res_async())?;
```
Mallets, line 44: Use `unwrap()` instead of `?`.

Mallets (changes requested):
See my comments above.

zenoh/tests/matching.rs
YuanYuYuan: Okay. Then I understand why we have our customized zerror since it prints more information than common errors.

zenoh/tests/matching.rs
YuanYuYuan: Done.

zenoh/tests/matching.rs
YuanYuYuan: Done

YuanYuYuan:
@Mallets I've reverted the changes, can you help review this?

Mallets (approved):

Issue #668, opened 2024-01-25
Expose completeness option in PublicationCach in Rust, C, C++


milyin:
### Describe the feature

- [x] zenoh https://github.com/eclipse-zenoh/zenoh/pull/688
- [x] zenoh-c https://github.com/eclipse-zenoh/zenoh-c/pull/243
- [ ] zenoh-cpp

Mallets:
`PublicationCache` is deprecated in favour of https://zenoh-cpp.readthedocs.io/en/1.3.0/ext.html#classzenoh_1_1ext_1_1AdvancedPublisher

PR #688, opened 2024-01-26
'complete' option in PublicationCache, refactoring

milyin:
- added 'complete' option to query of `PublicationCache`
- refactored access to `SessionRef` object: added mthods for creating publisher, queryable, etc to it. This allowed to access the `SessionRef` without code duplication just to satisfy borrow checker
- Simplified `SessionExt` trait

Closes 'rust' checkbox of https://github.com/eclipse-zenoh/zenoh/issues/668

### DIFF
--- a/zenoh/src/session.rs
+++ b/zenoh/src/session.rs
@@ -278,7 +278,7 @@ impl Resource {
     }
 }
 
-pub trait SessionDeclarations<'s> {
+pub trait SessionDeclarations<'s, 'a> {
     /// Create a [`Subscriber`](crate::subscriber::Subscriber) for the given key expression.
     ///
     /// # Arguments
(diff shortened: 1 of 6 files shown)


OlivierHecart:
This broke The usage of `SessionExt` with `Arc<Session>`.
```rust
#[async_std::main]
async fn main() {
    use zenoh_ext::SessionExt;

    let session = zenoh::open(Config::default())
        .res()
        .await
        .unwrap()
        .into_arc();

    let publication_cache = session
        .declare_publication_cache("key/expression")
        .res()
        .await
        .unwrap();

    async_std::task::spawn(async move {
        publication_cache.key_expr();
    });
}
```

milyin:
@OlivierHecart Fixed missing implementation of `SessionExt` for  `Arc<Session>`.
Also modified `SessionDeclaration` trait: now this trait is the only source of declaration functions. Scheme is exaclty the same as in `SessionExt`
@p-avital could you please look at it too? I hope this approach to using lifetimes is ok, but it's worth to check

OlivierHecart (approved):

zenoh/src/session.rs
```diff
@@ -499,135 +562,49 @@ impl Session {
     pub fn config(&self) -> &Notifier<Config> {
         self.runtime.config()
     }
+}
```
p-avital, line 565: This is actually a sneaky breaking change: the `SessionDeclarations` trait will have to be in scope for the methods to be called on `Session`, whereas this is currently unnecessary.

While I agree this makes a lot of sense, we should probably have a quick internal chat about whether or not we want to make that break (we have other more localized breaking changes in the next release, so I'm open to it)

p-avital (commented):
The lifetime handling looks correct to me (GATs could have been another way to reach the solution, but this pattern also allows some nice variance tricks if we ever need them).

There's a sneaky breaking change by moving some of `Session`'s method to a trait. I think moving to the trait makes _a lot_ of sense, but we might also want to re-add the intrinsic methods to avoid the breaking change.

This review will be updated to approved when a decision has been made regarding this breaking change.

milyin:
Michael Ilyin — Today at 11:04 AM
Thank you! Speaking on breaking change - I don't mind to repeat declaration functions in impl Session. I have only one argument against it:

If even core API, like declare_publisher is implemented in a trait outside the Session, this makes SessionExt extension first class citizen. It's now not a some hack to add new methods to Session, but completely valid and supported way to extend it.

Pierre 'Fox' Avital — Today at 11:08 AM
I agree, in general, I'm always in favour of traitifying, although that does mean more effort is necessary on docs to help the user navigate the API, as trait implementations are generally less obvious than methods

milyin:
Talked with @Mallets - the only concern from his point of view is navigation on documentation. Created task to not forget to take care of it https://github.com/eclipse-zenoh/zenoh/issues/710. Ohterwise he also thinks that this update is ok. Will integrate it today

PR #702, opened 2024-01-31
Examples: fix overwritten mode in config file (fix #700)

### LINKED ISSUE #700: Examples using a config file with `mode: "cient"` remain in peer mode
### Describe the bug

Using such a config file:
   ```json5
   { mode: "client" }
   ```
All example are not using the client mode, but the peer mode.

### To reproduce

1. create a `conf.json5` file as such:
   ```json5
   { mode: "client" }
   ```
2. Start a Zenoh router: `zenohd`
3. Start any example with the config file: `z_sub -c conf.json5`
4. Check the bridge mode as discovered by the bridge at http://localhost:8000/@/router/local (see `"whatami"` value)

### System info

- v0.10.0-rc

JEnoch:
Will fix #700

### DIFF
--- a/examples/src/lib.rs
+++ b/examples/src/lib.rs
@@ -53,7 +53,7 @@ impl From<&CommonArgs> for Config {
             Some(Wai::Peer) => config.set_mode(Some(zenoh::scouting::WhatAmI::Peer)),
             Some(Wai::Client) => config.set_mode(Some(zenoh::scouting::WhatAmI::Client)),
             Some(Wai::Router) => config.set_mode(Some(zenoh::scouting::WhatAmI::Router)),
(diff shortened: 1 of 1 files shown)


p-avital (approved):

examples/src/lib.rs
```diff
@@ -48,10 +48,12 @@ impl From<&CommonArgs> for Config {
             Some(path) => Config::from_file(path).unwrap(),
             None => Config::default(),
         };
+        println!("ARGS mode: {:?} ", value.mode);
```
Mallets, line 51: This line should be removed since now it's printed in very examples, including throughput and latency tests.

PR #566, opened 2023-10-11
Tokio Porting

YuanYuYuan:
- [x] Examples
- [x] IO
  - [x] zenoh-tranport
    - [x] Pass all tests
  - [x] zenoh-links
    - [x] unixsock_stream
    - [x] udp
    - [x] unixpipe
    - [x] ws
    - [x] quic
    - [x] tcp
    - [x] serial
    - [x] tls
- [x] Pass top-level tests
  - [x] Debug the tests with unstable feature
- [x] ~Upgrade QUIC link to rustls 0.22.2. Currently it's blocked by https://github.com/quinn-rs/quinn/pull/1715~ Maintain an older version of rustls in QUIC link. Upgrading the TLS link only.
- [x] Plugins
- [x] Fine-tune the `ZRuntime`
- [x] Documentation

PR #708, opened 2024-01-31
Fix memory leaks

### LINKED ISSUE #695: Memory leaks cleanup
### Describe the release item

There is a series of memory leaks, mostly due to circular dependencies (for example in Runtime) that should be addressed.

DenisBiryukov91:
Addresses a series of memory leaks detected by valgrind (Fixes #695 ):
 - due to cyclic references in Runtime and its dependents
 - due to cyclic references in Session and its dependents 
 - due to cyclic references in Resource and its children
 - due to cyclic reference in Mux->Face->State->Mux
 - due to non-terminated tasks after Session.close()

Valgrind still reports some leaks, but they seem to be due to async_std (no mention of any Zenoh code, and similar "leaks" are also observed when running a simple hello world-type async example).

### DIFF
--- a/commons/zenoh-task/Cargo.toml
+++ b/commons/zenoh-task/Cargo.toml
@@ -1,0 +1,36 @@
+#
+# Copyright (c) 2023 ZettaScale Technology
+#
+# This program and the accompanying materials are made available under the
+# terms of the Eclipse Public License 2.0 which is available at
+# http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+# which is available at https://www.apache.org/licenses/LICENSE-2.0.
+#
+# SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+#
+# Contributors:
+#   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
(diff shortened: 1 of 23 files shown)


Mallets:
#566 is expected to land soon. I'd suggest to rebase this PR on #566 and to coordinate with @YuanYuYuan regarding task handling and management.

commons/zenoh-task/Cargo.toml
Mallets, line 1: An attempt of finer grain executor and task management is being proposed in #566 .
Moreover, every time a new crate is added it should be also added in the right order for publication in the [CI](https://github.com/eclipse-zenoh/zenoh/blob/main/.github/workflows/crates_publish.sh).

imstevenpmwork:
@DenisBiryukov91 What's the latest status of this PR? Also, would be great if you can add the timeline to the respective issue

DenisBiryukov91:
@imstevenpmwork This was kind of finished for the main branch at the moment of last commit, but due to transition async-std->Tokyo, task termination part would likely need to be redone, so it is postponed until #566 is merged.

PR #714, opened 2024-02-02
Protocol update: Entity ID

OlivierHecart:


OlivierHecart:
Merged here: https://github.com/eclipse-zenoh/zenoh/pull/774

PR #715, opened 2024-02-05
Make cargo-nextest more resilient

YuanYuYuan:
This modification has already been adopted in the tokio-porting  branch. But this also helps the CI checks at this moment.
- Kill those endlessly running tests
- The test sometimes fails on routing::gossip, match_status_*, transport_tcp_concurrent, etc. And a simple re-run can make these errors disappear.  Let's introduce "retry" to solve this problem automatically.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,3 +1,14 @@
+# By default, retry a few times until pass the test within the specified timeout
+[profile.default]
+retries = 4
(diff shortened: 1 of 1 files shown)


YuanYuYuan:
@Mallets , can you please help review this? Thanks.

Mallets (approved):

PR #718, opened 2024-02-05
Add interfaces to the Link

sashacmc:
Add interfaces, which correspond to the address, to the Link.
This will allow interceptors to obtain information about the interface with which the message is associated.
Additionally, a small refactoring of the module std_only/net was made.

### DIFF
--- a/io/zenoh-link-commons/src/unicast.rs
+++ b/io/zenoh-link-commons/src/unicast.rs
@@ -11,7 +11,7 @@
 // Contributors:
 //   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
 //
-use alloc::{boxed::Box, sync::Arc, vec::Vec};
+use alloc::{boxed::Box, string::String, sync::Arc, vec::Vec};
 use async_trait::async_trait;
 use core::{
     fmt,
@@ -45,6 +45,7 @@ pub trait LinkUnicastTrait: Send + Sync {
     fn get_dst(&self) -> &Locator;
     fn is_reliable(&self) -> bool;
     fn is_streamed(&self) -> bool;
+    fn get_interface_names(&self) -> Vec<String>;
     async fn write(&self, buffer: &[u8]) -> ZResult<usize>;
     async fn write_all(&self, buffer: &[u8]) -> ZResult<()>;
     async fn read(&self, buffer: &mut [u8]) -> ZResult<usize>;
--- a/io/zenoh-links/zenoh-link-quic/src/unicast.rs
+++ b/io/zenoh-links/zenoh-link-quic/src/unicast.rs
@@ -144,6 +144,13 @@ impl LinkUnicastTrait for LinkUnicastQuic {
     }
 
     #[inline(always)]
+    fn get_interface_names(&self) -> Vec<String> {
+        // @TODO: Not supported for now
+        log::debug!("The get_interface_names for LinkUnicastQuic is not supported");
+        vec![]
+    }
+
+    #[inline(always)]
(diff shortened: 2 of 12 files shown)


io/zenoh-link-commons/src/unicast.rs
```diff
@@ -45,6 +45,7 @@ pub trait LinkUnicastTrait: Send + Sync {
     fn get_dst(&self) -> &Locator;
     fn is_reliable(&self) -> bool;
     fn is_streamed(&self) -> bool;
+    fn get_interfaces(&self) -> Vec<String>;
```
Mallets, line 48: I wonder wether a more appropriate name could be `get_interfaces_names` since what we return here is the symbolic name of a network interface.
Interfaces could also return their indexes as presented by the OS, so I believe it's good to disambiguate. Opinions?

io/zenoh-links/zenoh-link-tcp/src/unicast.rs
```diff
@@ -144,6 +144,28 @@ impl LinkUnicastTrait for LinkUnicastTcp {
         *TCP_DEFAULT_MTU
     }
 
+    #[inline(always)]
+    fn get_interfaces(&self) -> Vec<String> {
+        match zenoh_util::net::get_interfaces_by_addr(self.src_addr.ip()) {
+            Ok(interfaces) => {
+                log::debug!(
```
Mallets, line 151: I would put those kinds of log in `trace` and not in `debug`.

io/zenoh-link-commons/src/unicast.rs
sashacmc: Fine, but will be better `get_interface_names`

io/zenoh-links/zenoh-link-quic/src/unicast.rs
```diff
@@ -143,6 +143,12 @@ impl LinkUnicastTrait for LinkUnicastQuic {
         *QUIC_DEFAULT_MTU
     }
 
+    #[inline(always)]
+    fn get_interfaces(&self) -> Vec<String> {
+        // Not supported for now
+        vec![]
```
Mallets, line 149: As a general comment, we start having many links implementations that rely on `IPv4/IPv6`.
I wonder wether having some shared `ip utils` or even an `IPtrait` for IP-based links would be better.

io/zenoh-links/zenoh-link-tcp/src/unicast.rs
Mallets: Same comment applies to all the other link implementations.

Mallets (changes requested):
Please address the minor comments I left regarding logging and comments.
Apart of that, it LGTM.

io/zenoh-links/zenoh-link-quic/src/unicast.rs
```diff
@@ -143,6 +143,12 @@ impl LinkUnicastTrait for LinkUnicastQuic {
         *QUIC_DEFAULT_MTU
     }
 
+    #[inline(always)]
+    fn get_interfaces(&self) -> Vec<String> {
+        // Not supported for now
```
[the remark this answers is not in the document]
Mallets: I also think it's good to log a message in `debug` saying the functionality is not yet implemented. I'm already foreseeing the user changing the protocol for connectivity and their interceptors will not work as expected.
At least a log will provide some level of feedback.

Mallets (approved):

Mallets (approved):

PR #728, opened 2024-02-09
Add the downsampling interceptor

sashacmc:
Add the interceptor to reduce amount of sent messages based on specified filters:
- Key expressions
- Interfaces

At the moment, only one strategy is implemented: `ratelimit`,
the available parameter:
-  threshold_ms - the minimum amount of time between the messages

### DIFF
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -114,6 +114,20 @@
   //      ],
   //  },
 
+  //  /// The downsampling declaration.
+  //  downsampling: [
+  //    {
+  //      /// A list of network interfaces messages will be processed on, the rest will be passed as is.
+  //      interfaces: [ "wlan0" ],
+  //      /// Data flow messages will be processed on. ("egress" or "ingress")
+  //      flow: "egress",
+  //      /// A list of downsampling rules: key_expression and the rate (maximum frequency in Hertz)
+  //      rules: [
+  //        { key_expr: "demo/example/zenoh-rs-pub", rate: 0.1 }, 
+  //      ],
+  //    },
+  //  ],
+
   /// Configure internal transport parameters
   transport: {
     unicast: {
--- a/commons/zenoh-config/src/lib.rs
+++ b/commons/zenoh-config/src/lib.rs
@@ -70,6 +70,33 @@ impl Zeroize for SecretString {
 
 pub type SecretValue = Secret<SecretString>;
 
+#[derive(Debug, Deserialize, Serialize, Clone)]
+#[serde(rename_all = "lowercase")]
+pub enum DownsamplingFlow {
+    Egress,
+    Ingress,
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone)]
+pub struct DownsamplingRuleConf {
+    /// A list of key-expressions to which the downsampling will be applied.
+    /// Downsampling will be applied for all key extensions if the parameter is None
+    pub key_expr: OwnedKeyExpr,
+    /// The maximum frequency in Hertz;
+    pub rate: f64,
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone)]
+pub struct DownsamplingItemConf {
+    /// A list of interfaces to which the downsampling will be applied
+    /// Downsampling will be applied for all interfaces if the parameter is None
+    pub interfaces: Option<Vec<String>>,
+    /// A list of interfaces to which the downsampling will be applied.
+    pub rules: Vec<DownsamplingRuleConf>,
+    /// Downsampling flow direction: egress, ingress
+    pub flow: DownsamplingFlow,
+}
+
 pub trait ConfigValidator: Send + Sync {
     fn check_config(
         &self,
@@ -405,6 +432,10 @@ validated_struct::validator! {
             },
 
         },
+
+        /// Configuration of the downsampling.
+        downsampling: Vec<DownsamplingItemConf>,
+
         /// A list of directories where plugins may be searched for if no `__path__` was specified for them.
         /// The executable's current directory will be added to the search paths.
         plugins_search_dirs: Vec<String>, // TODO (low-prio): Switch this String to a PathBuf? (applies to other paths in the config as well)
--- a/zenoh/src/net/routing/interceptor/downsampling.rs
+++ b/zenoh/src/net/routing/interceptor/downsampling.rs
@@ -1,0 +1,166 @@
+//
+// Copyright (c) 2024 ZettaScale Technology
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+// which is available at https://www.apache.org/licenses/LICENSE-2.0.
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+//
+// Contributors:
+//   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
+//
+
+//! ⚠️ WARNING ⚠️
+//!
+//! This module is intended for Zenoh's internal use.
+//!
+//! [Click here for Zenoh's documentation](../zenoh/index.html)
+
+use crate::net::routing::interceptor::*;
+use std::sync::{Arc, Mutex};
+use zenoh_config::{DownsamplingFlow, DownsamplingItemConf, DownsamplingRuleConf};
+use zenoh_core::zlock;
+use zenoh_keyexpr::keyexpr_tree::impls::KeyedSetProvider;
+use zenoh_keyexpr::keyexpr_tree::IKeyExprTreeMut;
+use zenoh_keyexpr::keyexpr_tree::{support::UnknownWildness, KeBoxTree};
+use zenoh_protocol::network::NetworkBody;
+use zenoh_result::ZResult;
+
+pub(crate) fn downsampling_interceptor_factories(
+    config: &Vec<DownsamplingItemConf>,
+) -> ZResult<Vec<InterceptorFactory>> {
+    let mut res: Vec<InterceptorFactory> = vec![];
+
+    for ds in config {
+        res.push(Box::new(DownsamplingInterceptorFactory::new(ds.clone())));
+    }
+
+    Ok(res)
+}
+
+pub struct DownsamplingInterceptorFactory {
+    interfaces: Option<Vec<String>>,
+    rules: Vec<DownsamplingRuleConf>,
+    flow: DownsamplingFlow,
+}
+
+impl DownsamplingInterceptorFactory {
+    pub fn new(conf: DownsamplingItemConf) -> Self {
+        Self {
+            interfaces: conf.interfaces,
+            rules: conf.rules,
+            flow: conf.flow,
+        }
+    }
+}
+
+impl InterceptorFactoryTrait for DownsamplingInterceptorFactory {
+    fn new_transport_unicast(
+        &self,
+        transport: &TransportUnicast,
+    ) -> (Option<IngressInterceptor>, Option<EgressInterceptor>) {
+        log::debug!("New downsampler transport unicast {:?}", transport);
+        if let Some(interfaces) = &self.interfaces {
+            log::debug!(
+                "New downsampler transport unicast config interfaces: {:?}",
+                interfaces
+            );
+            if let Ok(links) = transport.get_links() {
+                for link in links {
+                    log::debug!(
+                        "New downsampler transport unicast link interfaces: {:?}",
+                        link.interfaces
+                    );
+                    if !link.interfaces.iter().any(|x| interfaces.contains(x)) {
+                        return (None, None);
+                    }
+                }
+            }
+        };
+
+        match self.flow {
+            DownsamplingFlow::Ingress => (
+                Some(Box::new(DownsamplingInterceptor::new(self.rules.clone()))),
+                None,
+            ),
+            DownsamplingFlow::Egress => (
+                None,
+                Some(Box::new(DownsamplingInterceptor::new(self.rules.clone()))),
+            ),
+        }
+    }
+
+    fn new_transport_multicast(
+        &self,
+        _transport: &TransportMulticast,
+    ) -> Option<EgressInterceptor> {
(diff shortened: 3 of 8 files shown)


zenoh/src/net/routing/interceptor/downsampling.rs
```diff
@@ -0,0 +1,160 @@
...
+
+impl EgressMsgDownsamplerRatelimit {
+    pub fn new(conf: DownsamplingItemConf) -> Self {
+        if let Some(threshold_ms) = conf.threshold_ms {
+            let threshold = std::time::Duration::from_millis(threshold_ms);
+            Self {
+                keyexprs: conf.keyexprs,
+                threshold,
+                // TODO (sashacmc): I need just := 0, but how???
+                latest_message_timestamp: Arc::new(Mutex::new(
```
sashacmc, line 86: I need some value in the past, e.g. zero, but I don't know how better do it in rust.
Any idea?

zenoh/src/net/routing/interceptor/downsampling.rs
```diff
@@ -0,0 +1,160 @@
...
+                    std::time::Instant::now() - threshold,
+                )),
+            }
+        } else {
+            // TODO (sashacmc): how correctly process an error?
+            panic!("Rate limit downsampler shoud have a threshold_ms parameter");
```
sashacmc, line 92: How better process error from the interceptor?
Do we have some policies for it?

zenoh/src/net/routing/mod.rs
```diff
@@ -168,4 +169,18 @@ impl RoutingContext<NetworkMessage> {
         }
         None
     }
+
+    #[inline]
+    pub(crate) fn full_key_expr(&self) -> Option<OwnedKeyExpr> {
```
sashacmc, line 174: I am not sure that it is most optimal way to generate KeyExpr.
Maybe during the message processing we have better time to create and cache it?

commons/zenoh-config/src/lib.rs
```diff
@@ -70,6 +70,18 @@ impl Zeroize for SecretString {
 
 pub type SecretValue = Secret<SecretString>;
 
+#[derive(Debug, Deserialize, Serialize, Clone)]
```
sashacmc, line 73: Do we have some tool for generate DEFAULT_CONFIG.json5 file or should sync/modify it manually?

Mallets: The sync is done manually for the time bein.

zenoh/src/net/routing/interceptor/downsampling.rs
Mallets: It's not clear to me what you are trying to achieve there... can you provide an example?

zenoh/src/net/routing/interceptor/downsampling.rs
Mallets: In general, instead of returning `Self` you should return a `ZResult<Self>`, by doing so you can return an error in case of wrong parameters.

For this specific case, I wonder why `ratelimit` is an `Option` instead of being mandatory. Maybe we should separate the parsing of the config from the initialization of the object.

zenoh/src/net/routing/mod.rs
Mallets: It could be a future optimization yes.

zenoh/src/net/routing/interceptor/downsampling.rs
OlivierHecart: Reading this: https://github.com/rust-lang/rust/issues/40910, it seems there isn't much better solution using the `Instant` type.

zenoh/src/net/routing/interceptor/downsampling.rs
OlivierHecart: As many checks as possible should be done at the factory instantiation.
Ultimately the `interceptor_factories` function should indeed return a `ZResult` as well as all the stack leading to it so that we might return an error to `zenoh::open`.

zenoh/src/net/routing/interceptor/downsampling.rs
sashacmc: @Mallets
I need to filter out the messages that arrive while the time elapsed since the last unfiltered one is less than a specified interval, and I do not want to filter out the first message.

The simplest way (in order not to do unnecessary checks every time) is to set an initial time value that will lag behind the current moment by obviously more than the specified interval.

For C and posix it will looks like
```
latest_timestamp = 0 // == 1970.01.01

void event_hadler(message) {
  timestamp = now();
  if (timestamp - latest_timestamp < threshold) {
    // skip message
  } else {
    latest_timestamp = timestamp;
    // process message
  }  
}
```

zenoh/src/net/routing/interceptor/downsampling.rs
```diff
@@ -0,0 +1,160 @@
...
+                    }
+                }
+            }
+            if !matched {
+                return Some(ctx);
+            }
+        }
+
+        let timestamp = std::time::Instant::now();
+        let mut latest_message_timestamp = self.latest_message_timestamp.lock().unwrap();
```
Mallets, line 65: In the rest of the code we use `zlock!` macro.

zenoh/src/net/routing/interceptor/downsampling.rs
```diff
@@ -0,0 +1,160 @@
...
+//                the rest will be passed as is.
+//   interfaces - optional, list of interfaces messages will be processed on,
+//                the rest will be passed as is.
+//
+// TODO: we can also add decimation strategy, with mandatory "factor" parameter
+
+pub(crate) struct EgressMsgDownsamplerRatelimit {
+    keyexprs: Option<Vec<OwnedKeyExpr>>,
+    threshold: std::time::Duration,
+    latest_message_timestamp: Arc<Mutex<std::time::Instant>>,
```
Mallets, line 42: I wonder wether we could use [AtomicInstant](https://docs.rs/atomic-instant/latest/atomic_instant/struct.AtomicInstant.html) in this case.

p-avital: Or we could use [this one](https://docs.rs/async-liveliness-monitor/0.1.1/async_liveliness_monitor/support/struct.AtomicInstant.html), it has way better resolution (Me? Shimming my own impls? Whatever do you mean?)

zenoh/src/net/routing/interceptor/downsampling.rs
```diff
@@ -0,0 +1,160 @@
...
+                return Some(ctx);
+            }
+        }
+
+        let timestamp = std::time::Instant::now();
+        let mut latest_message_timestamp = self.latest_message_timestamp.lock().unwrap();
+
+        if timestamp - *latest_message_timestamp >= self.threshold {
+            *latest_message_timestamp = timestamp;
+            log::debug!("Interceptor: Passed threshold, passing.");
```
Mallets, line 69: I'd move the log to `trace` level.

zenoh/src/net/routing/interceptor/downsampling.rs
```diff
@@ -0,0 +1,160 @@
...
+            Some(ctx)
+        } else {
+            log::debug!("Interceptor: Skipped due to threshold.");
```
Mallets, line 72: I'd move the log to `trace` level.

zenoh/src/net/routing/interceptor/downsampling.rs
```diff
@@ -0,0 +1,160 @@
...
+
+        if strategy == RATELIMIT_STRATEGY {
+            (
+                None,
+                Some(Box::new(EgressMsgDownsamplerRatelimit::new(
+                    self.conf.clone(),
+                ))),
+            )
+        } else {
+            panic!("Unsupported downsampling strategy: {}", strategy)
```
Mallets, line 146: I think we could log in debug and return `(None, None)`.

zenoh/src/net/routing/mod.rs
```diff
@@ -168,4 +169,18 @@ impl RoutingContext<NetworkMessage> {
...
+        match self.full_expr() {
+            Some(full_expr) => {
+                if let Ok(keyexpr) = OwnedKeyExpr::new(full_expr) {
+                    Some(keyexpr)
+                } else {
+                    None
+                }
+            }
+            None => None,
+        }
```
Mallets, line 184: To improve readability you may want to look into [?](https://doc.rust-lang.org/rust-by-example/std/result/question_mark.html) and [ok](https://doc.rust-lang.org/std/result/enum.Result.html#method.ok).

Code would then be reduced to:
```rust
let full_expr = self.full_expr()?;
OwnedKeyExpr::new(full_expr).ok()
```

zenoh/tests/interceptors.rs
```diff
@@ -0,0 +1,216 @@
...
+        }
+    }
+
+    fn tick(&mut self) {
+        let curr_time = std::time::Instant::now();
+        if self.first_tick {
+            self.first_tick = false;
+        } else {
+            self.total_time += curr_time - self.last_time;
+            self.count += 1;
```
Mallets, line 26: Is `IntervalCounter` be used in the code beside the tests?
`IntervalCounter` does not seem to provide a method to reset it and increment on `total_time` and `count` would lead to some overflow. As a side node, [wrapping_add](https://doc.rust-lang.org/std/primitive.usize.html#method.wrapping_add) could be used to deal with overflows.

zenoh/src/net/routing/interceptor/downsampling.rs
```diff
@@ -0,0 +1,160 @@
...
+//! This module is intended for Zenoh's internal use.
+//!
+//! [Click here for Zenoh's documentation](../zenoh/index.html)
+
+use crate::net::routing::interceptor::*;
+use std::sync::{Arc, Mutex};
+use zenoh_config::DownsamplingItemConf;
+use zenoh_protocol::core::key_expr::OwnedKeyExpr;
+
+const RATELIMIT_STRATEGY: &str = "ratelimit";
```
Mallets, line 26: Strings are often used in the config, but as general approach during config parsing we generate the corresponding `enum` in such a way no `String` comparison needs to be done in the internal code.

zenoh/src/net/routing/interceptor/downsampling.rs
sashacmc: I understand that I am poisoned by C/C++, but adding a new dependency for the sake of such a small improvement.. How normal is this?
(It seems that none of these crates have been used in zenoh yet)

zenoh/tests/interceptors.rs
sashacmc: I have no any intention to use it outside of the tests, this class was made only to improve the readability of tests and make it easier to write new ones.
Thanks for the wrapping_add, I'll know about it.

zenoh/src/net/routing/interceptor/downsampling.rs
sashacmc: Thanks, but it was there only for testing, I will remove it (my bad, forgot to leave TODO message)

zenoh/src/net/routing/interceptor/downsampling.rs
sashacmc: Debug? Maybe better error or at least warning? It seems to me that this is how we will "hide" the configuration error.

zenoh/src/net/routing/interceptor/downsampling.rs
sashacmc: Super! On some of my past projects, the introduction of a new dependency required approval from the CTO and the legal department 😆

zenoh/src/net/routing/interceptor/downsampling.rs
sashacmc: `ratelimit` is optional because it depends on downsampling strategy for the ratelimit strategy it is mandatory, for other may be not.

I cannot agree that all checks should be done in interceptor_factories; on the contrary, I would prefer to encapsulate all checks inside specific interceptors.
But I get your point with the error processing and I will rework configuration processing, thanks!

DEFAULT_CONFIG.json5
```diff
@@ -114,6 +114,27 @@
...
+  //          keyexprs: [
+  //            // key_expression
+  //          ],
+  //          // A list of network interfaces messages will be processed on, the rest will be passed as is.
+  //          interfaces: [
+  //            // network interface name
+  //          ],
+  //          // Threshold indicates the minimum interval between messages.
+  //          // All messages arriving with a smaller interval will be skipped
+  //          threshold_ms: 1000,
```
Mallets, line 133: I believe configuration should be exposed in a similar way as done in the [ros2dds bridge](https://github.com/eclipse-zenoh/zenoh-plugin-ros2dds/blob/fdf9a1af4f370da05f3a5c13225b296813c00339/DEFAULT_CONFIG.json5#L81), preferring `Hz` versus `time interval`.

From a configuration point of view, I'd prefer to have a set of key-expressions associated to an interface. With the proposed configuration format the same downsampling will be applied to all the interfaces. 

Moreover, with the proposed configuration format the same downsampling will be applied to all key-expressions while it should be specifiable per each key separately.

commons/zenoh-config/src/defaults.rs
```diff
@@ -211,3 +211,14 @@ impl Default for SharedMemoryConf {
         Self { enabled: false }
     }
 }
+
+impl Default for DownsamplingItemConf {
+    fn default() -> Self {
+        Self {
+            keyexprs: None,
+            interfaces: None,
+            strategy: Some(DownsamplingStrategy::Ratelimit),
```
Mallets, line 220: Should the `strategy` be `None` as well?

commons/zenoh-config/src/lib.rs
```diff
@@ -70,6 +70,24 @@ impl Zeroize for SecretString {
...
+#[serde(rename_all = "lowercase")]
+pub enum DownsamplingStrategy {
+    Ratelimit,
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone)]
+pub struct DownsamplingItemConf {
```
Mallets, line 80: See my previous comments on configuration. It may impact on how `DownsamplingItemConf` is defined.

Mallets (approved):

PR #730, opened 2024-02-09
Expose Quality of Service settings (Priority, Congestion Control, Express) the Sample was published with

### LINKED ISSUE #694: Priority in samples
### Describe the feature

For mapping UAttributes onto Zenoh. It should be added on all languages

DenisBiryukov91:
Expose Quality of Service settings (Priority, Congestion Control, Express flag) the sample was sent with (closes #694)

### DIFF
--- a/commons/zenoh-protocol/src/core/mod.rs
+++ b/commons/zenoh-protocol/src/core/mod.rs
@@ -431,11 +431,3 @@ pub enum QueryTarget {
     #[cfg(feature = "complete_n")]
     Complete(u64),
 }
-
-/// Structure containing quality of service data
-#[derive(Debug, Default, Copy, Clone, Eq, PartialEq)]
-pub struct QoS {
-    pub priority: Priority,
-    pub congestion_control: CongestionControl,
-    pub express: bool,
-}
--- a/commons/zenoh-protocol/src/core/mod.rs
+++ b/commons/zenoh-protocol/src/core/mod.rs
@@ -431,3 +431,11 @@ pub enum QueryTarget {
     #[cfg(feature = "complete_n")]
     Complete(u64),
 }
+
+/// Structure containing quality of service data
+#[derive(Debug, Default, Copy, Clone, Eq, PartialEq)]
+pub struct QoS {
+    pub priority: Priority,
+    pub congestion_control: CongestionControl,
+    pub express: bool,
+}
--- a/commons/zenoh-protocol/src/network/mod.rs
+++ b/commons/zenoh-protocol/src/network/mod.rs
@@ -263,6 +263,13 @@ pub mod ext {
             }
         }
 
+        pub fn set_is_express(&mut self, is_express: bool) {
+            match is_express {
+                true => self.inner = imsg::set_flag(self.inner, Self::E_FLAG),
+                false => self.inner = imsg::unset_flag(self.inner, Self::E_FLAG),
+            }
+        }
+
         pub const fn is_express(&self) -> bool {
             imsg::has_flag(self.inner, Self::E_FLAG)
         }
--- a/commons/zenoh-protocol/src/network/mod.rs
+++ b/commons/zenoh-protocol/src/network/mod.rs
@@ -331,6 +331,17 @@ pub mod ext {
         }
     }
 
+    type QoSPublic = crate::core::QoS;
+    impl<const ID: u8> From<QoSType<{ ID }>> for QoSPublic {
+        fn from(ext: QoSType<{ ID }>) -> Self {
+            QoSPublic {
+                priority: ext.get_priority(),
+                congestion_control: ext.get_congestion_control(),
+                express: ext.is_express(),
+            }
+        }
+    }
+
     /// ```text
     ///  7 6 5 4 3 2 1 0
     /// +-+-+-+-+-+-+-+-+
--- a/commons/zenoh-protocol/src/network/mod.rs
+++ b/commons/zenoh-protocol/src/network/mod.rs
@@ -331,17 +331,6 @@ pub mod ext {
         }
     }
 
-    type QoSPublic = crate::core::QoS;
-    impl<const ID: u8> From<QoSType<{ ID }>> for QoSPublic {
-        fn from(ext: QoSType<{ ID }>) -> Self {
-            QoSPublic {
-                priority: ext.get_priority(),
-                congestion_control: ext.get_congestion_control(),
-                express: ext.is_express(),
-            }
-        }
-    }
-
     /// ```text
     ///  7 6 5 4 3 2 1 0
     /// +-+-+-+-+-+-+-+-+
--- a/zenoh/src/publication.rs
+++ b/zenoh/src/publication.rs
@@ -859,12 +859,12 @@ fn resolve_put(
             timestamp,
             source_id: None,
             source_sn: None,
-            qos: ext::QoSType::new (
+            qos: ext::QoSType::new(
                 publisher.priority.into(),
                 publisher.congestion_control,
                 false,
-                )
-                .into(),
+            )
+            .into(),
         };
 
         publisher.session.handle_data(
--- a/zenoh/src/publication.rs
+++ b/zenoh/src/publication.rs
@@ -902,6 +902,12 @@ impl Priority {
     pub const NUM: usize = 1 + Self::MIN as usize - Self::MAX as usize;
 }
 
+impl From<zenoh_protocol::core::Priority> for Priority {
+    fn from(priority: zenoh_protocol::core::Priority) -> Self {
+        Self::try_from(priority as u8).unwrap()
+    }
+}
+
 impl TryFrom<u8> for Priority {
     type Error = zenoh_result::Error;
 
--- a/zenoh/src/publication.rs
+++ b/zenoh/src/publication.rs
@@ -860,7 +860,7 @@ fn resolve_put(
             timestamp,
             source_id: None,
             source_sn: None,
-            qos: QoS::from_or_default(ext::QoSType::new(
+            qos: QoS::from(ext::QoSType::new(
                 publisher.priority.into(),
                 publisher.congestion_control,
                 false,
--- a/zenoh/src/publication.rs
+++ b/zenoh/src/publication.rs
@@ -22,6 +22,7 @@ use crate::prelude::*;
 #[zenoh_macros::unstable]
 use crate::sample::Attachment;
 use crate::sample::DataInfo;
+use crate::sample::QoS;
 use crate::Encoding;
 use crate::SessionRef;
 use crate::Undeclarable;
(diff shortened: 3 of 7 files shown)


zenoh/tests/qos.rs
```diff
@@ -0,0 +1,45 @@
...
+            .priority(Priority::DataHigh)
+            .congestion_control(CongestionControl::Drop);
+
+        task::sleep(SLEEP).await;
+
+        let sub = session2.declare_subscriber("test/qos").res().await.unwrap();
+        task::sleep(SLEEP).await;
+
+        publisher.put("qos").res_async().await.unwrap();
+        let qos = sub.recv_async().await.unwrap().qos;
```
Mallets, line 31: It's better to wait with a timeout with e.g. a [ztimeout!()](https://github.com/eclipse-zenoh/zenoh/blob/84dc04acdcf794e71667e0e76e521f41b921eef8/io/zenoh-transport/tests/unicast_transport.rs#L228) macro.

zenoh/tests/qos.rs
```diff
@@ -0,0 +1,45 @@
+use std::time::Duration;
+
+use async_std::task;
+use zenoh::{publication::Priority, SessionDeclarations};
+use zenoh_core::{zasync_executor_init, AsyncResolve};
+use zenoh_protocol::core::CongestionControl;
```
Mallets, line 6: I believe we should only use imports from `zenoh` (including `zenoh::prelude`) in the tests.
The only exception should be `zenoh_core::zasync_executor_init`.

zenoh/tests/qos.rs
```diff
@@ -0,0 +1,45 @@
...
+        assert_eq!(qos.priority, Priority::DataHigh.into());
+        assert_eq!(qos.congestion_control, CongestionControl::Drop);
+
+        let publisher = publisher
+            .priority(Priority::DataLow)
+            .congestion_control(CongestionControl::Block);
+        publisher.put("qos").res_async().await.unwrap();
+        let qos = sub.recv_async().await.unwrap().qos;
+
+        assert_eq!(qos.priority, Priority::DataLow.into());
```
Mallets, line 42: At API level there shouldn't be any `.into()` necessary for comparison. That means that we are exposing two different types and it shouldn't be the case.

zenoh/tests/qos.rs
```diff
@@ -0,0 +1,45 @@
...
+
+        task::sleep(SLEEP).await;
+
+        let sub = session2.declare_subscriber("test/qos").res().await.unwrap();
+        task::sleep(SLEEP).await;
+
+        publisher.put("qos").res_async().await.unwrap();
+        let qos = sub.recv_async().await.unwrap().qos;
+
+        assert_eq!(qos.priority, Priority::DataHigh.into());
```
Mallets, line 33: At API level there shouldn't be any .into() necessary for comparison. That means that we are exposing two different types and it shouldn't be the case.

Mallets (changes requested):

zenoh/tests/qos.rs
```diff
@@ -0,0 +1,45 @@
...
+        zasync_executor_init!();
+        let session1 = zenoh::open(zenoh_config::peer()).res_async().await.unwrap();
+        let session2 = zenoh::open(zenoh_config::peer()).res_async().await.unwrap();
+
+        let publisher = session1
+            .declare_publisher("test/qos")
+            .res()
+            .await
+            .unwrap()
+            .priority(Priority::DataHigh)
```
Mallets, line 22: I find this very weird. 
`Priority` and `CongestionControl` should be set before calling `.res()` and not after.
I wonder if we should really allow that...

DenisBiryukov91: fixed

zenoh/tests/qos.rs
DenisBiryukov91: fixed

zenoh/tests/qos.rs
DenisBiryukov91: fixed

zenoh/tests/qos.rs
DenisBiryukov91: fixed

zenoh/tests/qos.rs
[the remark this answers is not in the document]
DenisBiryukov91: fixed

commons/zenoh-protocol/src/network/mod.rs
```diff
@@ -331,6 +331,17 @@ pub mod ext {
         }
     }
 
+    type QoSPublic = crate::core::QoS;
```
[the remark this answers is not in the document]
DenisBiryukov91: fixed

commons/zenoh-protocol/src/core/mod.rs
```diff
@@ -431,3 +431,11 @@ pub enum QueryTarget {
     #[cfg(feature = "complete_n")]
     Complete(u64),
 }
+
+/// Structure containing quality of service data
```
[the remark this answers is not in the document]
DenisBiryukov91: fixed

zenoh/src/publication.rs
```diff
@@ -902,6 +902,12 @@ impl Priority {
     pub const NUM: usize = 1 + Self::MIN as usize - Self::MAX as usize;
 }
 
+impl From<zenoh_protocol::core::Priority> for Priority {
```
Mallets, line 905: This is not safe. `zenoh_protocol::core::Priority::Control` does not exist in `Priority`, so the mapping will not be possible in that case. It should be a `TryFrom`.

DenisBiryukov91: fixed

zenoh/src/sample.rs
```diff
@@ -527,12 +527,26 @@ pub struct QoS {
     pub express: bool,
 }
 
-impl From<QoSType> for QoS {
-    fn from(ext: QoSType) -> Self {
-        QoS {
-            priority: ext.get_priority().into(),
+impl QoS {
+    /// Helper function to fallback to default QoS value and log a error in case of conversion failure
+    pub(crate) fn from_or_default(ext: QoSType) -> QoS {
```
Mallets, line 532: The only faillable conversion is `Priority`.
With current implementation the whole `QoS` is falling back to default while only the `Priority` field should be resetted to default. `congestion_control` and `express` fields shouldn't be changed.

DenisBiryukov91: fixed

zenoh/src/sample.rs
```diff
@@ -528,14 +528,19 @@ pub struct QoS {
...
+    /// Helper function to fallback to QoS with default priortiy value, in case we fail to extract it
     pub(crate) fn from_or_default(ext: QoSType) -> QoS {
-        match QoS::try_from(ext) {
-            Ok(qos) => return qos,
+        let priority = match Priority::try_from(ext.get_priority()) {
+            Ok(p) => p,
             Err(e) => {
-                log::error!("Failed to convert: {}", e.to_string());
-                QoS::default()
+                log::error!("Failed to convert priority: {}", e.to_string());
```
Mallets, line 536: I would put logging level to `trace` in this case. `error` may pollute too much.

DenisBiryukov91: Given that we can fail priority conversion only if something really wrong and unlikely happened, would not it make more sense to keep it as error ?

DenisBiryukov91: fixed

Mallets:
It LGTM. However, given the recent discussion on `Sample` and `SourceInfo` I'd like to have the feedback also from @p-avital and @milyin .

Mallets (approved):

milyin (approved):

Mallets:
Holding on in merging this PR since it will break the memory layout for zenoh-c and zenoh-cpp at least.
Sister PRs in zenoh-c and zenoh-cpp should be done pointing to this PR.

zenoh/src/sample.rs
```diff
@@ -326,6 +329,8 @@ pub struct Sample {
     pub kind: SampleKind,
     /// The [`Timestamp`] of this Sample.
     pub timestamp: Option<Timestamp>,
+    /// Quality of service settings this sample was sent with.
+    pub qos: QoS,
```
p-avital, line 333: I'm not a huge fan of how `Sample` keeps on getting more public fields. Maybe we can use the API break to make it a bit more private?

zenoh/src/sample.rs
```diff
@@ -509,3 +518,43 @@ impl TryFrom<Reply> for Sample {
         value.sample
     }
 }
+
+/// Structure containing quality of service data
+#[derive(Debug, Default, Copy, Clone, Eq, PartialEq)]
+pub struct QoS {
```
p-avital, line 524: Using accessors, this could be a bitfield and occupy a single byte instead of 3

p-avital (commented):
I think it might be time (and our last chance) to move `Sample` to accessors.

p-avital (changes requested):
@Mallets @OlivierHecart can you consider my proposal above of moving `Sample` to an accessor pattern?

zenoh/src/sample.rs
DenisBiryukov91: What is the benefit of reducing 3  bytes to one on the system with word size >= 4 bytes ? As a separate struct it will still tend to occupy 4 bytes at least, no ?

p-avital: It's all about breakpoints and, to be fair, I think this layout just barely fits in `Sample`'s current padding, but that also means the next field is guaranteed to make `Sample` fatter, which we've shown in previous layout improvements to be rather performances sensitive for us.

API-wise, it's tempting to have only public fields, but that turns any future change to the struct's fields an API break. Builder patterns are just as convenient and give us much more flexibility in the future, both for layout optimization and for field additions.

Mallets:
> @Mallets @OlivierHecart can you consider my proposal above of moving `Sample` to an accessor pattern?

I think the proposal is in principle ok. The question is: should this PR be blocked by the proposal or should we merge it while we rework `Sample` entirely? I'm more in favour to get it in alongside the sister PRs so as to unblock some work being done in https://github.com/eclipse-uprotocol/up-client-zenoh-rust/pull/2.

p-avital:
I'm fine with either, but I'd like at least `Qos` to come out with the accessor+builder API from the get-go.

Keep in mind that if we unblock them with a first branch, they'll have to deal with the breaking change immediately after.

Mallets:
> I'm fine with either, but I'd like at least `Qos` to come out with the accessor+builder API from the get-go.
> 
> Keep in mind that if we unblock them with a first branch, they'll have to deal with the breaking change immediately after.

I'm fine with your proposal for Rust.

Similar PRs are available in:
- C https://github.com/eclipse-zenoh/zenoh-c/pull/244 https://github.com/eclipse-zenoh/zenoh-pico/pull/348
- C++ https://github.com/eclipse-zenoh/zenoh-cpp/pull/99
Let's be consistent as well with other bindings. Should we have some kind of accessors also in C and C++?

PR #734, opened 2024-02-13
Fix Release workflow

fuzzypixelz:


.github/workflows/release.yml
```diff
@@ -48,6 +48,11 @@ on:
         description: "build with specified features (comma separated, no space)"
         type: string
 
+env:
+  VERSION_CROSS: 0.2.4
+  VERSION_CARGO_DEB: 1.44.1
+  VERSION_CARGO_NEXTEST: 0.9.67
```
Mallets, line 54: If versions are hardcoded in the `.release.yml` what would the update procedure be?
I.e.: when and how those version numbers are updated?

Mallets (approved):

PR #743, opened 2024-02-16
Ensure that the clap version is set to a version no newer than 4.5.0

### LINKED ISSUE #740: Clap MSRV incompatible
### Describe the bug

Zenoh 0.10.1-rc (and main) require Rust version 1.72.0 and specify a dependency on [Clap of "4.4.11"](https://github.com/eclipse-zenoh/zenoh/blob/f09fcc2c0423f4eb367db0f864031036f447a3e5/Cargo.toml#L85), which can now resolve to Clap version "4.5" ([as of February 7th](https://crates.io/crates/clap/versions)), which now has a [minimum supported rust version of 1.74](https://github.com/clap-rs/clap/blob/a751c5fe65cd33cb09e85ff3039b4fd0182cdb6e/Cargo.toml#L17C1-L17C10). 

### To reproduce

Run cargo update and then try to build zenoh:
```console
error: package `clap v4.5.0` cannot be built because it requires rustc 1.74 or newer, while the currently active rustc version is 1.72.0Either upgrade to rustc 1.74 or newer, or use
cargo update -p clap@4.5.0 --precise ver
where `ver` is the latest version of `clap` supporting rustc 1.72.0
```

### System info

N/A

YuanYuYuan:
Ensure that the clap version is set to a version no newer than 4.5.0 to fix #740 .
Closes #740 .

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -82,7 +82,7 @@ async-std = { version = "=1.12.0", default-features = false } # Default features
 async-trait = "0.1.60"
 base64 = "0.21.4"
 bincode = "1.3.3"
-clap = { version = "4.4.11", features = ["derive"] }
+clap = { version = "4.4.*", features = ["derive"] }
 const_format = "0.2.30"
 crc = "3.0.1"
 criterion = "0.5"


YuanYuYuan:
Close this since this PR introduces a package resolve error instead. See #740 for the conclusion.

PR #745, opened 2024-02-21
build plugins with default zenoh features

milyin:
fixes the issue:
```
cargo build --bin zenohd
cargo build -p zenoh-plugin-rest
cd target/debug
./zenohd
```
fails with `rest` plugin version incompatibility because zenohd is compiled with feature "zenoh/default", but pluigns are not


Mallets:
@milyin what's the status of this PR?

milyin:
I didn't merged it on time due to check failures unrelated to fix itself. 
Verified it again, fix is still actual. Merging

PR #747, opened 2024-02-21
Encoding mapping

Mallets:
This PR is an alternative to separate `Encoging` and `Mapping`.
It's an alternative solution of #746 .


Mallets:
#746 superseeds this PR

PR #755, opened 2024-02-23
Add possibility to binding on the interface

### LINKED ISSUE #742: Add possibility to restrict a data to the interface
### Describe the feature

It would be good to have a way to restrict data to interface:
    - Listening on `tcp/0.0.0.0#iface=eth0`
    - Connecting to `tcp/1.23.4.#iface=eth0`

sashacmc:
Add possibility to listen/connect for TCP/UDP connections on interface in place of (or together with) IP address.
Interface name provided via connection parameter `iface`.
Also can be used for restrict connection for specific interface.

Samples:
 - `tcp/0.0.0.0:7447#iface=wlan0`: Listen for tcp connection on any IP at the `wlan0` interface, connections from other interfaces will be rejected.
 - `udp/192.168.0.1:7447#iface=eth0`: Listen for udp connection on specific IP at and `eth0` interface, if `eth0` interface will be configured to other IP, connections will be rejected. 
 - `udp/192.168.0.1:7447#iface=eth0`: Connect by udp with specific IP at and `eth0` interface, if `eth0` interface will be configured to other IP, connections will be rejected.

Closes #742

### DIFF
--- a/commons/zenoh-util/src/std_only/net/mod.rs
+++ b/commons/zenoh-util/src/std_only/net/mod.rs
@@ -239,6 +239,11 @@ pub fn get_local_addresses(interface: Option<String>) -> ZResult<Vec<IpAddr>> {
             let mut result = vec![];
             let mut next_iface = (buffer.as_ptr() as *mut IP_ADAPTER_ADDRESSES_LH).as_ref();
             while let Some(iface) = next_iface {
+                if let Some(interface) = interface.clone() {
+                    if ffi::pstr_to_string(iface.AdapterName) != interface {
+                        continue;
+                    }
+                }
                 let mut next_ucast_addr = iface.FirstUnicastAddress.as_ref();
                 while let Some(ucast_addr) = next_ucast_addr {
                     if let Ok(ifaddr) = ffi::win::sockaddr_to_addr(ucast_addr.Address) {
@@ -427,8 +432,7 @@ pub fn get_ipv4_ipaddrs(interface: Option<String>) -> Vec<IpAddr> {
             IpAddr::V4(a) => Some(a),
             IpAddr::V6(_) => None,
         })
-        .filter(|x| !x.is_multicast())
-        // .filter(|x| !x.is_loopback() && !x.is_multicast()) // TODO(sashacmc): Why we exclude loopback?
+        .filter(|x| !x.is_loopback() && !x.is_multicast())
         .map(IpAddr::V4)
         .collect()
 }
@@ -448,8 +452,7 @@ pub fn get_ipv6_ipaddrs(interface: Option<String>) -> Vec<IpAddr> {
             IpAddr::V6(_) => None,
         })
         .filter(|x| {
-            // !x.is_loopback() && !x.is_link_local() && !x.is_multicast() && !x.is_broadcast() // TODO(sashacmc): Why we exclude loopback?
-            !x.is_multicast() && !x.is_broadcast()
+            !x.is_loopback() && !x.is_link_local() && !x.is_multicast() && !x.is_broadcast()
         });
 
     // Get next all IPv6 addresses
--- a/commons/zenoh-util/src/std_only/net/mod.rs
+++ b/commons/zenoh-util/src/std_only/net/mod.rs
@@ -210,12 +210,19 @@ pub fn get_multicast_interfaces() -> Vec<IpAddr> {
     }
 }
 
-pub fn get_local_addresses() -> ZResult<Vec<IpAddr>> {
+pub fn get_local_addresses(interface: Option<String>) -> ZResult<Vec<IpAddr>> {
     #[cfg(unix)]
     {
         Ok(pnet_datalink::interfaces()
             .into_iter()
-            .filter(|iface| iface.is_up() && iface.is_running())
+            .filter(|iface| {
+                if let Some(interface) = interface.clone() {
+                    if iface.name != interface {
+                        return false;
+                    }
+                }
+                iface.is_up() && iface.is_running()
+            })
             .flat_map(|iface| iface.ips)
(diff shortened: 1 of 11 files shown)


commons/zenoh-util/src/std_only/net/mod.rs
```diff
@@ -210,12 +210,19 @@ pub fn get_multicast_interfaces() -> Vec<IpAddr> {
...
 
-pub fn get_local_addresses() -> ZResult<Vec<IpAddr>> {
+pub fn get_local_addresses(interface: Option<String>) -> ZResult<Vec<IpAddr>> {
     #[cfg(unix)]
     {
         Ok(pnet_datalink::interfaces()
             .into_iter()
-            .filter(|iface| iface.is_up() && iface.is_running())
+            .filter(|iface| {
+                if let Some(interface) = interface.clone() {
```
Mallets, line 219: I think the `.clone()` can be avoided and `.as_ref()` could be used instead.

commons/zenoh-util/src/std_only/net/mod.rs
```diff
@@ -232,6 +239,11 @@ pub fn get_local_addresses() -> ZResult<Vec<IpAddr>> {
             let mut result = vec![];
             let mut next_iface = (buffer.as_ptr() as *mut IP_ADAPTER_ADDRESSES_LH).as_ref();
             while let Some(iface) = next_iface {
+                if let Some(interface) = interface.clone() {
```
Mallets, line 242: I think the `.clone()` can be avoided and `.as_ref()` could be used instead.

commons/zenoh-util/src/std_only/net/mod.rs
```diff
@@ -210,12 +210,19 @@ pub fn get_multicast_interfaces() -> Vec<IpAddr> {
     }
 }
 
-pub fn get_local_addresses() -> ZResult<Vec<IpAddr>> {
+pub fn get_local_addresses(interface: Option<String>) -> ZResult<Vec<IpAddr>> {
```
Mallets, line 213: It seems this function does not need to take ownership of the `String`, maybe pass `Option<&str>`?

commons/zenoh-util/src/std_only/net/mod.rs
```diff
@@ -412,8 +424,8 @@ pub fn get_interface_names_by_addr(addr: IpAddr) -> ZResult<Vec<String>> {
     }
 }
 
-pub fn get_ipv4_ipaddrs() -> Vec<IpAddr> {
-    get_local_addresses()
+pub fn get_ipv4_ipaddrs(interface: Option<String>) -> Vec<IpAddr> {
```
Mallets, line 427: It seems this function does not need to take ownership of the `String`, maybe pass `Option<&str>`?

commons/zenoh-util/src/std_only/net/mod.rs
```diff
@@ -425,12 +437,12 @@ pub fn get_ipv4_ipaddrs() -> Vec<IpAddr> {
         .collect()
 }
 
-pub fn get_ipv6_ipaddrs() -> Vec<IpAddr> {
+pub fn get_ipv6_ipaddrs(interface: Option<String>) -> Vec<IpAddr> {
```
Mallets, line 440: It seems this function does not need to take ownership of the `String`, maybe pass `Option<&str>`?

io/zenoh-link-commons/src/listener.rs
```diff
@@ -109,12 +111,17 @@ impl ListenersUnicastIP {
         let guard = zread!(self.listeners);
         for (key, value) in guard.iter() {
             let (kip, kpt) = (key.ip(), key.port());
+            let iface = value
+                .endpoint
+                .config()
+                .get(BIND_INTERFACE)
+                .map(|s| s.to_string());
```
Mallets, line 118: I think it is possible to avoid the `map(|s| to_string())` here if `get_local_addresses(..)` takes `Option<&str>`.

io/zenoh-links/zenoh-link-tcp/src/unicast.rs
```diff
@@ -212,15 +213,23 @@ impl LinkManagerUnicastTcp {
...
+        zenoh_util::net::set_bind_to_device_tcp_stream(&stream, iface);
+
         Ok((stream, src_addr, dst_addr))
     }
 
-    async fn new_listener_inner(&self, addr: &SocketAddr) -> ZResult<(TcpListener, SocketAddr)> {
+    async fn new_listener_inner(
+        &self,
+        addr: &SocketAddr,
+        iface: &Option<String>,
```
Mallets, line 224: In general is always better to pass `Option<&T>` rather than `&Option<T>`.
You can read more about the rationale behind [here](https://www.reddit.com/r/rust/comments/wqsxk2/is_it_better_to_pass_optiont_or_optiont/).

Mallets (changes requested):
Overall it LGTM. I've just left few minor comments that would lead to fewer memory allocations.

io/zenoh-links/zenoh-link-tcp/src/unicast.rs
sashacmc: Good to know! Thanks!

Mallets (approved):

biodranik:
Cyclone DDS also supports [subnets](https://github.com/eclipse-cyclonedds/cyclonedds/blob/master/docs/manual/options.md#cycloneddsdomaingeneralinterfacesnetworkinterfaceaddress) which is very useful for dynamic IP addresses on an interface with any name, e.g. 192.168.0.0 will match 192.168.1.100 and 192.168.0.100.

Is there a plan to support this feature?

PR #764, opened 2024-02-26
Revised Encoding API and wire format

Mallets:
Reviewed Encoding API and wire format:
- [x] Encoding ID resolution is changed from `8bit` to `16bit`.
- [x] A bit in the encoding prefix is used to indicate the presence of a suffix. Encoding/decoding of prefix happens as `32bit` to account for the bitflag. Encoding ID stays `16bit`.
- [x] Define a set of constants for common encodings.
- [x] Encodings can be parsed from a string and displayed as string.
- [x] Define a pre-defined and internal mapping of known encoding to compact wire representation.
- [ ] Provide `encoding::auto` for automatic encoding selection based on the information from the type system. This feature will be provided via a different PR. https://github.com/eclipse-zenoh/zenoh/issues/812.

### DIFF
--- a/zenoh/src/encoding.rs
+++ b/zenoh/src/encoding.rs
@@ -81,36 +81,50 @@ impl Encoding {
 
     // - Primitives types supported in all Zenoh bindings
     /// Just some bytes.
+    ///
+    /// Constant alias for string: `"zenoh/bytes"`.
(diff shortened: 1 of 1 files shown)


milyin (approved):
Looks good for me

PR #770, opened 2024-02-28
Add connection retry

sashacmc:
- Introduce connection retry config for listen and connect endpoints.
 - Listener initialisation reworked to add the connection retry.
 - Connection initialization reworked to be consistent with listening where it posible
 - Some configuration refactoring (ModeDependent related code moved to separate file)

### DIFF
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -22,9 +22,31 @@
   /// For TCP/UDP on Linux, it is possible additionally specify the interface to be connected to: 
   /// E.g. tcp/192.168.0.1:7447#iface=eth0, for connect only if the IP address is reachable via the interface eth0 
   connect: {
+    /// timeout waiting for all endpoints connected (0: no retry, -1: infinite timeout)
+    /// Accepts a single value or different values for router, peer and client.
+    timeout_ms: { router: -1, peer: -1, client: 0 },
+
     endpoints: [
       // "<proto>/<address>"
     ],
+
+    /// Global connect configuration, 
+    /// Accepts a single value or different values for router, peer and client.
+    /// The configuration can also be specified for the separate endpoint
+    /// it will override the global one
+    /// E.g. tcp/192.168.0.1:7447#retry_period_init_ms=20000;retry_period_max_ms=10000"
+     
+    /// exit from application, if timeout exceed
+    exit_on_failure: { router: false, peer: false, client: true },
+    /// connect establishing retry configuration 
+    retry: { 
+      /// intial wait timeout until next connect try
+      period_init_ms: 1000,
+      /// maximum wait timeout until next connect try
+      period_max_ms: 4000,
+      /// increase factor for the next timeout until nexti connect try
+      period_increase_factor: 2,
+    },
   },
 
   /// Which endpoints to listen on. E.g. tcp/localhost:7447.
@@ -33,9 +55,31 @@
   /// For TCP/UDP on Linux, it is possible additionally specify the interface to be listened to: 
   /// E.g. tcp/0.0.0.0:7447#iface=eth0, for listen connection only on eth0
   listen: {
+    /// timeout waiting for all listen endpoints (0: no retry, -1: infinite timeout)
+    /// Accepts a single value or different values for router, peer and client.
+    timeout_ms: 0,
+
     endpoints: [
       // "<proto>/<address>"
     ],
+
+    /// Global listen configuration, 
+    /// Accepts a single value or different values for router, peer and client.
+    /// The configuration can also be specified for the separate endpoint
+    /// it will override the global one
+    /// E.g. tcp/192.168.0.1:7447#exit_on_failure=false;retry_period_max_ms=1000"
+    
+    /// exit from application, if timeout exceed
+    exit_on_failure: true,
+    /// listen retry configuration 
+    retry: { 
+      /// intial wait timeout until next try
+      period_init_ms: 1000,
+      /// maximum wait timeout until next try
+      period_max_ms: 4000,
+      /// increase factor for the next timeout until next try
+      period_increase_factor: 2,
+    },
   },
   /// Configure the scouting mechanisms and their behaviours
   scouting: {
--- a/commons/zenoh-config/src/connection_retry.rs
+++ b/commons/zenoh-config/src/connection_retry.rs
@@ -1,0 +1,200 @@
+//
+// Copyright (c) 2024 ZettaScale Technology
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+// which is available at https://www.apache.org/licenses/LICENSE-2.0.
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+//
+// Contributors:
+//   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
+//
+
+use crate::{
+    defaults::{
+        self, DEFAULT_CONNECT_EXIT_ON_FAIL, DEFAULT_CONNECT_TIMEOUT_MS,
+        DEFAULT_LISTEN_EXIT_ON_FAIL, DEFAULT_LISTEN_TIMEOUT_MS,
+    },
+    Config,
+};
+use serde::{Deserialize, Serialize};
+use zenoh_core::zparse_default;
+use zenoh_protocol::core::WhatAmI;
+
+use crate::mode_dependent::*;
+
+#[derive(Debug, Deserialize, Serialize, Clone)]
+pub struct ConnectionRetryModeDependentConf {
+    // intial wait timeout until next try
+    pub period_init_ms: Option<ModeDependentValue<i64>>,
+    // maximum wait timeout until next try
+    pub period_max_ms: Option<ModeDependentValue<i64>>,
+    // increase factor for the next timeout until next try
+    pub period_increase_factor: Option<ModeDependentValue<f64>>,
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone, PartialEq)]
+pub struct ConnectionRetryConf {
+    pub exit_on_failure: bool,
+    pub period_init_ms: i64,
+    pub period_max_ms: i64,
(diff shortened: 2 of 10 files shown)


commons/zenoh-config/src/lib.rs
```diff
@@ -97,6 +97,19 @@ pub struct DownsamplingItemConf {
...
+pub struct ConnectionRetryConf {
+    //TODO(sashacmc): add comments, and infinit value for counter
+    // ModeDependentValue:
+    // 1) only for exit_on_fail
+    // 2) for all
+    // 3) for no one (we have connect and listen config, they already separate)
+    pub count: u32,
+    pub timeout_ms: u64,
+    pub timeout_increase_factor: f32,
+    pub exit_on_fail: ModeDependentValue<bool>,
```
sashacmc, line 110: Not sure why we need ModeDependentValue, it increase complexity, but I don't see profit if we have two separate retry config for listen and connect.

zenoh/src/net/runtime/orchestrator.rs
```diff
@@ -291,24 +291,119 @@ impl Runtime {
...
+        //TODO(sashacmc):
+        // "tcp/192.168.0.1:7447#retry_count=1;retry_timeout_ms=1000;retry_timeout_increase_factor=2;retry_exit_on_fail=false"
+        // seems too long, but I don't thick someone will configure it often like it
+        // alternatives:
+        //      "tcp/192.168.0.1:7447#retry{count=1;timeout_ms=1000;timeout_increase_factor=2;exit_on_fail=false}"
+        //              but we have no support for it now
+        //
+        //      "tcp/192.168.0.1:7447#retry=1;tm=1000;incfactor=2;exit_on_fail=false"
+        //              but it not consistent with global config
+        //
```
sashacmc, line 318: Need to decide about inline config

zenoh/src/net/runtime/orchestrator.rs
```diff
@@ -291,24 +291,119 @@ impl Runtime {
...
+                }
+                async_std::task::sleep(Duration::from_millis(delay)).await;
+                delay = (delay as f32 * retry_config.timeout_increase_factor) as u64;
+            }
+            if success {
+                this.print_locators();
+            } else {
+                if *retry_config.exit_on_fail.get(this.whatami()).unwrap() {
+                    //TODO(sashacmc): how to exit, this is don't work?
+                    panic!("Unable to connect");
```
sashacmc, line 382: How correctly terminate a program from the thread? Do we already have it somewhere?

zenoh/src/net/runtime/orchestrator.rs
```diff
@@ -499,6 +594,9 @@ impl Runtime {
                     );
                 }
             }
+            //TODO(sashacmc): Why it was implelneted like it?
+            //We agree rework it to the iteration count, or we should use this logic to connections
+            //listener too
             async_std::task::sleep(delay).await;
             delay *= CONNECTION_RETRY_PERIOD_INCREASE_FACTOR;
             if delay > CONNECTION_RETRY_MAX_PERIOD {
```
sashacmc, line 602: Currently we have different retry approach?
Why it was introduced like it?

DEFAULT_CONFIG.json5
```diff
@@ -25,6 +25,24 @@
     endpoints: [
       // "<proto>/<address>"
     ],
+    //  /// Global connect retry configuration,
```
Mallets, line 28: I believe some of those parameters are applied as default parameters, therefore they should be uncommented. The idea of the `DEFAULT_CONFIG.json5` is that everything applied by default should be uncommented. I.e., starting zenoh with no configuration or with the default configuration should lead to the same behaviour.

DEFAULT_CONFIG.json5
```diff
@@ -25,6 +25,24 @@
     endpoints: [
       // "<proto>/<address>"
     ],
+    //  /// Global connect retry configuration, 
+    //  /// The retry configuration can also be specified for the separate endpoint ("retry_" + field name)
+    //  /// it will override the global one
+    //  /// E.g. tcp/192.168.0.1:7447#retry_try_timeout_ms=20000;retry_period_max_ms=10000"
+    //  retry: { 
+    //    /// global timeout for full connect cycle (currently not supported)
```
Mallets, line 33: If it is not supported, let's remove it. We can always add it later.

sashacmc: Did you mean remove the comment? Because this structure shared between connect and listen, and make no sense to split them.

commons/zenoh-config/src/connection_retry.rs
```diff
@@ -0,0 +1,176 @@
+use crate::{defaults, Config};
```
Mallets, line 1: File header missing. Please add the header with the license etc.

DEFAULT_CONFIG.json5
```diff
@@ -25,6 +25,24 @@
...
+    //    /// timeout for one connect try
+    //    try_timeout_ms: 10000,
+    //    /// intial wait timeout until next connect try
+    //    period_init_ms: 1000,
+    //    /// maximum wait timeout until next connect try
+    //    period_max_ms: 4000,
+    //    /// increase factor for the next timeout until nexti connect try
+    //    period_increase_factor: 2,
+    //    /// if global timeout exceed, exit from application (currently not supported)
+    //    fail_exit: false,
```
OlivierHecart, line 44: Both the global_timeout and fail_exit only apply to the initial connection attempt and not for the reconnection case. We need a structure that reflects this. This would also allow to configure what to do at disconnection (retry or ignore). Then should we have different periods for both cases? I'm not sure.

OlivierHecart, line 44: fail_exit have different defaults for clients, peers and routers. This should be reflected here.

zenoh/src/net/runtime/orchestrator.rs
```diff
@@ -293,24 +290,102 @@ impl Runtime {
...
-            match self.manager().add_listener(endpoint).await {
-                Ok(listener) => log::debug!("Listener added: {}", listener),
-                Err(err) => {
-                    log::error!("Unable to open listener {}: {}", listener, err);
-                    return Err(err);
+            let retry_config = self.get_listen_retry_config(&endpoint);
+            log::debug!("Try to add listener: {:?}: {:?}", endpoint, retry_config);
+            if retry_config.timeout().is_zero() {
+                // try to add listener and exit immediately without retry
+                self.add_listener(endpoint).await?
```
OlivierHecart, line 315: This should not fail if `fail_exit` is `false`.

commons/zenoh-config/src/connection_retry.rs
```diff
@@ -0,0 +1,176 @@
...
+            std::time::Duration::MAX
+        }
+    }
+
+    pub fn period(&self) -> ConnectionRetryPeriod {
+        ConnectionRetryPeriod::new(self)
+    }
+}
+
+pub struct ConnectionRetryPeriod {
```
OlivierHecart, line 91: Should this logic be implemented here ?

DEFAULT_CONFIG.json5
sashacmc: Since we would not like to change the connection behavior in this PR, I will just remove this comments

DEFAULT_CONFIG.json5
```diff
@@ -33,9 +56,31 @@
...
+    //  exit_on_failure: true,
+    //  /// listen retry configuration 
+    //  retry: { 
+    //    /// intial wait timeout until next try
+    //    period_init_ms: 3000,
+    //    /// maximum wait timeout until next try
+    //    period_max_ms: 6000,
+    //    /// increase factor for the next timeout until next try
+    //    period_increase_factor: 1.5,
+    //  },
```
OlivierHecart, line 83: This should be set to the actual defaults (if not already) and be uncommented.

DEFAULT_CONFIG.json5
```diff
@@ -22,9 +22,32 @@
...
+    //  /// connect establishing retry configuration 
+    //  retry: { 
+    //    /// intial wait timeout until next connect try
+    //    period_init_ms: 1000,
+    //    /// maximum wait timeout until next connect try
+    //    period_max_ms: 4000,
+    //    /// increase factor for the next timeout until nexti connect try
+    //    period_increase_factor: 2,
+    //  },
+    //
```
OlivierHecart, line 50: This should be set to the actual defaults (if not already) and be uncommented.

zenoh/src/net/runtime/orchestrator.rs
```diff
@@ -87,25 +85,73 @@ impl Runtime {
...
-                        CONNECTION_TIMEOUT,
-                        self.manager().open_transport_unicast(locator.clone()),
-                    )
-                    .await
-                    {
-                        Ok(Ok(_)) => return Ok(()),
-                        Ok(Err(e)) => log::warn!("Unable to connect to {}! {}", locator, e),
-                        Err(e) => log::warn!("Unable to connect to {}! {}", locator, e),
-                    }
+            _ => self.connect_peers(&peers).await,
```
OlivierHecart, line 88: Clients cannot use the same `connect_peers` function as peers. Clients by definition can only be connected to at most one remote process at any point in time.

OlivierHecart (changes requested):

zenoh/src/net/runtime/orchestrator.rs
sashacmc: Splitted connect_peers_impl to two methods, one for single_link and other for multiply_links

OlivierHecart (approved):

commons/zenoh-config/src/mode_dependend.rs
Mallets, line 1: Type: the file should be called `mode_dependent`.

commons/zenoh-config/src/lib.rs
```diff
@@ -47,6 +44,12 @@ use zenoh_protocol::{
 use zenoh_result::{bail, zerror, ZResult};
 use zenoh_util::LibLoader;
 
+pub mod mode_dependend;
```
Mallets, line 47: The module should be called `mode_dependent`.

Mallets (approved):

Mallets (approved):

PR #796, opened 2024-03-07
refactor Query.reply()

### LINKED ISSUE #785: Refactor Query.reply api
### Describe the release item

- Reply API in Python and Rust should be
    - `query.reply(key_expr, value).res()`
    - `query.reply_err(value).res()`
    - `query.reply_del(key_expr).res()` → `reply_del` is not available in the bindings, it should be provided
- Reply sink (accepting a Sample) could be implemented in Rust as a way to simplify internal implementation. It is not meant to go to the public API.

DenisBiryukov91:
refactor Query.reply() into separate methods:reply, reply_del, reply_err (and reply_sample for backwards compatibility). Closes #785 .


zenoh/src/queryable.rs
```diff
@@ -100,17 +104,96 @@ impl Query {
     pub fn attachment(&self) -> Option<&Attachment> {
         self.inner.attachment.as_ref()
     }
+    /// Sends a reply in the form of [`Sample`] to this Query.
+    ///
+    /// By default, queries only accept replies whose key expression intersects with the query's.
+    /// Unless the query has enabled disjoint replies (you can check this through [`Query::accepts_replies`]),
+    /// replying on a disjoint key expression will result in an error when resolving the reply.
+    #[inline(always)]
```
Mallets, line 112: This function should be marked as unstable and only for internal usage really.

zenoh/src/queryable.rs
```diff
@@ -181,130 +282,66 @@ impl<'a> Resolvable for ReplyBuilder<'a> {
...
-                macro_rules! ext_attachment {
-                    () => {{
-                        #[allow(unused_mut)]
-                        let mut ext_attachment = None;
+        if !self.query._accepts_any_replies().unwrap_or(false)
+            && !self.query.key_expr().intersects(&self.key_expr)
+        {
+            bail!("Attempted to reply on `{}`, which does not intersect with query `{}`, despite query only allowing replies on matching key expressions", self.key_expr, self.query.key_expr())
+        }
+        #[allow(unused_mut)]
```
Mallets, line 290: A comment justifying the `#[allow(unused_mut)]` should be provided.

zenoh/src/queryable.rs
DenisBiryukov91: It is being used by Query.reply(...) in zenoh-backend-traits, which does not require "unstable" feature of zenoh.

zenoh/src/queryable.rs
DenisBiryukov91: done

zenoh/src/queryable.rs
Mallets: Then I believe we should add the "unstable" feature to the `zenoh-backend-traits`

Mallets (approved):

PR #811, opened 2024-03-11
Implement vsock link support

sashacmc:
Implement [vsock](https://man7.org/linux/man-pages/man7/vsock.7.html) link support.
Uses [tokio-vsock](https://github.com/rust-vsock/tokio-vsock) library.


io/zenoh-links/zenoh-link-vsock/src/lib.rs
```diff
@@ -0,0 +1,90 @@
...
+    // Amount of time in microseconds to throttle the accept loop upon an error.
+    // Default set to 100 ms.
+    static ref TCP_ACCEPT_THROTTLE_TIME: u64 = 100_000;
+}
+
+pub fn get_vsock_addr(address: Address<'_>) -> ZResult<VsockAddr> {
+    let parts: Vec<&str> = address.as_str().split(':').collect();
+
+    if parts.len() != 2 {
+        zerror!("Incorrect vsock address: {:?}", address);
```
Mallets, line 64: `zerror!(...)` should be used to create an error to be returned. In this case, you could use the `bail!(...)` macro to create a `ZError` and return.

imstevenpmwork:
@sashacmc Could you please link this to the relevant issue? (if there's no issue, then can you create one with the release tag and proper timelines?)

io/zenoh-links/zenoh-link-vsock/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+zenoh-core = { workspace = true }
+zenoh-link-commons = { workspace = true }
+zenoh-protocol = { workspace = true }
+zenoh-result = { workspace = true }
+zenoh-sync = { workspace = true }
+zenoh-util = { workspace = true }
+zenoh-runtime = { workspace = true }
+
+[target.'cfg(target_os = "linux")'.dependencies]
+tokio-vsock = "0.5.0"
```
Mallets, line 42: Dependencies versions are defined at workspace level to facilitate version checking.
I suggest to define the dependency in the `Cargo.toml` of the workspace and import it from there.

io/zenoh-links/zenoh-link-vsock/Cargo.toml
Mallets, line 1: Since using `vsock` requires loading of a Linux kernel module, it would be good to add a little readme on how to do it or alternatively a link to an existing guide.

Mallets (approved):

PR #819, opened 2024-03-13
remove Publisher::write

### LINKED ISSUE #731: Write in Rust Publisher should be moved to a trait

milyin:
This API method is redundant, the `put` and `delete` are enough
Related to https://github.com/eclipse-zenoh/zenoh/pull/736
Closes #731


Mallets:
@milyin I suggest to target https://github.com/eclipse-zenoh/zenoh/tree/protocol_changes to minimise impact on #566 merge.

PR #822, opened 2024-03-13
Sample accessors

### LINKED ISSUE #817: Make Sample fields private and provide corresponding accessors to the user
### Describe the release item

Make Sample fields private and provide corresponding accessors to the user.
This is to be done to avoid exposing internal structure of Sample.

DenisBiryukov91:
make Sample fields pub(crate),
provide accessors for external users,
mark existing sample-mutating methods as hidden and unstable. 
Closes #817.


zenoh/src/sample.rs
```diff
@@ -471,19 +451,67 @@ impl Sample {
...
+    #[doc(hidden)]
+    #[zenoh_macros::unstable]
     pub fn with_timestamp(mut self, timestamp: Timestamp) -> Self {
         self.timestamp = Some(timestamp);
         self
     }
 
+    /// Gets the quality of service settings this Sample was sent with.
+    #[inline]
+    pub fn qos(&self) -> &QoS {
```
Mallets, line 504: In the `put` and `delete` builders we have explicit options to set the `Priority` and `CongestionControl`. I understand that this PR only addresses the accessor pattern and it does not modify the types exposed. I wonder if we should have `priority()` and `congestion_control()` instead of a generic `QoS`?

Mallets: Why it makes a bit more sense on the `put(...)` than on the `sample`? Given the sample contains all the information associated to the data + publication?

DenisBiryukov91: put(ke, payload).with_priority(...).with_congestion_control(...) corresponds to the phrase : "_publish_ this _payload_ on this _ke_ with specified priority and congestion control values". 
sample.priority() - corresponds to the phrase: "Give me priority value this sample _was published with_. Compare this for example with sample.encoding() - which corresponds to "Give me encoding (or format) of this sample". As you can see in the latter case (and in case of put) we do not need to add any extra words (except prepositions/pronouns) beyond the ones present in the function call. This is different for sample.priority(). I believe as long as we need to add extra words to described what a line of code does, we ought to add extra layers of indirection.

Mallets: I'd like to quickly remark that the API and accessors should be the same on all bindings and languages, I wouldn't add a level of indirection in C for example.

DenisBiryukov91: maybe we can have something like z_sample_qos_priority(sample) ?

imstevenpmwork:
@DenisBiryukov91 Can you link this to its relevant issue? If there's no yet, then you can create one using the release template if this is meant to be included in the next release. Make sure also to add the proper timelines

milyin:
Integrating this. Further API changes, like flattening qos accessors, is easier to apply over this update

PR #831, opened 2024-03-14
fix(828): ensuring valid JSON response from REST API

### LINKED ISSUE #828: REST Plugin returing invalid JSON
### Describe the release item

If an `put` with emtpy value and `application/json`  is performed via REST API, then a subsequent `query` returns a not valid JSON.

## How to reproduce

Start a zenoh router with rest plugin and a storage.
Perform the malformed `put`
```
curl --location --request PUT 'http://127.0.0.1:8000/demo/temp/text-json-test' \
--header 'Content-Type: application/json'
```

perform a `query`
```
curl --location 'http://127.0.0.1:8000/demo/temp/text-json-test'
```

The output is:
```
[
    {
        "key": "demo/temp/text-json-test",
        "value": ,
        "time": "2024-03-14T11:31:46.980485536Z/ff29c02c6c1044c6a1b565c24566b760"
    }
]
```

while according to JSON format it should be.
```
[
    {
        "key": "demo/temp/text-json-test",
        "value": null,
        "time": "2024-03-14T11:31:46.980485536Z/ff29c02c6c1044c6a1b565c24566b760"
    }
]
```

I suggest to reject empty put with `application/json` encoding, or return the correct `null` value in the query

gabrik:
This PR ensures that the response of the REST plugin is a valid JSON, Closes #828 

If a sample has `application/json` or `text/json` or `text/json5` encoding then we try to convert it to a `serde_json::Value` if we fail then can be converted to a string via the `from_utf8` function then we return its payload as JSON string, otherwise  it is encoded as base64 and return as JSON string.
If the sample has a different encoding and can be converted to a string via the `from_utf8` function then we return its payload as JSON string , otherwise it is encoded as base64 and return as JSON string.

This is an example of a get with the different cases:

```
[
    // real json
    {
        "key": "demo/temp/json",
        "value": {
            "hello": "world",
            "this": {
                "is-a": "json"
            }
        },
        "time": "2024-03-14T13:36:57.440128661Z/ff29c02c6c1044c6a1b565c24566b760"
    },
    // empty json
    {
        "key": "demo/temp/empty",
        "value": null,
        "time": "2024-03-14T13:35:31.929630711Z/ff29c02c6c1044c6a1b565c24566b760"
    },
    // text that was inserted as it was a json
    {
        "key": "demo/temp/text",
        "value": "this is a text, but i said it is a json",
        "time": "2024-03-14T13:35:49.283543717Z/ff29c02c6c1044c6a1b565c24566b760"
    },
    // binary, image in this case, that was inserted as json
    {
        "key": "demo/temp/binary",
        "value": "/9j/4AAQSkZJRgABAQAAAAAAAAD/2wBDAAMCAgICAgMCAgIDAwMDBAYEBAQEBAgGBgUGCQgKCgkICQkKDA8MCgsOCwkJDRENDg8QEBEQCgwSExIQEw8QEBD/wAALCAAgACABAREA/8QAFQABAQAAAAAAAAAAAAAAAAAAAAn/xAAUEAEAAAAAAAAAAAAAAAAAAAAA/9oACAEBAAA/AKpgAAAAP//Z",
        "time": "2024-03-14T13:41:08.952893450Z/ff29c02c6c1044c6a1b565c24566b760"
    },
    // image inserted as image
        {
        "key": "demo/temp/image2",
        "value": "/9j/4AAQSkZJRgABAQAAAAAAAAD/2wBDAAMCAgICAgMCAgIDAwMDBAYEBAQEBAgGBgUGCQgKCgkICQkKDA8MCgsOCwkJDRENDg8QEBEQCgwSExIQEw8QEBD/wAALCAAgACABAREA/8QAFQABAQAAAAAAAAAAAAAAAAAAAAn/xAAUEAEAAAAAAAAAAAAAAAAAAAAA/9oACAEBAAA/AKpgAAAAP//Z",
        "encoding": "image/jpeg",
        "time": "2024-03-14T13:46:20.142042767Z/ff29c02c6c1044c6a1b565c24566b760"
    }
]
```

### DIFF
--- a/plugins/zenoh-plugin-rest/src/lib.rs
+++ b/plugins/zenoh-plugin-rest/src/lib.rs
@@ -29,6 +29,7 @@ use std::sync::Arc;
 use tide::http::Mime;
 use tide::sse::Sender;
 use tide::{Request, Response, Server, StatusCode};
+use zenoh::payload::StringOrBase64;
 use zenoh::plugins::{RunningPluginTrait, ZenohPlugin};
 use zenoh::prelude::r#async::*;
 use zenoh::query::{QueryConsolidation, Reply};
@@ -60,9 +61,9 @@ pub fn base64_encode(data: &[u8]) -> String {
     general_purpose::STANDARD.encode(data)
 }
 
-fn payload_to_string(payload: &Payload) -> String {
-    String::from_utf8(payload.contiguous().to_vec()).unwrap_or(base64_encode(&payload.contiguous()))
-}
+// fn payload_to_string(payload: &Payload) -> String {
+//     String::from_utf8(payload.contiguous().to_vec()).unwrap_or(base64_encode(&payload.contiguous()))
+// }
 
 fn payload_to_json(payload: Payload, encoding: &Encoding) -> serde_json::Value {
     match payload.len() {
@@ -73,11 +74,12 @@ fn payload_to_json(payload: Payload, encoding: &Encoding) -> serde_json::Value {
             match encoding {
                 // If it is a JSON try to deserialize as json, if it fails fallback to base64
                 &Encoding::APPLICATION_JSON | &Encoding::TEXT_JSON | &Encoding::TEXT_JSON5 => {
-                    serde_json::from_slice::<serde_json::Value>(&payload.contiguous())
-                        .unwrap_or(serde_json::Value::String(payload_to_string(&payload)))
+                    serde_json::from_slice::<serde_json::Value>(&payload.contiguous()).unwrap_or(
+                        serde_json::Value::String((*StringOrBase64::from(payload)).clone()),
+                    )
                 }
                 // otherwise convert to JSON string
-                _ => serde_json::Value::String(payload_to_string(&payload)),
+                _ => serde_json::Value::String((*StringOrBase64::from(payload)).clone()),
             }
         }
     }
--- a/plugins/zenoh-plugin-rest/src/lib.rs
+++ b/plugins/zenoh-plugin-rest/src/lib.rs
@@ -61,10 +61,6 @@ pub fn base64_encode(data: &[u8]) -> String {
     general_purpose::STANDARD.encode(data)
 }
 
-// fn payload_to_string(payload: &Payload) -> String {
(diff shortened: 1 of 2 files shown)


plugins/zenoh-plugin-rest/src/lib.rs
```diff
@@ -46,47 +47,71 @@ lazy_static::lazy_static! {
...
-            ts.to_string()
-        } else {
-            "None".to_string()
+pub fn base64_encode(data: &[u8]) -> String {
+    use base64::engine::general_purpose;
+    general_purpose::STANDARD.encode(data)
+}
+
+fn payload_to_string(payload: &Payload) -> String {
+    String::from_utf8(payload.contiguous().to_vec()).unwrap_or(base64_encode(&payload.contiguous()))
```
Mallets, line 64: Why not using the `.deserialize()` as initially done?

Mallets: Or even https://github.com/eclipse-zenoh/zenoh/blob/d808ba29321c00e60329806ec5e7eafde25acfdf/zenoh/src/payload.rs#L561?

gabrik: Didn't knew about it

gabrik: Updated code to use `StringOrBase64::from` and then dereferecing to `String`.

A `clone()` becomes necessary but should be fine.

plugins/zenoh-plugin-rest/src/lib.rs
```diff
@@ -46,47 +48,68 @@ lazy_static::lazy_static! {
...
-            ts.to_string()
-        } else {
-            "None".to_string()
+pub fn base64_encode(data: &[u8]) -> String {
+    use base64::engine::general_purpose;
+    general_purpose::STANDARD.encode(data)
+}
+
+fn payload_to_json(payload: Payload, encoding: &Encoding) -> serde_json::Value {
+    match payload.len() {
```
Mallets, line 65: Rust prefers the usage of `is_empty()` rather than checking if the `len() == 0`

PR #838, opened 2024-03-15
Optimize zint encode/decode

Mallets:
This PR improves ZInt encoder/decoder via avoding unnecessary checks.

### DIFF
--- a/commons/zenoh-codec/src/core/zint.rs
+++ b/commons/zenoh-codec/src/core/zint.rs
@@ -17,7 +17,7 @@ use zenoh_buffers::{
     writer::{DidntWrite, Writer},
 };
 
-const VLE_LEN: usize = 10;
+const VLE_LEN: usize = 9;
 
 impl LCodec<u64> for Zenoh080 {
     fn w_len(self, x: u64) -> usize {
@@ -29,7 +29,6 @@ impl LCodec<u64> for Zenoh080 {
         const B6: u64 = u64::MAX << (7 * 6);
         const B7: u64 = u64::MAX << (7 * 7);
         const B8: u64 = u64::MAX << (7 * 8);
-        const B9: u64 = u64::MAX << (7 * 9);
 
         if (x & B1) == 0 {
             1
@@ -47,10 +46,8 @@ impl LCodec<u64> for Zenoh080 {
             7
         } else if (x & B8) == 0 {
             8
-        } else if (x & B9) == 0 {
-            9
         } else {
-            10
+            9
         }
     }
 }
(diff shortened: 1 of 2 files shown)


Mallets:
@DenisBiryukov91 can you help in reviewing this PR?

commons/zenoh-codec/src/core/zint.rs
```diff
@@ -112,14 +112,24 @@ where
     fn write(self, writer: &mut W, mut x: u64) -> Self::Output {
         writer.with_slot(VLE_LEN, move |buffer| {
             let mut len = 0;
-            let mut b = x as u8;
-            while x > 0x7f {
-                buffer[len] = b | 0x80;
+            while (x & !0x7f) != 0 {
```
DenisBiryukov91, line 115: I would suggest to specify explicitly the type of 0x7f literal as 0x7fu64.

commons/zenoh-codec/src/core/zint.rs
```diff
@@ -112,14 +112,24 @@ where
...
+                // SAFETY: buffer is guaranteed to be VLE_LEN long where VLE_LEN is
+                //         the maximum number of bytes a VLE can take once encoded.
+                //         I.e.: x is shifted 7 bits to the right every iteration,
+                //         the loop is at most VLE_LEN iterations.
+                unsafe {
+                    *buffer.get_unchecked_mut(len) = (x as u8) | 0x80;
```
DenisBiryukov91, line 121: same here, I would suggest specifying explicitly the type of 0x80 literal as 0x80u8

commons/zenoh-codec/src/core/zint.rs
```diff
@@ -134,18 +144,20 @@ where
 
     fn read(self, reader: &mut R) -> Result<u64, Self::Error> {
         let mut b = reader.read_u8()?;
+        if (b & 0x80) == 0 {
```
DenisBiryukov91, line 147: do we really need this piece of code, given that even without it the result would be the same ?

Mallets: Yes, it's an early return that avoids useless shift and improves the performance for common case when a `zint` is smaller than `128`.

commons/zenoh-codec/src/core/zint.rs
Mallets: Done

commons/zenoh-codec/src/core/zint.rs
Mallets: Done

commons/zenoh-codec/src/core/zint.rs
```diff
@@ -134,18 +144,20 @@ where
...
+            return Ok(b as u64);
+        }
 
         let mut v = 0;
         let mut i = 0;
-        let mut k = VLE_LEN;
-        while b > 0x7f && k > 0 {
+        // 7 * VLE_LEN is beyond the maximum number of shift bits
+        while (b & 0x80) != 0 && i != 7 * VLE_LEN {
```
DenisBiryukov91: should VLE_LEN be set to 9, and shall we keep 8 bits of data in the 8-th byte of buffer ? This would allow us to use 9 bytes at most for 8 + 7 * 8 = 64 bits, since we really don't care about continuation bit in the msB: we know it should be the last. Then the check  i != 7 * VLE_LEN can be replaced with I < 7* (VLE_LEN - 1) (i.e. I < 56).  As a bonus this would also allow us to never overflow u64, because we are using exactly 64 bits rather than 70.

commons/zenoh-codec/src/core/zint.rs
DenisBiryukov91: Ok, if it is indeed the case. I just thought that an extra if branch might have a higher impact overall than an extra bitshift.

Mallets: Counterintuitively that's not the case.

commons/zenoh-codec/src/core/zint.rs
Mallets: An update has been just committed to properly handle this case.

DenisBiryukov91 (approved):

PR #840, opened 2024-03-18
Adding TLS authentication

snehilzs:
Adding TLS authentication for ACL support


Mallets:
@oteffahi could you please review this PR as well?

Cargo.toml
```diff
@@ -75,14 +75,17 @@ description = "Zenoh: Zero Overhead Pub/sub, Store/Query and Compute."
 # DEFAULT-FEATURES NOTE: Be careful with default-features and additivity!
 #                        (https://github.com/rust-lang/cargo/issues/11329)
 [workspace.dependencies]
+
+
+async-rustls = "0.4.0"
```
Mallets, line 80: Why reintroducing `async-rustls` while the implementation moved to `tokio-tls`?

oteffahi: Good catch. It doesn't seem to be used in the code, and it is not present in the lockfile.

Mallets (approved):

PR #846, opened 2024-03-19
Fix/memory leaks

### LINKED ISSUE #695: Memory leaks cleanup

DenisBiryukov91:
Addresses a series of memory leaks detected by valgrind (Closes #695 ):

due to cyclic references in Runtime and its dependents
due to cyclic references in Session and its dependents
due to cyclic references in Resource and its children
due to cyclic reference in Mux->Face->State->Mux
due to non-terminated tasks after Session.close()
Valgrind still reports leaks, related to static variable ZRUNTIME_POOL. This is because rust never calls drop on static variables by design. This leaks can be removed by calling zruntime_pool_free() or using ZRuntimePoolGuard.

Also adds a valgrind memory leaks test into ci.


zenoh/src/net/routing/hat/linkstate_peer/network.rs
```diff
@@ -336,6 +337,7 @@ impl Network {
 
     pub(super) fn link_states(&mut self, link_states: Vec<LinkState>, src: ZenohId) -> Changes {
         log::trace!("{} Received from {} raw: {:?}", self.name, src, link_states);
+        let strong_runtime = self.runtime.upgrade().unwrap();
```
milyin, line 340: Same question: why not to just immediately return empty `Changes` instead of panic?

milyin (changes requested):
Looks good for me. But I don't have enough experience with zenoh deep internals and with tokyo, so it would be good to have this PR reviewed by someone from protocol team. @OlivierHecart ?
The only concern: I'd prefer graceful failure instead of `unwrap()` panics. In most cases it's possible to just return empty result as far as I see

zenoh/src/net/routing/hat/linkstate_peer/network.rs
```diff
@@ -247,7 +248,7 @@ impl Network {
             whatami: self.graph[idx].whatami,
             locators: if details.locators {
                 if idx == self.idx {
-                    Some(self.runtime.get_locators())
+                    Some(self.runtime.upgrade().unwrap().get_locators())
```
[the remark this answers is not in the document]
DenisBiryukov91: It is mostly done to be able to detect issues with arc-weak management. It would be way more complicated to detect them if I return just an empty result. In this not being able to acquire a strong runtime is likely a non-recoverable error.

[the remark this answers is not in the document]
milyin: Maybe then use `debug_assert!` and `log::error!` with "internal error" message? This will allow to detect the problem and avoid panic. Anyway this may happen on shutdown stage only. I think it's better to shutdown with error message than abort the whole user application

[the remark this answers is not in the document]
DenisBiryukov91: It will allow to detect the problem, but will not give you stack trace and a such it will severely complicate the debugging.
If we agree that this is an impossible situation (which it should be, given that runtime is expected to terminate all the tasks upon its close) I suggest we just panic and not overload the code with unnecessary if/match statements and error logs, which likely nobody will ever look at.

commons/zenoh-task/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+        futures::executor::block_on(task);
+    }
+
+    /// Async version of [`TaskController::terminate_all()`].
+    pub async fn terminate_all_async(&self) {
+        self.tracker.close();
+        self.token.cancel();
+        let task = async move {
+            tokio::select! {
+                _ = tokio::time::sleep(Duration::from_secs(10)) => {
```
OlivierHecart, line 130: I would not define an hardcoded timeout down there. I'd rather have it defined from the caller.
(Note you can use tokio::time::timeout to timeout a future more elegantly.)

commons/zenoh-task/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// The caller must ensure that all tasks spawned with [`TaskController::spawn()`]
+    /// or [`TaskController::spawn_with_rt()`] can yield in finite amount of time either because they will run to completion
+    /// or due to cancellation of token acquired via [`TaskController::get_cancellation_token()`].
+    /// Tasks spawned with [`TaskController::spawn_abortable()`] or [`TaskController::spawn_abortable_with_rt()`] will be aborted (i.e. terminated upon next await call).
+    pub fn terminate_all(&self) {
+        self.tracker.close();
+        self.token.cancel();
+        let task = async move {
+            tokio::select! {
+                _ = tokio::time::sleep(Duration::from_secs(10)) => {
```
OlivierHecart, line 115: I would not define an hardcoded timeout down there. I'd rather have it defined from the caller.

zenoh/src/net/routing/dispatcher/resource.rs
```diff
@@ -306,6 +307,7 @@ impl Resource {
                         }
                     }
                 }
+                mutres.nonwild_prefix.take();
```
OlivierHecart, line 310: This should not be necessary as mutes will be dropped.

zenoh/tests/routing.rs
```diff
@@ -27,7 +27,7 @@ use zenoh_core::ztimeout;
 use zenoh_protocol::core::{WhatAmI, WhatAmIMatcher};
 use zenoh_result::bail;
 
-const TIMEOUT: Duration = Duration::from_secs(10);
+const TIMEOUT: Duration = Duration::from_secs(60);
```
OlivierHecart, line 30: Why do we need to increase this ?

DenisBiryukov91: Because waiting until the task terminate gracefully requires some time, with 10 seconds I observe a lot of tracked tasks still alive by the end of the test.

YuanYuYuan:
@DenisBiryukov91 , I'd like to leave some comments since you're addressing some issues that I wanted to solve a few weeks ago on Tokio porting. Here're some suggestions

- It's better not to mix an additional executor with Tokio. Any Tokio `spawn` , `spawn_blocking`, etc. require to be inside a  Tokio context. The tests used for valgrind don't fail since we use `#[tokio::main]`. It would panic if we run it with different runtime or in a blocking manner. It's suggested to either explicitly use our `ZRuntime` or use `std::thread`.
- Regarding the `Drop` of `ZRUNTIME_POOL` , we could use fewer unsafe codes and tackle the runtime properly. Can you have a look at this  [PR](https://github.com/DenisBiryukov91/zenoh/pull/3) please?
- I once tried to implement something similar to `TaskController`before but gave up then. The reasons are 
  1. The variance of the using scope. Wrapping every task with a `CancellationToken` and hooking it to a `TaskTracker` is not always the optimal solution. Furthermore, we may want to collect the results of spawning tasks, then we need to use `JoinSet` instead of `TaskTracker`.
  2. You may find that some implementations in zenoh-links doesn't fit the scenario of `TaskTracker` and we simply use a `JoinHandle`  to manage it. This is more efficient for handling the tasks for that specific zenoh-link. 
  Anyway, I do not object to the additional zenoh-task crate. But it might not be a complete solution and will suffer to change in the future. A better one is to review and improve our async usage. Like do we really need to spawn this task that never awaits on a ZRuntime? https://github.com/DenisBiryukov91/zenoh/blob/fix/memory_leaks/zenoh/src/session.rs#L1510-L1537

zenoh/tests/routing.rs
DenisBiryukov91: fixed back to 10 seconds. Termination seems to work faster once I switched  to Tokio executor.

commons/zenoh-task/src/lib.rs
DenisBiryukov91: fixed.

commons/zenoh-task/src/lib.rs
DenisBiryukov91: fixed

zenoh/src/net/routing/dispatcher/resource.rs
DenisBiryukov91: mutres is a reference here, so no actual drop operation will happen.  Based on my tests, this line is necessary, since non_wild prefix might contain an Arc to a parent, that will fail to clean due to higher than expect Arc reference count. I.e. we can have following situation: mutres.non_wild prefix holds parent, grandparent.childs hold parent, mutres.parent holds parent, together with reclone it will lead to Arc::strong_count of 4, and thus we will not be able to clean the parent.

OlivierHecart: Well that was the point of the code. You can't clean a resource that has childs. You have to clean the childs first.

OlivierHecart: But we call clean(parent) just after and so need to clean nonwild_prefix Arc before that. Ok

milyin (approved):

PR #848, opened 2024-03-19
Add configurable wait_before_drop time in case of CongestionControl::Drop

### LINKED ISSUE #847: CongestionControl::Drop is too agressive
### Describe the bug

`CongestionControl::Drop` setting allows Zenoh to drop some messages in case of congestion when sending messages.
More precisely, when sending messages, Zenoh has a queue of batches in preparation to be send. If this queue is empty because all batches are full and about to be sent by another task, this setting tells Zenoh to drop the message.

Now, whenever a burst of messages (>3000Hz) is to be sent, we can see that Zenoh drops lot of messages, even if the global throughput could be handled by Zenoh.
The problem is that the scheduling of tasks may penalise the task sending complete batches and giving back empty batches to the queue with regard to the task trying to send new messages and requiring an empty batch.

It would be good to relax this drop mechanism, waiting a very short time for the sending task to have a chance to give back an empty batch into the queue. This should not have a negative impact on the global throughput.

### To reproduce

1. Start a publisher at 3000Hz payloads of 100 bytes, using ``CongestionControl::Drop`
2. Check in a subscriber (connected via TCP on loopback) how many publications are lost

### System info

- Platform: all
- Version: 0.10.1-rc

JEnoch:
Fix #847 

Add in configuration within `transport/link/tx/queue` a configuration section for CongestionControl:
```json5
          /// Congestion occurs when the queue is empty (no available batch).
          /// Using CongestionControl::Block the caller is blocked until a batch is available and re-insterted into the queue.
          /// Using CongestionControl::Drop the message might be dropped, depending on conditions configured here.
          congestion_control: {
            /// The maximum time in nanosecond to wait for an available batch before dropping the message if still no batch is available.
            wait_before_drop: 1000000
          },
```

This `wait_before_drop` setting is used by Zenoh in case of `CongestionControl::Drop` to wait for this amount of time for an  available batch in the batches queue.

### DIFF
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -261,6 +261,13 @@
             data_low: 4,
             background: 4,
           },
+          /// Congestion occurs when the queue is empty (no available batch).
+          /// Using CongestionControl::Block the caller is blocked until a batch is available and re-insterted into the queue.
+          /// Using CongestionControl::Drop the message might be dropped, depending on conditions configured here.
+          congestion_control: {
+            /// The maximum time in nanosecond to wait for an available batch before dropping the message if still no batch is available.
+            wait_before_drop: 1000000
+          },
           /// The initial exponential backoff time in nanoseconds to allow the batching to eventually progress.
           /// Higher values lead to a more aggressive batching but it will introduce additional latency.
           backoff: 100,
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -265,8 +265,8 @@
           /// Using CongestionControl::Block the caller is blocked until a batch is available and re-insterted into the queue.
           /// Using CongestionControl::Drop the message might be dropped, depending on conditions configured here.
           congestion_control: {
-            /// The maximum time in nanosecond to wait for an available batch before dropping the message if still no batch is available.
-            wait_before_drop: 1000000
+            /// The maximum time in microseconds to wait for an available batch before dropping the message if still no batch is available.
+            wait_before_drop: 1000
           },
           /// The initial exponential backoff time in nanoseconds to allow the batching to eventually progress.
           /// Higher values lead to a more aggressive batching but it will introduce additional latency.
--- a/io/zenoh-transport/src/common/pipeline.rs
+++ b/io/zenoh-transport/src/common/pipeline.rs
@@ -19,10 +19,12 @@ use super::{
 };
 use flume::{bounded, Receiver, Sender};
 use ringbuffer_spsc::{RingBuffer, RingBufferReader, RingBufferWriter};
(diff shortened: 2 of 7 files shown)


DEFAULT_CONFIG.json5
```diff
@@ -261,6 +261,13 @@
             data_low: 4,
             background: 4,
           },
+          /// Congestion occurs when the queue is empty (no available batch).
+          /// Using CongestionControl::Block the caller is blocked until a batch is available and re-insterted into the queue.
+          /// Using CongestionControl::Drop the message might be dropped, depending on conditions configured here.
+          congestion_control: {
+            /// The maximum time in nanosecond to wait for an available batch before dropping the message if still no batch is available.
```
Mallets, line 268: Should it be in microsends? I believe it's more user friendly and I don't really see the need to expose such granurality that may not even be guaranteed by the system timer.

io/zenoh-transport/src/common/pipeline.rs
```diff
@@ -140,19 +148,24 @@ impl StageIn {
...
-                                    // The yield is to avoid the writing task to spin
-                                    // indefinitely and monopolize the CPU usage.
-                                    thread::yield_now();
-                                    return false;
-                                } else {
-                                    if !self.s_ref.wait() {
+                                match deadline_before_drop {
+                                    Some(deadline) if !$fragment => {
                                         // Restore the sequence number
                                         $restore_sn;
```
Mallets, line 154: `$restore_sn` should be called only if we are dropping the message, i.e. inside the 

```rust
if !self.s_ref.wait_deadline(deadline) {
    $restore_sn
    return false;
}
```

Mallets (changes requested):
Overall LGTM, just left a couple of comments that should be addressed before merging.

DEFAULT_CONFIG.json5
JEnoch: I choose nanoseconds to be consistent with the `backoff` setting which is in nanoseconds.
From a user perspective, I feel weird and possibly confusing that different time units have to be used in a same config file.

io/zenoh-transport/src/common/pipeline.rs
JEnoch: Fixed

DEFAULT_CONFIG.json5
Mallets: We already have different time units depending on the configuration variable and on the context, e.g.: https://github.com/eclipse-zenoh/zenoh/blob/5f6037293a6c425d5769797010257271924228c6/DEFAULT_CONFIG.json5#L233

JEnoch: Fair enough, I changed the unit to microseconds

Mallets (approved):

PR #853, opened 2024-03-20
Enable queries timeout

sashacmc:


io/zenoh-transport/tests/unicast_authenticator.rs
```diff
@@ -638,7 +638,7 @@ async fn run_with_lowlatency_transport(endpoint: &EndPoint) {
 #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
 async fn authenticator_tcp() {
     let _ = env_logger::try_init();
-    let endpoint: EndPoint = format!("tcp/127.0.0.1:{}", 8000).parse().unwrap();
+    let endpoint: EndPoint = format!("tcp/127.0.0.1:{}", 8001).parse().unwrap();
```
Mallets, line 641: Why changing the port in the tests?

sashacmc: It consistently conflicts (100% fails when running locally on my PC) with some other test, perhaps with unicast_authenticator.

Mallets: Maybe it's conflicting with the REST API enabled somewhere? From the code it seems it's the only other thing that makes use of port 8000.

sashacmc: Mybe, but the conflict appears only when I enable the queries timeout, seems some concurrency timeshift

Mallets: Query timeout should have nothing to do with port conflict if the port is uniquely chosen... Can you do an `ss -pl` to check what are the processes using it?

sashacmc: The port is free, so I’m still inclined to believe that the conflict is with this:
https://github.com/eclipse-zenoh/zenoh/blob/3a2645588ed739a2fbfdc6dc00d5f24442017613/io/zenoh-transport/tests/unicast_authenticator.rs#L641

Mallets: Yes, but that's the exact line you changed in this PR. If that line is the only one opening the por `800` and IF there is a conflict, what's the other thing opening the same port? The only other option I found is the REST API. Is that one creating the conflict?

sashacmc: Seems, it was some side effect, from zeno pico tests, reverted.

Mallets (approved):

PR #858, opened 2024-03-24
Sample api rework

### LINKED ISSUE #837: `Sample` rework
After latest discussions with @Mallets, @DenisBiryukov91 the following approach to `Sample` structure was made:

- For all languages (in so called "core" API) Sample is a data structure received from queryable or publisher, i.e. input parameter for callbacks of `get` and `subscriber`.
- Sample is Zenoh internal structure, it's actual representation is hidden from user. Sample data is accessed by getter methods. These getters **are** part of "core" API.
- Sample received by `get` or `subscriber` may be owned by user. Use can store it in memory and modify it using setters. This behavior is **not** a part of core API, i.e. some bindings may provide only R/O access to received sample.
- Sample structure can be 
   - sent by publisher's "write" method (maybe it's worth to rename it to "send_sample" ?)
   - sent from queriable with "reply_sample" method. 

   These features are **not** part of core API and user is should not be encouraged to use it. But these features are useful for some storages/caches where we resending the same sample which we received. So maybe it makes sense to include them to stable Rust API. For API completeness user is allowed to create the new sample.

I.e. in other words, we have API separated to two parts:

# Core API

This is API obligatory for all language bindings. 

Operations "put", "delete", "reply" are builders which technically builds the sample from pieces and sends it, but they doesn't give user access to this sample. Example:
```Rust
publisher
   .put("payload")
   .with_attachment(attachment)
   .with_priority(High)
   .sync();
```

Operations 'get', 'subscribe' receives 'Sample' - some entity with set of accessor methods, i.e. `get_payload()`, `get_attachment()`, etc.

# Optional API

I don't know how to better name it. "Extended" API is not good because it reminds about zenoh-ext, which is unrelated to this case. "Rust" API is wrong too: many of this functionality is provided not in Rust only and some functionality may exist in other languages, but not in Rust - like synchronization primitives. "Expanded" API maybe?

The idea of this "optional" API is that this API is not obligatory for all bindings, but if it's provided, it must be coherent in all bindings.

In Rust API (and supposedly in C and C++ APIs) the sample can be owned by user. This allows him to receive sample, make modification to it with setter methods (like "ensure_timestamp") and then send this sample. The storages will use this feature for example.

The user can also create sample (with constructor interface similar to "put", "delete"), depending on the language, i.e. builders for Rust, `z_create_sample` in C with structure for options.

milyin:
Closes #837 

The API changes proposed by this request are:

1. Implemented set of traits for unified way of constructing sample in publisher and queryable. These traits are:
```Rust
pub trait QoSBuilderTrait {
    /// Change the `congestion_control` to apply when routing the data.
    fn congestion_control(self, congestion_control: CongestionControl) -> Self;
    /// Change the priority of the written data.
    fn priority(self, priority: Priority) -> Self;
    /// Change the `express` policy to apply when routing the data.
    /// When express is set to `true`, then the message will not be batched.
    /// This usually has a positive impact on latency but negative impact on throughput.
    fn express(self, is_express: bool) -> Self;
}

pub trait TimestampBuilderTrait {
    /// Sets of clears timestamp
    fn timestamp<T: Into<Option<Timestamp>>>(self, timestamp: T) -> Self;
}

pub trait SampleBuilderTrait {
    /// Attach source information
    #[zenoh_macros::unstable]
    fn source_info(self, source_info: SourceInfo) -> Self;
    /// Attach user-provided data in key-value format
    #[zenoh_macros::unstable]
    fn attachment<T: Into<Option<Attachment>>>(self, attachment: T) -> Self;
}

pub trait ValueBuilderTrait {
    /// Set the [`Encoding`]
    fn encoding<T: Into<Encoding>>(self, encoding: T) -> Self;
    /// Sets the payload
    fn payload<T: Into<Payload>>(self, payload: T) -> Self;
    /// Sets both payload and encoding at once.
    /// This is convenient for passing user type which supports `Into<Value>` when both payload and encoding depends on user type
    fn value<T: Into<Value>>(self, value: T) -> Self;
}
```
Depending on the component, specific traits are implemented. For example builder for `Session::put` operation mplements `QoSBuilderTrait + SampleBuilderTrait + ValueBuilderTrait`, but builder for `Publisher::delete` implements only `SampleBuilderTrait`, because priority is set in publisher itself (trait `QoSBuilderTrait` is implemented by the `PublisherBuilder`) and `encoding` and `payload` are not sent on delete operation.

This allows to guarantee consistent support of sample data in different components. 

`TimestampBuilderTrait` is separate because `get` operation doesn't support sending timestamp, so it was necessary to detach this function from others

2. removed write access to `Sample` structure. Implemented builders `SampleBuilder`, `PutSampleBuilder`, `DeleteSampleBuilder` instead. These builders may be constructed either empty or from existing sample. They are implemented as a wrappers around `Sample`, so sample can be edited on place.

Example:
```Rust
// ensure the sample has a timestamp, thus it will always be sorted into the MergeQueue
// after any timestamped Sample possibly coming from a fetch reply.
let s = if s.timestamp().is_none() {
  SampleBuilder::from(s).with_timestamp(new_reception_timestamp()).into()
} else {
  s
};
```


zenoh/src/sample_builder.rs
Mallets, line 1: I would rather make `sample` as a module and add the builder inside it. E.g.:

```
zenoh/sample/mod.rs
zenoh/sample/builder.rs
```

examples/examples/z_ping.rs
```diff
@@ -16,6 +16,7 @@ use std::time::{Duration, Instant};
 use zenoh::config::Config;
 use zenoh::prelude::sync::*;
 use zenoh::publication::CongestionControl;
+use zenoh::sample_builder::QoSBuilderTrait;
```
Mallets, line 19: I suggest to add the builder traits directly in the prelude.

plugins/zenoh-plugin-rest/src/lib.rs
```diff
@@ -34,6 +34,7 @@ use zenoh::plugins::{RunningPluginTrait, ZenohPlugin};
 use zenoh::prelude::r#async::*;
 use zenoh::query::{QueryConsolidation, Reply};
 use zenoh::runtime::Runtime;
+use zenoh::sample_builder::ValueBuilderTrait;
```
Mallets, line 37: I suggest to add the builder traits directly in the prelude.

plugins/zenoh-plugin-rest/src/lib.rs
```diff
@@ -419,7 +420,7 @@ async fn query(mut req: Request<(Arc<Session>, String)>) -> tide::Result<Respons
                 .content_type()
                 .map(|m| Encoding::from(m.to_string()))
                 .unwrap_or_default();
-            query = query.with_value(Value::from(body).with_encoding(encoding));
+            query = query.with_value(Value::from(body).encoding(encoding));
```
Mallets, line 423: `with_value` should be `value`

examples/examples/z_pong.rs
```diff
@@ -15,6 +15,7 @@ use clap::Parser;
 use zenoh::config::Config;
 use zenoh::prelude::sync::*;
 use zenoh::publication::CongestionControl;
+use zenoh::sample_builder::QoSBuilderTrait;
```
Mallets, line 18: Similar. Trait should be part of the prelude.

examples/examples/z_pub.rs
```diff
@@ -16,6 +16,7 @@ use clap::Parser;
 use std::time::Duration;
 use zenoh::config::Config;
 use zenoh::prelude::r#async::*;
+use zenoh::sample_builder::SampleBuilderTrait;
```
Mallets, line 19: Similar. Trait should be part of the prelude.

examples/examples/z_pub_shm_thr.rs
```diff
@@ -15,6 +15,7 @@ use clap::Parser;
 use zenoh::config::Config;
 use zenoh::prelude::r#async::*;
 use zenoh::publication::CongestionControl;
+use zenoh::sample_builder::QoSBuilderTrait;
```
Mallets, line 18: Similar. Trait should be part of the prelude.

examples/examples/z_pub_thr.rs
```diff
@@ -16,6 +16,7 @@ use clap::Parser;
 use std::convert::TryInto;
 use zenoh::prelude::sync::*;
 use zenoh::publication::CongestionControl;
+use zenoh::sample_builder::QoSBuilderTrait;
```
Mallets, line 19: Similar. Trait should be part of the prelude.

plugins/zenoh-plugin-rest/examples/z_serve_sse.rs
```diff
@@ -15,6 +15,7 @@ use clap::{arg, Command};
 use std::time::Duration;
 use zenoh::prelude::r#async::*;
 use zenoh::publication::CongestionControl;
+use zenoh::sample_builder::QoSBuilderTrait;
```
Mallets, line 18: Similar. Trait should be part of the prelude.

plugins/zenoh-plugin-storage-manager/src/replica/align_queryable.rs
```diff
@@ -20,6 +20,8 @@ use std::str;
 use std::str::FromStr;
 use zenoh::payload::StringOrBase64;
 use zenoh::prelude::r#async::*;
+use zenoh::sample_builder::TimestampBuilderTrait;
+use zenoh::sample_builder::ValueBuilderTrait;
```
Mallets, line 24: Similar. Traits should be part of the prelude.

plugins/zenoh-plugin-storage-manager/src/replica/aligner.rs
```diff
@@ -21,6 +21,7 @@ use std::str;
 use zenoh::key_expr::{KeyExpr, OwnedKeyExpr};
 use zenoh::payload::StringOrBase64;
 use zenoh::prelude::r#async::*;
+use zenoh::sample_builder::{PutSampleBuilder, TimestampBuilderTrait, ValueBuilderTrait};
```
Mallets, line 24: Similar. Traits should be part of the prelude.

plugins/zenoh-plugin-storage-manager/src/replica/aligner.rs
```diff
@@ -108,9 +109,10 @@ impl Aligner {
                 let Value {
                     payload, encoding, ..
                 } = value;
-                let sample = Sample::new(key, payload)
-                    .with_encoding(encoding)
-                    .with_timestamp(ts);
+                let sample = PutSampleBuilder::new(key, payload)
```
Mallets, line 112: From and API point of view, in my opinion would be better to have something like `SampleBuilder::put(key, payload) -> PutSampleBuilder`. But the user never explicitly uses a `PutSampleBuilder`. Opinions?

plugins/zenoh-plugin-storage-manager/src/replica/storage.rs
```diff
@@ -290,23 +298,31 @@ impl StorageService {
...
+                            .into()
                     }
-                    None => Sample::new(KeyExpr::from(k.clone()), sample.payload().clone())
-                        .with_encoding(sample.encoding().clone())
-                        .with_timestamp(*sample.timestamp().unwrap())
-                        .with_kind(sample.kind()),
+                    Some(Update {
+                        kind: SampleKind::Delete,
+                        data,
+                    }) => DeleteSampleBuilder::new(KeyExpr::from(k.clone()))
```
Mallets, line 320: From and API point of view, in my opinion would be better to have something like `SampleBuilder::delete(key) -> DeleteSampleBuilder`. But the user never explicitly uses a `DeleteSampleBuilder`. Opinions?

zenoh/src/publication.rs
```diff
@@ -293,25 +406,22 @@ impl<'a> Publisher<'a> {
 
     /// Change the `congestion_control` to apply when routing the data.
     #[inline]
-    pub fn congestion_control(mut self, congestion_control: CongestionControl) -> Self {
+    pub fn set_congestion_control(&mut self, congestion_control: CongestionControl) {
```
Mallets, line 409: Why the `set_`?

zenoh/src/sample_builder.rs
```diff
@@ -0,0 +1,341 @@
...
+            timestamp: None,
+            qos: QoS::default(),
+            #[cfg(feature = "unstable")]
+            source_info: SourceInfo::empty(),
+            #[cfg(feature = "unstable")]
+            attachment: None,
+        }))
+    }
+    /// Allows to change keyexpr of [`Sample`]
+    pub fn with_keyexpr<IntoKeyExpr>(self, key_expr: IntoKeyExpr) -> Self
```
Mallets, line 171: Remove `with_`?

zenoh/src/sample_builder.rs
```diff
@@ -0,0 +1,341 @@
...
```
Mallets, line 258: Remove `with_`?

examples/examples/z_ping.rs
milyin: done

plugins/zenoh-plugin-rest/src/lib.rs
milyin: done

plugins/zenoh-plugin-rest/src/lib.rs
milyin: Agree. I reworked `Value` API: added `value` method to `ValueBuilderTrait`. I.e. each object whcih accepts payload/encoding accepts value also. For `GetBuilder` added special treatment of emply value. When value is empty, setter behaves like value is set to `None`. This allows to write code like in current "z_get" example:
```Rust
    let (config, selector, value, target, timeout) = parse_args();
...
    let replies = session
        .get(&selector)
        .value(value)
        .target(target)
        .timeout(timeout)
        .res()
        .await
        .unwrap();
 ```
instead of 
```Rust
    let replies = match value {
        Some(value) => session.get(&selector).with_value(value),
        None => session.get(&selector),
    }
    .target(target)
    .timeout(timeout)
    .res()
    .await
    .unwrap();
```

examples/examples/z_pong.rs
milyin: doen

examples/examples/z_pub.rs
milyin: done

plugins/zenoh-plugin-rest/examples/z_serve_sse.rs
milyin: done

plugins/zenoh-plugin-storage-manager/src/replica/aligner.rs
milyin: done

plugins/zenoh-plugin-storage-manager/src/replica/aligner.rs
milyin: agree, done

zenoh/src/publication.rs
milyin: Because these methods are not builder ones, but setters for object itself. Otherwise it's unclear, do they set or return value

zenoh/src/sample_builder.rs
milyin: done

zenoh/src/sample_builder.rs
milyin: done

plugins/zenoh-plugin-storage-manager/src/replica/storage.rs
milyin: done

plugins/zenoh-plugin-storage-manager/src/replica/align_queryable.rs
milyin: done

examples/examples/z_pub_shm_thr.rs
milyin: done

zenoh/src/sample_builder.rs
milyin: done

zenoh/src/publication.rs
Mallets: Do we really want to allow override of values once the publisher is already created? I'm skeptical about it because it allows the user to override at a later stage what has been already configured with the `declare_publisher`. I'm more inclined to not having such `set_` functions.

milyin: I think If this functionality is not really useful, better to remove it to give us more freedom for future updates. So I don't mind to remove it.
In C API it's possible to only set options on declaration stage, so this change will not affect the C/C++ bindings

zenoh/src/value.rs
```diff
@@ -36,23 +36,36 @@ impl Value {
...
 
-    /// Sets the encoding of this [`Value`]`.
-    #[inline(always)]
-    pub fn with_encoding<IntoEncoding>(mut self, encoding: IntoEncoding) -> Self
-    where
-        IntoEncoding: Into<Encoding>,
-    {
-        self.encoding = encoding.into();
-        self
+impl ValueBuilderTrait for Value {
```
Mallets, line 53: `Value` has not been ported yet to accessor pattern.
By implementing `ValueBuilderTrait` directly on `Value` we will have a conflict in terms of functions.

With accessor pattern:
```
let payload: &Payload = value.payload();
```

With builder pattern:
```
value = value.payload("foo");
```

Do we really need to implement `ValueBuilderTrait` on `Value` itself or could we have a `ValueBuilder` as we do with a `SampleBuilder`?

PR #870, opened 2024-03-28
Protocol interest

Mallets:
Add the interest id in the declare message to disambiguate wether the declaration is sent in a response of a declare interest with `future == false` or not.


commons/zenoh-protocol/src/network/declare.rs
```diff
@@ -25,27 +25,30 @@ pub use subscriber::*;
 pub use token::*;
 
 pub mod flag {
-    // pub const X: u8 = 1 << 5; // 0x20 Reserved
-    // pub const X: u8 = 1 << 6; // 0x40 Reserved
+    pub const I: u8 = 1 << 5; // 0x20 Interest      if I==1 then the declare is in a response of an Interest with future==false
```
fuzzypixelz, line 28:
```suggestion
    pub const I: u8 = 1 << 5; // 0x20 Interest      if I==1 then the declare is in a response to an Interest with future==false
```

commons/zenoh-protocol/src/network/declare.rs
```diff
@@ -25,27 +25,30 @@ pub use subscriber::*;
...
+                              // pub const X: u8 = 1 << 6; // 0x40 Reserved
     pub const Z: u8 = 1 << 7; // 0x80 Extensions    if Z==1 then an extension will follow
 }
 
 /// Flags:
-/// - X: Reserved
+/// - I: Interest       If I==1 then the declare is in a response of an Interest with future==false
```
fuzzypixelz, line 34:
```suggestion
/// - I: Interest       If I==1 then the declare is in a response to an Interest with future==false
```

fuzzypixelz (approved):

PR #871, opened 2024-03-28
Add protocol version to error message

sashacmc:


io/zenoh-transport/src/unicast/establishment/accept.rs
```diff
@@ -167,9 +167,11 @@ impl<'a, 'b: 'a> AcceptFsm for &'a mut AcceptLink<'b> {
         // Check if the version is supported
         if init_syn.version != input.mine_version {
             let e = zerror!(
-                "Rejecting InitSyn on {} because of unsupported Zenoh version from peer: {}",
+                "Rejecting InitSyn on {} because of unsupported Zenoh protocol version (expected: {}, peer: {}) from peer: {}",
```
Mallets, line 170: I believe the word `peer` comes from the past where we used to call everything peer. Now a peer has a very specific meaning in Zenoh, so I suggest to change the log message as follows:

`Rejecting InitSyn on {} because of unsupported Zenoh protocol version (expected: {}, received: {}) from: {}`

Mallets (approved):

PR #874, opened 2024-03-29
Query timeout

Mallets:
This PR adds a tombstone for query timeout to actually disable any timeout and any associated task spawning.
In practice, it is now possible to configure no timeout at all in both queries and infrastructure.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,6 +102,16 @@ jobs:
       - name: Run doctests
         run: cargo test --doc
 
+      - name: Install valgrind
+        if: ${{ matrix.os == 'ubuntu-latest' }}
+        run: sudo apt-get install -y valgrind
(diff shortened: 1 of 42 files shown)


Mallets:
@JEnoch I'm not very fond of using `u64::MAX` (i.e., `18446744073709551615`) to disable timeout. A better approach may be to use `0`? This of course will require some change on `rmw_zenoh`. What do you think?

JEnoch:
That makes sense. I don't think that's an issue to change this in rmw_zenoh at this stage.

However I think the comment in `DEFAULT_CONFIG.json5` (and therefore in rmw_zenoh's default config files) should be more explicit wrt. the relationship and/or difference between the timeout in config and the timeout in Query API.
And similarly in the doc for Query timeout API.

Mallets:
Superseded by #881

PR #875, opened 2024-03-29
Review tokio runtime being used

Mallets:
It seems that some parts of the code wasn't using the right tokio runtime.
This PR fixes that.

@YuanYuYuan could you review it?


YuanYuYuan:
Thanks! This awesome fix resolves the issue https://github.com/ros2/rmw_zenoh/pull/134#issuecomment-2024955800. But I need to rebase to the main branch to test rmw_zenoh.  Maybe we can directly merge the branch [here](https://github.com/ZettaScaleLabs/zenoh/tree/fix/tokio_runtime). I'll delve deeper into this fix next week to gain a better understanding of the root cause.

YuanYuYuan:
LGTM

PR #878, opened 2024-03-29
Query.reply and reply_del, now accept TryIntoKeyExpr instead of IntoKeyExpr

DenisBiryukov91:
Query.reply and reply_del, now accept TryIntoKeyExpr instead of IntoKeyExpr


Mallets (approved):

milyin:
This change breaks merging of https://github.com/eclipse-zenoh/zenoh/pull/858 where builders are reworked. 
I'll keep this comment to not forget to restore this "TryIntoKeyExpr" functionality in #858

PR #879, opened 2024-03-29
Correct generic name IntoSelector to TryIntoSelector in Session.get

DenisBiryukov91:
Correct generic name IntoSelector to TryIntoSelector in Session.get


Mallets (approved):

Mallets:
Clippy fix landing via #880.

PR #885, opened 2024-03-30
Zenoh: Fix badly behaved build.rs scripts.

chachi:
build.rs should only ever modify OUT_DIR. These scripts were modifying local files which caused a git index update which caused them to rebuild due to the git-version appearing out of date.

### DIFF
--- a/plugins/zenoh-plugin-rest/build.rs
+++ b/plugins/zenoh-plugin-rest/build.rs
@@ -27,13 +27,13 @@ fn main() {
     );
     // Generate config schema
     let schema = schema_for!(Config);
-    std::fs::write(
-        "config_schema.json5",
-        serde_json::to_string_pretty(&schema).unwrap(),
-    )
-    .unwrap();
+
+    let schema_file = std::path::PathBuf::from(std::env::var("OUT_DIR").unwrap())
+        .join("config_schema.json5");
+    std::fs::write(&schema_file, serde_json::to_string_pretty(&schema).unwrap()).unwrap();
+
     // Check that the example config matches the schema
-    let schema = std::fs::read_to_string("config_schema.json5").unwrap();
+    let schema = std::fs::read_to_string(schema_file).unwrap();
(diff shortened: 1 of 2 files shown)


J-Loudet:
Thanks for your contribution @chachi!
This looks good to me, if you could fix the formatting issue (by running and committing the result of `cargo fmt`) I would then be able to merge.

J-Loudet (changes requested):
LGTM 👍
Please run `cargo fmt` and commit the changes.

chachi:
Fixed!

J-Loudet:
Hi @chachi, thanks for the fix.
It turns out I did not fully grasp the purpose of these `build.rs` files, they not only validate the `config.json5` but also update the content of `config_schema.json5`.

This is, as you pointed out, not a recommended practice which led to some internal issues: #897 
However I am not sure of the course of action.

@milyin @fuzzypixelz what's your take?

fuzzypixelz:
> Hi @chachi, thanks for the fix. It turns out I did not fully grasp the purpose of these `build.rs` files, they not only validate the `config.json5` but also update the content of `config_schema.json5`.
> 
> This is, as you pointed out, not a recommended practice which led to some internal issues: #897 However I am not sure of the course of action.
> 
> @milyin @fuzzypixelz what's your take?

I would either:

1. Remove the code that read/writes `config_schema.json5` (in this PR it's a no-op)
2. Assert that `config_schema.json5` conforms to `config::Config` without actually overriding it with the correct value

chachi:
No problem. Let me know if I should make any changes to this PR or you'd like to take it from here. Happy to help!

fuzzypixelz:
> No problem. Let me know if I should make any changes to this PR or you'd like to take it from here. Happy to help!

Yes, please do update this PR if you can.

I suggest going with option (2), failing the build when the config schema file doesn't match the `Config` structure.

J-Loudet:
See #939

PR #886, opened 2024-03-30
Storage Replication Align Queryable improvements

chachi:
2 commits:
1. Incrementally replies with digest replies in order to avoid queries timing out w/o sending any data, especially in the case of very outdated storages.
2. Replace the unbounded queue for digests with a bounded queue to prevent infinitely growing digest queues when digest updates take a long time.

#1 is definitely valuable for long queries to make incremental progress. #2 is a bit less obvious if it's what y'all desire upstream. I'd strongly suggest something that's not unbounded as that's a recipe for disaster, but the sizing of that is where I don't know what y'all are thinking.

### DIFF
--- a/plugins/zenoh-plugin-storage-manager/src/replica/mod.rs
+++ b/plugins/zenoh-plugin-storage-manager/src/replica/mod.rs
@@ -119,7 +119,7 @@ impl Replica {
 
         // Create channels for communication between components
         // channel to queue digests to be aligned
-        let (tx_digest, rx_digest) = flume::unbounded();
+        let (tx_digest, rx_digest) = flume::bounded(10);
         // channel for aligner to send missing samples to storage
         let (tx_sample, rx_sample) = flume::unbounded();
         // channel for storage to send logging information back
@@ -247,9 +247,12 @@ impl Replica {
                 .await;
             if to_be_processed {
                 log::trace!("[DIGEST_SUB] sending {} to aligner", digest.checksum);
-                match tx.send_async((from.to_string(), digest)).await {
+                match tx.try_send((from.to_string(), digest)) {
                     Ok(()) => {}
-                    Err(e) => log::error!("[DIGEST_SUB] Error sending digest to aligner: {}", e),
+                    Err(e) => {
+                        // Trace because this can happen _a lot_ on busy channels.
+                        log::trace!("[DIGEST_SUB] Error sending digest to aligner: {}", e)
+                    }
                 }
             };
             received.insert(from.to_string(), ts);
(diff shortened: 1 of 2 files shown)


plugins/zenoh-plugin-storage-manager/src/replica/mod.rs
```diff
@@ -247,9 +247,12 @@ impl Replica {
...
             if to_be_processed {
                 log::trace!("[DIGEST_SUB] sending {} to aligner", digest.checksum);
-                match tx.send_async((from.to_string(), digest)).await {
+                match tx.try_send((from.to_string(), digest)) {
                     Ok(()) => {}
-                    Err(e) => log::error!("[DIGEST_SUB] Error sending digest to aligner: {}", e),
+                    Err(e) => {
+                        // Trace because this can happen _a lot_ on busy channels.
+                        log::trace!("[DIGEST_SUB] Error sending digest to aligner: {}", e)
+                    }
```
J-Loudet, line 255: I need to further investigate the implications of this change. As of today, I do not know if skipping digests could have unforeseen consequences on the replication (my first guess is that it doesn't but I want to make sure).

J-Loudet (changes requested):

plugins/zenoh-plugin-storage-manager/src/replica/mod.rs
```diff
@@ -119,7 +119,7 @@ impl Replica {
 
         // Create channels for communication between components
         // channel to queue digests to be aligned
-        let (tx_digest, rx_digest) = flume::unbounded();
+        let (tx_digest, rx_digest) = flume::bounded(10);
```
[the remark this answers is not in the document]
chachi: No, this is entirely a speculative first stab at an improvement. I certainly agree that a configurable value would be better than hardcoding anything. 

Truthfully, it feels like it would be better for this buffer management to be done entirely at the Zenoh level so that the backpressure happens when a Publisher tries to send its digest and it gets dropped because there's no space to receive. 

Even better, frankly, would be to remove this middle queue entirely and just have all the digest processing happen on recv from the digest-sub. I'm not entirely sure what the value of doing just the JSON parsing and `is_processed` checking separately from the rest of the handling is.

plugins/zenoh-plugin-storage-manager/src/replica/mod.rs
chachi: Yup, it's not a small change. Ultimately this system needs to have some sort of backpressure and dropping because as a network of storages grows, if anything is out of sync it's impossible to parse and process _every_ digest w/o dropping.

chachi:
Overall, this is more of a proof-of-concept PR than anything else. @J-Loudet if you'd like to just close it and make an issue, feel free to.

J-Loudet:
See #937

PR #888, opened 2024-04-01
Fix issue #887

### LINKED ISSUE #887: Enable default features in zenoh-ext/examples
### Describe the bug

Similar issue as #854. If we run examples under _zenoh-ext_ instead of workspace, then the default features of zenoh are missing to include.

### To reproduce

```bash
cd zenoh-ext
cargo build --example z_pub_cache --example z_query_sub
cd ..
# Terminal 1
./target/debug/examples/z_pub_cache
# Terminal 2
./target/debug/examples/z_query_sub
```

Symptom: The querying side never receives a message.

### System info

- OS: Arch Linux x86_64
- Kernel: 6.8.2-arch2-1

YuanYuYuan:
Closes #887.

### DIFF
--- a/zenoh-ext/Cargo.toml
+++ b/zenoh-ext/Cargo.toml
@@ -38,7 +38,7 @@ flume = { workspace = true }
 futures = { workspace = true }
 log = { workspace = true }
 serde = { workspace = true, features = ["default"] }
-zenoh = { workspace = true, features = ["unstable"], default-features = false }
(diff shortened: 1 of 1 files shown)


Mallets (approved):

Mallets:
It seems this PR makes zenoh-c CI fail: https://github.com/eclipse-zenoh/zenoh-c/actions/runs/8562959546/job/23467191384

PR #902, opened 2024-04-04
Declare message can be Push/Request/RequestContinuous/Response

Mallets:


commons/zenoh-codec/src/network/declare.rs
```diff
@@ -104,9 +108,13 @@ where
 
         // Header
         let mut header = id::DECLARE;
-        if x.interest_id.is_some() {
-            header |= declare::flag::I;
-        }
+        header |= match mode {
```
OlivierHecart, line 111: Just a suggestion: Could having both `Request` and `RequestContinuous` have the heigh weight bit to 1 allow to easily distinguish Requests in some cases ?

OlivierHecart (changes requested):

commons/zenoh-protocol/src/network/declare.rs
```diff
@@ -668,60 +726,66 @@ pub mod interest {
...
     ///     |                   |
     /// ```
     ///
     /// The DECLARE INTEREST message structure is defined as follows:
     ///
     /// ```text
     /// Flags:
-    /// - C: Current        if C==1 then the interest refers to the current declarations.
+    /// - X: Reserved
     /// - F: Future         if F==1 then the interest refers to the future declarations. Note that if F==0 then:
```
fuzzypixelz, line 782: What's the difference between setting `F==1` on `DeclareInterest` and setting `Mode==Push` on `Declare`?

fuzzypixelz: (Meant to select the next line, can't edit it)

Mallets: That's a good point.

DeclareInterest has 3 combinations:
1. Only current
2. Only future
3. Current and future

These could be achieved via:
1. Declare(Request)::Interest -> Only current
2. Declare(Push)::Interest -> Only future
3. Declare(RequestContinous) -> Current and future

If that's the case, we might not even need the `F` flag for future and it can be removed. 
Chiming in @OlivierHecart. What do you think?

OlivierHecart: Indeed we should not need Current neither Future flags any more.

Mallets: Ok, flag removed.

PR #903, opened 2024-04-04
RingChannel sync/async/blocking/non-blocking

Mallets:
Improve the current RingChannel handler implementation to be sync/async/blocking/non-blocking.


sashacmc (approved):
LGTM, what about examples for query/reply?

Mallets:
I've added a comment in both `z_get` and `z_queryable` examples. I don't think there is the need to add dedicate examples.

PR #906, opened 2024-04-05
Declare message can be Push/Request/RequestContinuous/Response 2

OlivierHecart:


commons/zenoh-protocol/src/network/declare.rs
```diff
@@ -760,6 +760,26 @@ pub mod interest {
     ///     |                   |
     ///     |       FINAL       |
     ///     |<------------------| -- Sent in Declare::Response
+    /// ```
+    ///
+    ///
+    /// And the [`DeclareInterest`] message flow is the following for a ContinuousRequest:
```
Mallets, line 766:
```suggestion
    /// And the [`DeclareInterest`] message flow is the following for a RequestContinuous:
```

Mallets (approved):

PR #929, opened 2024-04-12
make Reply fields pub(crate) and add accessors

DenisBiryukov91:
make Reply fields pub(crate) and add accessors

### DIFF
--- a/zenoh/src/query.rs
+++ b/zenoh/src/query.rs
@@ -131,7 +131,7 @@ pub(crate) struct QueryState {
 ///     .await
 ///     .unwrap();
 /// while let Ok(reply) = replies.recv_async().await {
-///     println!("Received {:?}", reply.sample())
+///     println!("Received {:?}", reply.result())
 /// }
 /// # }
 /// ```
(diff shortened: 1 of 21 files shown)


zenoh/src/query.rs
```diff
@@ -77,10 +77,33 @@ impl Default for QueryConsolidation {
...
+        self.sample.as_ref()
+    }
+
+    /// Gets the id of the zenoh instance that answered this Reply.
+    pub fn replier_id(&self) -> ZenohId {
+        self.replier_id
+    }
+
+    /// Attempts to convert reply into sample.
+    pub fn try_into_sample(self) -> Result<Sample, Value> {
```
Mallets, line 96: I believe this method is not needed since `impl TryFrom<Reply> for Sample` is already provided.

DenisBiryukov91: Yes, but it becomes too verbose to write, when compiler requires explicit type annotation like TryInto::<Sample>::try_into(reply) instead of just reply.try_into_sample()

Mallets: Correct, but IMHO it's a proliferation of unnecessary methods. Should we add all `try_into_x` implementations? This specific case would be an exception in the whole codebase.

In general the `try_from` is preferred to `try_into` in the same way implementing `From` is preferred to implementing `Into` (and `From` derives `Into`). Having the user doing `Sample::try_from(reply)` doesn't seem much of a burden.

However, the main problem I have with this is the asymmetry between `.sample()` that returns `Result<&Sample, &Value>` and  `.try_into_sample()` which returns `Result<Sample, Value>`. Should then `.sample()` renamed `.try_sample()`? and `.try_into_sample()` as `.try_sample_into()`?

Mallets: As agreed offline with @DenisBiryukov91 , @sashacmc , and @J-Loudet `.sample()` and `.try_into_sample()` have been renamed to `.result()` and `.into_result()`.

Mallets (approved):

PR #930, opened 2024-04-12
make receiver fields pub(crate) in Subscriber, Queryable, Scout and MatchingListener

DenisBiryukov91:
make receiver fields pub(crate) in Subscriber, Queryable, Scout and MatchingListener

### DIFF
--- a/plugins/zenoh-plugin-rest/examples/z_serve_sse.rs
+++ b/plugins/zenoh-plugin-rest/examples/z_serve_sse.rs
@@ -12,7 +12,6 @@
 //   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
 //
 use clap::{arg, Command};
-use std::ops::Deref;
 use std::time::Duration;
 use zenoh::prelude::r#async::*;
 use zenoh::publication::CongestionControl;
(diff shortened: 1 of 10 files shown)


plugins/zenoh-plugin-rest/examples/z_serve_sse.rs
```diff
@@ -46,7 +47,7 @@ async fn main() {
     let queryable = session.declare_queryable(key).res().await.unwrap();
 
     async_std::task::spawn({
-        let receiver = queryable.receiver.clone();
+        let receiver = queryable.deref().clone();
```
Mallets, line 50: I find odd for the user to actually call `.deref()` to get the `Handler` of the subscriber.
In this particular case, the default `Handler` is a `flume::Receiver` but it could be something else.

IMHO we should add `.handler()` and `.handler_mut()` methods to `subscriber`, `queryable`, etc. that allows the user to explicitly retrieve the actual handler.

Mallets: As discussed offline with @DenisBiryukov91 and https://github.com/wyfo, I've added `.handler()` and `.handler_mut()` to both `subscriber` and `queryable`.

Mallets (approved):

PR #932, opened 2024-04-12
Protocol properties

milyin:


zenoh/src/selector.rs
```diff
@@ -66,377 +66,200 @@ use std::{
...
-    }
-    /// Extracts the selector parameters' name-value pairs into a hashmap, returning an error in case of duplicated parameters.
-    pub fn parameters_stringmap(&'a self) -> ZResult<HashMap<String, String>> {
-        self.decode_into_map()
-    }
+
     /// Gets a mutable reference to the parameters as a String.
     ///
     /// Note that calling this function may cause an allocation and copy if the selector's parameters wasn't
     /// already owned by `self`. `self` owns its parameters as soon as this function returns.
```
milyin, line 99: This comment isn't true anymore

zenoh/src/selector.rs
```diff
@@ -66,377 +66,200 @@ use std::{
...
-        let it = parameters.into_iter();
-        let selector = self.parameters_mut();
-        let mut encoder = form_urlencoded::Serializer::new(selector);
-        encoder.extend_pairs(it).finish();
+    #[zenoh_macros::unstable]
+    /// Extracts the standardized `_time` argument from the selector parameters.
+    ///
+    /// The default implementation still causes a complete pass through the selector parameters to ensure that there are no duplicates of the `_time` key.
+    pub fn time_range(&self) -> ZResult<Option<TimeRange>> {
+        self.parameters().time_range()
```
milyin, line 137: Do we really need these functions now, as the publicly available `.parameters()`/`.parameters_mut()` supports them too? Same about `accept_any_keyexpr`

Mallets: I've marked them as `unstable` since it's unclear yet the impact on the rest. `accept_any_keyexpr` could be removed since it's `pub(crate)`, `time_range()` also performs deserialization of the parameter if present. The impact of removing `time_range()` should be further investigated.

PR #935, opened 2024-04-15
fix: Replace macos-latest with macos-13 and macos-14 runners

fuzzypixelz:
Clippy from Rust 1.77.2 has been failing on maco-latest for a while. I suspect it might have something to do with https://github.com/rust-lang/rust/issues/92173. Except in our case the failure happens 100% of the time and is not sporadic.

What I know so far:

1. Even though we specify macos-latest, GitHub simply uses macos-12 images, for some reason I'm not too keen on knowing.
2. The command that fails is `cargo +stable clippy --all-targets` and not `cargo +stable build --all-targets` and not `cargo +stable clippy`.
3. Every time clippy fails out of a SIGABRT, it's because of failure to compile some zenoh crate, often for test targets. It seems that this can be any crate, which helps the hypothesis that this is a sporadic failure made worse by zenoh having 34 crates.
3. I cannot reproduces this on macOS 13, neither on AArch64 nor on x86_64.

This pull request tries to force GitHub to give us macOS 13/14 to see what happens. The idea is that Rust 1.77.2 doesn't play too nicely with the C runtime in macOS 12, which would explain why it doesn't fail on macOS 13 or 14 for me (locally).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macOS-latest, windows-latest]
+        os: [ubuntu-latest, windows-latest, macos-12, macos-13, macos-14]
 
     steps:
       - name: Clone this repository
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-12, macos-13, macos-14]
+        os: [ubuntu-latest, windows-latest, macos-13, macos-14]
 
     steps:
       - name: Clone this repository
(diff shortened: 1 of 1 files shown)


fuzzypixelz:
Okay. Clippy 1.77.2 fails on macOS 12 but works on macOS 13 and 14. [GitHub is not supposed to be swapping macOS 14 for macOS 12](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories). They should update their documentation.

I'm just going to keep both macOS 13 and 14 because __if we can trust the GitHub docs__ macos-14 runners are M1 machines, which would be useful for running tests on AArch64 in the CI.

fuzzypixelz:
EDIT: [GitHub is in the process of migrating macos-latest to use macos-14](https://github.blog/changelog/2024-04-01-macos-14-sonoma-is-generally-available-and-the-latest-macos-runner-image/). There was no mention of this in the docs though...

Mallets:
@fuzzypixelz should we merge this PR for the time being or should we wait for the migration?

fuzzypixelz:
> @fuzzypixelz should we merge this PR for the time being or should we wait for the migration?

I think we can just merge it. We'd like to have both macOS 13 and 14 anyway.

Mallets (approved):

PR #936, opened 2024-04-15
replaced unwraps for plugin configurations insert to correct error check

### LINKED ISSUE #834: zenoh crashes on write access to adminspace for control zenoh-backend-filesystem
### Describe the bug

Copied from discord message from @oteffahi:

The backend tests follow the README instructions to setup the zenoh-backend-filesystem at runtime via curl commands on the admin space
so it runs zenohd with `zenohd --adminspace-permissions rw`
and then runs the curl command
```
curl -X PUT -H 'content-type:application/json' -d '{}' http://localhost:8000/@/router/local/config/plugins/storage_manager/
volumes/fs
```
 
to load the library 
However, the zenohd instance crashes after receiving this request, with the following error:
```
thread '<unnamed>' panicked at /root/temp/backend-inv/zenoh/commons/zenoh-config/src/lib.rs:1173:35:
called `Option::unwrap()` on a `None` value
fatal runtime error: Rust cannot catch foreign exceptions
Aborted
 ```
As for the curl command, it returns this error:
curl: (52) Empty reply from server
```
Oussama Teffahi — Today at 5:10 PM
We have found a workaround to load the library correctly, which consists of running this curl command before the other one
```
curl -X PUT -H 'content-type:application/json' -d "{}" http://localhost:8000/@/router/local/config/plugins
```

The output zenohd when receiving this request is:
```
[2024-03-14T16:10:44Z INFO  tide::log::middleware] <-- Request received
[2024-03-14T16:10:44Z INFO  tide::log::middleware] --> Response sent
[2024-03-14T16:10:44Z DEBUG zenoh_plugin_trait::manager::dynamic_plugin] Plugin `rest` stopped
```


### To reproduce

-

### System info

-

milyin:
Closes #834

### DIFF
--- a/commons/zenoh-config/src/lib.rs
+++ b/commons/zenoh-config/src/lib.rs
@@ -1192,19 +1192,17 @@ impl validated_struct::ValidatedMap for PluginsConfig {
             // Note: it's ok if original "new_value" is not an object: this can be some subkey of the plugin configuration. But the result of the merge should be an object.
             // Buf this may happen if original plugin configuration is not an object itself (e.g. null).
             let Some(new_plugin_config) = new_value.as_object() else {
-                return Err(format!("Attempt to provide non-object value as configuration for plugin `{plugin}`").into());
+                return Err(format!(
+                    "Attempt to provide non-object value as configuration for plugin `{plugin}`"
+                )
+                .into());
             };
             // Original plugin configuration for compare with new configuration.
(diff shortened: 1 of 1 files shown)


Mallets (approved):

milyin:
Fixing this issue uncovered this one https://github.com/eclipse-zenoh/zenoh/issues/944

PR #942, opened 2024-04-17
refactor: replace `AsyncResolve` with `IntoFuture`

wyfo:
Into the future with `IntoFuture`! :D

The issue was that using `IntoFuture` alone was not possible because there is no bound on `IntoFuture::IntoFuture`, and we need to have a `Send` bound, particularly with `impl Resolve<...>` returns. See #349.

Adding `where <Self as IntoFuture>::IntoFuture: Send` bound to `Resolve` was not working, because `where` bound must be repeated anywhere it is used, and there is no way to add `where` bound to returned `impl Resolve<...>` – surprisingly, I've not found any resource on the internet about adding `where` bound to a returned `impl Trait`...

The solution used here is to add this bound using the additional `IntoSendFuture` trait, which add it to its associated `Future` type, and this one into `IntoFuture` bound like this `IntoFuture<IntoFuture = <Self as IntoSendFuture>::Future, Output = Output>`.


gabrik:
@wyfo can you run a `cargo fmt` so that the CI can run?

gabrik:
apparently `clippy` is not happy

wyfo:
By simply replacing `AsyncResolve` by `IntoFuture` (and adding `where <Self as IntoFuture>::IntoFuture: Send` bound to `Resolve`), I faced *`Send` propagation issues*.
The trick was to reintroduce `AsyncResolve` in this new form, to bind its associated future with `Send` bound to `IntoFuture` one.

wyfo:
Sorry, I have not set up `pre-commit` for now on this project... 😓

Mallets:
This PR is a followup of an old PR: https://github.com/eclipse-zenoh/zenoh/pull/349. 
It seems that this PR solves the trait bounds issue of the original PR. 

However, before merging it is necessary to check at least the impact on the various language bindings and plugins in the Zenoh project. Issues with the original PR where highlighted by applications using Zenoh as a library, like indeed bindings and plugins.

wyfo:
Internal plugins like REST have been updated in this PR, and have been tested: they load fine.

YuanYuYuan:
If this proposal is accepted, then we may remove `AsyncResolve` and modify the calling of `SyncResolve` by `.wait()`, so that we have the consistency as below.
```rust
let sync_session = Session::open().wait();
let async_session = Session::open().await;

**Comment by wyfo** [comment]
Does sync operations support timeouts? In that case, it could be interesting to add some `wait_timeout`, similarly to [`Condvar::wait_timeout`](https://doc.rust-lang.org/std/sync/struct.Condvar.html#method.wait_timeout).

**Comment by wyfo** [comment]
I've edited the PR to add an explanation.

**Comment by Mallets** [comment]
I would still like to keep the `.res()`, `.res_sync`, and `.res_async` and add a `#[deprecated]` attribute for the user.

**Comment by wyfo** [comment]
I've added deprecation attributes in the last commit, as well as updated the documentation.
I've tested the branch into `zenoh-plugin-dds`, see the POC [commit](https://github.com/eclipse-zenoh/zenoh-plugin-dds/commit/5fc5d015379862de27cfa837e6d7e0047d7e8062). Everything seems to work fine, I've let some `res_sync`/`res_async` to observe deprecation messages.

**Comment by wyfo** [comment]
There is an issue with shared memory part, which directly defines `res`/`res_sync`/`res_async` methods. I assume I've to update them to implement the `Resolvable` protocol.

**Comment by wyfo** [comment]
@Mallets It seems to be ok to be merged

File: examples/examples/z_get.rs [hunk 1]
```diff
@@ -24,19 +24,18 @@ async fn main() {
...
 
     println!("Sending Query '{selector}'...");
     let replies = session
         .get(&selector)
-        // // By default get receives replies from a FIFO. 
-        // // Uncomment this line to use a ring channel instead. 
+        // // By default get receives replies from a FIFO.
+        // // Uncomment this line to use a ring channel instead.
         // // More information on the ring channel are available in the z_pull example.
         .with(zenoh::handlers::RingChannel::default())
```
Mallets, line 35: I think it slipped in from some other PR but this line should be commented out.

```suggestion
        // .with(zenoh::handlers::RingChannel::default())
```

wyfo: I've found the culprit 😛 https://github.com/eclipse-zenoh/zenoh/pull/903/files#diff-de144b4bbbdf20af6b0b2ed3a9fc9301218263abb726d71700859c1c270caeecR36

Mallets (approved):

PR #949, opened 2024-04-18
Introduce rust-cache

YuanYuYuan:
### Spending time of two consecutive runs

| Job                          | 1st Run | 2nd Run |
| -                            | -       | -       |
| Run checks on macos-13       | 5m48s   | 2m23s   |
| Run checks on macos-14       | 2m50s   | 1m23s   |
| Run checks on ubuntu-latest  | 3m28s   | 2m31s   |
| Run checks on windows-latest | 7m24s   | 5m12s   |
| Run tests on macos-13        | 15m8s   | 13m37s  |
| Run tests on macos-14        | 10m23s  | 10m21s  |
| Run tests on ubuntu-latest   | 14m54s  | 14m59s  |
| Run tests on windows-latest  | 21m26s  | 19m13s  |

P.S. This cache trick is also used in tokio.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,6 +81,9 @@ jobs:
       - name: Install latest Rust toolchain
         run: rustup show
 
+      - name: Setup rust-cache
+        uses: Swatinem/rust-cache@v2
+
       - name: Set rustflags
         shell: bash
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,9 @@ jobs:
           rustup show
           rustup component add rustfmt clippy
 
+      - name: Setup rust-cache
+        uses: Swatinem/rust-cache@v2
+
       - name: Code format check
         run: cargo fmt --check
 


Mallets (approved):

YuanYuYuan:
## 1st Run

X  Run tests on ubuntu-latest in 12m40s 
✓ Run tests on macos-13 in 13m54s 
✓ Run tests on macos-14 in 10m49s 
✓ Run tests on windows-latest in 21m46s 

## 2nd Run

X  Run tests on ubuntu-latest in 12m49s
✓ Run tests on macos-13 in 17m43s
✓ Run tests on macos-14 in 15m57s
✓ Run tests on windows-latest in 18m38s

P.S. The failures are caused by valgrind installation.

- The compilation time of cargo nextest is improved.
- The overall speedup on tests is not obvious because the tests sometimes run longer with slow cases.

Anyway, let's add the cache mechanism to test as well.

YuanYuYuan:
The check required by auto-merge is blocked by #955 .

fuzzypixelz:
@YuanYuYuan [Looking at the cache usage so far, we have already exceeded the limit (10GB). Each cache miss means 100MB-400MB of space usage, so you can see how that quickly adds up.](https://github.com/eclipse-zenoh/zenoh/actions/caches)

I think what's hurting us the most are `Cargo.toml` changes. There are many manifests (41 to be precise) so each time a pull request commit changes one of them we suffer a cache miss, even if the workspace dependencies have not changed.

We might want to fine-tune the strategy further to get more stable keys. For example, if we enforce dependency declaration in the workspace virtual manifest (i.e. in the root `Cargo.toml`) and nowhere else, we can be sure that any change to other manifests does not change our dependencies (however, I'm not sure if rust-cache caches dependencies outside the workspace or if it caches zenoh crates as well...)

PR #951, opened 2024-04-18
CI: add tests on Ubuntu 24.04 to pre-release workflow

oteffahi:


.github/workflows/pre-release.yml
```diff
@@ -77,6 +77,95 @@ jobs:
           CARGO_REGISTRIES_CRATES_IO_PROTOCOL: sparse
           ASYNC_STD_THREAD_COUNT: 4
 
+  checks-ubuntu-24:
+    name: Code checks on Ubuntu 24.04
+    runs-on: ubuntu-latest
+    container: ubuntu:24.04
```
fuzzypixelz, line 83: To avoid all the code duplication, you can add a matrix over multiple container images. This way we can have less code and more flexibility. I suggest putting the following images:

- `ubuntu:24.04`
- `ubuntu:22.04`
- `ubuntu:20.04`

You can also experiment some Alpine, RedHat etc.

.github/workflows/pre-release.yml
```diff
@@ -77,6 +77,95 @@ jobs:
...
+        
+      - name: Install Rust toolchain
+        run: curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+
+      - name: Clone this repository
+        uses: actions/checkout@v4
+      
+      - name: Add Zenoh Rust toolchain version
+        run: |
+          . "$HOME/.cargo/env"
```
fuzzypixelz, line 143: You can replace this with `echo "$HOME/.cargo/bin" >> $GITHUB_PATH`

.github/workflows/pre-release.yml
```diff
@@ -77,6 +77,95 @@ jobs:
...
+      - name: Install Rust toolchain
+        run: curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+
+      - name: Clone this repository
+        uses: actions/checkout@v4
+
+      - name: Add Zenoh Rust toolchain version
+        run: |
+          . "$HOME/.cargo/env"
+          rustup override set $(cat rust-toolchain.toml | grep '^channel' | sed 's/[^"]*"//' | sed 's/"//')
```
fuzzypixelz, line 97: Is this really necessary? Won't `rustup` install the toolchain from the file by itself?

.github/workflows/pre-release.yml
oteffahi: The issue is that if we add a matrix it would mean that all runs will occur on containers, but we want the base workflow to run on baremetal, which is why I made a copy.

.github/workflows/pre-release.yml
oteffahi: I remember testing and noticing that it does not install it. But I just tested again and it works 😅

.github/workflows/pre-release.yml
fuzzypixelz: [GitHub-hosted runners are VMs anyway.](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#overview-of-github-hosted-runners)

fuzzypixelz: And AFAICT, there is no additional virtualization imposed by Docker in this case, so the only real difference is the OS.

oteffahi:
@fuzzypixelz could you please review?

fuzzypixelz (approved):
LGTM.

Mallets (approved):

oteffahi:
CI failed because of an issue with Ubuntu mirrors. Had to push another commit to trigger CI, which disabled auto-merge...
@Mallets could you please re-enable it?

PR #954, opened 2024-04-19
Fix tuple deserialization lifetime

Mallets:
@DenisBiryukov91 can you please check it works as you would expect in zenoh-c?


Mallets:
@DenisBiryukov91 and I have checked offline this PR solves the problem of in the new zenoh-c related to generic lifetimes.

PR #959, opened 2024-04-22
Admin space: Show known origins of Subscribers and Queryables.

OlivierHecart:
When reporting it's known Subscribers and Queryables, a router also reports the known source nodes of those Subscribers and Queryables. 
The reported information is local. The list of sources may not be exhaustive in the whole system and some sources may be intermediate (typically a router may be reported as the source of a subscription while in fact one or more of the clients served by this router are the real source of the subscription).
Those informations are valuable in a monitoring/debugging tool for example in the PaaS.

Known limitation: linkstate peers will report router sources in the list of peers sources. They don't necessarily have the information to distinguish peers and routers in the local mesh. Linkstate peers don't have an admin space yet anyway.

### DIFF
--- a/zenoh/src/net/routing/hat/linkstate_peer/pubsub.rs
+++ b/zenoh/src/net/routing/hat/linkstate_peer/pubsub.rs
@@ -19,7 +19,7 @@ use crate::net::routing::dispatcher::pubsub::*;
 use crate::net::routing::dispatcher::resource::{NodeId, Resource, SessionContext};
 use crate::net::routing::dispatcher::tables::Tables;
 use crate::net::routing::dispatcher::tables::{Route, RoutingExpr};
-use crate::net::routing::hat::HatPubSubTrait;
+use crate::net::routing::hat::{HatPubSubTrait, Sources};
 use crate::net::routing::router::RoutesIndexes;
 use crate::net::routing::{RoutingContext, PREFIX_LIVELINESS};
 use petgraph::graph::NodeIndex;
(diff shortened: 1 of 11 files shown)


gabrik (approved):
LGTM

zenoh/src/net/runtime/adminspace.rs
```diff
@@ -682,11 +689,18 @@ fn queryables_data(context: &AdminContext, query: Query) {
...
-            qabl.expr()
+            qabl.0.expr()
         ))
         .unwrap();
         if query.key_expr().intersects(&key) {
-            if let Err(e) = query.reply(Ok(Sample::new(key, Value::empty()))).res() {
+            if let Err(e) = query
+                .reply(Ok(Sample::new(
+                    key,
+                    Value::from(serde_json::to_string(&qabl.1).unwrap_or("{}".to_string()))
```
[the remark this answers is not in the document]
Mallets: `unwrap_or` eagerly evaluates the expression, that means the `"{}".to_string()` is always called.

zenoh/src/net/routing/hat/router/pubsub.rs
```diff
@@ -925,8 +925,38 @@ impl HatPubSubTrait for HatCode {
...
+    fn get_subscriptions(&self, tables: &Tables) -> Vec<(Arc<Resource>, Sources)> {
+        hat!(tables)
+            .router_subs
+            .iter()
+            .map(|s| {
+                (
+                    s.clone(),
+                    Sources {
+                        routers: Vec::from_iter(res_hat!(s).router_subs.iter().cloned()),
+                        peers: if hat!(tables).full_net(WhatAmI::Peer) {
```
Mallets, line 937: Please add comments and what's happening here. The higher-level logic is not clear from this code.

zenoh/src/net/routing/hat/linkstate_peer/pubsub.rs
```diff
@@ -605,8 +605,28 @@ impl HatPubSubTrait for HatCode {
...
-    fn get_subscriptions(&self, tables: &Tables) -> Vec<Arc<Resource>> {
-        hat!(tables).peer_subs.iter().cloned().collect()
+    fn get_subscriptions(&self, tables: &Tables) -> Vec<(Arc<Resource>, Sources)> {
+        hat!(tables)
+            .peer_subs
+            .iter()
+            .map(|s| {
+                (
+                    s.clone(),
+                    Sources {
```
Mallets, line 615: Please add comments and what's happening here. The higher-level logic is not clear from this code.

Mallets (approved):

PR #962, opened 2024-04-22
Fix the downsampling test on Windows

### LINKED ISSUE #957: The downsampling test often fails on Windows CI.
### Describe the bug

Check https://github.com/eclipse-zenoh/zenoh/actions/runs/8778673304/job/24085495910?pr=956#step:6:975

### To reproduce

Skip

### System info

Skip

YuanYuYuan:
Closes #957


YuanYuYuan:
@sashacmc and I  had a discussion and concluded the cause might be related to the runtime performance. This PR makes the downsampling test run individually. But it's worth noting that it still has the chance to fail on some slow tests on Windows (depending on the status of GitHub runner).

Mallets (approved):

PR #971, opened 2024-04-25
Reduce open time

OlivierHecart:


fuzzypixelz:
@OlivierHecart You can add me as a reviewer.

zenoh/tests/open_time.rs
```diff
@@ -0,0 +1,426 @@
...
+    ztimeout_expected!(router.close().into_future()).unwrap();
+    println!(
+        "close(mode:{}, listen_endpoint:{}, lowlatency:{}): {:#?}",
+        WhatAmI::Router,
+        listen_endpoint.as_str().split('#').next().unwrap(),
+        lowlatency,
+        start.elapsed()
+    );
+
+    // Wait a little bit
```
fuzzypixelz, line 122: Why?

oteffahi: @fuzzypixelz I believe it's considered common practice to have some delay after closing a session, we do this is all tests. This is to avoid issues like https://github.com/eclipse-zenoh/zenoh/issues/1019

OlivierHecart: Yes. Btw it's copy pasted from there: https://github.com/eclipse-zenoh/zenoh/blob/912bd0c0fb728e2c66a40df0ec3b64d26239fc0b/io/zenoh-transport/tests/unicast_time.rs#L218

oteffahi (approved):
Ran [these tests](https://github.com/ZettaScaleLabs/zenoh-perf/tree/open-session/protocol-changes-tests/throughput) locally. As per our exchanges through private channels, the behavior for peers was changed in https://github.com/eclipse-zenoh/zenoh/pull/971/commits/ad0362c2b44ae03191ba380e8ab6d8c2bf761f09, and the absence of delay for routers was fixed in https://github.com/eclipse-zenoh/zenoh/pull/971/commits/70b9cb75128603a2b92842b45200744ee192b275.
The remaining cases in the behavior table were confirmed through the same tests.

LGTM.

PR #972, opened 2024-04-25
fix(zenoh-runtime): zenoh-c DLL crash in `libc::atexit` handler

YuanYuYuan:
See #973 for the background of the issue.

In short, we need to disable the `atexit` callback while building to DLL on Windows.

It's hard to do conditional compilation on `zenoh-runtime` crate only since `zenoh-runtime` is a static lib to `zenoh`. To prevent introducing too complicated feature flags affecting both `zenoh` and `zenoh-c`, let's catch the panic error and process it accordingly.

To be clear,

1. On Windows, `zenoh-c` in DLL: panic is expected.
2. On Windows, `zenoh-c` in lib: panic shouldn't happen.
3. Any other cases should report the error.


YuanYuYuan:
This PR has been discussed in the chat. @Mallets can you help review this?

Mallets (approved):

PR #974, opened 2024-04-25
refactor: use `Into` bound for `KeyExpr::with_parameters`

wyfo:
@Mallets @milyin

PR #979, opened 2024-04-25
Remove keyexpr with_parameters

Mallets:
Related to https://github.com/eclipse-zenoh/zenoh/pull/974

### DIFF
--- a/plugins/zenoh-plugin-storage-manager/src/replica/storage.rs
+++ b/plugins/zenoh-plugin-storage-manager/src/replica/storage.rs
@@ -638,7 +638,7 @@ impl StorageService {
             // with `_time=[..]` to get historical data (in case of time-series)
             let replies = match self
                 .session
-                .get(Selector::new(KeyExpr::from(&self.key_expr), "_time=[..]"))
+                .get(Selector::new(&self.key_expr, "_time=[..]"))
                 .target(QueryTarget::All)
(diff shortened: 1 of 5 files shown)


plugins/zenoh-plugin-storage-manager/src/replica/storage.rs
```diff
@@ -638,7 +638,7 @@ impl StorageService {
             // with `_time=[..]` to get historical data (in case of time-series)
             let replies = match self
                 .session
-                .get(KeyExpr::from(&self.key_expr).with_parameters("_time=[..]"))
+                .get(Selector::new(KeyExpr::from(&self.key_expr), "_time=[..]"))
```
wyfo, line 641:
```suggestion
                .get(Selector::new(&self.key_expr, "_time=[..]"))
```

wyfo (commented):
Honestly, I would prefer to keep this API, because it can be more convenient to chain calls, for example if you retrieve the keyexpr from another object.
But I've no experience about API use, so my opinion doesn't count a lot.

Mallets:
It's more about removing unnecessary API that needs to be supported in the various bindings. I don't see much difference in the actual API usage in terms of complexity (at least as it was used in the zenoh codebase) and it will remove the need of eexposing the method on all other languages.

PR #981, opened 2024-04-26
fix(zenoh_runtime): disable atexit on windows

YuanYuYuan:
As discussed today, we found that the previous is still problematic. Let's disable the atexit cleanup for the time being.


YuanYuYuan:
The test on zenoh-c seems good. https://github.com/YuanYuYuan/zenoh-c/actions/runs/8846318062/job/24291948644.

@fuzzypixelz can you help review this PR?

fuzzypixelz (approved):
LGTM. We shouldn't spawn threads in the `atexit` handler on Windows. Until we figure out how implement it correctly, this'll have to do :)

Mallets (approved):

PR #982, opened 2024-04-26
fix bug in acl config validation

snehilzs:
This PR fixes input validation issue in access control config.

### DIFF
--- a/zenoh/src/net/routing/interceptor/access_control.rs
+++ b/zenoh/src/net/routing/interceptor/access_control.rs
@@ -66,7 +66,7 @@ pub(crate) fn acl_interceptor_factories(
                     enforcer: Arc::new(policy_enforcer),
                 }))
             }
-            Err(e) => tracing::error!("Access control not enabled due to: {}", e),
+            Err(e) => bail!("Access control not enabled due to: {}", e),
         }
     } else {
         tracing::debug!("Access control is disabled");
--- a/zenoh/src/net/routing/interceptor/access_control.rs
+++ b/zenoh/src/net/routing/interceptor/access_control.rs
@@ -66,7 +66,7 @@ pub(crate) fn acl_interceptor_factories(
                     enforcer: Arc::new(policy_enforcer),
                 }))
             }
-            Err(e) => tracing::error!("Access control inizialization error: {}", e),
+            Err(e) => tracing::error!("Access control not enabled due to: {}", e),
         }
     } else {
         tracing::debug!("Access control is disabled");
--- a/zenoh/src/net/routing/interceptor/authorization.rs
+++ b/zenoh/src/net/routing/interceptor/authorization.rs
@@ -200,12 +200,21 @@ impl PolicyEnforcer {
         let mut policy_rules: Vec<PolicyRule> = Vec::new();
         for config_rule in config_rule_set {
             // config validation
-            if config_rule.interfaces.is_empty()
-                || config_rule.actions.is_empty()
-                || config_rule.flows.is_empty()
-                || config_rule.key_exprs.is_empty()
-            {
-                bail!("error from bad config");
+            let mut validation_err = String::new();
+            if config_rule.interfaces.is_empty() {
+                validation_err.push_str("ACL config interfaces list is empty. ");
+            }
+            if config_rule.actions.is_empty() {
+                validation_err.push_str("ACL config rules list is empty. ");
+            }
+            if config_rule.flows.is_empty() {
+                validation_err.push_str("ACL config flows list is empty. ");
+            }
+            if config_rule.key_exprs.is_empty() {
+                validation_err.push_str("ACL config key_exprs list is empty. ");
+            }
+            if !validation_err.is_empty() {
+                bail!("{}", validation_err);
             }
             for subject in &config_rule.interfaces {
                 if subject.trim().is_empty() {
--- a/zenoh/src/net/routing/interceptor/authorization.rs
+++ b/zenoh/src/net/routing/interceptor/authorization.rs
@@ -200,11 +200,12 @@ impl PolicyEnforcer {
         config_rule_set: &Vec<AclConfigRules>,
     ) -> ZResult<PolicyInformation> {
         let mut policy_rules: Vec<PolicyRule> = Vec::new();
-        for config_rule in config_rule_set {
+        for rule_offset in 0..config_rule_set.len() {
             // config validation
-            let mut validated_rule = config_rule.clone();
+            let mut validated_rule = config_rule_set[rule_offset].clone();
             let mut validation_err = String::new();
             if validated_rule.interfaces.is_empty() {
+                tracing::warn!("ACL config interfaces list is empty. Applying rule #{} to all network interfaces", rule_offset);
                 if let Ok(all_interfaces) =
                     get_interface_names_by_addr(Ipv4Addr::UNSPECIFIED.into())
                 {
@@ -212,6 +213,7 @@ impl PolicyEnforcer {
                 }
             }
             if validated_rule.flows.is_empty() {
+                tracing::warn!("ACL config flows list is empty. Applying rule #{} to both Ingress and Egress flows", rule_offset);
                 validated_rule
                     .flows
                     .extend([InterceptorFlow::Ingress, InterceptorFlow::Egress]);
--- a/zenoh/src/net/routing/interceptor/authorization.rs
+++ b/zenoh/src/net/routing/interceptor/authorization.rs
@@ -224,7 +224,7 @@ impl PolicyEnforcer {
                     for action in &config_rule.actions {
                         for key_expr in &config_rule.key_exprs {
                             if key_expr.trim().is_empty() {
-                                bail!("error from bad key-expression value");
+                                bail!("found an empty key-expression value in key_exprs list");
                             }
                             policy_rules.push(PolicyRule {
                                 subject: Subject::Interface(subject.clone()),
--- a/zenoh/src/net/routing/interceptor/authorization.rs
+++ b/zenoh/src/net/routing/interceptor/authorization.rs
@@ -218,7 +218,7 @@ impl PolicyEnforcer {
             }
             for subject in &config_rule.interfaces {
                 if subject.trim().is_empty() {
-                    bail!("error from bad interface value");
+                    bail!("found an empty interface value in interfaces list");
                 }
                 for flow in &config_rule.flows {
                     for action in &config_rule.actions {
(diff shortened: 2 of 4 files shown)


zenoh/src/net/routing/interceptor/authorization.rs
```diff
@@ -199,10 +199,24 @@ impl PolicyEnforcer {
     ) -> ZResult<PolicyInformation> {
         let mut policy_rules: Vec<PolicyRule> = Vec::new();
         for config_rule in config_rule_set {
+            // config validation
+            if config_rule.interfaces.is_empty()
```
Mallets, line 203: It seems that with this PR the interface list becomes always required, am I correct?
If that's the case, then I believe this is not an acceptable  behaviour.

snehilzs: for 0.11 it should be required in the `rules` field, since it will not make logical sense to have a rule without a subject to implement it for. For 1.0, this will be modified to have any one of : `cert_name`, `username` or `interface` fields present and non-empty.

snehilzs: For that we will require explicit keywords like `ALL` (this lies in the territory of Groups and Roles). Equating empty list in our _explicit_ rules to `ALL` is bad access control logic. For example, we already implement similar behavior for key-expressions where ALL is signified specifically using `**` and not empty list.

Mallets:
@oteffahi can you help in reviewing this PR?

zenoh/src/net/routing/interceptor/access_control.rs
```diff
@@ -66,7 +66,7 @@ pub(crate) fn acl_interceptor_factories(
                     enforcer: Arc::new(policy_enforcer),
                 }))
             }
-            Err(e) => tracing::error!("Access control inizialization error: {}", e),
+            Err(e) => tracing::error!("Access control not enabled due to: {}", e),
```
oteffahi, line 69: Since a security feature that the user wished to enable did not go through correctly, in my opinion it seems dangerous to proceed. Shouldn't we stop execution?

```suggestion
            Err(e) => panic!("Access control not enabled due to: {}", e),
```

zenoh/src/net/routing/interceptor/authorization.rs
```diff
@@ -199,10 +199,24 @@ impl PolicyEnforcer {
...
+                || config_rule.actions.is_empty()
+                || config_rule.flows.is_empty()
+                || config_rule.key_exprs.is_empty()
+            {
+                bail!("error from bad config");
```
oteffahi, line 208: Error message is not specific enough.

```suggestion
                bail!("one or multiple ACL config lists are empty");
```

zenoh/src/net/routing/interceptor/authorization.rs
```diff
@@ -199,10 +199,24 @@ impl PolicyEnforcer {
...
+            }
             for subject in &config_rule.interfaces {
+                if subject.trim().is_empty() {
+                    bail!("error from bad interface value");
```
oteffahi, line 212: Error message is not specific enough

```suggestion
                    bail!("found an empty interface value in interfaces list");
```

zenoh/src/net/routing/interceptor/authorization.rs
```diff
@@ -199,10 +199,24 @@ impl PolicyEnforcer {
...
+                }
                 for flow in &config_rule.flows {
                     for action in &config_rule.actions {
                         for key_expr in &config_rule.key_exprs {
+                            if key_expr.trim().is_empty() {
+                                bail!("error from bad key-expression value");
```
oteffahi, line 218: Error message is not specific enough

```suggestion
                                bail!("found an empty key-expression value in key_exprs list");
```

oteffahi (changes requested):
@Mallets @snehilzs I suggested some changes, apply them if you agree.
Else it LGTM.

zenoh/src/net/routing/interceptor/access_control.rs
Mallets: I think the upper logic should decide what to do with the execution. The interceptor has the freedom to return an error here and it should do so. I wouldn't panic directly at this level.

zenoh/src/net/routing/interceptor/authorization.rs
Mallets: I would separate the chained `||` checks in different ifs where the final log message is returned in case of error. In this way the user knows exactly what went wrong.

zenoh/src/net/routing/interceptor/authorization.rs
Mallets: Log should start with capital letters.

oteffahi: @Mallets this log is concatenated to another one, forming the full log that starts with a capital letter. Same applies for all the logs from this function.

zenoh/src/net/routing/interceptor/authorization.rs
```diff
@@ -199,10 +199,24 @@ impl PolicyEnforcer {
...
         let mut policy_rules: Vec<PolicyRule> = Vec::new();
         for config_rule in config_rule_set {
+            // config validation
+            if config_rule.interfaces.is_empty()
+                || config_rule.actions.is_empty()
+                || config_rule.flows.is_empty()
+                || config_rule.key_exprs.is_empty()
+            {
+                bail!("error from bad config");
+            }
```
oteffahi, line 209:
```suggestion
            let mut validation_err = "".to_owned();
            if config_rule.interfaces.is_empty() {
                validation_err.push_str("ACL config interfaces list is empty. ");
            }
            if config_rule.actions.is_empty() {
                validation_err.push_str("ACL config rules list is empty. ");
            }
            if config_rule.flows.is_empty() {
                validation_err.push_str("ACL config flows list is empty. ");
            }
            if config_rule.key_exprs.is_empty() {
                validation_err.push_str("ACL config key_exprs list is empty. ");
            }
            if validation_err != "" {
                bail!("{}", validation_err);
            }
```

oteffahi (changes requested):
@Mallets As you suggested, I put each list in a seperate if statement and return the aggregated error message at the end.

zenoh/src/net/routing/interceptor/authorization.rs
Mallets: I suggest to:
```
let mut validation_err = String::new(); // Does not allocate
...
if !validation_err.is_empty() {
    ...
}
```

oteffahi: Updated my suggestion above with your changes. Commiting it.

zenoh/src/net/routing/interceptor/authorization.rs
oteffahi, line 221: @Mallets please also apply this, we maked it as resolved above without applying it. 
```suggestion
                    bail!("found an empty interface value in interfaces list");
```

oteffahi (changes requested):

zenoh/src/net/routing/interceptor/authorization.rs
sashacmc: I don't think that we need some special keyword for all, IMHO, if filter filed absent it means that filter not applied for this field.
If filed interfaces absent - apply for all interfaces
If filed flows absent - apply for all flows
And for key-expressions will be better to have the same logic

Mallets: I agree for interfaces and flows.
I'm still inclined to having `key_exprs` mandatory though...

oteffahi:
@Mallets as requested, I added a way to allow empty `flows` and `interfaces` lists in ACL rules, and interpet them as "all interfaces" and "all flows" respectively.

I'd like to mention that I maintain my position regarding this change: implicit wildcards are in my opinion a bad practice for security-related features, and that we should not proceed with this change. I added warning logs to alert the user whenever these implicit wildcards are enabled. In the future it would be better to replace this with explicit wildcards that the user has to set for each list to be populated automatically, preferably in alternatively-named configuration fields for each list (i.e: `all-flows: true` and `all-interfaces: true`).

zenoh/src/net/routing/interceptor/authorization.rs
```diff
@@ -195,19 +197,51 @@ impl PolicyEnforcer {
...
-            for subject in &config_rule.interfaces {
-                for flow in &config_rule.flows {
-                    for action in &config_rule.actions {
-                        for key_expr in &config_rule.key_exprs {
+        for (rule_offset, config_rule) in config_rule_set.iter().enumerate() {
+            // config validation
+            let mut validated_rule = config_rule.clone();
+            let mut validation_err = String::new();
+            if validated_rule.interfaces.is_empty() {
+                tracing::warn!("ACL config interfaces list is empty. Applying rule #{} to all network interfaces", rule_offset);
```
Mallets, line 208: I believe this is not the right place to implement such behaviour. `policy_information_point` should work on valid configuration and it's completely fine to consider an empty interface list as invalid here. Proper enforcement should be done at configuration file parsing.

Mallets:
> @Mallets as requested, I added a way to allow empty `flows` and `interfaces` lists in ACL rules, and interpret them as "all interfaces" and "all flows" respectively.
> 
> I'd like to mention that I maintain my position regarding this change: implicit wildcards are in my opinion a bad practice for security-related features, and that we should not proceed with this change. I added warning logs to alert the user whenever these implicit wildcards are enabled. In the future it would be better to replace this with explicit wildcards that the user has to set for each list to be populated automatically, preferably in alternatively-named configuration fields for each list (i.e: `all-flows: true` and `all-interfaces: true`).

There is a difference between implicit and not-specified. Here we are tackling the configuraiton-aspect of the problem and not the implementation of it. To me, interfaces are not mandatory as part of the ACL rules, especially looking at the evolution of this configuration. Let's assume we make the evolve this ACL towards a `x509` certificate check. It goes without saying that the interface may be not relevant since I'd like to always apply it. If interfaces are something that make sense, then I can restrict the configuration to the specific interface. Same thing goes for the flows (egress vs ingress). Moreover, the more we add as mandatory fields to the config files, the more the config file becomes verbose, hard to debug, and more costly to transmit over the network.

oteffahi:
@Mallets I changed it so the `interfaces` and `flows` lists are optional in the config file, they behave as follows:
- If they are not provided they will be considered wildcards (apply to all).
- If they are provided as empty: error case.
I also made it so we error out when ACL cannot be enabled (the function returns an Error which as you suggested is handled by upper logic, which was already setup to exit in case of a returned Error from interceptor factories)

Mallets (approved):

PR #986, opened 2024-04-27
Shm refine

yellowhatter:


Mallets:
As part of the SHM refine, I believe we should rename `zsliceshm` and `zsliceshmmut` to something more compact.
For example:
- `zsliceshm` -> `zshm` | `shmbuf`
- `zsliceshmmut` -> `zshmmut` | `shmbufmut`
- 
I tend to prefer `shmbuf` and `shmbufmut` because it does not present the double `mm` as in `zshmmut`.
In any case, any other proposal is well accepted.

Mallets (approved):

PR #989, opened 2024-04-29
missing API items reexported

milyin:
Exported public instances into API

### DIFF
--- a/zenoh/src/lib.rs
+++ b/zenoh/src/lib.rs
@@ -178,6 +178,7 @@ pub mod key_expr {
         pub use zenoh_keyexpr::keyexpr_tree::{IKeyExprTree, IKeyExprTreeMut};
     }
     pub use crate::api::key_expr::KeyExpr;
+    pub use crate::api::key_expr::KeyExprUndeclaration;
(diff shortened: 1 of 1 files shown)


zenoh/src/lib.rs
```diff
@@ -199,24 +200,42 @@ pub mod session {
...
     pub use crate::api::session::SessionRef;
+    pub use crate::api::session::Undeclarable;
+}
+
+/// Tools to access information about the current zenoh [`Session`](crate::Session).
+pub mod info {
+    pub use crate::api::info::PeersZidBuilder;
+    pub use crate::api::info::RoutersZidBuilder;
+    pub use crate::api::info::SessionInfo;
+    pub use crate::api::info::ZidBuilder;
```
Mallets, line 218: Not in the scope of this PR but I believe `ZidBuilder` should be called `ZenohIdBuilder` since the correct type is `ZenohId`.

Mallets (approved):

zenoh/src/lib.rs
milyin: Added issue to not forget https://github.com/eclipse-zenoh/zenoh/issues/996

Issue #973, opened 2024-04-25
zenoh-c DLL crahes in `libc::atexit` handler on Windows


fuzzypixelz:
### Describe the bug

[See this workflow run failure for context.](https://github.com/eclipse-zenoh/zenoh-c/actions/runs/8800625076/job/24152200914)

The `z_api_double_drop_test` fails when syncing when syncing with zenoh, starting from commit 0283aaae480d0c0608802a6fbfb79f7fc681469f.

I've observed this crash only when zenoh-c is linked dynamically to an application and not when linked statically. Weirdly enough, this crash still happens when one of (or both of) the `z_drop` calls are removed.

<details>
<summary>Unfold this line to see the backtrace of the crash</summary>

```console
thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be\library\std\src\thread\mod.rs:1439:40
stack backtrace:
   0:     0x7ffd8557aee3 - std::backtrace_rs::backtrace::dbghelp::trace
                               at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library\std\src\..\..\backtrace\src\backtrace\dbghelp.rs:98
   1:     0x7ffd8557aee3 - std::backtrace_rs::backtrace::trace_unsynchronized
                               at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library\std\src\..\..\backtrace\src\backtrace\mod.rs:66
   2:     0x7ffd8557aee3 - std::sys_common::backtrace::_print_fmt
                               at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library\std\src\sys_common\backtrace.rs:65
   3:     0x7ffd8557aee3 - std::sys_common::backtrace::_print::impl$0::fmt
                               at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library\std\src\sys_common\backtrace.rs:44
   4:     0x7ffd852eca2b - core::fmt::rt::Argument::fmt
                               at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library\core\src\fmt\rt.rs:138
   5:     0x7ffd852eca2b - core::fmt::write
                               at /rustc/5680fa18feaa87f3ff04063800aec256c3d4b4be/library\core\src\fmt\mod.rs:1094
[...]

PR #1004, opened 2024-05-02
fix: `zenoh_session_multicast` test

### LINKED ISSUE #1002: Sporadic test failure of `zenoh_session_multicast `
### Describe the bug

For example, https://github.com/eclipse-zenoh/zenoh/actions/runs/8916537158/job/24488119417 reported a timeout.

### To reproduce

On Github CI runners.

### System info

- Zenoh: 371ca6b

YuanYuYuan:
Closes #1002.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,8 +1,8 @@
-# # By default, retry a few times until pass the test within the specified timeout
-# [profile.default]
-# retries = 1
-# slow-timeout = { period = "60s", terminate-after = 2 }
-#
+# By default, retry a few times until pass the test within the specified timeout
+[profile.default]
+retries = 1
+slow-timeout = { period = "60s", terminate-after = 1 }
+
 # # Run the following tests exclusively with longer timeout
 # [[profile.default.overrides]]
 # filter = """
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,16 +1,16 @@
 # By default, retry a few times until pass the test within the specified timeout
 [profile.default]
-# retries = 1
-slow-timeout = { period = "60s", terminate-after = 1 }
+retries = 1
+slow-timeout = { period = "60s", terminate-after = 2 }
 
-# # Run the following tests exclusively with longer timeout
-# [[profile.default.overrides]]
-# filter = """
-# test(=zenoh_session_unicast) |
-# test(=zenoh_session_multicast) |
-# test(=transport_tcp_intermittent) |
-# test(=transport_tcp_intermittent_for_lowlatency_transport) |
-# test(=three_node_combination)
-# """
-# threads-required = 'num-cpus'
-# slow-timeout = { period = "60s", terminate-after = 6 }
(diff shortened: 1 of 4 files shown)


YuanYuYuan:
In the above repeated tests, an unexpectedly slow run is caught. https://github.com/eclipse-zenoh/zenoh/actions/runs/8921424650/job/24501502386?pr=1004#step:7:113

YuanYuYuan:
As the timeout of nextest is disabled, the `ztimeout!` within the test appears to be ineffective. We might fail into the issue of `tokio::timeout(fut).await` never yields where the future `fut` is a blocking call per se.

YuanYuYuan:
Two issues have been identified:

1. Passed with panicking error.

```txt
2024-05-06T04:23:29.255530Z TRACE                    rx-0 ThreadId(10) zenoh_link_udp::multicast: Read error on UDP link 192.168.68.114:55606 => 224.0.0.1:17448: A Tokio 1.x context was found, but it is being shutdown. at io/zenoh-links/zenoh-link-udp/src/multicast.rs:97.
2024-05-06T04:23:29.255557Z DEBUG                    rx-0 ThreadId(10) zenoh_transport::multicast::link: Read error on UDP link 192.168.68.114:55606 => 224.0.0.1:17448: A Tokio 1.x context was found, but it is being shutdown. at io/zenoh-links/zenoh-link-udp/src/multicast.rs:97.
```

2. Failed with timeout error.

`zenoh::session::close` -> `zenoh::net::routing::dispatcher::tables::close_face` deadlock found.

https://github.com/ZettaScaleLabs/zenoh/blob/936bde67d95d102d6d582e884a1f931956141bd3/zenoh/src/net/routing/dispatcher/tables.rs?plain=1#L177

YuanYuYuan:
Regarding the first issue, the removal of [atexit callback](https://github.com/ZettaScaleLabs/zenoh/blob/936bde67d95d102d6d582e884a1f931956141bd3/commons/zenoh-runtime/src/lib.rs#L163-L165) is proposed after the discussion. The reasons are listed below,

1. Previously with async-std, we don't have such kind of cleanup on the global runtime.
2. Since Rust doesn't drop the static variable, we need to add this cleanup callback to pass the memory leak check with Valgrind.
However, we can workaround it by the previous solution "a runtime guard"  as well.
The downside of the previous solution is users may find the memory leak reported by Valgrind in our examples or the bindings that no runtime guard is used by default.
3. We found that the unsafe `atexit` is unsound. For example, we got the following error if we tried to log with `tracing` while dropping `ZRuntimePool`.
```txt
thread '<unnamed>' panicked at 'use of std::thread::current() is not possible after the thread's local data has been destroyed', library/std/src/thread/mod.rs:711:35
```
Moreover, we still have an unresolved issue with Windows DLL. See https://github.com/eclipse-zenoh/zenoh/issues/973 for details.

Once we remove the `atexit` cleanup, the first issue should disappear including the panics found on rmw_zenoh.

YuanYuYuan:
The first issue has been addressed via #1015

PR #1007, opened 2024-05-02
refactor: keep only trait in prelude

wyfo:
The current prelude exposes all the types of the API. This is considered as a bad practice, as many of these names are quite common and could conflict. Most of the popular crate using prelude are only putting traits in it (plus some types with particular names).
These PR removes all types from the prelude, and put the main ones directly in the root.

### DIFF
--- a/zenoh/src/lib.rs
+++ b/zenoh/src/lib.rs
@@ -204,7 +204,7 @@ pub mod session {
     #[doc(hidden)]
     pub use crate::api::session::InitBuilder;
     pub use crate::api::{
-        builders::publication::{SessionDeleteBuilder, SessionPutBuilder},
+        builders::publisher::{SessionDeleteBuilder, SessionPutBuilder},
         session::{open, OpenBuilder, Session, SessionDeclarations, SessionRef, Undeclarable},
     };
 }
@@ -282,7 +282,7 @@ pub mod publisher {
     #[zenoh_macros::unstable]
     pub use crate::api::publisher::PublisherRef;
     pub use crate::api::{
-        builders::publication::{
+        builders::publisher::{
             PublicationBuilder, PublicationBuilderDelete, PublicationBuilderPut, PublisherBuilder,
             PublisherDeleteBuilder, PublisherPutBuilder,
         },
--- a/zenoh/src/lib.rs
+++ b/zenoh/src/lib.rs
@@ -38,7 +38,7 @@
 //!
 //! #[tokio::main]
 //! async fn main() {
-//!     let session = zenoh::open(config::default()).await.unwrap();
+//!     let session = zenoh::open(zenoh::config::default()).await.unwrap();
 //!     session.put("key/expression", "value").await.unwrap();
 //!     session.close().await.unwrap();
 //! }
@@ -52,7 +52,7 @@
 //!
 //! #[tokio::main]
 //! async fn main() {
-//!     let session = zenoh::open(config::default()).await.unwrap();
+//!     let session = zenoh::open(zenoh::config::default()).await.unwrap();
 //!     let subscriber = session.declare_subscriber("key/expression").await.unwrap();
 //!     while let Ok(sample) = subscriber.recv_async().await {
 //!         println!("Received: {:?}", sample);
@@ -69,7 +69,7 @@
 //!
 //! #[tokio::main]
 //! async fn main() {
-//!     let session = zenoh::open(config::default()).await.unwrap();
+//!     let session = zenoh::open(zenoh::config::default()).await.unwrap();
 //!     let replies = session.get("key/expression").await.unwrap();
 //!     while let Ok(reply) = replies.recv_async().await {
 //!         println!(">> Received {:?}", reply.result());
@@ -111,23 +111,27 @@ pub const FEATURES: &str = zenoh_util::concat_enabled_features!(
     ]
 );
 
-// Expose some functions directly to root `zenoh::`` namespace for convenience
-pub use crate::api::{scouting::scout, session::open};
+#[doc(inline)]
+pub use crate::{
+    config::Config,
+    core::{Error, Result},
+    key_expr::{kedefine, keformat, kewrite},
+    scouting::scout,
+    session::{open, Session},
+};
 
 pub mod prelude;
 
 /// Zenoh core types
 pub mod core {
     #[allow(deprecated)]
-    pub use zenoh_core::AsyncResolve;
-    #[allow(deprecated)]
-    pub use zenoh_core::SyncResolve;
+    pub use zenoh_core::{AsyncResolve, SyncResolve};
     pub use zenoh_core::{Resolvable, Resolve, Wait};
+    pub use zenoh_result::ErrNo;
     /// A zenoh error.
     pub use zenoh_result::Error;
     /// A zenoh result.
     pub use zenoh_result::ZResult as Result;
-    pub use zenoh_util::{core::zresult::ErrNo, try_init_log_from_env};
 }
 
 /// A collection of useful buffers used by zenoh internally and exposed to the user to facilitate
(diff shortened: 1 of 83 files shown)


wyfo:
Following a discussion with @milyin.
Don't hesitate to ping other people interested in it.

@Mallets

wyfo:
See https://www.reddit.com/r/rust/comments/1cle18j/to_prelude_or_to_not_to_prelude/ as well as https://users.rust-lang.org/t/to-use-prelude-or-to-not-to-use-prelude-that-is-the-question/110855

fuzzypixelz:
@wyfo Please change the base branch to dev/1.0.0 as protocol_changes is no longer the development branch.

zenoh/tests/attachments.rs
Mallets, line 1: `use` slipped before the header.

zenoh/src/lib.rs
```diff
@@ -109,8 +109,24 @@ pub const FEATURES: &str = zenoh_util::concat_enabled_features!(
     ]
 );
 
-// Expose some functions directly to root `zenoh::`` namespace for convenience
-pub use crate::api::{scouting::scout, session::open};
+#[doc(inline)]
+pub use crate::{
```
Mallets, line 113: I'm not against but not 100% convinced wether it's a good idea to export all those types directly under `zenoh::*`. @milyin any opinion?

zenoh/tests/acl.rs
```diff
@@ -19,8 +19,10 @@ mod test {
     };
 
     use tokio::runtime::Handle;
-    use zenoh::prelude::*;
+    use zenoh::{config, prelude::*, Session};
+    use zenoh_config::Config;
```
Mallets, line 23: Shouldn't `Config` imported directly from `zenoh::Config`?

zenoh/tests/acl.rs
```diff
@@ -19,8 +19,10 @@ mod test {
...
     use zenoh_core::{zlock, ztimeout};
+    use zenoh_protocol::core::{EndPoint, WhatAmI};
```
Mallets, line 25: Shouldn't `Endpoint, WhatAmI` imported directly from `zenoh::*`?

zenoh/tests/attachments.rs
```diff
@@ -1,3 +1,6 @@
+use zenoh::ZBytes;
+use zenoh_config::Config;
```
Mallets, line 2: Shouldn't `Config` imported directly from `zenoh::*`?

zenoh/tests/liveliness.rs
Mallets, line 1: `use` slipped before the header.

zenoh/tests/connection_retry.rs
```diff
@@ -12,6 +12,8 @@
 //   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
 //
 use zenoh::prelude::*;
+use zenoh_config::{Config, ConnectionRetryConf};
```
Mallets, line 15: Shouldn't `Config` imported directly from `zenoh::*`?

zenoh/tests/handler.rs
```diff
@@ -13,7 +13,8 @@
 //
 use std::{thread, time::Duration};
 
-use zenoh::prelude::*;
+use zenoh::{handlers::RingChannel, prelude::*};
+use zenoh_config::Config;
```
Mallets, line 17: Shouldn't `Config` imported directly from `zenoh::*`?

zenoh/tests/interceptors.rs
```diff
@@ -14,6 +14,7 @@
 use std::sync::{Arc, Mutex};
 
 use zenoh::{internal::zlock, prelude::*};
+use zenoh_config::Config;
```
Mallets, line 17: Shouldn't `Config` imported directly from `zenoh::*`?

zenoh/tests/matching.rs
```diff
@@ -11,6 +11,9 @@
 // Contributors:
 //   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
 //
+use zenoh::{sample::Locality, Session};
+use zenoh_config::peer;
+use zenoh_protocol::core::Locator;
```
Mallets, line 16: Imports from `zenoh_config` and `zenoh_protocol` should be from `zenoh::*`?

zenoh/tests/qos.rs
```diff
@@ -13,7 +13,8 @@
 //
 use std::time::Duration;
 
-use zenoh::{internal::ztimeout, prelude::*};
+use zenoh::{internal::ztimeout, prelude::*, Priority};
+use zenoh_protocol::core::CongestionControl;
```
Mallets, line 17: Check imports from non `zenoh::*`.

zenoh/tests/routing.rs
```diff
@@ -25,7 +25,10 @@ use zenoh::{
     core::Result,
     internal::{bail, ztimeout},
     prelude::*,
+    Session,
 };
+use zenoh_config::{Config, ModeDependentValue};
+use zenoh_protocol::core::{CongestionControl, WhatAmI, WhatAmIMatcher};
```
Mallets, line 31: Check imports from non `zenoh::*`.

wyfo:
About imports from zenoh dependencies in tests, I didn't write them but used my IDE auto-import. I don't think it matters a lot in tests, as it's not public code, but I can modify them if wanted

zenoh/tests/session.rs
```diff
@@ -21,7 +21,8 @@ use std::{
 
 #[cfg(feature = "unstable")]
 use zenoh::runtime::{Runtime, RuntimeBuilder};
-use zenoh::{internal::ztimeout, prelude::*};
+use zenoh::{config, internal::ztimeout, prelude::*, sample::SampleKind, KeyExpr, Session, Value};
+use zenoh_protocol::core::{CongestionControl, Reliability};
```
Mallets, line 25: Check imports from non `zenoh::*`.

zenoh/tests/shm.rs
```diff
@@ -21,7 +21,15 @@ mod tests {
...
-    use zenoh::{internal::ztimeout, prelude::*};
+    use zenoh::{config, internal::ztimeout, prelude::*, Session};
+    use zenoh_protocol::core::{CongestionControl, Reliability};
+    use zenoh_shm::api::{
+        protocol_implementations::posix::{
+            posix_shared_memory_provider_backend::PosixSharedMemoryProviderBackend,
+            protocol_id::POSIX_PROTOCOL_ID,
+        },
+        provider::shared_memory_provider::{BlockOn, GarbageCollect, SharedMemoryProviderBuilder},
+    };
```
Mallets, line 32: Check imports from non `zenoh::*`.

zenoh/tests/unicity.rs
```diff
@@ -20,7 +20,8 @@ use std::{
 };
 
 use tokio::runtime::Handle;
-use zenoh::{internal::ztimeout, prelude::*};
+use zenoh::{config, internal::ztimeout, prelude::*, KeyExpr, Session};
+use zenoh_protocol::core::{CongestionControl, EndPoint, WhatAmI};
```
Mallets, line 24: Check imports from non `zenoh::*`.

Mallets:
> About imports from zenoh dependencies in tests, I didn't write them but used my IDE auto-import. I don't think it matters a lot in tests, as it's not public code, but I can modify them if wanted

I agree that it doesn't matter a lot in the tests but it would be good to verify they can actually be imported from `zenoh` without requiring importing the types from other crates.

zenoh/src/lib.rs
wyfo: Being able to write `zenoh::Result`/`zenoh::Error` is kind of mandatory IMO. 
Also a lot of common english names, you may not want to import them directly, but use the module instead, for example `zenoh::Session`, or `zenoh::Priority`. I think it's more convenient for the user (in my own code, I think I would use `zenoh::Session` more than `Session`).

Mallets: I agree that some user may prefer `zenoh::Session` to just `Session`. My point is more on what do we decide to put in the root? And what not? E.g., `Session`, `Result`, `Error` make sense to be in the root but I'm not sure about `Priority` (which is just taken as an example here). Is `Priority` really a first class type to be exported in the root? Are we going to put all types in the root loosing any kind of hierarchy/grouping?

wyfo: I can split the PR in two, one for the prelude, and one draft for the putting the types into the root. The only issue is about imports in the examples.

wyfo:
I've rebased the PR and updated test imports

wyfo:
The error in the CI is a false positive, `Value` is used in `zenoh/tests/session.rs` in line 172 https://github.com/eclipse-zenoh/zenoh/pull/1007/files#diff-535d5571c1b372d9e23a72ae9944449c512e46aa82eea002eb0ef0681b66562fR172
I don't know what to do.

fuzzypixelz:
> The error in the CI is a false positive, `Value` is used in `zenoh/tests/session.rs` in line 172 I don't know what to do. When I execute the command on my computer, there is no error.

@wyfo You're experiencing the wonderful phenomenon of "Shadow Merges" in GitHub pull requests. The CI does not run on your branch as such, but on different version of it where a `git merge origin/dev/1.0.0` has been executed, yielding a different `HEAD`. This is done in the `actions/checkout` action.

I can confirm that someone on `dev/1.0.0` removed the usage of `Value` in line 172. You should rebase/merge.

wyfo:
I don't know why the CI fails this time. @fuzzypixelz do you have an idea?

zenoh/tests/shm.rs
```diff
@@ -11,188 +11,194 @@
 // Contributors:
 //   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
 //
-#[cfg(all(feature = "unstable", feature = "shared-memory"))]
```
Mallets, line 14: This test shouldn't be run at all if not with `#[cfg(all(feature = "unstable", feature = "shared-memory"))]`. 
What's the reason of removing this?

wyfo: I didn't removed it. I used a inner attribute instead of a simple attribute on a module, so I removed the useless module.

Mallets: Rigth, I had missed the `!` :)

Mallets (approved):

Issue #1008, opened 2024-05-02
Performance issues with current "contiguous" implementation


kydos:
The implementation for the contiguous operation is sub-optimal for the case in which we have multiple slices. 

This code:
https://github.com/eclipse-zenoh/zenoh/blob/7a4744576e21c85d31b1527b2c4bf2230e5152e8/commons/zenoh-buffers/src/lib.rs#L105

Should be rewritten to look something like:

```
_ => {
                    let mut l = 0;
                    for s in slices.by_ref() {
                        l += s.len();
                    }
                    let mut vec = Vec::with_capacity(l);
                    for slice in slices {
                        vec.extend_from_slice(slice);
                    }
                    Cow::Owned(vec)
                },

````

This will ensure a single allocation and will reduce to the minimum the total number of bytes actually copied.

YuanYuYuan:
Has been completed via #1009 and #1022.

PR #1009, opened 2024-05-02
Fixed performance issue with contiguous (#1008)

kydos:
Proposed changes to fix the performance issue with the current `contiguous` implementation.

### DIFF
--- a/commons/zenoh-buffers/src/lib.rs
+++ b/commons/zenoh-buffers/src/lib.rs
@@ -112,7 +112,7 @@ pub mod buffer {
                         vec.extend_from_slice(slice);
                     }
                     Cow::Owned(vec)
-                },
+                }
             }
         }
     }
--- a/commons/zenoh-buffers/src/lib.rs
+++ b/commons/zenoh-buffers/src/lib.rs
@@ -102,10 +102,17 @@ pub mod buffer {
             match slices.len() {
                 0 => Cow::Borrowed(b""),
                 1 => Cow::Borrowed(slices.next().unwrap()),
-                _ => Cow::Owned(slices.fold(Vec::new(), |mut acc, it| {
-                    acc.extend(it);
-                    acc
-                })),
+                _ => {
+                    let mut l = 0;
+                    for s in slices.by_ref() {
+                        l += s.len();
+                    }
+                    let mut vec = Vec::with_capacity(l);
+                    for slice in slices {
+                        vec.extend_from_slice(slice);
+                    }
+                    Cow::Owned(vec)
+                },
             }
         }
     }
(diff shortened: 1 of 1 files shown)


YuanYuYuan:
This is an interesting fix. I tried to benchmark the effect and here are my observations.

* Code

```rust
use zenoh_buffers::{buffer::SplitBuffer, ZBuf, ZSlice};
use std::time::Instant;


fn run() -> u128 {
    const SLICE_LEN: usize = 65536;
    const NUM_SLICES: usize = 1_000;

    let mut zbuf = ZBuf::default();

    for _ in 0..NUM_SLICES {
        zbuf.push_zslice(ZSlice::from([0; SLICE_LEN]));
    }

    let timer = Instant::now();
    let _ = zbuf.contiguous();
    timer.elapsed().as_micros()
}

fn main() {
    const REPEAT: usize = 100;
    let total: u128 = (0..REPEAT).into_iter().map(|_| run()).sum();
    println!("Time on average: {}(us)", total as f64 / REPEAT as f64);
}
```

* Competitors (run in the release mode)

```rust
        fn contiguous(&self) -> Cow<'_, [u8]> {
            let mut slices = self.slices();
            match slices.len() {
                0 => Cow::Borrowed(b""),
                1 => Cow::Borrowed(slices.next().unwrap()),
                _ => {
                    // Style 1: Time on average: 123872.52(us)
                    let vec = slices.into_iter().flatten().cloned().collect();

                    // Style 2: Time on average: 15222.12(us)
                    let vec = slices.fold(Vec::new(), |mut acc, it| {
                        acc.extend(it);
                        acc
                    });

                    // Style 3: Time on average: 11271.38(us)
                    let vec: Vec<_> = slices.collect::<Vec<_>>().concat();

                    // Style 4: Time on average: 4.77(us)
                    let mut vec = Vec::with_capacity(slices.by_ref().map(|s| s.len()).sum());
                    for slice in slices {
                        vec.extend_from_slice(slice);
                    }

                    Cow::Owned(vec)
                }
            }
        }
```

The lesson is functional programming way is not always better. Rust compiler seems not to fetch the total size in any way.

kydos:
Indeed... The original code is _cute_  but in practice quadratic in the amount of bytes copied.

commons/zenoh-buffers/src/lib.rs
```diff
@@ -102,10 +102,17 @@ pub mod buffer {
             match slices.len() {
                 0 => Cow::Borrowed(b""),
                 1 => Cow::Borrowed(slices.next().unwrap()),
```
Mallets, line 104:
```suggestion
                1 => {
                    // SAFETY: unwrap here is safe because we have explicitly checked
                    //         the iterator has 1 element.
                    Cow::Borrowed(unsafe { slices.next().unwrap_unchecked() })
                }
```

Mallets (approved):

PR #1011, opened 2024-05-06
Fix release workflow to use docker image attribute

diogomatsubara:
In https://github.com/eclipse-zenoh/ci/pull/129/files, the action to release docker images was changed to use an image attribute instead of tags, so we can properly tag latest and nightly releases.


fuzzypixelz (approved):
LGTM. @Mallets can you please approve the workflow run?

PR #1013, opened 2024-05-06
SHM refine: Allocation API

yellowhatter:


yellowhatter:
@Mallets this is a new shm alloc API

examples/examples/z_alloc_shm.rs
```diff
@@ -34,20 +34,51 @@ async fn run() -> ZResult<()> {
...
-        // OPTION 2: Comprehensive configuration:
+        // OPTION: Comprehensive configuration:
         let _comprehensive_layout = provider
-            .alloc_layout()
-            .size(512)
-            .alignment(AllocAlignment::new(2))
-            .res()
+            .alloc(512)
+            .with_alignment(AllocAlignment::new(2))
+            .make_layout()
```
Mallets, line 81: What about calling it `into_layout`?

commons/zenoh-shm/src/api/provider/shared_memory_provider.rs
```diff
@@ -64,83 +67,69 @@ impl BusyChunk {
...
+    pub fn with_policy<Policy>(self) -> AllocBuilder2<'a, IDSource, Backend, Policy> {
+        AllocBuilder2 {
+            data: self.0,
+            _phantom: PhantomData,
+        }
+    }
+
+    /// Get the result
+    #[zenoh_macros::unstable_doc]
+    pub fn res(self) -> BufLayoutAllocResult {
```
Mallets, line 126: https://github.com/eclipse-zenoh/zenoh/pull/942 has been recently merged. I believe we should use either `.wait()` or `.await` from now on.

commons/zenoh-shm/src/api/provider/shared_memory_provider.rs
```diff
@@ -511,6 +501,75 @@ unsafe impl<'a, Policy: AllocPolicy, IDSource, Backend: SharedMemoryProviderBack
...
+// Async Alloc policy
+impl<'a, IDSource, Backend, Policy> AllocBuilder2<'a, IDSource, Backend, Policy>
+where
+    IDSource: ProtocolIDSource + Send + Sync,
+    Backend: SharedMemoryProviderBackend + Sync,
+    Policy: AsyncAllocPolicy,
+{
+    /// Get the async result
+    #[zenoh_macros::unstable_doc]
+    pub async fn res_async(self) -> BufLayoutAllocResult {
```
Mallets, line 561: Same comment as above. We should use either `.await` or `.wait()` by implementing `Wait` or `IntoFuture` traits.

commons/zenoh-shm/src/api/provider/shared_memory_provider.rs
yellowhatter: I know about our new `Wait` and `IntoFuture` traits, I was just concerned that `.wait()` will confuse our users that it blocks when it doesn't.

Practically it blocks only if BlockOn policy is used, so we have three types of behaviors:
- non-blocking (any policy except BlockOn)
- blocking (only BlockOn policy):
- - sync
- - async (available for `Backend: Sync`)

I think I need to split allocation API into these three forms:
- `res()` for non-blocking policies
- `Wait` for sync BlockOn
- `IntoFuture` for async BlockOn

Mallets: Regardless it's blocking or non-blocking, `.wait()` is the common method for resolving builders in `sync` context in the same way you call `.await` in `async` context. I don't see much benefit for adding another way to resolve the builder...

yellowhatter:
macos-13 test failure is a regression that is fixed in main

PR #1015, opened 2024-05-06
fix: remove the unsound atexit cleanup

YuanYuYuan:


commons/zenoh-runtime/src/lib.rs
```diff
@@ -146,23 +146,26 @@ lazy_static! {
 }
 
 // To drop the data mannually since Rust does not drop static variables.
-pub extern "C" fn cleanup() {
+pub fn cleanup() {
```
Mallets, line 149: I think it's better to keep the `cleanup` function marked as `unsafe` and remove the `unsafe` block internally.

Mallets (approved):

commons/zenoh-runtime/src/lib.rs
YuanYuYuan: That's a good point. Then I think we could hide this unsafe function and let the `ZRuntimeGuard` be the only way to manipulate it.

PR #1020, opened 2024-05-13
Apply Clippy lints from Rust 1.78

fuzzypixelz:
With the release of Rust 1.78 comes new lints and previously undetected issues.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,7 @@ members = [
 exclude = ["ci/nostd-check", "ci/valgrind-check"]
 
 [workspace.package]
-rust-version = "1.66.1"
+rust-version = "1.72.0"
 version = "0.11.0-dev" # Zenoh version
 repository = "https://github.com/eclipse-zenoh/zenoh"
 homepage = "http://zenoh.io"
@@ -91,7 +91,7 @@ crc = "3.0.1"
 criterion = "0.5"
 derive_more = "0.99.17"
 derive-new = "0.6.0"
-tracing-subscriber = {version = "0.3", features = ["json", "env-filter"]}
+tracing-subscriber = { version = "0.3", features = ["json", "env-filter"] }
 tracing-loki = "0.2"
 event-listener = "4.0.0"
 flume = "0.11"
--- a/commons/zenoh-keyexpr/src/key_expr/fuzzer.rs
+++ b/commons/zenoh-keyexpr/src/key_expr/fuzzer.rs
@@ -50,6 +50,9 @@ fn make(ke: &mut Vec<u8>, rng: &mut impl rand::Rng) {
 }
 
 pub struct KeyExprFuzzer<Rng: rand::Rng>(pub Rng);
+// NOTE: Denying `clippy::assigning_clones` causes:
+// error[E0502]: cannot borrow `next` as mutable because it is also borrowed as immutabl
+#[allow(clippy::assigning_clones)]
 impl<Rng: rand::Rng> Iterator for KeyExprFuzzer<Rng> {
     type Item = OwnedKeyExpr;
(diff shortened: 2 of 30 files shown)


Cargo.toml
```diff
@@ -56,7 +56,7 @@ members = [
 exclude = ["ci/nostd-check", "ci/valgrind-check"]
 
 [workspace.package]
-rust-version = "1.66.1"
+rust-version = "1.72.0"
```
OlivierHecart, line 59: This indicates the minimal rust version supported by Zenoh (https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field). If Zenoh still supports 1.66.1 we should keep this to 1.66.1.

commons/zenoh-keyexpr/src/key_expr/utils.rs
```diff
@@ -228,12 +205,15 @@ impl<const N: usize> Split<[u8; N]> for [u8] {
     }
 }
 
+#[allow(dead_code)]
```
OlivierHecart, line 208: If this is unused shouldn't we just remove it ?

io/zenoh-transport/src/unicast/transport_unicast_inner.rs
```diff
@@ -80,7 +80,6 @@ pub(crate) trait TransportUnicastTrait: Send + Sync {
     /*************************************/
     /*            TERMINATION            */
     /*************************************/
-    async fn close_link(&self, link: Link, reason: u8) -> ZResult<()>;
```
OlivierHecart, line 83: I don't know the Transport layer enough to tell if this should be removed or not. Maybe add a note to ask Luca when he comes back ?

Cargo.toml
fuzzypixelz: Zenoh hasn't supported 1.66.1 for a long time. I checked with the cargo-msrv tool which uses binary search to find the MSRV. Although this is actually just the MSRV of the zenoh crate, other crates have a lower MSRV. But we keep one MSRV for all crates.

commons/zenoh-keyexpr/src/key_expr/fuzzer.rs
```diff
@@ -50,6 +50,9 @@ fn make(ke: &mut Vec<u8>, rng: &mut impl rand::Rng) {
 }
 
 pub struct KeyExprFuzzer<Rng: rand::Rng>(pub Rng);
+// NOTE: Denying `clippy::assigning_clones` causes:
+// error[E0502]: cannot borrow `next` as mutable because it is also borrowed as immutabl
+#[allow(clippy::assigning_clones)]
```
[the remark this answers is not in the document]
fuzzypixelz: That would require borrowing `next` while already immutably borrowed in `n`.

commons/zenoh-keyexpr/src/key_expr/utils.rs
fuzzypixelz: That `Utf` trait is used by the `utfdbg` macro. And said macro is kept because it's "useful when introducing new matchers to debug them".

commons/zenoh-keyexpr/src/key_expr/fuzzer.rs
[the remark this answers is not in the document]
OlivierHecart: This should do the trick:
```
if let Some(n) = next.strip_prefix('/').map(ToOwned::to_owned) {
    next = n
}
```

io/zenoh-transport/src/unicast/transport_unicast_inner.rs
yellowhatter: @OlivierHecart @fuzzypixelz I've checked the code, it seems that this method is not needed, we can remove it

Issue #1018, opened 2024-05-13
Query timeout >10 seconds is not applied


JEnoch:
### Describe the bug

Whatever the timeout set for a query, the queryable always applies a 10 seconds timeout, displaying such log if its callback lasts more than 10 seconds:
```log
2024-05-13T11:33:16.140489Z  WARN net-0 ThreadId(11) zenoh::net::routing::dispatcher::queries: Didn't receive final reply Face{1, d5d7972809c555fa5cb45c556803218}:1 from Face{0, 1b3f024aa4d55f411f6ed98f91124625}: Timeout!
2024-05-13T11:33:16.140860Z DEBUG net-0 ThreadId(11) zenoh::net::routing::dispatcher::queries: Propagate final reply Face{1, d5d7972809c555fa5cb45c556803218}:1
```

### To reproduce

1. Modify `z_queryable` example adding `tokio::time::sleep(Duration::from_secs(15)).await;` before reply
2. Run `RUST_LOG=zenoh::net::routing::dispatcher::queries=debug ./target/debug/examples/z_queryable`
3. Run `RUST_LOG=zenoh::net::routing::dispatcher::queries=debug ./target/debug/examples/z_get --timeout 20000`

After 10 seconds `z_queryable` logs:
```log
2024-05-13T11:33:16.140489Z  WARN net-0 ThreadId(11) zenoh::net::routing::dispatcher::queries: Didn't receive final reply Face{1, d5d7972809c555fa5cb45c556803218}:1 from Face{0, 1b3f024aa4d55f411f6ed98f91124625}: Timeout!
2024-05-13T11:33:16.140860Z DEBUG net-0 ThreadId(11) zenoh::net::routing::dispatcher::queries: Propagate final reply Face{1, d5d7972809c555fa5cb45c556803218}:1
```
meaning the 20 seconds timeout requested by the querier have not been applied.

After 15 seconds `z_queryable` logs:
```log
>> [Queryable ] Responding ('demo/example/zenoh-rs-queryable': 'Queryable from Rust!')
2024-05-13T11:33:18.140702Z  WARN  main ThreadId(01) zenoh::net::routing::dispatcher::queries: Route reply Face{0, 1b3f024aa4d55f411f6ed98f91124625}:1 from Face{0, 1b3f024aa4d55f411f6ed98f91124625}: Query nof found!
[...]

PR #1021, opened 2024-05-13
Fix bug in Query Timeout

OlivierHecart:


JEnoch (approved):
LGTM.
I confirm this fixes #1018

PR #1022, opened 2024-05-13
Fix: prevent iterator consumption when computing the length of slices during conversion to contiguous

YuanYuYuan:
Closes https://github.com/eclipse-zenoh/zenoh-c/issues/359.


kydos (approved):
Looks good.

kydos (approved):

PR #1026, opened 2024-05-14
Check crate MSRVs 

fuzzypixelz:
This add a check for the MSRV of all crates in the workspace. Each crate has its own MSRV specified through `package.rust-version`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -144,14 +144,20 @@ jobs:
         uses: crate-ci/typos@master
 
   msrvs:
-    name: MSRV checks on ${{ matrix.os }}
-    runs-on: ${{ matrix.os }}
-    # NOTE: It is necessary to run the check-msrv action on multiple platforms since a crate's MSRV
+    name: MSRV checks for ${{ matrix.target }}
+    runs-on: macos-14
+    # NOTE: It is necessary to run the check-msrv action for multiple targets since a crate's MSRV
     # depends on the target it's compiled for.
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-13, macos-14]
+        target:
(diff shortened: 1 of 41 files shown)


Mallets:
@fuzzypixelz what's the status of this PR? Is it ready for review?

fuzzypixelz:
> @fuzzypixelz what's the status of this PR? Is it ready for review?

Right now it seems to report incorrect MSRVs. I would have to investigate that first.

fuzzypixelz:
The issue wasn't that the job was reporting incorrect MSRVs. But that the MSRV of the crate `zenohd` depends on the target platform it's compiled for. 

On the ubuntu-latest runner (i.e. `x86_64-unknown-linux-gnu`) the MSRV is `1.70`. But on my Apple Silicon machine (i.e. `aarch64-apple-darwin`) the MSRV is `1.72`.

fuzzypixelz:
The problem now is handling platform-specific MSRVs. Right now we check that the `rust-version` field is equal to the MSRV.

If crate C1 on platform P1 has an MSRV of V1 and an MSRV of V2 on platform P2. Then the checks on platforms P1 and P2 cannot both succeed. 

We cannot tolerate a `rust-version` lower than the MSRV (here MSRV means the minimum version that builds). So we should instead tolerate `rust-version`s higher than the MSRV and emit a warning for good measure.

fuzzypixelz:
The other problem is that cargo-msrv fails to install toolchains on windows-latest with error:

```console
error: could not create link from 'C:\\Users\\runneradmin\\.cargo\\bin\\rustup.exe' to 'C:\\Users\\runneradmin\\.cargo\\bin\\cargo.exe'
```

It seemingly cannot create a (symbolic? hard?) link. This is very odd since the `check` job installs the Stable toolchain successfully on windows-latest.

PR #1032, opened 2024-05-16
added multicast scouting ttl option

jeremiah-wooten:
adds the ability to change the ttl on the multicast scouting socket from the default of 1

### DIFF
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -95,7 +95,7 @@
       address: "224.0.0.224:7446",
       /// The network interface which should be used for multicast scouting
       interface: "auto", // If not set or set to "auto" the interface if picked automatically
-      /// the time-to-live on multicast scouting packets
+      /// The time-to-live on multicast scouting packets
       ttl: 1,
       /// Which type of Zenoh instances to automatically establish sessions with upon discovery on UDP multicast.
       /// Accepts a single value or different values for router, peer and client.
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -95,6 +95,8 @@
       address: "224.0.0.224:7446",
       /// The network interface which should be used for multicast scouting
       interface: "auto", // If not set or set to "auto" the interface if picked automatically
+      /// the time-to-live on multicast scouting packets
+      ttl: 1,
       /// Which type of Zenoh instances to automatically establish sessions with upon discovery on UDP multicast.
       /// Accepts a single value or different values for router, peer and client.
       /// Each value is bit-or-like combinations of "peer", "router" and "client".
--- a/zenoh/src/net/runtime/orchestrator.rs
+++ b/zenoh/src/net/runtime/orchestrator.rs
@@ -614,6 +614,10 @@ impl Runtime {
         socket.set_nonblocking(true)?;
         socket.set_multicast_ttl_v4(multicast_ttl)?;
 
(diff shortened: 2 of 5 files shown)


DEFAULT_CONFIG.json5
```diff
@@ -95,6 +95,8 @@
       address: "224.0.0.224:7446",
       /// The network interface which should be used for multicast scouting
       interface: "auto", // If not set or set to "auto" the interface if picked automatically
+      /// the time-to-live on multicast scouting packets
```
fuzzypixelz, line 98:
```suggestion
      /// The time-to-live on multicast scouting packets
```

zenoh/src/net/runtime/orchestrator.rs
```diff
@@ -603,14 +612,15 @@ impl Runtime {
         // Must set to nonblocking according to the doc of tokio
         // https://docs.rs/tokio/latest/tokio/net/struct.UdpSocket.html#notes
         socket.set_nonblocking(true)?;
+        socket.set_multicast_ttl_v4(multicast_ttl)?;
```
fuzzypixelz, line 615: The documentation of [`UdpSocket`](https://doc.rust-lang.org/std/net/struct.UdpSocket.html#method.set_multicast_ttl_v4) states that "this might not have any effect on IPv6 sockets.". Perhaps we can add something like:

```rust
if sockaddr.is_ipv6() && multicast_ttl > 1 {
  tracing::warn!("UDP Multicast TTL has been set to a value greater than 1 on a socket bound to an IPv6 address. This might not have the desired effect");
}

```

zenoh/src/net/runtime/orchestrator.rs
```diff
@@ -637,6 +647,7 @@ impl Runtime {
...
```
fuzzypixelz, line 650: Here we don't join any multicast group. Isn't this a no-op?

zenoh/src/scouting.rs
```diff
@@ -312,6 +312,7 @@ fn scout(
     tracing::trace!("scout({}, {})", what, &config);
     let default_addr = SocketAddr::from(zenoh_config::defaults::scouting::multicast::address);
     let addr = config.scouting.multicast.address().unwrap_or(default_addr);
+    let multicast_ttl = config.scouting.multicast.ttl.unwrap_or(1);
```
fuzzypixelz, line 315:
```suggestion
    let default_multicast_ttl = zenoh_config::defaults::scouting::multicast::ttl;
    let multicast_ttl = config.scouting.multicast.ttl.unwrap_or(default_multicast_ttl);
```

fuzzypixelz (changes requested):

fuzzypixelz:
@jeremiah-wooten Thank you for the contribution :)

zenoh/src/net/runtime/orchestrator.rs
jeremiah-wooten: If I understand multicast properly, only listening for multicast packets requires joining a multicast group, but sending packets only requires sending to the correct IP address, and this is the function used in the scout function [here](https://github.com/eclipse-zenoh/zenoh/blob/25f06bd9e71b4dbed5ba8d75eaa1708693e7b9ec/zenoh/src/scouting.rs#L323C1-L323C70) to create the socket that send out multicast packets, so if the TTL is not set on this socket, then all the scouting messages sent will still have a TTL of 1. I tried taking this line out and saw that packet TTL was still 1.

PR #1036, opened 2024-05-20
Fixed stabby ( https://github.com/rust-lang/rust/issues/123281 )

yellowhatter:


yellowhatter:
@OlivierHecart this is a fix for `protocol_changes` compilation error caused by `stabby`

PR #1042, opened 2024-05-22
Disable generation of multilink keys if not used

oteffahi:
The multilink RSA keypair is used to authentify zenoh instances connected through multiple links. It is only used if the `transport_multilink` feature is compiled in, and the config parameter `transport.unicast.max_links` is set to a value superior to 1. This PR disables the keypair initialization if multilink is compiled in and not used in the config, saving some time when opening the session.


io/zenoh-transport/src/unicast/manager.rs
```diff
@@ -243,7 +243,7 @@ impl TransportManagerBuilderUnicast {
             protocols: Arc::new(AsyncMutex::new(HashMap::new())),
             transports: Arc::new(AsyncMutex::new(HashMap::new())),
             #[cfg(feature = "transport_multilink")]
-            multilink: Arc::new(MultiLink::make(prng)?),
+            multilink: Arc::new(MultiLink::make(prng, config.max_links > 1)?),
```
yellowhatter, line 246: I'd rather suggest not to create Multilink at all if multilink is not used: `Option<Multilink>` ....

yellowhatter (changes requested):
This is discussional

io/zenoh-transport/src/unicast/manager.rs
Mallets: I agree with @yellowhatter 's suggestion. It's easier to have `Option<Multilink>` and to create `Some(multilink)` only when `config.max_links > 1`.

oteffahi: Okay. I will revert the previous commits and change to this approach.

oteffahi:
@Mallets @yellowhatter could you please review this new version?

io/zenoh-transport/src/unicast/establishment/accept.rs
```diff
@@ -660,11 +675,10 @@ pub(crate) async fn accept_link(link: LinkUnicast, manager: &TransportManager) -
                 resolution: manager.config.resolution,
                 ext_qos: ext::qos::StateAccept::new(manager.config.unicast.is_qos),
                 #[cfg(feature = "transport_multilink")]
-                ext_mlink: manager
-                    .state
-                    .unicast
-                    .multilink
-                    .accept(manager.config.unicast.max_links > 1),
+                ext_mlink: match &manager.state.unicast.multilink {
```
Mallets, line 678: Should `ext_multilink` be an `Option` here as well? If that's the case, maybe `disabled()` is no longer required?

oteffahi: Making it an `Option` would require also changing the `Cookie.ext_mlink` to `Option<ext::multilink::StateAccept>`, but then it's unclear to me how [this line](https://github.com/eclipse-zenoh/zenoh/blob/dev/1.0.0/io/zenoh-transport/src/unicast/establishment/cookie.rs#L64-L65) should be adapted.

oteffahi:
@Mallets This PR is ready to be merged

PR #1060, opened 2024-05-29
fix(1059): adding explicit plugin id in PluginStatus and PluginManager

### LINKED ISSUE #1059: [BUG] Spawning more volumes with the same backend is not reflected in admin space
### Describe the bug

When configuring zenoh to use the same backend for more than a volume, only one of the volume is exposed in the admin space.

### To reproduce

Start a zenoh router with two volumes of the same backend configured:

```json
 "file-system-volume": {
                    "backend": "filesystem_backend"
                },
                "file-system-volume2": {
                    "backend": "filesystem_backend"
                },
```

perform a get on `@/router/*/status/plugins/storage_manager/volumes/*`

the result will show only one `filesystem_backend` volume instead of two.

```json

[
    {
        "key": "@/router/8f2aef5a89c14100b70184094f5bdea9/status/plugins/storage_manager/volumes/filesystem_backend",
        "value": {
            "root": "/home/vscode/.zenoh/zenoh_backend_fs",
            "version": "0.11.0-dev"
        },
        "time": "None"
    }
]
```

the issue comes from this line: https://github.com/eclipse-zenoh/zenoh/blob/main/plugins/zenoh-plugin-storage-manager/src/lib.rs#L322 where the `name` of the plugin (the backend in this case), is used to craft the kexpr.

There is no way currently to get the `id` of the plugin(backend) from the current PluginsManager.

### System info

Any platform
zenoh 0.11.0-rc.3 and previous

gabrik:
This PR adds the `id()` method to the `PluginStatus` trait and allows passing it as a parameter in the `PluginManager` when dynamic loading or statically linking a plugin.
It exposes plugins in the admin space via their ID (see below what the ID is), and allows multiple instances of the same plugin with different configurations, essential for backends, and useful for other plugins.

Changes to configuration:
```json
        "rest2": { // plugin id, used in admin space, to retrieve plugin config
            "__required__": true,
            "http_port": 8000,
            "__plugin__":"rest", // plugin name used in libloading
        },
```

Sister PRs:
- [ ] MQTT: https://github.com/eclipse-zenoh/zenoh-plugin-mqtt/pull/110
- [ ] DDS: https://github.com/eclipse-zenoh/zenoh-plugin-dds/pull/245
- [ ] ROS2 https://github.com/eclipse-zenoh/zenoh-plugin-ros2dds/pull/156
- [ ] WebServer https://github.com/eclipse-zenoh/zenoh-plugin-webserver/pull/85

Sister PRs for testing, do not need a merge.
- [ ] FS Backend https://github.com/eclipse-zenoh/zenoh-backend-filesystem/pull/113
- [ ] RocksDB Backend https://github.com/eclipse-zenoh/zenoh-backend-rocksdb/pull/100
- [ ] S3 https://github.com/eclipse-zenoh/zenoh-backend-s3/pull/82
- [ ] Influx https://github.com/eclipse-zenoh/zenoh-backend-influxdb/pull/121

Closes #1059

### DIFF
--- a/commons/zenoh-config/src/lib.rs
+++ b/commons/zenoh-config/src/lib.rs
@@ -1021,6 +1021,7 @@ fn load_external_plugin_config(title: &str, value: &mut Value) -> ZResult<()> {
 #[derive(Debug, Clone)]
 pub struct PluginLoad {
     pub name: String,
+    pub id: String,
     pub paths: Option<Vec<String>>,
     pub required: bool,
 }
@@ -1045,15 +1046,20 @@ impl PluginsConfig {
                 Some(Value::Bool(b)) => *b,
                 _ => panic!("Plugin '{}' has an invalid '__required__' configuration property (must be a boolean)", name)
             };
+            let id = match value.get("__plugin__") {
+                Some(Value::String(p)) => p,
+                _ => name,
+            };
+
             if let Some(paths) = value.get("__path__"){
                 let paths = match paths {
                     Value::String(s) => vec![s.clone()],
                     Value::Array(a) => a.iter().map(|s| if let Value::String(s) = s {s.clone()} else {panic!("Plugin '{}' has an invalid '__path__' configuration property (must be either string or array of strings)", name)}).collect(),
                     _ => panic!("Plugin '{}' has an invalid '__path__' configuration property (must be either string or array of strings)", name)
                 };
-                PluginLoad {name: name.clone(), paths: Some(paths), required}
+                PluginLoad {id: id.clone(), name: name.clone(), paths: Some(paths), required}
             } else {
-                PluginLoad {name: name.clone(), paths: None, required}
+                PluginLoad {id: id.clone(), name: name.clone(), paths: None, required}
             }
         })
     }
--- a/commons/zenoh-config/src/lib.rs
+++ b/commons/zenoh-config/src/lib.rs
@@ -1020,8 +1020,8 @@ fn load_external_plugin_config(title: &str, value: &mut Value) -> ZResult<()> {
 
 #[derive(Debug, Clone)]
 pub struct PluginLoad {
-    pub name: String,
     pub id: String,
+    pub name: String,
     pub paths: Option<Vec<String>>,
     pub required: bool,
 }
@@ -1054,8 +1054,8 @@ impl PluginsConfig {
             if let Some(paths) = value.get("__path__"){
                 let paths = match paths {
                     Value::String(s) => vec![s.clone()],
-                    Value::Array(a) => a.iter().map(|s| if let Value::String(s) = s {s.clone()} else {panic!("Plugin '{}' has an invalid '__path__' configuration property (must be either string or array of strings)", name)}).collect(),
-                    _ => panic!("Plugin '{}' has an invalid '__path__' configuration property (must be either string or array of strings)", name)
+                    Value::Array(a) => a.iter().map(|s| if let Value::String(s) = s {s.clone()} else {panic!("Plugin '{}' has an invalid '__path__' configuration property (must be either string or array of strings)", id)}).collect(),
+                    _ => panic!("Plugin '{}' has an invalid '__path__' configuration property (must be either string or array of strings)", id)
                 };
                 PluginLoad {id: id.clone(), name: name.clone(), paths: Some(paths), required}
             } else {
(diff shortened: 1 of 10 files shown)


gabrik:
FYI: I see too many unwraps in the `PluginManager` implementation, it is true that they should never fail as they are calls to `Vec::last` after a `Vec::insert` but I think they should not be there anyway.

gabrik:
Note: Lint is failing on a file that has not been touched by the PR.

Mallets (approved):

Mallets:
The PR looks good to me. 
@gabrik should we wait merging it till all sister PRs are made?

gabrik:
> The PR looks good to me. @gabrik should we wait merging it till all sister PRs are made?

I'm opening the sister PRs and tagging people for review, the ones not to be merged are going to be closed.

gabrik:
Once the sisters are approved then I'm going to merge everything.

plugins/zenoh-plugin-trait/src/manager.rs
```diff
@@ -99,7 +106,7 @@ impl<StartArgs: PluginStartArgs, Instance: PluginInstance> DeclaredPlugin<StartA
 pub struct PluginsManager<StartArgs: PluginStartArgs, Instance: PluginInstance> {
     default_lib_prefix: String,
     loader: Option<LibLoader>,
-    plugins: Vec<PluginRecord<StartArgs, Instance>>,
+    plugins: HashMap<String, PluginRecord<StartArgs, Instance>>,
```
milyin, line 109: `Vec` instead of `HashMap` was chosen because each plugin already knows it's unique name so it was excessive to store this name outside. As far as I see nothing changed, there is just `id` instead of `name`. Is there any particular reason to switch to `HashMap` ?
Storing plugins in `Vec` may become useful in hypothetical situation when plugins depends on each other and loading order is important. Though yes, at this moment choice between `Vec` and `HashMap` here is just a question of personal taste.

commons/zenoh-config/src/lib.rs
```diff
@@ -1021,6 +1021,7 @@ fn load_external_plugin_config(title: &str, value: &mut Value) -> ZResult<()> {
 #[derive(Debug, Clone)]
 pub struct PluginLoad {
     pub name: String,
```
milyin, line 1023: In fact, `name` and `paths`have the same purpose: it's a information for loader where to look for corresponding plugin DLL. Probably the `name` should near the `paths` and be optional also

plugins/zenoh-plugin-trait/src/manager.rs
```diff
@@ -155,38 +170,46 @@ impl<StartArgs: PluginStartArgs + 'static, Instance: PluginInstance + 'static>
         tracing::debug!("Declared dynamic plugin {} by name {}", &name, &plugin_name);
         let loader = DynamicPlugin::new(
             name,
+            id.clone(),
             DynamicPluginSource::ByName((libloader, plugin_name)),
             required,
         );
-        self.plugins.push(PluginRecord::new(loader));
-        Ok(self.plugins.last_mut().unwrap())
+        self.plugins.insert(id.clone(), PluginRecord::new(loader));
```
milyin, line 177: No protection against adding plugin with the same name. Before it was wrong also: same-named plugin was added, but it wasn't possible to access it by name later. Now the old one is just silently removed

plugins/zenoh-plugin-trait/src/manager.rs
```diff
@@ -60,6 +62,11 @@ impl<StartArgs: PluginStartArgs, Instance: PluginInstance> PluginStatus
     fn name(&self) -> &str {
```
milyin, line 62: I'm not sure that after this update we need this `name` getter at all. If plugin is dynamic, the `path()` getter gives us full information where this plugin came from. If it's static, name and id is the same thing.

milyin (changes requested):
This will work ok, but later some refactoring is needed I think, as the `name` became mostly unnecessary.
The parameters of methods should be renamed (`name` -> `id`)

plugins/zenoh-plugin-trait/src/manager.rs
gabrik: The former is true, while not the latter, the user can statically link the plugin once, but instantiate it multiple times with different IDs.

milyin (approved):

PR #1062, opened 2024-05-29
fix: remove zenoh::scouting::WhatAmI reexport

wyfo:
`WhatAmI` is already present in `zenoh::config`.

This PR is based on #1007, so it will be draft until its base is merged.


wyfo:
@milyin

Mallets (approved):

PR #1073, opened 2024-05-31
Add certificate and user/password authentication methods to ACL filters

oteffahi:
Reopening PR as followup to https://github.com/eclipse-zenoh/zenoh/pull/840 and https://github.com/eclipse-zenoh/zenoh/pull/1072.


io/zenoh-transport/src/unicast/lowlatency/transport.rs
```diff
@@ -187,6 +188,10 @@ impl TransportUnicastTrait for TransportUnicastLowlatency {
         self.config.zid
     }
 
+    fn get_auth_ids(&self) -> Vec<AuthId> {
+        vec![]
```
oteffahi, line 192: @Mallets Does this seem correct to you?

Mallets: No, it seems an incomplete implementation.

oteffahi:
@Mallets This is ready for review.
Regarding the issue we discussed this morning with `pest_meta`, do we bump the version for the crates or do we wait for the rust-toolchain bump before merging this PR?

Mallets:
We need first to bump rust-toolchain

Mallets:
rust-toolchain has just bumped to `1.75`

oteffahi:
@Mallets ready to be merged.

- Made some changes to auth config validation in regards to its current state and use-cases.
- Authentication test is slow (60s). Will be broken down to concurrent tests in a later PR.

Don't mind the CI issues, clippy checks are not passing because the runners bumped to Rust 1.79.

io/zenoh-link-commons/src/unicast.rs
```diff
@@ -118,3 +120,71 @@ pub fn get_ip_interface_names(addr: &SocketAddr) -> Vec<String> {
...
+
+impl LinkAuthIdBuilder {
+    pub fn new() -> LinkAuthIdBuilder {
+        LinkAuthIdBuilder {
+            auth_type: LinkAuthType::None,
+            auth_value: None,
+        }
+    }
+
+    pub fn auth_type(&mut self, auth_type: LinkAuthType) -> &mut Self {
```
Mallets, line 175: While different options exist for builders, e.g. to accept `&mut self` *vs* `self`. In Zenoh we tend to use the `self` version (see [here](https://github.com/eclipse-zenoh/zenoh/blob/7adad9482f8d27fab56ca54aa053de4cc2f63557/zenoh/src/subscriber.rs#L458)).

io/zenoh-link-commons/src/unicast.rs
```diff
@@ -118,3 +120,71 @@ pub fn get_ip_interface_names(addr: &SocketAddr) -> Vec<String> {
...
+    None,
+}
+
+#[derive(Clone, Debug, Serialize, Hash, PartialEq, Eq)]
+pub struct LinkAuthId {
+    auth_type: LinkAuthType,
+    auth_value: Option<String>,
+}
+
+impl LinkAuthId {
```
Mallets, line 137: What about adding `fn builder() -> LinkAuthIdBuilder` function?

io/zenoh-links/zenoh-link-quic/Cargo.toml
```diff
@@ -30,13 +30,13 @@ base64 = { workspace = true }
...
 rustls-native-certs = { workspace = true }
-rustls-pki-types =  { workspace = true }
+rustls-pki-types = { workspace = true }
 rustls-webpki = { workspace = true }
 secrecy = { workspace = true }
 tokio = { workspace = true, features = [
-  "fs",
   "io-util",
   "net",
+  "fs",
```
Mallets, line 39: Why reordering `fs` out of alphabetical order?

io/zenoh-links/zenoh-link-quic/src/unicast.rs
```diff
@@ -156,6 +160,11 @@ impl LinkUnicastTrait for LinkUnicastQuic {
     fn is_streamed(&self) -> bool {
         true
     }
+
+    #[inline(always)]
+    fn get_auth_identifier(&self) -> LinkAuthId {
```
Mallets, line 165: Does it really need to be always cloned? Couldn't return `&LinkAuthId` and let to the caller the decision wether to call it or not?

io/zenoh-links/zenoh-link-quic/src/unicast.rs
```diff
@@ -418,3 +434,37 @@ async fn accept_task(
     }
     Ok(())
 }
+
+fn get_cert_common_name(conn: quinn::Connection) -> ZResult<QuicAuthId> {
```
Mallets, line 438: I think this function should accept `&quinn::Connection` as parameter and to not take ownership of the connection.

io/zenoh-links/zenoh-link-quic/src/unicast.rs
```diff
@@ -385,7 +397,10 @@ async fn accept_task(
                             }
                         };
 
+                        // Get Quic auth identifier
                         let dst_addr = quic_conn.remote_address();
+                        let auth_id = get_cert_common_name(quic_conn.clone())?;
```
Mallets, line 402: See my comment below on passing the ownership of `QuicConn`.

io/zenoh-links/zenoh-link-tcp/src/unicast.rs
```diff
@@ -164,6 +164,11 @@ impl LinkUnicastTrait for LinkUnicastTcp {
     fn is_streamed(&self) -> bool {
         true
     }
+
+    #[inline(always)]
+    fn get_auth_identifier(&self) -> LinkAuthId {
+        LinkAuthId::default()
```
Mallets, line 170: See my comment on `Default` implementation of `LinkAuthId`. `serial` does not have any level of `LinkAuthId`, so it should be stated clear that is `empty`/`none`. By simply looking at this code it's not apparent.

io/zenoh-transport/src/unicast/authentication.rs
Mallets, line 1: License header missing.

io/zenoh-transport/src/unicast/establishment/ext/auth/mod.rs
```diff
@@ -571,7 +571,10 @@ impl<'a> AcceptFsm for &'a AuthFsm<'a> {
     }
 
     type RecvOpenSynIn = (&'a mut StateAccept, Option<open::ext::Auth>);
-    type RecvOpenSynOut = ();
+
+    #[cfg(feature = "auth_usrpwd")]
```
Mallets, line 575: I suggest to have one single type as return type that has multiple fields, one per extension. The inner field than could be conditionally available.

io/zenoh-transport/src/unicast/establishment/ext/auth/mod.rs
```diff
@@ -604,13 +607,17 @@ impl<'a> AcceptFsm for &'a AuthFsm<'a> {
...
+                }
+                (None, None) => {
+                    return Ok(UsrPwdId(None));
                 }
-                (None, None) => {}
                 _ => bail!("{S} Invalid UsrPwd configuration."),
             }
         }
-
+        #[cfg(not(feature = "auth_usrpwd"))]
```
Mallets, line 620: Always return the same type, conditionally enable/disable fields inside the type.

io/zenoh-transport/src/unicast/mod.rs
```diff
@@ -117,6 +122,11 @@ impl TransportUnicast {
         Ok(transport.get_links())
     }
 
+    pub fn get_auth_ids(&self) -> ZResult<Vec<AuthId>> {
```
Mallets, line 125: In some part of the code we use `get_auth_identifiers` while here we use the shorter version. I'd keep the naming 
consistent and I'm more inclined to keep the short version `get_auth_ids`.

zenoh/src/net/routing/interceptor/authorization.rs
```diff
@@ -177,6 +177,18 @@ impl PolicyEnforcer {
                                 );
                             }
                         }
+                        match rule.usernames {
```
Mallets, line 180: What's the point of having `rule.usernames` as `Option<Vec<_>>` if `None` is set to `Some` with an empty vector? Couldn't we just use a `Vec<_>`?

In any case, the `match` could be rewritten as:
```rust
if rule.usernames.is_none() {
    rule.usernames = Some(Vec::new());
}
```

zenoh/tests/authentication.rs
```diff
@@ -0,0 +1,1269 @@
...
+    use zenoh_core::{zlock, ztimeout};
+
+    const TIMEOUT: Duration = Duration::from_secs(60);
+    const SLEEP: Duration = Duration::from_secs(1);
+    const KEY_EXPR: &str = "test/demo";
+    const VALUE: &str = "zenoh";
+    #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+    async fn test_authentication() {
+        zenoh_util::try_init_log_from_env();
+        let path = "./tests/testfiles";
```
Mallets, line 38: What about using `https://doc.rust-lang.org/std/env/fn.temp_dir.html`?

io/zenoh-transport/src/unicast/mod.rs
oteffahi: The long version `get_auth_identifier` is implemented at link level and returns a single `LinkAuthId`.
The short version `get_auth_ids` is implemented a transport level and returns multiple AuthIds (since the transport can control multiple links and have user/password auth enabled).
I believe it's better to have long and short versions instead of having `get_auth_id` and `get_auth_ids` which look too similar to the reader.

.gitignore
```diff
@@ -21,4 +21,6 @@
 
 cargo-timing*.html
 
+#ignore test data
+testfiles
```
Mallets, line 25: This change could be reverted if `std::env::temp_dir()` is used in the tests

Mallets:
[DEFAULT_CONFIG](https://github.com/eclipse-zenoh/zenoh/blob/main/DEFAULT_CONFIG.json5) should be updated to reflect the new configuration possibility for access control.

Documentation should also be updated for [TLS](https://zenoh.io/docs/manual/tls/), [QUIC](https://zenoh.io/docs/manual/quic/) and [UsrPwd](https://zenoh.io/docs/manual/user-password/).

zenoh/src/net/routing/interceptor/authorization.rs
oteffahi: `Option<Vec<_>>` is just to allow the user to not specify it in the config, instead of having to explicitely set it to an empty vector when the filter is not a concern for the rule in question.

io/zenoh-transport/src/unicast/mod.rs
oteffahi: Renamed all `get_auth_identifier` to `get_auth_id`.

io/zenoh-links/zenoh-link-quic/Cargo.toml
```diff
@@ -54,5 +54,6 @@ zenoh-sync = { workspace = true }
 zenoh-util = { workspace = true }
 # Lock due to quinn not supporting rustls 0.22 yet
 rustls = { version = "0.21", features = ["dangerous_configuration", "quic"] }
-tokio-rustls = "0.24.1"
 rustls-pemfile = { version = "1" }
+tokio-rustls = "0.24.1"
```
Mallets, line 58: The dependency should be moved to the workspace.

io/zenoh-links/zenoh-link-quic/Cargo.toml
```diff
@@ -54,5 +54,6 @@ zenoh-sync = { workspace = true }
 zenoh-util = { workspace = true }
 # Lock due to quinn not supporting rustls 0.22 yet
 rustls = { version = "0.21", features = ["dangerous_configuration", "quic"] }
```
Mallets, line 56: Can this dependency moved also to the workspace?

io/zenoh-links/zenoh-link-quic/Cargo.toml
Mallets: I see there are incompatibility with quinn at the moment. Let's leave it as it is for now and let's update all the dependencies in a separate PR.

Mallets (approved):

PR #1075, opened 2024-06-03
feat: add `RingChannel.recv_deadline`

wyfo:
It's needed for Python binding. In fact, `recv` cannot be called in Python because it prevents Python signals, e.g. `KeyboardInterrupt` to be handled. The solution is to call `recv_deadline` within in a loop with a short deadline, e.g. 100ms.

### DIFF
--- a/zenoh/src/api/handlers/ring.rs
+++ b/zenoh/src/api/handlers/ring.rs
@@ -13,7 +13,10 @@
 //
 
 //! Callback handler trait.
-use std::sync::{Arc, Weak};
+use std::{
+    sync::{Arc, Weak},
+    time::Instant,
+};
 
 use zenoh_collections::RingBuffer;
 use zenoh_result::ZResult;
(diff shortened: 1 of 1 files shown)


wyfo:
@Mallets

Mallets:
From a user point of view, should we add also a `recv_timeout(duration: Duration)`? It will simply call `recv_deadline()` under the hood.

zenoh/src/api/handlers/ring.rs
```diff
@@ -62,7 +67,29 @@ impl<T> RingChannelHandler<T> {
...
+    /// Receive from the ring channel with a deadline.
+    ///
+    /// If the ring channel is empty, this call will block until an element is available in the channel,
+    /// or return `None` if the deadline has passed.
+    pub fn recv_deadline(&self, deadline: Instant) -> ZResult<Option<T>> {
+        let Some(channel) = self.ring.upgrade() else {
+            bail!("The ringbuffer has been deleted.");
+        };
+
+        loop {
```
Mallets, line 79: Why the loop?

wyfo: Because of half-backed PR 😓  it should be better now.

wyfo: I mean, there is a loop because there is one in `recv`, but I forgot to catch the timeout error.

Issue #806, opened 2024-03-10
remove internal APIs from public interface


milyin:
### Describe the release item

We should not expose internal APIs or at least should explicitly mark them as internal in the documentation (e.g. APIs needed for plugin support only)

The list of APIs to be checked

keyexpr -> get_nonwild_prefix
keyexpr -> is_wild
keyexpr -> chunks
keyexpr -> strip_prefix

publication.rs: Priority::MIN, Priority::MAX, Priority::NUM

~~WhatAmI matcher~~

ZenohId -> size, max_size, rand, into_keyexpr, to_le_bytes

milyin:
Consider using `#[doc(hidden)]` attribute for them

wyfo:
`WhatAmIMatcher` is used as argument for [`zenoh::scout`](https://docs.rs/zenoh/latest/zenoh/fn.scout.html), so I think it can be removed from the list?

milyin:
All functionality which is needed by some external crates but should not be part of public API should be moved into traits available through "internal" mod. Here is an example: https://github.com/eclipse-zenoh/zenoh/pull/1046/files

milyin:
from @Mallets:
- [ ]  `buffers` goes to internal
- [ ]  `ZBytes` impl for `ZSlice` and `ZBuf` goes to internal

milyin:
> All functionality which is needed by some external crates but should not be part of public API should be moved into traits available through "internal" mod. Here is an example: [#1046](https://github.com/eclipse-zenoh/zenoh/pull/1046)

Seems that #[cfg(feature="internal")] is enough, no need to complicate with traits

wyfo:
Instead of `#[cfg(all(feature = "unstable", feature="internal"))]`, I think we should annotate items with `#[cfg(any(feature = "unstable", feature="internal"))]`. It would allow to use "internal" items without needing to build with `unstable` feature. 
(Of course, `zenoh::internal` module would only have `#[cfg(feature = "internal)]`)

PR #1082, opened 2024-06-03
feat: add `zenoh_util::init_log_with_callback`

wyfo:
It will notably use in bindings, to redirect tracing events into native logs.


wyfo:
@Mallets @milyin

milyin:
This is going to be used by python binding right now and other bindings may benefit from it too. Merging  it and after that need to add internal/unstable features as part of task #806

Issue #1081, opened 2024-06-03
ZSlice and ZBytes are internal now. Can we avoid their usage in shm examples


milyin:
### Describe the release item

It was decided to move `ZSlice` and `ZBuf` into `internal` mod. But these structs are used in shm examples.
Should we fix examples or the decision to make these things internal doesn't work for shm?
@yellowhatter can you check?

wyfo:
I think you wrote `ZBbytes` instead `ZBuf`.  `ZBytes` is meant to be public, it appears everywhere in the API. `ZBuf` is the internal implementation of `ZBytes`.

yellowhatter:
@milyin examples fixed: #1129

Issue #1078, opened 2024-06-03
[dev/1.0.0]`Encoding::from(&str)` skip unsupported encoding


wyfo:
### Describe the release item

https://github.com/eclipse-zenoh/zenoh/blob/9bcfdc78ba73957cb49fa475e9192221c75baa58/zenoh/src/api/encoding.rs#L642-L644

This behavior is quite unintuitive. I think we should remove this `From<&str>` implementation and replace it with a `FromStr`.

@Mallets @milyin 

EDIT: we should remove this `From<&str>` implementation and raise an error in `FromStr` if the encoding is not supported.

wyfo:
> FromStr is already implemented. 

I know, my point is that we should not skip unsupported encoding silently, we should raise and error instead, so only use a "raising" `FromStr` implementation.

wyfo:
If I read the code correctly, `Encoding::from("my_encoding")` will return `Encoding::ZENOH_BYTES`. Is that really what we want?

wyfo:
@milyin think we maybe should then put the unrecognized string in schema instead of discarding it. I also think that's a better solution than what we have now.

Mallets:
Closed by #1102

PR #1084, opened 2024-06-04
minor api updates

milyin:
- build feature `internal` added
- `kedefine`, `keformat`, `kewrite` removed from root, made unstable
- `Priority` in core
- `buffers` made `internal::buffers`
- `key_expr::keyexpr_tree` unstable
- `SetIntersectionLevel`, `relation_to` unstable
- `key_expr::format` unstable
- `session::init` internal
- `GetBuilder` now `session::SessionGetBuilder` , like `PutBuilder`
- `querier` mod added, empty for now, later `Quierier` object will be added to it
- `Query`, `Reply` and corresponding builders/trait in mod `query` now. In `queryable` only `Queriable` remained
- `ZenohId` internal methods hidden now, wrapper `config::ZenohId` made instead of reexporting
- keyexpr's `get_nonwild_prefix`, `is_wild`, `chunks`, `strip_prefix` unstable and internal
- `unstable` feature is required for `plugins`, `internal` and `shared-memory`. This restiction added in lib.rs by firing copmpile error if feature is not set. This makes not necessary to add `unstable` near all `plugins`, `internal` and `shared-memory` items
- removed `EncodingInternals` trait, `internal` feature is used instead


Mallets (approved):

examples/examples/z_ping_shm.rs
```diff
@@ -15,7 +15,7 @@ use std::time::{Duration, Instant};
 
 use clap::Parser;
 use zenoh::{
-    buffers::ZSlice,
+    internal::buffers::ZSlice,
```
wyfo, line 18: I'm not sure it its normal to use internal API in examples

zenoh/Cargo.toml
```diff
@@ -50,7 +50,8 @@ transport_udp = ["zenoh-transport/transport_udp"]
 transport_unixsock-stream = ["zenoh-transport/transport_unixsock-stream"]
 transport_ws = ["zenoh-transport/transport_ws"]
 transport_vsock = ["zenoh-transport/transport_vsock"]
-unstable = []
+unstable = ["zenoh-keyexpr/unstable"]
+internal = ["zenoh-keyexpr/internal"]
```
wyfo, line 54: features seems to be sorted alphabetically

zenoh/src/api/encoding.rs
```diff
@@ -836,24 +836,17 @@ impl EncodingMapping for serde_pickle::Value {
...
-    fn id(&self) -> u16;
-
-    fn schema(&self) -> Option<&ZSlice>;
-
-    fn new(id: u16, schema: Option<ZSlice>) -> Self;
-}
-
-impl EncodingInternals for Encoding {
-    fn id(&self) -> u16 {
+impl Encoding {
```
wyfo, line 839: So, if I understand, there is no way for the user to retrieve the id or the schema of its encoding. Is there some interest to expose it at all if use cannot work with it?

zenoh/src/api/info.rs
```diff
@@ -93,7 +94,7 @@ impl<'a> Wait for RoutersZenohIdBuilder<'a> {
                     s.get_whatami()
                         .ok()
                         .and_then(|what| (what == WhatAmI::Router).then_some(()))
-                        .and_then(|_| s.get_zid().ok())
+                        .and_then(|_| s.get_zid().map(Into::<ZenohId>::into).ok())
```
wyfo, line 97: `map(Into::into)` should work, because `map(|zid| zid.into())` work below.

zenoh/src/api/info.rs
```diff
@@ -143,7 +144,7 @@ impl<'a> Wait for PeersZenohIdBuilder<'a> {
                     s.get_whatami()
                         .ok()
                         .and_then(|what| (what == WhatAmI::Peer).then_some(()))
-                        .and_then(|_| s.get_zid().ok())
+                        .and_then(|_| s.get_zid().map(|zid| zid.into()).ok())
```
wyfo, line 147: Same, you could use `map(Into::into)`

zenoh/src/lib.rs
```diff
@@ -135,16 +134,9 @@ pub mod core {
...
-/// A collection of useful buffers used by zenoh internally and exposed to the user to facilitate
-/// reading and writing data.
-pub mod buffers {
-    pub use zenoh_buffers::{
-        buffer::SplitBuffer,
-        reader::{HasReader, Reader},
-        ZBuf, ZBufReader, ZSlice, ZSliceBuffer,
-    };
+    /// Zenoh message priority
+    pub use crate::api::publisher::Priority;
```
wyfo, line 139: Why did you move Priority?

zenoh/src/lib.rs
```diff
@@ -289,29 +285,33 @@ pub mod publisher {
...
         },
-        publisher::{Priority, Publisher, PublisherUndeclaration},
+        publisher::{Publisher, PublisherUndeclaration},
     };
 }
 
-/// Query primitives
+/// Get operation primitives
+pub mod querier {
+    // Later the `Querier` with `get`` operation will be added here, in addition to `Session::get`,
```
wyfo, line 294: If the `Querier` is not ready, I don't think we should have a module for now.
Also, the same way than `put` relative types are in the module `publisher`, `query` module itself should be renamed `queryier`, shouldn't it?

zenoh/src/lib.rs
```diff
@@ -364,26 +355,58 @@ pub mod config {
...
 }
 
-#[doc(hidden)]
-#[cfg(all(feature = "unstable", feature = "plugins"))]
-pub mod plugins {
-    pub use crate::api::plugins::{
-        PluginsManager, Response, RunningPlugin, RunningPluginTrait, ZenohPlugin, PLUGIN_PREFIX,
-    };
-}
+#[cfg(all(feature = "internal", not(feature = "unstable")))]
```
wyfo, line 358: If you want to do something like that, you should add `"unstable"` as a dependency of the `"internal"` feature.
But I don't see why you should do that. *internal* and *unstable* are orthogonal concepts. I think every plugins/bindings should be able to compile with *internal* only.

zenoh/src/lib.rs
```diff
@@ -364,26 +355,58 @@ pub mod config {
...
+#[cfg(all(
+    feature = "plugins",
+    not(all(feature = "unstable", feature = "internal"))
+))]
+compile_error!(
+    "The plugins support is internal and unstable. The `unstable` and `internal` features must be enabled to use `plugins`."
+);
 
-#[doc(hidden)]
+#[zenoh_macros::internal]
```
wyfo, line 370: `session::init` is not in the `internal` module. Is it wanted to have *internal* items spreads across modules?

zenoh/src/lib.rs
```diff
@@ -364,26 +355,58 @@ pub mod config {
...
+    #[cfg(feature = "plugins")]
+    pub mod plugins {
+        pub use crate::api::plugins::{
+            PluginsManager, Response, RunningPlugin, RunningPluginTrait, ZenohPlugin, PLUGIN_PREFIX,
+        };
+    }
 }
 
-#[cfg(all(feature = "unstable", feature = "shared-memory"))]
+#[cfg(all(feature = "shared-memory", not(feature = "unstable")))]
```
wyfo, line 404: Do you want to require user putting both `shared_memory` and `unstable` features in their `Cargo.toml`, or could `unstable` be a dependency of `shared_memory`? I'm ok with the first solution.

zenoh/src/lib.rs
```diff
@@ -364,26 +355,58 @@ pub mod config {
...
+compile_error!(
+    "The shared-memory support is unstable. The `unstable` feature must be enabled to use `shared-memory`."
+);
+
+#[cfg(feature = "shared-memory")]
```
wyfo, line 409:
```suggestion
#[zenoh_macros::unstable]
#[cfg(feature = "shared-memory")]
```

commons/zenoh-config/src/lib.rs
```diff
@@ -51,6 +52,39 @@ pub use mode_dependent::*;
 pub mod connection_retry;
 pub use connection_retry::*;
 
+/// The global unique id of a zenoh peer.
```
wyfo, line 55: Why not just use `#[doc(hidden)]` on `zenoh_protocol::core::ZenohId` internal methods, instead of writing a new type?

zenoh/src/lib.rs
milyin: Good point. Maybe we should do it. But this is to be discussed additionally I think. `session::init` is a way to do `session::open` for plugins and it accepts runtime as a parameter. Maybe it should be put into `internal::runtime` mod, maybe not. Let's leave it as is for now

zenoh/src/lib.rs
milyin: Same as above, the "unstable" feature is a red flag for user, he should explicitly agree, that he is stepping on shaky ground

zenoh/src/api/encoding.rs
wyfo: I mean, if the Rust user cannot have any info about the encoding present in the sample he receives, why exposing it?

zenoh/src/lib.rs
wyfo: Ok, but I still think we should just rename the `query` module to be consistent with `publisher` module.

zenoh/src/lib.rs
wyfo: I disagree, everything *internal* should not be treated by definition, as it is `doc(hidden)`. User should not care, because user should not use. And the user who cares because writing plugins/bindings already knows that *internal* API is *unstable*, because he cares.

zenoh/src/lib.rs
wyfo: `session::init` is not the only one, there is also `query::ReplySample` for example.
I've no particular objection having *internal* items spreads across modules, but then, maybe we should put back root `buffers` module and mark it with `#[zenoh_macros::internal]`.

zenoh/src/lib.rs
wyfo: If it is such a red flag, you may be interested in [`tokio` approach](https://docs.rs/tokio/latest/tokio/#unstable-features).

zenoh/src/api/info.rs
wyfo:
```suggestion
                        .and_then(|_| s.get_zid().map(Into::into).ok())
```

zenoh/src/api/info.rs
wyfo:
```suggestion
                        .and_then(|_| s.get_zid().map(Into::into).ok())
```

commons/zenoh-keyexpr/src/key_expr/borrowed.rs
```diff
@@ -92,6 +93,7 @@ impl keyexpr {
     /// Returns the relation between `self` and `other` from `self`'s point of view ([`SetIntersectionLevel::Includes`] signifies that `self` includes `other`).
     ///
     /// Note that this is slower than [`keyexpr::intersects`] and [`keyexpr::includes`], so you should favor these methods for most applications.
+    #[cfg(feature = "unstable")]
```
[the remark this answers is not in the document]
wyfo: I think I prefer having it done in this issue.

zenoh/src/lib.rs
wyfo: Ok

commons/zenoh-config/src/lib.rs
wyfo: `#[doc(hidden)]` methods are not visible in autocompletion either in RustRovers, or in VSCode.

zenoh/Cargo.toml
wyfo: Ok

commons/zenoh-config/src/lib.rs
milyin: And technically these methods are still available to the user, why they should not be. They are even not in "internal" part of api - it's just some internal methods which leaked due to our quite arbitrary crate separation. So better to hide them correctly.

commons/zenoh-keyexpr/src/key_expr/borrowed.rs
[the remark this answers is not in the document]
milyin: This is not the purpose of this pr. And adding new crate affects publising process, so this will not be done here

examples/examples/z_ping_shm.rs
milyin: https://github.com/eclipse-zenoh/zenoh/issues/1081

zenoh/src/api/encoding.rs
milyin: This was exposed by request from @DariusIMP for kotlin binding. We don't provide schema ids as part of user API, but bindings may need this internal detail.
User can get/set encoding as string or as predefined constant. Internal representation of this encoding string is not exposed

zenoh/src/lib.rs
milyin: `query` module logically now similar to `sample`. I.e. 
`publisher` is a server, `subscriber` is a client, `sample::Sample`, etc is data which circulates between them
`queryable` is a server, `querier` is a client, `query::Query`, `query::Reply` is data

zenoh/src/lib.rs
milyin: I prefer to classify it from user point of view like this:
- `unstable` - normal end user API which just not stabilized yet
- `internal` - end user API which exposes internal implementation details which are necessary for performing some tasks. This is unstable by definition and forever, the internal implementation is that we are free to change. But this doesn't mean that it's always will be `#[doc(hidden)]`, maybe it will make sense to document it, for plugins for example.

So I want to keep this scheme. If you want to make separation between `internal` and `unstable`, let's do it in separate update. This update just formalizes existing practice: now practically all internal api is marked unstable

zenoh/src/lib.rs
milyin: maybe yes, maybe no. I'm agnostic about that. Moving buffers into "internal" was proposed by @Mallets , I had no objections to that. And I think it's better as it immediately made clear that some examples uses internal API, why they should not

zenoh/src/lib.rs
milyin: As far as I see our approach is exactly the same: "some feature flags and some api parts are available only if `unstable` flag is set"

zenoh/src/api/encoding.rs
Mallets: The encoding is created/converted to/from `String`. That's the interface exposed to the user. As it stands now, that interface is implemented. Am I missing something?

zenoh/src/lib.rs
wyfo: `query` is not similar to `sample`, btw `Query` is a member of `queryable`, not `query`.
`query` module is here to provide the types specifically used by `Session::get`, and when `get` will be abstracted into a `Querier` object, `query` module should be renamed to `querier` to stay consistent with `publisher`.

zenoh/src/api/encoding.rs
wyfo: No, I was missing it, sorry. You can close the thread. By the way the discussion about https://github.com/eclipse-zenoh/zenoh/issues/1078#issuecomment-2145224412 is still open.

zenoh/src/lib.rs
wyfo: Ok.

zenoh/src/lib.rs
wyfo: Our approach uses compilation feature, while tokio is using compiler *cfg* flag. This is radically different, as explained in the [post](https://internals.rust-lang.org/t/feature-request-unstable-opt-in-non-transitive-crate-features/16193#why-not-a-crate-feature-2) linked in tokio documentation quoted above.
I honestly think we should follow the same convention, but that's for an other PR. I will open an issue to discuss it.

commons/zenoh-config/src/lib.rs
wyfo: Ok.

zenoh/src/lib.rs
wyfo: Ok.

PR #1086, opened 2024-06-05
bump quinn & rustls

JLerxky:
bump quinn to 0.11.1, rustls to 0.23.9, tokio-rustls to 0.26.0


Mallets:
@YuanYuYuan could you please help in reviewing this PR?

YuanYuYuan:
> @YuanYuYuan could you please help in reviewing this PR?

Sure.

Hi @JLerxky, it seems that the new dependence "aws-lc-rs" introduced by "rustls" 0.23.9 failed to pass the Windows Check. https://github.com/eclipse-zenoh/zenoh/actions/runs/9378878353/job/25937352481?pr=1086#step:7:143

image

JLerxky:
Okey @YuanYuYuan , I fixed it

io/zenoh-links/zenoh-link-tls/src/utils.rs
```diff
@@ -188,6 +188,11 @@ impl TlsServerConfig {
             bail!("No private key found for TLS server.");
         }
 
+        // Install rustls provider
+        rustls::crypto::ring::default_provider()
+            .install_default()
+            .ok();
```
Mallets, line 194: I think it's better to return an error if this operation fails.
```suggestion
            .map_err(|err| zerror!("Error installing TLS crypto provider: {err}."))?;
```

io/zenoh-links/zenoh-link-tls/src/utils.rs
```diff
@@ -269,6 +274,11 @@ impl TlsClientConfig {
             root_cert_store.extend(custom_root_cert.roots);
         }
 
+        // Install rustls provider
+        rustls::crypto::ring::default_provider()
+            .install_default()
+            .ok();
```
Mallets, line 280: I think it's better to return an error if this operation fails.
```suggestion
            map_err(|err| zerror!("Error installing TLS crypto provider: {err}."))?;
```

io/zenoh-links/zenoh-link-quic/src/utils.rs
```diff
@@ -271,93 +265,78 @@ impl TlsClientConfig {
...
         if let Some(custom_root_cert) = load_trust_anchors(config)? {
             tracing::debug!("Loading user-generated certificates.");
-            root_cert_store.roots.extend(custom_root_cert.roots);
+            root_cert_store.extend(custom_root_cert.roots);
         }
 
+        // Install rustls provider
+        rustls::crypto::ring::default_provider()
+            .install_default()
+            .ok();
```
Mallets, line 279: I think it's better to return an error if this operation fails.
```suggestion
            .map_err(|err| zerror!("Error installing TLS crypto provider: {err}."))?;
```

io/zenoh-links/zenoh-link-quic/src/utils.rs
```diff
@@ -160,40 +158,40 @@ impl TlsServerConfig {
...
         }
 
         if keys.is_empty() {
             bail!("No private key found for TLS server.");
         }
 
+        // Install rustls provider
+        rustls::crypto::ring::default_provider()
+            .install_default()
+            .ok();
```
Mallets, line 193: I think it's better to return an error if this operation fails.
```suggestion
            .map_err(|err| zerror!("Error installing TLS crypto provider: {err}."))?;
```

io/zenoh-links/zenoh-link-quic/src/unicast.rs
```diff
@@ -276,8 +270,17 @@ impl LinkManagerUnicastTrait for LinkManagerUnicastQuic {
             .map_err(|e| zerror!("Cannot create a new QUIC listener on {addr}: {e}"))?;
         server_crypto.server_config.alpn_protocols =
             ALPN_QUIC_HTTP.iter().map(|&x| x.into()).collect();
-        let mut server_config =
-            quinn::ServerConfig::with_crypto(Arc::new(server_crypto.server_config));
+
+        // Install rustls provider
+        rustls::crypto::ring::default_provider()
+            .install_default()
+            .ok();
```
Mallets, line 277: I think it's better to return an error if this operation fails.
```suggestion
            .map_err(|err| zerror!("Error installing TLS crypto provider: {err}."))?;
```

JLerxky:
``` rust
...
static PROCESS_DEFAULT_PROVIDER: OnceBox<Arc<CryptoProvider>> = OnceBox::new();
...

pub fn install_default(self) -> Result<(), Box<Arc<Self>>> {
    PROCESS_DEFAULT_PROVIDER.set(Box::new(Arc::new(self)))
}
```

JLerxky: After testing, maybe the warning shouldn't be added, because it will be printed whenever try to connect

Mallets: I think it's still safer to map and return the error. The behaviour you observed may actually be true for the current version of `quinn` and `rustls` but no guarantee is provided it will behave the same in future versions.
By simply converting and returning the error, everything will be handled by upper logic.

JLerxky: Hi @Mallets
I agree that it is only true for the current version，but can't throw error here, because when listening on the quic and connecting to the quic endpoint at the same time, install_default() will inevitably be executed multiple times, and all but the first execution will fail. If exec install_default() in only one place, and it has to install before ClientConfig::builder() or ServerConfig::builder(). 
That can only be done before the program is initialized and the listener begins. This will lose the effect of lazy loading crypto providers

Mallets: Right, I didn't have the knowledge about the fact that `install_default` will succeed the first time and it will fail any subsequent attempt. In this case it's fine to ignore the error but I'd add the explanation in a comment in the code on why we are ignoring the error, i.e. the explanation you just provided in your comment :)

Beside this point, the PR LGTM.

Mallets (approved):

PR #1097, opened 2024-06-07
Improve pipeline backoff

Mallets:
This PR improves the pipeline backoff implementation and reduces the possibility of spinning tasks.

Without this patch, it may happen before that in high-load scenario we could observe the following logs:
```
2024-06-07T07:56:30.783760Z  WARN tx-0 ThreadId(16) zenoh_transport::common::pipeline: Pipeline pull backoff overflow detected! Retrying in 4294967295ns. (tcp/127.0.0.1:7447 => tcp/127.0.0.1:51696)
2024-06-07T07:56:30.783771Z  WARN tx-0 ThreadId(16) zenoh_transport::common::pipeline: Pipeline pull backoff overflow detected! Retrying in 4294967295ns. (tcp/127.0.0.1:7447 => tcp/127.0.0.1:51696)
2024-06-07T07:56:30.783772Z  WARN tx-0 ThreadId(16) zenoh_transport::common::pipeline: Pipeline pull backoff overflow detected! Retrying in 4294967295ns. (tcp/127.0.0.1:7447 => tcp/127.0.0.1:51696)
2024-06-07T07:56:30.783774Z  WARN tx-0 ThreadId(16) zenoh_transport::common::pipeline: Pipeline pull backoff overflow detected! Retrying in 4294967295ns. (tcp/127.0.0.1:7447 => tcp/127.0.0.1:51696)
2024-06-07T07:56:30.783775Z  WARN tx-0 ThreadId(16) zenoh_transport::common::pipeline: Pipeline pull backoff overflow detected! Retrying in 4294967295ns. (tcp/127.0.0.1:7447 => tcp/127.0.0.1:51696)
2024-06-07T07:56:30.783830Z  WARN tx-0 ThreadId(16) zenoh_transport::common::pipeline: Pipeline pull backoff overflow detected! Retrying in 4294967295ns. (tcp/127.0.0.1:7447 => tcp/127.0.0.1:51696)
```

With this patch, in my tests I didn't observe those logs anymore.
What I expect is that this patch heavily mitigates the problem although it might not solve it in 100% of the cases.

### DIFF
--- a/io/zenoh-transport/src/common/pipeline.rs
+++ b/io/zenoh-transport/src/common/pipeline.rs
@@ -674,6 +674,9 @@ impl TransmissionPipelineConsumer {
                 }
             }
 
+            // Since tokio::timeout may not sleep immediately, yield the task first.
+            tokio::task::yield_now().await;
+
             // Wait for the backoff to expire or for a new message
             let _ =
                 tokio::time::timeout(Duration::from_nanos(bo as u64), self.n_out_r.recv_async())
--- a/io/zenoh-transport/src/common/pipeline.rs
+++ b/io/zenoh-transport/src/common/pipeline.rs
@@ -49,7 +49,6 @@ use zenoh_protocol::{
 type NanoSeconds = u32;
 
 const RBLEN: usize = QueueSizeConf::MAX;
-const TSLOT: NanoSeconds = 100;
 
 // Inner structure to reuse serialization batches
 struct StageInRefill {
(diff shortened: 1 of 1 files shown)


yellowhatter:
@Mallets looks nice!

I've been in parallel working on a little bit different design. Our queue implementation uses `mutex` to achieve synchronization around `current` batch. This is a weak point, because it makes a producer thread to go to sleep when `current` batch is being processed by a consumer task:
```
pipeline.rs: 138:
// Lock the current serialization batch.
let mut c_guard = self.mutex.current();
```
My idea is that the producer thread should take care about always having "one batch behind": after each serialization check atomic count of batches in `s_out` and move current batch (potentially not full) to `s_out` if it is empty. In this case, there won't be any producer/consumer contention around `current` batch. Regarding the performance in terms of batch size utilization, this approach will essentially regulate batch utilization, because misperforming (due to sending non-full batches) consumer will cause producer to manage to completely fill `current`, restoring consumer performance....

yellowhatter (approved):

PR #1098, opened 2024-06-07
Start link tx_task before notifying router

Mallets:
The routing logic is notified of a new link available upon new transport or link establishment.
However, the task consuming the transmission queue for that link is started only after the routing callback returns.
In case the routing logic needs to propagate many declarations, it will be filling up the queue and it will block since no transmission task has been started yet. 

This PR modifies this behaviour by starting the transmission task before calling the routing logic callback.


Mallets:
@yellowhatter can you please help in reviewing this PR?

yellowhatter (approved):
Looks good!

Issue #1070, opened 2024-05-30
Not clear the difference between `selector::Properties` and `selector::Parameters`


milyin:
### Describe the release item

mod selector exports `Properties` and `Parameters`which is wrapper for properties. This is hard to understand, why

Mallets:
`Properties` should not be exposed and kept internal. Only `Parameters` should be exposed.

wyfo:
@Mallets I think the issue is that `Properties` is currently exposed as the `Deref` target of `Parameters`, and as the only way to construct a `Parameters` instance via the `From<impl Into<Properties>>` implementation.

Mallets:
It is being reworked in #1100

Issue #1067, opened 2024-05-30
correct stable/unsable features


milyin:
### Describe the release item

- [x]  some `keyexpr` public functions could be marked as unstable, like `relation_to`. To be checked for all API.
- [x]  `keyexprtree` as `unstable`
- [x]  `attachment` should be stable (and not `unstable`)
- [x]  `selector::time_range` goes to `unstable`

milyin:
time_range is going to be fixed in #1100

PR #1100, opened 2024-06-09
properties renamed to parameters, selector rework

milyin:
This PR fixes issue #1070 and reduces importance of `Selector` type which is now just convenient wrapper for (KeyExpr, Parameters) pair without much usage throughout API. This also fixes the last item (time_range issue) in #1067

- empty type `zenoh_protocol::Parameters` removed, added bunch of standalone functions in module `zenoh_protocol::parameters` instead
- type `zenoh_protocol::Properties` renamed to `zenoh_protocol:Parameters`
- `zenoh_protocol:Parameters` reexported from `zenoh` without wrapper type
-  additional trait `PredefinedParameters` implemented for `Parameters` type. This trait adds getters/setters and constant names for specific parameters used by `zenoh`


zenoh-ext/src/querying_subscriber.rs
```diff
@@ -179,8 +178,8 @@ impl<'a, 'b, Handler> QueryingSubscriberBuilder<'a, 'b, crate::UserSpace, Handle
     #[inline]
     pub fn query_selector<IntoSelector>(mut self, query_selector: IntoSelector) -> Self
     where
-        IntoSelector: TryInto<Selector<'b>>,
-        <IntoSelector as TryInto<Selector<'b>>>::Error: Into<Error>,
+        IntoSelector: TryInto<KeyExpr<'b>>,
```
Mallets, line 181: I'm not sure this will work out. A string like `myresource?_time=[..]` will not be valid anymore.

zenoh/src/api/selector.rs
```diff
@@ -60,157 +55,125 @@ use super::{key_expr::KeyExpr, queryable::Query};
...
+    /// Sets the time range targeted by the selector parameters.
+    fn set_time_range<T: Into<Option<TimeRange>>>(&mut self, time_range: T);
+    /// Sets the parameter allowing to receieve replies from queryables not matching
+    /// the requested key expression. This may happen in this scenario:
+    /// - we are requesting keyexpr `a/b`.
+    /// - queryable is declared to handle `a/*` requests and contains data for `a/b` and `a/c`. 
+    /// - queryable receives our request and sends two replies with data for `a/b` and `a/c`
+    /// 
+    /// Normally only `a/b` reply would be accepted, but with `_anyke` parameter set, both replies are accepted.
+    fn set_reply_key_expr_any(&mut self);
```
milyin, line 131: @Mallets could you please check if I understand the logic of `_anyke` parameter correctly?

zenoh-ext/src/querying_subscriber.rs
milyin: Seems like an error after several renamings :-( Fixing

milyin: Thank you!

zenoh/src/api/selector.rs
Mallets: What you wrote is correct. In addition to that, when `_anyke` is defined, literally *any key expression* is allowed in the reply. I.e., the queryable could reply with `x/y/z` and that will be accepted.

Mallets (approved):

PR #1104, opened 2024-06-10
feat: remove `alive` flag in session/undeclarable objects

wyfo:
This flag was just used to prevent double close/undeclaration, i.e. object being drop after `close`/`undeclare`. Using `ManuallyDrop` instead in `close`/`undeclare` solve this issue, and save some space in structs (often one word counting the padding, which is not negligible).

### DIFF
--- a/zenoh/src/api/session.rs
+++ b/zenoh/src/api/session.rs
@@ -530,6 +530,8 @@ impl Session {
     /// # }
     /// ```
     pub fn close(self) -> impl Resolve<ZResult<()>> {
+        // ManuallyDrop wrapper prevents the drop code to be executed,
+        // which would lead to a double close
         let session = ManuallyDrop::new(self);
         ResolveFuture::new(async move {
             trace!("close()");
--- a/zenoh/src/api/session.rs
+++ b/zenoh/src/api/session.rs
@@ -2484,6 +2484,7 @@ impl Primitives for SessionClone {
 
 impl Drop for Session {
     fn drop(&mut self) {
+        // Use clone inner session, as it will be rewrapped in ManuallyDrop inside Session::close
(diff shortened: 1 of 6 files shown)


wyfo:
@Mallets @OlivierHecart

Mallets:
In general the approach LGTM.

However, could you please provide additional comments in the code on how those `ManuallyDrop` inter-dependency should be handled to avoid memory-leaks or misbehaviour? Is there a test covering this case?

wyfo:
> However, could you please provide additional comments in the code on how those ManuallyDrop inter-dependency should be handled to avoid memory-leaks or misbehaviour?

Not sure to have understood, but I've added comments; is it better now?

> Is there a test covering this case?

A lot of tests are already calling `undeclare()`. I cannot really test more. 
I mean, the only real test we could add would be to parse the logs after a call to `undeclare()`, and check there is no error caused by a double undeclaration. But I don't think we have any tests of this kind, and if we want to add it.

zenoh/src/api/session.rs
```diff
@@ -824,16 +827,25 @@ impl Session {
     }
 }
 
+/// Like a [`Session`], but not closed on drop
+pub(crate) struct SessionClone(ManuallyDrop<Session>);
```
Mallets, line 831: It's unclear to me the need of such `SessionClone`.  What real problem is it solved here?
Since `SessionClone` contains clones of `Runtime`, `State`, etc. how the state is cleaned up?
How is this change robust to internal changes, e.g. `Runtime` will required to be properly cleaned? If that happens, it seems we will be leaking memory or not freeing resources.

wyfo: Yes, that was a mistake of mine.

wyfo:
Closed in favor of #1106

PR #1106, opened 2024-06-10
feat: add unstable background method to subscriber/queryable/matching listeners

wyfo:
Following #1012, but completely reworked (no more builder method), hence a new PR.

~~Based on #1104, should wait to have it merge before merging this PR.~~

### DIFF
--- a/zenoh/src/api/publisher.rs
+++ b/zenoh/src/api/publisher.rs
@@ -13,11 +13,13 @@
 //
 
 use std::{
+    collections::HashSet,
     convert::TryFrom,
     fmt,
     future::{IntoFuture, Ready},
     mem::ManuallyDrop,
     pin::Pin,
+    sync::{Arc, Mutex},
     task::{Context, Poll},
 };
 
@@ -134,6 +136,7 @@ pub struct Publisher<'a> {
     pub(crate) priority: Priority,
     pub(crate) is_express: bool,
     pub(crate) destination: Locality,
+    pub(crate) matching_listeners: Arc<Mutex<HashSet<Id>>>,
 }
 
 impl<'a> Publisher<'a> {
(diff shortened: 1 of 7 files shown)


wyfo:
@Mallets @milyin Do you think the `unstable` feature is required for this API, as it is just a shortcut for `mem::forget`? 
I would like to remove `unstable` cfg and directly add into the documentation: "This method is just a convenient shortcut for calling `std::mem::forget`, it does nothing else."

wyfo:
@OlivierHecart you might be interested

wyfo:
@Mallets I've integrated our discussion (see the comments) and fixed my stupid mistakes. I've also renamed all the `alive` flags into `close_on_drop`/`undeclare_on_drop`, because explicit is better, and because this spelling is better with background feature.

zenoh/src/api/publisher.rs
```diff
@@ -160,28 +163,33 @@ impl<'a> Publisher<'a> {
...
+    pub fn priority(&self) -> Priority {
+        self.priority
     }
 
-    /// Restrict the matching subscribers that will receive the published data
-    /// to the ones that have the given [`Locality`](crate::prelude::Locality).
-    #[zenoh_macros::unstable]
+    /// Change the priority of the written data.
     #[inline]
-    pub fn set_allowed_destination(&mut self, destination: Locality) {
```
milyin, line 183: what happened with `set_allowed_destination`. It it's removal planned?

milyin (changes requested):

zenoh/src/api/publisher.rs
wyfo: I discussed about it with @OlivierHecart weeks ago, this thing is currently broken (the modification is not transmitted to the `PublisherState` stored in session, only the creation destination matter).
There was maybe another reason, but I forgot. @OlivierHecart do you remember?

zenoh/src/api/publisher.rs
```diff
@@ -350,6 +358,13 @@ impl<'a> Publisher<'a> {
     pub fn undeclare(self) -> impl Resolve<ZResult<()>> + 'a {
         Undeclarable::undeclare_inner(self, ())
     }
+
+    fn undeclare_matching_listeners(&self) -> ZResult<()> {
+        for id in zlock!(self.matching_listeners).drain() {
```
Mallets, line 363: This code would be safer w.r.t. mutex deadlock if:

1. it takes the lock
2. it drains the iterator in a supporting vec
3. it releases the lock
4. it undeclares the matching listeners

wyfo: #1116

Issue #1107, opened 2024-06-10
QoS parameters on queries


DariusIMP:
### Describe the bug

On `dev/1.0.0` branch, when doing a query operation, as well as when replying to the query, if we specify QoS parameters, those aren't properly loaded.

### To reproduce

Take the [z_get](https://github.com/eclipse-zenoh/zenoh/blob/dev/1.0.0/examples/examples/z_get.rs) and the [z_queryable](https://github.com/eclipse-zenoh/zenoh/blob/dev/1.0.0/examples/examples/z_queryable.rs) examples. 

I haven't reproduced this issue on Pub/Sub examples.

## Query

Edit the z_get example to change the QoS parameters, for instance:

```
diff --git a/examples/examples/z_get.rs b/examples/examples/z_get.rs
index d4fc416f9..2d2bf16a0 100644
--- a/examples/examples/z_get.rs
+++ b/examples/examples/z_get.rs
@@ -37,7 +37,10 @@ async fn main() {
         .value(value)
         .target(target)
         .timeout(timeout)
-        .await
+        .congestion_control(zenoh::publisher::CongestionControl::Block)
+        .priority(zenoh::core::Priority::Background)
+        .express(true)
+        .await
         .unwrap();
     while let Ok(reply) = replies.recv_async().await {
         match reply.result() {
```

Run the example:

```
RUST_LOG=trace ./z_get -e tcp/localhost:7447 --no-multicast-scouting
```

And see the log, where the QoS parameters do not correspond to what we specified:
```
[...]

PR #1108, opened 2024-06-10
fix forgotten QoS and Timestamp ext processing for query/reply

yellowhatter:
Issue #1107


yellowhatter:
@Mallets @DariusIMP

DariusIMP:
LGTM, I have tested it and it solves the issue.

However there is still another issue which may not be related (and I guess we can open a different issue for it) in which the priority is automatically set to Data despite specifying different values, both while doing a query and when replying to it.

DariusIMP (approved):

Mallets:
If the priority is automatically set to Data despite specifying different values then I wonder wether the proposed changes are enough. I think also at API level `QoS` parameters should be mapped.

yellowhatter:
Fixed bug in QoS bitfield. It affected ALL our QoS Priority in all the messages.

DariusIMP:
Tested it and it works fine now! 👍

Mallets (approved):

PR #1113, opened 2024-06-10
feat: add `auto_encoding` method to builder to set encoding based from payload type

### LINKED ISSUE #812: Enable auto filling of encoding based on type information
### Describe the feature

Provide `encoding::auto` for automatic encoding selection based on the information from the type system.
I.e.: the feature would allow to have a publisher with `auto` encoding and any following `put()` will automatically set the right encoding.

Desired behaviour:
```rust
let pub = session.declare_publisher("foo/test").with_encoding("auto").res().await.unwrap();
pub.put(1).res().await.unwrap();
pub.put(-1).res().await.unwrap();
pub.put(1.01).res().await.unwrap();
pub.put("mystring").res().await.unwrap();
```
Equivalent to existing behaviour:
```rust
let pub = session.declare_publisher("foo/test").res().await.unwrap();
pub.put(1).with_encoding("zenoh/uint").res().await.unwrap();
pub.put(-1).with_encoding("zenoh/int").res().await.unwrap();
pub.put(1.01).with_encoding("zenoh/float").res().await.unwrap();
pub.put("mystring").with_encoding("zenoh/string").res().await.unwrap();
```
More information about the reworked encoding can be found in https://github.com/eclipse-zenoh/zenoh/pull/764.

wyfo:
Fixes #812 

@Mallets @milyin

### DIFF
--- a/plugins/zenoh-plugin-rest/src/lib.rs
+++ b/plugins/zenoh-plugin-rest/src/lib.rs
@@ -34,7 +34,7 @@ use zenoh::{
     },
     key_expr::{keyexpr, KeyExpr},
     query::{QueryConsolidation, Reply},
-    sample::{Sample, SampleKind, ValueBuilderTrait},
+    sample::{Sample, SampleKind},
     selector::{Selector, TIME_RANGE_KEY},
     session::{Session, SessionDeclarations},
     value::Value,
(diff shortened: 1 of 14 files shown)


wyfo:
After discussion with @Mallets, this PR can be closed for now, as we are not sure about wether or not this feature is really wanted.

PR #1118, opened 2024-06-11
rename wrapped zenoh-protocol structures to `...Proto`, wrap and expose `EntityGlobalId`

milyin:
There are structures with public fields and/or public methods which should not be in the API in the internal crates. These structures should be also exported in the API without these public items. This update proposes common rule how tho handle such structures:
- structures which are completely internal for zenoh crate should have suffix `Inner` (this is the case for some shm structs)
- structures which are used by extra libraries / tools, like structures exposed by zenoh-protocol should have suffix corresponding to crate name, like `...Proto`
- corresponding wrapper API structures reexported by zenoh have no suffix and located now in `zenoh-config::wrappers` crate

This PR proposes updates above for `ZenohId`, `EntityGlobalId` and `Hello`

Also it moves `ZenohId` from `config::` to `info::`. This seems more logical

### DIFF
--- a/commons/zenoh-protocol/src/core/mod.rs
+++ b/commons/zenoh-protocol/src/core/mod.rs
@@ -341,8 +341,8 @@ impl EntityGlobalIdProto {
 pub struct EntityGlobalId(EntityGlobalIdProto);
 
 impl EntityGlobalId {
-    pub fn zid(&self) -> ZenohIdProto {
-        self.0.zid
+    pub fn zid(&self) -> ZenohId {
+        self.0.zid.into()
     }
 
     pub fn eid(&self) -> EntityId {
--- a/commons/zenoh-protocol/src/core/mod.rs
+++ b/commons/zenoh-protocol/src/core/mod.rs
@@ -319,12 +319,12 @@ pub type EntityId = u32;
 
 /// The global unique id of a zenoh entity.
 #[derive(Debug, Default, Copy, Clone, Eq, Hash, PartialEq)]
-pub struct EntityGlobalId {
+pub struct EntityGlobalIdInner {
     pub zid: ZenohIdInner,
     pub eid: EntityId,
 }
 
-impl EntityGlobalId {
+impl EntityGlobalIdInner {
     #[cfg(feature = "test")]
     pub fn rand() -> Self {
         use rand::Rng;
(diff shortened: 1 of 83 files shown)


commons/zenoh-protocol/src/core/mod.rs
```diff
@@ -77,21 +77,60 @@ impl ZenohId {
...
 }
 
-impl Default for ZenohId {
+impl Default for ZenohIdProto {
     fn default() -> Self {
         Self::rand()
     }
 }
 
+/// The global unique id of a zenoh peer.
```
Mallets, line 95: I would expect this type to be defined in the API and not in `zenoh-protocol`.

commons/zenoh-protocol/src/core/mod.rs
```diff
@@ -278,22 +320,42 @@ pub type EntityId = u32;
...
         }
     }
 }
 
+#[derive(Debug, Default, Copy, Clone, Eq, Hash, PartialEq)]
+#[repr(transparent)]
+pub struct EntityGlobalId(EntityGlobalIdProto);
+
+impl EntityGlobalId {
+    pub fn zid(&self) -> ZenohIdProto {
```
Mallets, line 344: Shouldn't return `ZenohId`?

commons/zenoh-protocol/src/scouting/hello.rs
```diff
@@ -129,3 +130,39 @@ impl Hello {
         }
     }
 }
+
+/// A zenoh Hello message.
+pub struct Hello(HelloProto);
```
Mallets, line 135: I would expect this type to be defined in the API and not in `zenoh-protocol`.

milyin: On the one hand this is right, but sometimes it doesn't work. Especially this doesn't work for `ZenohId`: it's a wrapper for `ZenohIdProto` and logically should be in `zenoh`, but it's also used by `zenoh-config` so it should be outside of `zenoh` to avoid circular dependency.
So the approach when all wrappers for reexported structures should be located near the original structure is systematic and guarantees no dependency problems. Another solution would be to add special crate for wrapped structures which can be reused by other crates. But 1) in some rare cases this also can cause circular dependency and 2) this would require adding one more crate, increasing complexity of our project

So I think that the approach when reexported wrappers are declared near the original structures is the most acceptable

commons/zenoh-protocol/src/core/mod.rs
milyin: see below

commons/zenoh-protocol/src/core/mod.rs
milyin: thank you, fixed

commons/zenoh-protocol/src/scouting/hello.rs
Mallets: I think the real problem is that `zenoh-config` uses *API-level* types directly from `zenoh-protocol`. But `zenoh-protocol` per-se does not need to care about those wrapper. 

In `zenoh-protocol` everything is defined and exported as public to detect as many breaking changes as possible throughout the codebase as soon as one minimal change happens. Adding the wrapper inside `zenoh-protocol` may lead to more misusage to me than anything else. Could the wrapper defined in `zenoh-config` and then re-exported by `zenoh` and then leave untouched the `zenoh-protocol`?

milyin: Ok, that's another solution. Let's put all such wrappers to `zenoh-config`. This will cause reexport duplication at this moment because we reexporting `zenoh-config::*` in `config` crate and also should reexport `Hello` in `scoiting`, etc. But I think this is not a big issue at this moment, but better to make reexport from `zenoh-config` explicit in 1.0.0

Mallets (approved):

PR #1119, opened 2024-06-11
feat: remove `Value` from the public API, make it internal for plugins

wyfo:
@Mallets @milyin

### DIFF
--- a/zenoh/src/api/builders/sample.rs
+++ b/zenoh/src/api/builders/sample.rs
@@ -23,7 +23,6 @@ use crate::api::{
     key_expr::KeyExpr,
     publisher::Priority,
     sample::{QoS, QoSBuilder, Sample, SampleKind},
-    value::Value,
 };
 #[cfg(feature = "unstable")]
 use crate::sample::SourceInfo;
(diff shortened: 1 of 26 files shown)


zenoh/src/api/query.rs
```diff
@@ -79,27 +79,63 @@ impl Default for QueryConsolidation {
...
+impl ReplyError {
+    /// Gets the payload of this ReplyError.
+    #[inline]
+    pub fn payload(&self) -> &ZBytes {
+        &self.payload
+    }
+
+    /// Gets the payload of this ReplyError.
+    #[inline]
+    pub fn payload_mut(&mut self) -> &mut ZBytes {
```
Mallets, line 98: Is `payload_mut` really necessary?

zenoh/src/api/builders/sample.rs
```diff
@@ -99,6 +93,14 @@ impl SampleBuilder<SampleBuilderPut> {
             _t: PhantomData::<SampleBuilderPut>,
         }
     }
+
+    pub fn payload<IntoZBytes>(mut self, payload: IntoZBytes) -> Self
```
Mallets, line 97: I don't think this is needed. Actual payload is already provided in the `put()` function.

zenoh/src/api/query.rs
wyfo: It was copied from `Value`, but I think we can remove it yes, so I will do it.

zenoh/src/api/builders/sample.rs
Mallets: You are right. In that case it makes sense to keep it.

Mallets:
As it stands this PR LGTM. However CI should be green before approving and merging (modulo the memory leak test that still needs to be fixed).

Mallets (approved):

PR #1122, opened 2024-06-12
refactor: make `Encoding` take only one word in memory

wyfo:
The only visible side-effect is that encoding constants can no more be used in pattern matching.

### DIFF
--- a/commons/zenoh-codec/src/core/encoding.rs
+++ b/commons/zenoh-codec/src/core/encoding.rs
@@ -73,7 +73,7 @@ where
         let mut encoding = Encoding::new(id);
         if has_schema {
             let zodec = Zenoh080Bounded::<u8>::new();
-            let schema: String = zodec.read(&mut *reader)?;
+            let schema: alloc::string::String = zodec.read(&mut *reader)?;
             encoding = encoding.with_schema(schema);
         }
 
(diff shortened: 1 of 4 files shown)


wyfo:
@Mallets @OlivierHecart This is a "crazy" idea I had yesterday, and it seems to work well. Do we have a benchmark suite to test the impact seriously?

Mallets:
As far as I understand from this PR, I believe there is a misconception on `Encoding`. The encoding is not either an `ID` or a `String`. It can be an `ID` plus a `String` at any point of time. So I don't think this proposal is valid.

wyfo:
> As far as I understand from this PR, I believe there is a misconception on Encoding. The encoding is not either an ID or a String. It can be an ID plus a String at any point of time

I think you misunderstood the PR, `Encoding` always contains an id. `Encoding::id_and_schema` returns `(EncodingId, Option<&str>)`.

wyfo:
On my computer, comparing `dev/1.0.0` branch with this one, running `z_pub_thr 8`/`z_sub_thr --number 10000000` (with `--release` obviously), I obtain almost 5% performance increase, from 4.15Mmsg/s to 4.30Mmsg/s. I was expecting a theoretical increase, but not such a big one, I'm even doubting about it, but for the moment, when I execute both benchmark one after the other, the increase remains constant.

That's why I'm asking for more serious benchmark suite, because I'm doubtful.

Issue #1093, opened 2024-06-06
Listening on ANY interface, Link.interfaces contains all the interfaces rather than the source one


JEnoch:
### Describe the bug

This leads access control to misbehave, the interceptor being called for a wrong interface.

### To reproduce

1. Run `zenohd` with this conf file that denies put on WiFi interface:
```json5
{
  access_control:{
    enabled:true,
    default_permission:"allow",
    rules:[
      {
        permission:"deny", actions:["put"],
        flows:["ingress"], interfaces: ["wlp1s0"],
        key_exprs:["demo/**"]
      },
    ]
  },
}
```
2. Run `z_sub -m client -e tcp/localhost:7447`
3. Run `z_pub -m client -e tcp/localhost:7447`

The subscriber doesn't receive the publications. It should receive because the traffic only goes via `lo` interface, and the deny rule is only for WiFi interface.

### System info

- Zenoh [c279982](https://github.com/eclipse-zenoh/zenoh/commit/c279982c131c556ebd474f4ffa9fd6c096d13830)

Mallets:
Fix provided in:
- [x] TCP https://github.com/eclipse-zenoh/zenoh/pull/1123
- [x] TLS https://github.com/eclipse-zenoh/zenoh/pull/1123
- [x] QUIC https://github.com/eclipse-zenoh/zenoh/pull/1123
- [x] WS https://github.com/eclipse-zenoh/zenoh/pull/1123
- [ ] UDP - fix not yet provided

PR #1123, opened 2024-06-12
Fix interface name scanning when listening on IP unspecified for TCP/TLS/QUIC/WS

Mallets:
Mostly fixes https://github.com/eclipse-zenoh/zenoh/issues/1093.

While the fix is quite straightforward for connection-oriented links, i.e. TCP/TLS/QUIC/WS, a major work may be needed in the UDP implementation to retrieve the right interface a message was received on.
For this reason, UDP is left out of this PR and a warning is added.

### DIFF
--- a/io/zenoh-links/zenoh-link-udp/src/unicast.rs
+++ b/io/zenoh-links/zenoh-link-udp/src/unicast.rs
@@ -499,7 +499,7 @@ async fn accept_read_task(
     tracing::trace!("Ready to accept UDP connections on: {:?}", src_addr);
 
     if src_addr.ip().is_unspecified() {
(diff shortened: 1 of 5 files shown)


io/zenoh-links/zenoh-link-udp/src/unicast.rs
```diff
@@ -498,6 +498,10 @@ async fn accept_read_task(
 
     tracing::trace!("Ready to accept UDP connections on: {:?}", src_addr);
 
+    if src_addr.ip().is_unspecified() {
+        tracing::warn!("Interceptors (e.g. Access Control, Downsampling) are not guaranteed to work on UDP when listening on 0.0.0.0 or [::]. Their usage is discouraged. See https://github.com/eclipse-zenoh/zenoh/issues/1093.");
```
JEnoch: Refer `https://github.com/eclipse-zenoh/zenoh/issues/1126` rather than `https://github.com/eclipse-zenoh/zenoh/issues/1093`

JEnoch (approved):

JEnoch:
#1126 created for UDP case.

PR #1128, opened 2024-06-12
Fix memory leak CI

Mallets:
Import zenoh with unstable features


wyfo:
LGTM, and CI passes, so the goal is achived.

PR #1130, opened 2024-06-13
Fix interest protocol when client connected to peer

OlivierHecart:
Fixed 2 problems:
1. When peer was receiving interest from client it was not sending it's known declarations (from other peers).
2. When peer was receiving interest from client it was sending DeclareFinal back immediately without waiting for the DeclareFinals from the routers it is eventually connected to.

To fix point 2 I had to implement interest bread crumbing. I did so through a code reorganization that would allow liveliness token refactor to also benefit from it.


zenoh/src/net/routing/dispatcher/tables.rs
```diff
@@ -175,7 +175,9 @@ pub fn close_face(tables: &TablesLock, face: &Weak<FaceState>) {
             tracing::debug!("Close {}", face);
             face.task_controller.terminate_all(Duration::from_secs(10));
             finalize_pending_queries(tables, &mut face);
-            zlock!(tables.ctrl_lock).close_face(tables, &mut face);
+            let ctrl_lock = zlock!(tables.ctrl_lock);
+            finalize_pending_interests(tables, &mut face);
```
fuzzypixelz, line 179: Why do we need the CTRL lock finalize pending Interests but not Queries?

zenoh/src/net/routing/dispatcher/face.rs
```diff
@@ -212,57 +215,17 @@ impl Primitives for Face {
     fn send_interest(&self, msg: zenoh_protocol::network::Interest) {
         let ctrl_lock = zlock!(self.tables.ctrl_lock);
         if msg.mode != InterestMode::Final {
```
fuzzypixelz, line 217: Maybe change this to `msg.mode.final()` and invert the branches?

```suggestion
        if msg.mode.final() {
```

zenoh/src/net/routing/hat/p2p_peer/interests.rs
```diff
@@ -0,0 +1,224 @@
...
+                    wire_expr,
+                    ext_qos: ext::QoSType::DECLARE,
+                    ext_tstamp: None,
+                    ext_nodeid: ext::NodeIdType::DEFAULT,
+                },
+                res.as_ref().map(|res| res.expr()).unwrap_or_default(),
+            ));
+        }
+
+        if mode.current() {
```
fuzzypixelz, line 156: Maybe I'm missing something here. But shouldn't we send the `DeclareFinal`s only after the above Current Interests finalize (either with replies or a timeout)? Because if a client declares an Interest to a peer we would immediatly reply with `DeclareFinal` and thus the client would do writer side filtering without full knowledge of Subscribers from the Peer subsystem's Routers.

zenoh/src/net/routing/hat/router/interests.rs
```diff
@@ -0,0 +1,96 @@
...
+        res: Option<&mut Arc<Resource>>,
+        mode: InterestMode,
+        mut options: InterestOptions,
+    ) {
+        if options.aggregate() && face.whatami == WhatAmI::Peer {
+            tracing::warn!(
+                "Received Interest with aggregate=true from peer {}. Not supported!",
+                face.zid
+            );
+            options -= InterestOptions::AGGREGATE;
```
fuzzypixelz, line 55: Shouldn't we print an error and return instead? We author the code that might get these options wrong. I think if this happens it's a bug, not a sign that the user is doing something fishy.

fuzzypixelz: I just realized that this code seems to predate this pull request. We don't really have to discuss it here.

zenoh/src/net/routing/hat/client/interests.rs
```diff
@@ -0,0 +1,173 @@
...
+};
+
+pub(super) fn interests_new_face(tables: &mut Tables, face: &mut Arc<FaceState>) {
+    for mut src_face in tables
+        .faces
+        .values()
+        .cloned()
+        .collect::<Vec<Arc<FaceState>>>()
+    {
+        if face.whatami != WhatAmI::Client {
```
Mallets, line 44: The filter could be applied to the iterator above. In this way you avoid some unnecessary cloning when the `face.whatami != WhatAmI::Client`.

zenoh/src/net/routing/dispatcher/tables.rs
OlivierHecart: In both cases, we need a lock to prevent concurrent insertion and removal of pending queries/interests. Queries have their dedicated lock to prevent that. I considered interest as control messages and so reused the control lock (already used at insertion anyway) for interests.

zenoh/src/net/routing/hat/p2p_peer/interests.rs
OlivierHecart: You are right. That's the point of the `if let Some(interest) = Arc::into_inner(interest)` check just after which will only be true if the interest has not been propagated anywhere (when the peer is not connected to any router) and the `Arc` `strong_count` is 1.

zenoh/src/net/routing/dispatcher/face.rs
fuzzypixelz: Yes there is no `final` function (we can add it easily anyway...). I still think it easier to read if the branches are inverted. This is mostly a nitpick.

zenoh/src/net/routing/hat/client/interests.rs
OlivierHecart: Well, even better, as this check only applies on the function parameter we can do the check once before iterating. Fixed.

Mallets (approved):

PR #1134, opened 2024-06-13
fix: fix proto wrappers debug impl

wyfo:
@milyin

### DIFF
--- a/commons/zenoh-config/src/wrappers.rs
+++ b/commons/zenoh-config/src/wrappers.rs
@@ -25,12 +25,15 @@ use zenoh_protocol::{
 };
 
 /// The global unique id of a zenoh peer.
-#[derive(
-    Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Serialize, Deserialize, Debug, Default,
-)]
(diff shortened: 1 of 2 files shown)


commons/zenoh-config/src/wrappers.rs
```diff
@@ -99,7 +103,13 @@ impl fmt::Display for Hello {
     }
 }
 
-#[derive(Debug, Default, Copy, Clone, Eq, Hash, PartialEq)]
+impl fmt::Display for Hello {
```
wyfo, line 106: @Mallets @milyin Is there a reason to have a `Display` implementation that looks like a `Debug` one on `Hello`.
When the previous `Hello` type contained the `version` in its `Debug` implementation, why not, but now, shouldn't we remove this `Display` implementation?

milyin (approved):

commons/zenoh-config/src/wrappers.rs
Mallets: I would keep `Debug` printing all information and the `Display` a simplified version of it.

Mallets: To me the `Debug` should print the whole internal state.

wyfo: Ok, so I will fix the `Debug` impl.

Issue #709, opened 2024-02-01
Is replier_id always needed in Reply?


Mallets:
### Describe the release item

[Reply](https://docs.rs/zenoh/0.10.1-rc/zenoh/query/struct.Reply.html) includes a mandatory `replier_id` of type [ZenohId](https://docs.rs/zenoh/0.10.1-rc/zenoh/prelude/struct.ZenohId.html). 
However, the [Sample](https://docs.rs/zenoh/0.10.1-rc/zenoh/sample/struct.Sample.html) accepted in the [reply()](https://docs.rs/zenoh/0.10.1-rc/zenoh/queryable/struct.Query.html#method.reply) contains a [SourceInfo](https://docs.rs/zenoh/0.10.1-rc/zenoh/sample/struct.SourceInfo.html) containing an optional [ZenohId](https://docs.rs/zenoh/0.10.1-rc/zenoh/prelude/struct.ZenohId.html).

In the light of this, shouldn't `replier_id` be of type `Option<ZenohId>` instead of `ZenohId`? If `ZenohId` is always present in the [Reply](https://docs.rs/zenoh/0.10.1-rc/zenoh/query/struct.Reply.html), then who is setting it all the time? Is this behaviour compatible with zenoh-pico?

Mallets:
It seems the `ZenohId` is set in https://github.com/eclipse-zenoh/zenoh/blob/5601669ae4f1409e2080682acdb54a8f1d700ee8/zenoh/src/session.rs#L2400C58-L2400C65.
I believe the current behaviour is not the expected one and the `replier_id` should be an `Option<ZenohId>`.

Mallets:
Linking C structure in zenoh-pico: [z_reply_data_t](https://zenoh-pico.readthedocs.io/en/0.10.1-rc/api.html?highlight=z_owned_reply_t#c.z_reply_data_t)

milyin:
@Mallets @OlivierHecart 
@DariusIMP noticed, that there is a problem with this issue. Before the fix the `reply_id` was set to random, which was definitely wrong: https://github.com/eclipse-zenoh/zenoh/pull/1138/files#diff-f0aac9d83f0d53d0be9a6ee02269c26605c3920dbd6782f744e59c21aec235d5L2311

But it seems that this field is unrelated to `SourceId` field of the sample. In case of error this field is set to zid of the replying session. I see no reason to not to do the same for normal replies. So maybe we should remove this option and just always set this field with current session zid ?

[...]

PR #1138, opened 2024-06-13
replier_id() returns Option<ZenohId>

milyin:
Fix for issue #709 
Also leak of `ZenohIdProto` fixed


wyfo:
LGTM, also, you can trigger the CI again, it should work

Mallets (approved):

PR #1139, opened 2024-06-13
fix: fix clippy warnings for 1.79

wyfo:
@Mallets

### DIFF
--- a/commons/zenoh-buffers/src/slice.rs
+++ b/commons/zenoh-buffers/src/slice.rs
@@ -77,7 +77,7 @@ impl Writer for &mut [u8] {
         // SAFETY: this doesn't compile with simple assignment because the compiler
         // doesn't believe that the subslice has the same lifetime as the original slice,
         // so we transmute to assure it that it does.
-        *self = unsafe { mem::transmute(lhs) };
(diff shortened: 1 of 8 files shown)


wyfo:
I'm honestly worried about having so much transmute in the code (I've replaced a few of them), but for now, I just obeyed to the warning.

As the problematic transmute are about lifetime, particularly for mutable reference writes, you may be interested in https://doc.rust-lang.org/src/std/io/impls.rs.html#362-399 or https://github.com/tokio-rs/bytes/blob/fa1daac3ae1dcb07dffe3a41a041dffd6edf177b/src/buf/buf_mut.rs#L1493. I didn't try to modify this unsafe code myself in this PR.

Mallets (approved):

Mallets:
> I'm honestly worried about having so much transmute in the code (I've replaced a few of them), but for now, I just obeyed to the warning.
> 
> As the problematic transmute are about lifetime, particularly for mutable reference writes, you may be interested in https://doc.rust-lang.org/src/std/io/impls.rs.html#362-399 or https://github.com/tokio-rs/bytes/blob/fa1daac3ae1dcb07dffe3a41a041dffd6edf177b/src/buf/buf_mut.rs#L1493. I didn't try to modify this unsafe code myself in this PR.

Interesting, it is definitely worth a try.

PR #1142, opened 2024-06-14
feat: add `Selector::into_owned`

wyfo:
@milyin


wyfo:
The real fix I need for the Python binding is the `Query::parameter` one, but I thought `Selector::into_owned` would be nice.

Mallets (approved):

milyin (approved):

PR #1144, opened 2024-06-14
Update lib.rs

yellowhatter:
This is a completely wrong error, we need to remove it

### DIFF
--- a/zenoh/src/lib.rs
+++ b/zenoh/src/lib.rs
@@ -398,11 +398,6 @@ pub mod internal {
     pub use crate::api::value::Value;
 }
 
-#[cfg(all(feature = "shared-memory", not(feature = "unstable")))]
-compile_error!(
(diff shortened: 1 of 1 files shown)


yellowhatter:
@milyin @Mallets

Mallets:
Why this error is wrong? `shared-memory` support is `unstable` at the moment.

yellowhatter:
> Why this error is wrong? `shared-memory` support is `unstable` at the moment.

`shared-memory` support is not unstable, `shm API` is. It is ok (and intended) to compile zenoh with `shared-memory` but without `unstable` and have transparent shm buffers reception and reading support without having access to specific shm API. During further development we might stabilize shm API leaving some of it's pieces unstable (for example, currently not-exising allocator api), and the fact that there is no strict relation between `shared-memory` and `unstable` features becomes 100% logical. We just stand in the beginning of this path, having the whole shm API `unstable`.....

PR #1149, opened 2024-06-16
Dev/1.0.0 plugin update

milyin:
updates necessary for external plugins:
- conversion of `ZenohId` to keyexpr exposed - it's used by plugins for making adminspace path
- exported function `init_log_from_env_or`

### DIFF
--- a/commons/zenoh-config/src/wrappers.rs
+++ b/commons/zenoh-config/src/wrappers.rs
@@ -20,7 +20,7 @@ use std::str::FromStr;
 
 use serde::{Deserialize, Serialize};
 use zenoh_protocol::{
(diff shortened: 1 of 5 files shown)


commons/zenoh-config/src/wrappers.rs
```diff
@@ -29,6 +29,14 @@ use zenoh_protocol::{
 #[repr(transparent)]
 pub struct ZenohId(ZenohIdProto);
 
+impl ZenohId {
+    /// Used by plugins for crating adminspace path
+    #[zenoh_macros::unstable]
+    pub fn into_keyexpr(self) -> OwnedKeyExpr {
```
evshary, line 35: This is exactly what I need :+1:

Issue #1054, opened 2024-05-27
``PublicationCache is not dropped


YuanYuYuan:
### Describe the bug

A new session creation hangs if a `PublicationCache` of another closed session is not dropped.

### To reproduce

```rust
use config::ModeDependentValue;
use zenoh_ext::SessionExt;
use zenoh::prelude::r#async::*;

#[tokio::main]
async fn main() {
    {
        let mut config = config::peer();
        config
            .timestamping
            .set_enabled(Some(ModeDependentValue::Unique(true)))
            .unwrap();

        println!("Open a zenoh session.");
        let session = zenoh::open(config).res().await.unwrap();

        println!("Declare a publisher.");
        session.declare_publisher("asdfasdfas").res().await.unwrap();

        // Comment out this block can pass the test
        println!("Declare a publisher cache.");
        session
            .declare_publication_cache("asdfasdf")
            .res()
            .await
            .unwrap();

        println!("Drop the session.");
    }
    println!("Open a new zenoh session.");
    zenoh::open(config::peer()).res().await.unwrap();

    println!("Done.");
}
```

### System info

- zenoh: [25f06bd](https://github.com/eclipse-zenoh/zenoh/commit/25f06bd9e71b4dbed5ba8d75eaa1708693e7b9ec)

YuanYuYuan:
It seems that we need to replace the `close` function [here](https://github.com/eclipse-zenoh/zenoh/blob/ed6c636a84c00552ca7f7c07d075d5783a40de44/zenoh-ext/src/publication_cache.rs?plain=1#L255) with the `Undeclare` trait to align the undeclaration used in `Queryable`, `Subscriber`, etc.

YuanYuYuan:
Possibly affect https://github.com/eclipse-zenoh/zenoh-c/blob/5a1728e09d60a093b2abee8b876925184e636067/src/publication_cache.rs?plain=1#L173

sashacmc:
PublicationCache was deprecated

PR #1151, opened 2024-06-17
feat: make `TerminatableTask` terminate itself when dropped

YuanYuYuan:
This PR introduces an auto termination when dropped and resolves the bug in #1054.

### DIFF
--- a/commons/zenoh-task/src/lib.rs
+++ b/commons/zenoh-task/src/lib.rs
@@ -130,8 +130,14 @@ impl TaskController {
 }
 
 pub struct TerminatableTask {
-    handle: JoinHandle<()>,
+    handle: Option<JoinHandle<()>>,
     token: CancellationToken,
+}
+
+impl Drop for TerminatableTask {
+    fn drop(&mut self) {
+        self.terminate(std::time::Duration::from_secs(10));
+    }
 }
 
 impl TerminatableTask {
@@ -147,7 +153,7 @@ impl TerminatableTask {
         T: Send + 'static,
     {
         TerminatableTask {
-            handle: rt.spawn(future.map(|_f| ())),
+            handle: Some(rt.spawn(future.map(|_f| ()))),
             token,
         }
     }
@@ -168,24 +174,26 @@ impl TerminatableTask {
         };
 
         TerminatableTask {
-            handle: rt.spawn(task),
+            handle: Some(rt.spawn(task)),
             token,
         }
     }
 
     /// Attempts to terminate the task.
     /// Returns true if task completed / aborted within timeout duration, false otherwise.
-    pub fn terminate(self, timeout: Duration) -> bool {
+    pub fn terminate(&mut self, timeout: Duration) -> bool {
         ResolveFuture::new(async move { self.terminate_async(timeout).await }).res_sync()
     }
 
     /// Async version of [`TerminatableTask::terminate()`].
-    pub async fn terminate_async(self, timeout: Duration) -> bool {
+    pub async fn terminate_async(&mut self, timeout: Duration) -> bool {
         self.token.cancel();
-        if tokio::time::timeout(timeout, self.handle).await.is_err() {
-            tracing::error!("Failed to terminate the task");
-            return false;
-        };
+        if let Some(handle) = self.handle.take() {
+            if tokio::time::timeout(timeout, handle).await.is_err() {
+                tracing::error!("Failed to terminate the task");
+                return false;
+            };
+        }
         true
     }
 }
--- a/zenoh-ext/src/publication_cache.rs
+++ b/zenoh-ext/src/publication_cache.rs
@@ -257,7 +257,7 @@ impl<'a> PublicationCache<'a> {
             let PublicationCache {
                 _queryable,
                 local_sub,
-                task,
+                mut task,
             } = self;
             _queryable.undeclare().res_async().await?;
             local_sub.undeclare().res_async().await?;
--- a/zenoh/src/net/routing/hat/linkstate_peer/mod.rs
+++ b/zenoh/src/net/routing/hat/linkstate_peer/mod.rs
@@ -47,7 +47,6 @@ use std::{
     any::Any,
     collections::{HashMap, HashSet},
     sync::Arc,
-    time::Duration,
 };
 use zenoh_config::{unwrap_or_default, ModeDependent, WhatAmI, WhatAmIMatcher, ZenohId};
 use zenoh_protocol::{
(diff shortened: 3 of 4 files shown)


YuanYuYuan:
@wyfo could you please review this? Thanks.

commons/zenoh-task/src/lib.rs
```diff
@@ -168,24 +174,26 @@ impl TerminatableTask {
...
-            handle: rt.spawn(task),
+            handle: Some(rt.spawn(task)),
             token,
         }
     }
 
     /// Attempts to terminate the task.
     /// Returns true if task completed / aborted within timeout duration, false otherwise.
-    pub fn terminate(self, timeout: Duration) -> bool {
+    pub fn terminate(&mut self, timeout: Duration) -> bool {
```
wyfo, line 184: I don't think you should change the receiver type of the method

commons/zenoh-task/src/lib.rs
```diff
@@ -168,24 +174,26 @@ impl TerminatableTask {
...
         ResolveFuture::new(async move { self.terminate_async(timeout).await }).res_sync()
     }
 
     /// Async version of [`TerminatableTask::terminate()`].
-    pub async fn terminate_async(self, timeout: Duration) -> bool {
+    pub async fn terminate_async(&mut self, timeout: Duration) -> bool {
```
wyfo, line 189: Idem about the receiver type

wyfo (commented):
I see why you changed the receiver type, but I think you should not try to reuse `terminate` in `drop`.
Instead, you can use some intermediate function:
```rust
async fn terminate_inner(handle: JoinHandle<()>, timeout: Duration) -> bool {
    if tokio::time::timeout(timeout, handle).await.is_err() {
        tracing::error!("Failed to terminate the task");
        return false;
    }
    true
}
```
And use it like
```rust
fn drop(&mut self) {
    if let Some(handle) = self.handle.take() {
        ResolveFuture::new(terminate_inner(handle, std::time::Duration::from_secs(10))).res_sync();
    }
}

fn terminate(self, timeout: Duration) -> bool {
    ResolveFuture::new(terminate_inner(self.handle.take().unwrap(), timeout)).res_sync();
}

fn terminate_async(self, timeout: Duration) -> bool {
    terminate_inner(self.handle.take().unwrap(), timeout).await
}
```

YuanYuYuan:
If I'm not mistaken, to take out the value inside an `Option` requires `&mut self`. So we need to modify `terminate` and `terminate_async` in any way. Furthermore, we discussed with @Mallets today that moving out the ownership is not necessary as the borrow checker only lives in Rust, and mixing deconstruction and undeclaration could cause some conflicts. (You can check `Undeckarable` for more details.) We could follow the common pattern like https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html#method.close. Closing a socket or receiver simply takes `&mut self` and hence `Drop` can coexist. I will create another draft PR on this topic further. 

In conclusion, I suggest splitting the termination and the deconstruction of the task. @wyfo  what do you think?

wyfo:
> If I'm not mistaken, to take out the value inside an Option requires &mut self. So we need to modify terminate and terminate_async in any way. 

Just use `mut self` instead of `self`, and you can still use value semantic. 

> We could follow the common pattern like https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.Receiver.html#method.close.

This is not the same thing here, as `terminate` returns a boolean. Do you want to store this boolean to return it on following calls?

> Closing a socket or receiver simply takes &mut self and hence Drop can coexist.

There is no issue to coexist with drop and having `terminate` with `self` receiver.

YuanYuYuan:
Hi @wyfo , thanks for you feedback.

Yes. That's what I mean we need to make some change in any way. Since I saw you wrote this in your first version, I thought you want to avoid any modification on `self`.
```rust
fn terminate(self, timeout: Duration) -> bool {
    ResolveFuture::new(terminate_inner(self.handle.take().unwrap(), timeout)).res_sync();
}
```

And I'm fine with the change of `fn terminate(mut self)`.


The remaining issue is do we want to enforce a descrtuction here? This is related to the potential conflict with `Drop`. Let me explain it.

Originally, we define our `TerminatableTask` without `Option`

```rust
pub struct TerminatableTask {
    handle: JoinHandle<()>,
    token: CancellationToken,
}
```
This forbids us from implementing `Drop` for it. The conflict comes from

```rust
async fn terminate_inner(handle: JoinHandle<()>, timeout: Duration) -> bool {
    if tokio::time::timeout(timeout, handle).await.is_err() {
        tracing::error!("Failed to terminate the task");
        return false;
    }
    true
}
impl TerminatableTask {
    pub async fn terminate_async(self, timeout: Duration) -> bool {
        terminate_inner(self.handle, timeout).await  // ERROR: cannot move out of type `TerminatableTask`, which implements the `Drop` trait
    }
}
```

Passing `self` into the function potentially trigger a deconstruction or move out itself. There are same issues in the `Undeclarable` trait.
For instance, [we implement `Undeclarable` trait for the `Subscriber`](https://github.com/eclipse-zenoh/zenoh/blob/main/zenoh/src/subscriber.rs#L803-L807), then we can't implement `Drop` for it anymore.

```rust
impl<R> Drop for Subscriber<'_, R> {
    fn drop(&mut self) {
        todo!()
    }
}
```

(You can check this does cause lots of "cannot move out ..." errors)

So that's why I need to wrap the `JoinHandle` with a `Option` to not take out the handle but the inner value.

Back to this PR, either using `&mut self` or `mut self` can compile. But we still need to think twice: do we need to embed a deconstruction in `terminate`? As the example I pasted, closing a socket multiple times is acceptable in real use cases. In this PR, a 2nd termination on a closed task is simply doing nothing. This is exactly the same with the previous behavior in zenoh library (before the tokio porting).

Moreover, it can avoid any conflict with `Drop` if we implment/modify the `TerminatableTask` as below one day (which unfortunately has already existed in the case of `Undeclarable`).

```rust
impl TerminatableTask {
    pub async fn terminate_async(mut self, timeout: Duration) -> bool {
        let Self { token, mut handle } = self;  // ERROR: cannot move out of type `TerminatableTask`, which implements the `Drop` trait
        token.cancel();
        if let Some(handle) = handle.take() {
            if tokio::time::timeout(timeout, handle).await.is_err() {
                tracing::error!("Failed to terminate the task");
                return false;
            };
        }
        true
    }
}
```

BTW, I'm little bit confused with your point "should not try to reuse terminate in drop". TBH, I don't care about the return boolean especially it would report an error once it failed. And it seems that your example don't process the boolean result in `Drop` as well. Am I missing something?

wyfo:
I give way to your arguments, it's good to me.

Mallets (approved):

PR #1160, opened 2024-06-18
Remove some duplicate imports

evshary:
Now there are some duplicate imports in `lib.rs`

```rust
use zenoh::{Error, Result};
use zenoh::core::{Error, Result};

use zenoh::{Session, open};
use zenoh::session::{Session, open};

use zenoh::scout;
use zenoh::scouting::scout;
```

For session one, it's used commonly and perhaps it makes sense to have a shorter path. However, I would suggest not to have duplicate paths for others.

### DIFF
--- a/examples/examples/z_scout.rs
+++ b/examples/examples/z_scout.rs
@@ -11,7 +11,7 @@
 // Contributors:
 //   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
 //
-use zenoh::{config::WhatAmI, scout, Config};
+use zenoh::{config::WhatAmI, scouting::scout, Config};
 
 #[tokio::main]
 async fn main() {
--- a/zenoh/src/api/scouting.rs
+++ b/zenoh/src/api/scouting.rs
@@ -39,7 +39,7 @@ use crate::{
 /// # async fn main() {
 /// use zenoh::{config::WhatAmI, prelude::*};
 ///
-/// let receiver = zenoh::scout(WhatAmI::Peer | WhatAmI::Router, zenoh::config::default())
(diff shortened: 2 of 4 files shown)


Mallets:
I'm not sure about this change... @milyin ?

wyfo:
@Mallets you should have tagged me also 🙂 

@evshary Reexporting in root is a common practice across Rust ecosystem; you never use `serde::ser::Serialize` but `serde::Serialize`. Having root reexport of `Result` overload (and custom `Error`) is also quite a convention, see `serde_json::Result`/`serde_json::Error`.

You may be interested in #1007 where the discussion about root reexporting took place.

evshary:
@wyfo You convince me now why putting these under the root.
Then, the question might be why we choose these to re-export. Is there any rule for this?

wyfo:
There is no "rule" as far as I know. Some crates like [`rand`](https://docs.rs/rand/0.8.5/rand/#traits) choose to not reexport, others like [`serde_json`](https://docs.rs/serde_json/latest/serde_json/) do it. So you can find both in fundamental crates of the ecosystem.

IMO, it is a matter of taste. When I wrote the reexports, I had `serde_json` way of doing in mind. This is obviously a debatable decision, and we solve and close that debate quickly before 1.0. 

That's also something we could ask on Reddit like  folks of `ratatui` [did](https://www.reddit.com/r/rust/comments/1cle18j/to_prelude_or_to_not_to_prelude/). Thanks to those Reddit/Rust forum discussion, we already closed the prelude discussion by keeping only traits in it. 
I think I will open a Reddit discussion on Monday.

Mallets:
Superseded by https://github.com/eclipse-zenoh/zenoh/pull/1193

PR #1174, opened 2024-06-19
Improve bytes

Mallets:
- Refine trait bounds
- Implement support for `HashMap`
- Implement support for `char`
- Define explicit `Encoding` for `u8`, `u16`, `u32`, `u64`, `u128`, `i8`, `i16`, `i32`, `i64`, `i128`, `f32`, `f64`
- Improve docs

### DIFF
--- a/zenoh/src/api/bytes.rs
+++ b/zenoh/src/api/bytes.rs
@@ -2085,10 +2085,10 @@ where
 
 impl<A, B> Deserialize<(A, B)> for ZSerde
 where
-    A: TryFrom<ZBytes> + 'static,
-    <A as TryFrom<ZBytes>>::Error: Debug + 'static,
-    B: TryFrom<ZBytes> + 'static,
-    <B as TryFrom<ZBytes>>::Error: Debug + 'static,
+    A: TryFrom<ZBytes>,
+    <A as TryFrom<ZBytes>>::Error: Debug,
+    B: TryFrom<ZBytes>,
+    <B as TryFrom<ZBytes>>::Error: Debug,
 {
     type Input<'a> = &'a ZBytes;
     type Error = ZBytesTupleReadError<A, B>;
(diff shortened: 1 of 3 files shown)


zenoh/src/api/bytes.rs
```diff
@@ -2345,75 +2925,54 @@ mod tests {
         hm.insert(0, 0);
         hm.insert(1, 1);
         println!("Serialize:\t{:?}", hm);
-        let p = ZBytes::from_iter(hm.clone().drain());
+        let p = ZBytes::from(hm.clone());
```
evshary, line 2928: I love it. It's easier to use and understandable now.

zenoh/src/api/bytes.rs
```diff
@@ -2094,12 +2222,464 @@ where
...
-    type Error = ZError;
+    type Error = ZBytesReadErrorTuple2<A, B>;
+
+    fn try_from(value: &mut ZBytes) -> Result<Self, Self::Error> {
+        ZSerde.deserialize(&*value)
+    }
+}
+
+// Tuple (a, b, c)
+macro_rules! impl_tuple3 {
```
evshary, line 2238: I guess the reason we use macro instead of inline here is type issue, right?

zenoh/src/api/bytes.rs
```diff
@@ -2094,12 +2222,464 @@ where
...
+    <D as TryFrom<ZBytes>>::Error: Debug,
+{
+    type Error = ZBytesReadErrorTuple4<A, B, C, D>;
+
+    fn try_from(value: &mut ZBytes) -> Result<Self, Self::Error> {
+        ZSerde.deserialize(&*value)
+    }
+}
+
+// HashMap
```
evshary, line 2592: Do you think we can provide something similar for Vec? Now we only support `Vec<u8>`

evshary (approved):
I really love the changes :+1:

zenoh/src/api/bytes.rs
Mallets: Correct and trait recursiveness...

zenoh/src/api/bytes.rs
Mallets: Unfortunately not. Rust does not support serialization and `Vec<u8>` should have a special treatment compared to `Vec<T> where ZSerde: Serialize<T>`.
An iterator of `T` is serialized in this way:
- Serialize `T` as standalone
- Write length of serialized `T` into final buffer
- Append serialized `T`

This allows to iterate over the elements `T` because we know where the serialized `T` starts and end.
`u8` is a special case of `T` where we do not want to prepend the serialized length but just copy everything.
But in order to treat it differently, we need specialization in Rust... which is not supported :(

Mallets:
@evshary could you please review this PR?

evshary (approved):
LGTM for adding more specific int/uint type in encoding

PR #1175, opened 2024-06-19
Draft: Add `new_timestamp` to Session

Charles-Schleich:


zenoh/src/api/session.rs
```diff
@@ -657,6 +657,28 @@ impl Session {
...
+    /// # async fn main() {
+    /// use zenoh::prelude::*;
+    ///
+    /// let session = zenoh::open(zenoh::config::peer()).await.unwrap();
+    /// let timestamp = session.new_timestamp();
+    /// # }
+    /// ```
+    pub fn new_timestamp(&self) -> Timestamp {
+        let id = self.runtime.zid();
+        // TODO: Should we make this an Result return type ?
```
Mallets, line 678: `duration_since(UNIX_EPOCH)` should never fail since `SystemTime` is not allowed to be earlier than `UNIX_EPOCH`.

zenoh/src/api/time.rs
```diff
@@ -15,6 +15,7 @@ use std::time::{SystemTime, UNIX_EPOCH};
 
 use zenoh_protocol::core::{Timestamp, TimestampId};
 
+// TODO: Shall we remove this new_timestamp in favoir of the src/api/session::Session::new_timestamp();
```
milyin, line 18: Yes, there was discussion about it, created issue https://github.com/eclipse-zenoh/zenoh/issues/1179

milyin (approved):

Issue #1125, opened 2024-06-12
Suggestion: `internal::Value` destructor into components. 


Charles-Schleich:
### Describe the feature

Once https://github.com/eclipse-zenoh/zenoh/pull/1119 lands, It May be useful to be able to destruct a `Value` into its `payload` and `encoding`, taking ownership of both upon destruction.
Currently the only option is to take a reference to the `payload` and `encoding`, forcing a `.clone()` in circumstances where destructing would be preferred.

wyfo:
Because `Value` becomes *internal*, it could also make sense to make its fields public.

Mallets:
@Charles-Schleich can you submit a PR?

Charles-Schleich:
Resolved by [1178](https://github.com/eclipse-zenoh/zenoh/pull/1178)

PR #1178, opened 2024-06-20
Value fields public

Charles-Schleich:
Make fields of `Value` public now that now that value is behind feature flag `internal`
https://github.com/eclipse-zenoh/zenoh/issues/1125

This would remove need to clone fields when  you want to separate out `payload` and `encoding`


milyin (approved):

milyin:
Approved, but it's better to finally get rid of value completely

skreborn:
> Approved, but it's better to finally get rid of value completely

Apologies if this is not the place to enquire, but is that planned for the 1.0.0 release? I'm using `Value` in various places in our codebase and would love to know what's coming so I can better prepare. Happy to move this to a separate discussion if that's preferred.

PR #1180, opened 2024-06-24
Add z_bytes in the examples.

evshary:
Now we use ZBytes to replace Value, and we should provide some examples for users to know how to do the transformation.
Add z_bytes.rs in the example to help users know how it works.
Waiting for https://github.com/eclipse-zenoh/zenoh/pull/1174 to be merged and need some update.

### DIFF
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -43,6 +43,10 @@ zenoh-collections = { workspace = true }
 tracing = { workspace = true }
 zenoh = { workspace = true, default-features = true }
 zenoh-ext = { workspace = true }
+serde_json = { workspace = true }
+serde_yaml = { workspace = true }
+prost = "0.12.6"
+prost-types = "0.12.6"
 
 [dev-dependencies]
 rand = { workspace = true, features = ["default"] }
--- a/examples/examples/z_bytes.rs
+++ b/examples/examples/z_bytes.rs
@@ -12,8 +12,7 @@
 //   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
 //
 
-use std::{borrow::Cow, collections::HashMap};
-
+use std::{borrow::Cow, collections::HashMap, io::Cursor};
 use zenoh::bytes::ZBytes;
 
 fn main() {
@@ -88,4 +87,56 @@ fn main() {
     let payload = ZBytes::from(input.clone());
     let output = payload.deserialize::<HashMap<usize, String>>().unwrap();
     assert_eq!(input, output);
+
+    // JSON
+    let data = r#"
+    {
+        "name": "John Doe",
+        "age": 43,
+        "phones": [
+            "+44 1234567",
+            "+44 2345678"
+        ]
+    }"#;
+    let input: serde_json::Value = serde_json::from_str(data).unwrap();
+    let payload = ZBytes::try_serialize(input.clone()).unwrap();
+    let output: serde_json::Value = payload.deserialize().unwrap();
+    assert_eq!(input, output);
+    // Corresponding encoding to be used in operations like `.put()`, `.reply()`, etc.
+    // let encoding = Encoding::APPLICATION_JSON;
+
+    // YAML
+    let data = r#"
+        name: "John Doe"
+        age: 43
+        phones:
+          - "+44 1234567"
+          - "+44 2345678"
+    "#;
+    let input: serde_yaml::Value = serde_yaml::from_str(data).unwrap();
+    let payload = ZBytes::try_serialize(input.clone()).unwrap();
+    let output: serde_yaml::Value = payload.deserialize().unwrap();
+    assert_eq!(input, output);
+    // Corresponding encoding to be used in operations like `.put()`, `.reply()`, etc.
+    // let encoding = Encoding::APPLICATION_YAML;
+
+    // Protobuf
+    use prost::Message;
+    #[derive(Message, Eq, PartialEq)]
+    struct EntityInfo {
+        #[prost(uint32)]
+        id: u32,
+        #[prost(string)]
+        name: String,
+    }
+    let input = EntityInfo {
+        id: 1234,
+        name: String::from("John Doe"),
+    };
+    let payload = ZBytes::from(input.encode_to_vec());
+    let output =
+        EntityInfo::decode(Cursor::new(payload.deserialize::<Vec<u8>>().unwrap())).unwrap();
+    assert_eq!(input, output);
+    // Corresponding encoding to be used in operations like `.put()`, `.reply()`, etc.
+    // let encoding = Encoding::APPLICATION_PROTOBUF;
 }
--- a/examples/examples/z_bytes.rs
+++ b/examples/examples/z_bytes.rs
@@ -1,0 +1,83 @@
+//
+// Copyright (c) 2024 ZettaScale Technology
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+// which is available at https://www.apache.org/licenses/LICENSE-2.0.
+//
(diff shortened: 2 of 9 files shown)


examples/examples/z_put.rs
```diff
@@ -26,6 +26,7 @@ async fn main() {
     let session = zenoh::open(config).await.unwrap();
 
     println!("Putting Data ('{key_expr}': '{payload}')...");
+    // Refer to z_bytes.rs to see how to put different type of messages
```
Mallets, line 29:
```suggestion
    // Refer to z_bytes.rs to see how to put different types of message
```

examples/examples/z_bytes.rs
```diff
@@ -0,0 +1,86 @@
...
+// Contributors:
+//   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
+//
+
+use std::{borrow::Cow, collections::HashMap};
+
+use zenoh::bytes::ZBytes;
+
+fn main() {
+    // Numeric: u8, u16, u32, u128, usize, i8, i16, i32, i128, isize, f32, f64
```
Mallets, line 20: The corresponding default encoding should be added in the example.

evshary: Do you mean we should have the actual put and sub here? ZBytes doesn't include the encoding information unless we use Sample.

Mallets: I mean something like:

```rust
let input = 1234_u32;
let payload = ZBytes::from(input);
let output: u32 = payload.deserialize().unwrap();
assert_eq!(input, output); 
// Corresponding encoding
// let encoding = Encoding::ZENOH_INT32;
```

evshary: Done

JEnoch:
[examples/README.md](https://github.com/eclipse-zenoh/zenoh/blob/main/examples/README.md) is describing each example. I think `z_bytes.rs` must be added here.

JEnoch:
In the past we were often ask how to publish/subscribe with Protobuf encoding.
Could be worth to add an example here, also to show any encoding is supported.
@Mallets, your opinion ?

wyfo:
I would like to add a protobuf encoding directly to zenoh, because I imagine that's a fairly common used format, but there is no normalized content-type for it. See https://stackoverflow.com/questions/30505408/what-is-the-correct-protobuf-content-type.
We can still use `application/x-protobuf` like Cloudflare does, as well as `application/x-messagepack` for example.
The issue with encodings like protobuf or flatbuffer is that you need the associated schema, so users will have to add schema to their `Encoding`.

People also suggest to simply use `application/octet-stream` with the schema, like in https://stackoverflow.com/questions/55383771/mime-type-for-flatbuffers.

JEnoch:
Sure, any encoding that is not self-descriptive requires the user to take care of the associates schema. Either making assumptions (my key expr `a/b/c` will always have `MyType` values), either passing the type name or full description as encoding suffix: `application/x-protobuf;schema=MyType` (but this increases metadata size).

But aside encoding name, my point was more on providing code examples on how to use for instance [`prost`](https://crates.io/crates/prost) or `serde-json` with `ZBytes`.

Mallets:
> I would like to add a protobuf encoding directly to zenoh, because I imagine that's a fairly common used format, but there is no normalized content-type for it. See https://stackoverflow.com/questions/30505408/what-is-the-correct-protobuf-content-type. We can still use `application/x-protobuf` like Cloudflare does, as well as `application/x-messagepack` for example. The issue with encodings like protobuf or flatbuffer is that you need the associated schema, so users will have to add schema to their `Encoding`.
> 
> People also suggest to simply use `application/octet-stream` with the schema, like in https://stackoverflow.com/questions/55383771/mime-type-for-flatbuffers.

Protobuf is already present as part of the automatically optimized encodings: https://github.com/eclipse-zenoh/zenoh/blob/fc18f90310cd451a2f3d6e579dd9cfd9dba4bda6/zenoh/src/api/encoding.rs#L305

wyfo:
How did I miss it... 🤦‍♂️

wyfo:
`application/protobuf` doesn't exist in https://www.iana.org/assignments/media-types/media-types.xhtml.
(but `application/vnd.msgpack` was [added](https://github.com/msgpack/msgpack/issues/194#issuecomment-2111445791) a few weeks ago)
I'm not sure we should use unregistered media type...

Mallets:
We can't support all types nor be limited to only registered types. Moreover, *vendor* media types are kind out-of-scope in the sense they could be whatever... `application/protobuf` was the proposal from IETF and I'd rather stick to that. In my opinion no changes are needed on that regards.

evshary:
> @evshary I agree we should add in the example how to serialize/deserialize complex formats like JSON, YAML and PROTOBUF.

Sure. Working on it now.

evshary:
@JEnoch @Mallets I think it's ready for another review now.

examples/examples/z_bytes.rs
```diff
@@ -0,0 +1,143 @@
...
+        #[prost(string)]
+        name: String,
+    }
+    let input = EntityInfo {
+        id: 1234,
+        name: String::from("John Doe"),
+    };
+    let payload = ZBytes::from(input.encode_to_vec());
+    let output =
+        EntityInfo::decode(Cursor::new(payload.deserialize::<Vec<u8>>().unwrap())).unwrap();
```
Mallets, line 139: Could we use a `payload.reader()`?

Mallets: In addition to that it would make sense to show in the example how to use a `ZBytesReader` and `ZBytesWriter`.

evshary: I'm not sure whether it's better or not.

```rust
// serialize
let mut bytes = ZBytes::empty();
let mut writer = bytes.writer();
writer.serialize(input.encode_to_vec());
// deserialize
let mut reader = bytes.reader();
let output: EntityInfo =
    EntityInfo::decode(Cursor::new(reader.deserialize::<Vec<u8>>().unwrap())).unwrap();
```

BTW, we've already had `ZBytesReader` and `ZBytesWriter` here.
https://github.com/eclipse-zenoh/zenoh/pull/1180/files#diff-3d491564591b569630ebbbe501a1d6331086fc9b3c17399285f3e97cbb3f4a5dR52

Mallets: I was referring to the fact of using something like 

```rust
let output: EntityInfo =
    EntityInfo::decode(payload.reader()).unwrap();
```

To avoid potential memory allocation.

evshary: I think we need to have `impl TryFrom<ZBytes> for bytes::Buf` to support this. I can give it a try.

evshary: Hmmm, after having a further look, `bytes:Buf` is the trait that needs to be implemented in `ZBytes`. We can rely on `ZBufReader` to implement it, but it seems a little difficult to implement `chunk()` function since it needs to return the all remaining data. Our data might be stored in several ZSlices, so I'm not sure whether there is a way to do so.
https://docs.rs/bytes/latest/bytes/buf/trait.Buf.html#tymethod.chunk

Mallets: Ok, then let's just deserialize it to 

```rust
let output: EntityInfo = EntityInfo::decode(Cursor::new(reader.deserialize::<Cow<[u8]>>().unwrap())).unwrap();
```

Mallets (approved):

examples/Cargo.toml
```diff
@@ -43,6 +43,10 @@ zenoh-collections = { workspace = true }
 tracing = { workspace = true }
 zenoh = { workspace = true, default-features = true }
 zenoh-ext = { workspace = true }
+serde_json = { workspace = true }
+serde_yaml = { workspace = true }
+prost = "0.12.6"
```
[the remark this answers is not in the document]
evshary: I'm not sure. Since these crates are only used by examples now, do you think we should move it outside and use `{ workspace = true }` here?

PR #1182, opened 2024-06-24
fix: adminspace encodings

gabrik:
After the removal of `Value` the encoding of the Admin Space have been lost.

So for instance this:

```
{
        "key": "@/router/8f2aef5a89c14100b70184094f5bdea9/status/plugins/storage_manager/volumes/file-system-volume",
        "value": {"root":"/home/vscode/.zenoh/zenoh_backend_fs","version":"0.11.0-dev"},
        "time": null
    }
```

becomes this:

```
{
        "key": "@/router/8f2aef5a89c14100b70184094f5bdea9/status/plugins/storage_manager/volumes/file-system-volume",
        "value": "eyJyb290IjoiL2hvbWUvdnNjb2RlLy56ZW5vaC96ZW5vaF9iYWNrZW5kX2ZzIiwidmVyc2lvbiI6IjAuMTEuMC1kZXYifQ==",
        "encoding": "zenoh/bytes",
        "time": null
    }
```

This PR aims at solving this bug.

### DIFF
--- a/plugins/zenoh-plugin-rest/src/lib.rs
+++ b/plugins/zenoh-plugin-rest/src/lib.rs
@@ -67,6 +67,7 @@ pub fn base64_encode(data: &[u8]) -> String {
 }
 
 fn payload_to_json(payload: &ZBytes, encoding: &Encoding) -> serde_json::Value {
+    println!("Encoding is: {encoding:?}");
     match payload.is_empty() {
         // If the value is empty return a JSON null
         true => serde_json::Value::Null,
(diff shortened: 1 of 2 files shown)


gabrik:
There are still some Key that provide a base64 response:

- [ ] @/router/<id>/status/<plugin>/__path__
- [ ] @/router/<id>/linkstate/routers
- [ ] @/router/<id>/metrics 

still investigating why

gabrik:
So a `z_get` gets them correct, the REST API does not...

gabrik:
Updated the `Encoding` matching in the REST plugin.

plugins/zenoh-plugin-rest/src/lib.rs
```diff
@@ -67,6 +67,7 @@ pub fn base64_encode(data: &[u8]) -> String {
 }
 
 fn payload_to_json(payload: &ZBytes, encoding: &Encoding) -> serde_json::Value {
+    println!("Encoding is: {encoding:?}");
```
Mallets, line 70:
```suggestion
```

zenoh/src/net/runtime/adminspace.rs
```diff
@@ -622,15 +622,8 @@ fn local_data(context: &AdminContext, query: Query) {
...
-    let payload = match ZBytes::try_from(json) {
-        Ok(p) => p,
-        Err(e) => {
-            tracing::error!("Error serializing AdminSpace reply: {:?}", e);
-            return;
-        }
-    };
     if let Err(e) = query
-        .reply(reply_key, payload)
+        .reply(reply_key, json.to_string())
```
Mallets, line 626: `to_string()` should not be used for encoding a json. The previous code performs the correct `try` for serialization.

Mallets (approved):

PR #1184, opened 2024-06-25
fix: replace some unsafe code with safe version

wyfo:
Compiler is able to optimize bound checks based on previous checks See https://godbolt.org/z/oGesnb6a4 or https://godbolt.org/z/c6c41bvE5

`Writer::with_slot` has been made unsafe, because its implementations rely on a precondition on the write callback

### DIFF
--- a/commons/zenoh-buffers/src/bbuf.rs
+++ b/commons/zenoh-buffers/src/bbuf.rs
@@ -127,7 +127,7 @@ impl Writer for &mut BBuf {
         self.capacity() - self.len()
     }
 
-    fn with_slot<F>(&mut self, len: usize, f: F) -> Result<NonZeroUsize, DidntWrite>
+    unsafe fn with_slot<F>(&mut self, len: usize, f: F) -> Result<NonZeroUsize, DidntWrite>
     where
         F: FnOnce(&mut [u8]) -> usize,
     {
--- a/commons/zenoh-buffers/src/bbuf.rs
+++ b/commons/zenoh-buffers/src/bbuf.rs
@@ -127,7 +127,7 @@ impl Writer for &mut BBuf {
         self.capacity() - self.len()
     }
 
-    unsafe fn with_slot<F>(&mut self, len: usize, f: F) -> Result<NonZeroUsize, DidntWrite>
+    unsafe fn with_slot<F>(&mut self, len: usize, write: F) -> Result<NonZeroUsize, DidntWrite>
     where
         F: FnOnce(&mut [u8]) -> usize,
     {
@@ -135,7 +135,7 @@ impl Writer for &mut BBuf {
             return Err(DidntWrite);
         }
 
-        let written = f(self.as_writable_slice());
+        let written = write(self.as_writable_slice());
         self.len += written;
 
         NonZeroUsize::new(written).ok_or(DidntWrite)
--- a/commons/zenoh-buffers/src/bbuf.rs
+++ b/commons/zenoh-buffers/src/bbuf.rs
@@ -135,7 +135,8 @@ impl Writer for &mut BBuf {
             return Err(DidntWrite);
         }
 
-        let written = write(self.as_writable_slice());
+        // SAFETY: self.remaining() >= len
+        let written = write(unsafe { self.as_writable_slice().get_unchecked_mut(..len) });
         self.len += written;
 
         NonZeroUsize::new(written).ok_or(DidntWrite)
--- a/commons/zenoh-buffers/src/lib.rs
+++ b/commons/zenoh-buffers/src/lib.rs
@@ -139,7 +139,7 @@ pub mod writer {
         }
         /// Provides a buffer of exactly `len` uninitialized bytes to `f` to allow in-place writing.
         /// `f` must return the number of bytes it actually wrote.
-        fn with_slot<F>(&mut self, len: usize, f: F) -> Result<NonZeroUsize, DidntWrite>
+        unsafe fn with_slot<F>(&mut self, len: usize, f: F) -> Result<NonZeroUsize, DidntWrite>
         where
             F: FnOnce(&mut [u8]) -> usize;
     }
--- a/commons/zenoh-buffers/src/lib.rs
+++ b/commons/zenoh-buffers/src/lib.rs
@@ -137,9 +137,14 @@ pub mod writer {
         fn can_write(&self) -> bool {
             self.remaining() != 0
         }
-        /// Provides a buffer of exactly `len` uninitialized bytes to `f` to allow in-place writing.
-        /// `f` must return the number of bytes it actually wrote.
-        unsafe fn with_slot<F>(&mut self, len: usize, f: F) -> Result<NonZeroUsize, DidntWrite>
+        /// Provides a buffer of exactly `len` uninitialized bytes to `write` to allow in-place writing.
+        /// `write` must return the number of bytes it actually wrote.
+        ///
+        /// # Safety
+        ///
+        /// Caller must ensure that `write` return an integer lesser than or equal to the length of
+        /// the slice passed in argument
+        unsafe fn with_slot<F>(&mut self, len: usize, write: F) -> Result<NonZeroUsize, DidntWrite>
         where
             F: FnOnce(&mut [u8]) -> usize;
     }
--- a/commons/zenoh-buffers/src/slice.rs
+++ b/commons/zenoh-buffers/src/slice.rs
@@ -61,26 +61,13 @@ impl HasWriter for &mut [u8] {
 
 impl Writer for &mut [u8] {
     fn write(&mut self, bytes: &[u8]) -> Result<NonZeroUsize, DidntWrite> {
-        let len = bytes.len().min(self.len());
-        if len == 0 {
+        let Some(len) = NonZeroUsize::new(bytes.len().min(self.len())) else {
             return Err(DidntWrite);
-        }
-
-        // SAFETY: len is guaranteed to be the minimum between lhs and rhs length.
-        //         We early return if length is 0.
-        let lhs = crate::unsafe_slice_mut!(self, ..len);
-        let rhs = crate::unsafe_slice!(bytes, ..len);
-        lhs.copy_from_slice(rhs);
-
-        // SAFETY: len is guaranteed to be the minimum between lhs and rhs length.
-        let lhs = crate::unsafe_slice_mut!(self, len..);
-        // SAFETY: this doesn't compile with simple assignment because the compiler
-        // doesn't believe that the subslice has the same lifetime as the original slice,
-        // so we transmute to assure it that it does.
-        *self = unsafe { mem::transmute::<&mut [u8], &mut [u8]>(lhs) };
-
-        // SAFETY: this operation is safe since we check if len is non-zero.
-        Ok(unsafe { NonZeroUsize::new_unchecked(len) })
+        };
+        let (to_write, remain) = mem::take(self).split_at_mut(len.get());
+        to_write.copy_from_slice(&bytes[..len.get()]);
+        *self = remain;
+        Ok(len)
     }
 
     fn write_exact(&mut self, bytes: &[u8]) -> Result<(), DidntWrite> {
@@ -88,19 +75,7 @@ impl Writer for &mut [u8] {
         if self.len() < len {
             return Err(DidntWrite);
         }
-
-        // SAFETY: len is guaranteed to be the smaller than lhs length.
-        let lhs = crate::unsafe_slice_mut!(self, ..len);
-        let rhs = crate::unsafe_slice!(bytes, ..len);
-        lhs.copy_from_slice(rhs);
-
-        // SAFETY: len is guaranteed to be the minimum between lhs and rhs length.
-        let lhs = crate::unsafe_slice_mut!(self, len..);
-        // SAFETY: this doesn't compile with simple assignment because the compiler
-        // doesn't believe that the subslice has the same lifetime as the original slice,
-        // so we transmute to assure it that it does.
-        *self = unsafe { mem::transmute::<&mut [u8], &mut [u8]>(lhs) };
-
+        let _ = self.write(bytes);
         Ok(())
     }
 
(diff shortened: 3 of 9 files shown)


wyfo:
@Mallets @kydos

wyfo:
@fuzzypixelz

commons/zenoh-buffers/src/lib.rs
```diff
@@ -137,9 +137,14 @@ pub mod writer {
...
-        /// `f` must return the number of bytes it actually wrote.
-        fn with_slot<F>(&mut self, len: usize, f: F) -> Result<NonZeroUsize, DidntWrite>
+        /// Provides a buffer of exactly `len` uninitialized bytes to `write` to allow in-place writing.
+        /// `write` must return the number of bytes it actually wrote.
+        ///
+        /// # Safety
+        ///
+        /// Caller must ensure that `write` return an integer lesser than or equal to the length of
+        /// the slice passed in argument
+        unsafe fn with_slot<F>(&mut self, len: usize, write: F) -> Result<NonZeroUsize, DidntWrite>
```
Mallets, line 147: I don't think `with_slot` should be marked as `unsafe`. If `with_slot()` is `unsafe` then also `len()` should be marked as `unsafe`?

commons/zenoh-buffers/src/slice.rs
```diff
@@ -165,39 +133,30 @@ impl<'a> HasReader for &'a [u8] {
...
             return Err(DidntRead);
         }
-        // SAFETY: len is guaranteed to be the smaller than lhs length.
-        let lhs = crate::unsafe_slice_mut!(into, ..len);
-        let rhs = crate::unsafe_slice!(self, ..len);
-        lhs.copy_from_slice(rhs);
-        // SAFETY: len is guaranteed to be smaller than slice.len().
-        *self = crate::unsafe_slice!(self, len..);
+        let (to_write, remain) = self.split_at(len);
+        into[..len].copy_from_slice(to_write);
```
Mallets, line 151: In general you are re-introducing a lot of *slice boundaries* checks that have an impact at runtime while guarantees are held by the contract and covered by tests. Since these operations are very hot in the code, impact is not negligible and assessed in the past on few `100k msg/s`.

commons/zenoh-buffers/src/zslice.rs
```diff
@@ -122,7 +122,7 @@ impl ZSlice {
     }
 
     pub fn empty() -> Self {
-        unsafe { ZSlice::new_unchecked(Arc::new([]), 0, 0) }
+        Self::new(Arc::new([]), 0, 0).unwrap()
```
Mallets, line 125: `unwrap()` add runtime code to handle panics. Here it seems unnecessary to me.

commons/zenoh-buffers/src/lib.rs
wyfo: Which `len()` are you talking about?

commons/zenoh-buffers/src/slice.rs
wyfo: I've checked before on Godbolt than bound checks are elided. See the links in the PR description. If you're not satisfied, I can also run `cargo asm` to ensure that bound checks are removed.

commons/zenoh-buffers/src/zslice.rs
wyfo: I adds it only if the check is not elided, see https://godbolt.org/z/5G1795dn1

commons/zenoh-buffers/src/slice.rs
wyfo: Btw, we are already using elided checks, because `copy_from_slice` have an assertion about the slices' length being equal. So to be consistent with this full unsafe approach, we should directly use `ptr::copy_nonoverlapping` instead of `copy_from_slice`.

wyfo:
By the way, since 1.78, [Rust has enabled debug of the standard library](https://blog.rust-lang.org/2024/05/02/Rust-1.78.0.html#asserting-unsafe-preconditions), so `<[T]>::get_unchecked` now trigger an error in tests if they are not built with `--release`.

That's why we may want to remove `zenoh_buffers::unsafe_slice` and other macros of the same kind, as they are no more "necessary". There are two caveats:
- as written in the 1.78 release, these debug assertion are not stable (but I think we can still rely on `<[T]>::get_unchecked` one)
- they are removed from `--release` build, contrary to the `"test"` feature

Do we care having the check on release build for tests?

wyfo:
Also, I've one (big) issue with `unsafe_slice` macro, which is that it doesn't require unsafe block. It make it harder to track unsafe code. I know that it is more convenient to write, but I don't think it's a good argument when it comes to unsafe code, which requires more cautious care than convenience.

So, if we keep `unsafe_slice`, I would like to remove the unsafe block from it to for us writing `unsafe { crate::unsafe_slice!(s, ..) }`.

commons/zenoh-buffers/src/lib.rs
Mallets: I could implement 
```rust
fn len(&self) -> usize {
    usize::MAX
}
```
and may buffer implementation is broken in the same way. It's a logical bug, not really an `unsafe` implementation.

Mallets:
> By the way, since 1.78, [Rust has enabled debug of the standard library](https://blog.rust-lang.org/2024/05/02/Rust-1.78.0.html#asserting-unsafe-preconditions), so `<[T]>::get_unchecked` now trigger an error in tests if they are not built with `--release`.
> 
> That's why we may want to remove `zenoh_buffers::unsafe_slice` and other macros of the same kind, as they are no more "necessary". There are two caveats:
> 
> * as written in the 1.78 release, these debug assertion are not stable (but I think we can still rely on `<[T]>::get_unchecked` one)
> * they are removed from `--release` build, contrary to the `"test"` feature
> 
> Do we care having the check on release build for tests? If we do, I will put back `unsafe_slice!` slice in the places where I used `get_unchecked` instead.

At the moment Zenoh uses Rust `1.75`. I agree with your comment once we bump Rust to `1.78`.

commons/zenoh-buffers/src/lib.rs
wyfo: Marking a function `unsafe` or not obey to a simple rule: does calling this function with dumb parameters may corrupt the program? if yes, it must be marked as unsafe.
Current `with_slot` implementation must be marked unsafe.

For `Buffer::len`, we are not in this case, but we may fall into the case of unsafe traits.
For unsafe trait, the rule is: is it possible to implement this trait in a way that may corrupt the program? if yes, the trait must be marked unsafe.
I have not looked in detail to the `Buffer` trait yet, but maybe it should be marked as unsafe too.

Mallets: Calling `len()` on a slice is not `unsafe` but clearly a wrong implementation `len()` will lead to a buggy behaviour. In any case, the wrong behaviour may be a well-defined and not undefined behaviour. Calling `with_slot()` with wrong parameters will lead to a buggy behaviour but it will be deterministic behaviour, in the same way an `unwrap()` may crash your program. Therefore it doesn't need to be `unsafe` for me.

wyfo: It's not about undeterministic behavior, it's about memory corruption and undefined behavior; both don't necessarily make program crash. 

Here is the current implementation of `<&mut [u8]>::with_slot`:
```rust
unsafe fn with_slot<F>(&mut self, len: usize, write: F) -> Result<NonZeroUsize, DidntWrite>
    where
        F: FnOnce(&mut [u8]) -> usize,
    {
        if len > self.len() {
            return Err(DidntWrite);
        }
        let written = write(&mut self[..len]);
        // SAFETY: `written` < `len` is guaranteed by function contract
        *self = unsafe { mem::take(self).get_unchecked_mut(written..) };
        NonZeroUsize::new(written).ok_or(DidntWrite)
    }
 ```
 The reason why `with_slot` must be unsafe is straightforward: the result of `write` is used with `get_unchecked_mut`, and `get_unchecked_mut` has a precondition that must be respected. The only way to do it without changing the implementation is to add a precondition to `with_slot`.

Comment (wyfo) [reply commons/zenoh-buffers/src/lib.rs:147]: > Calling len() on a slice is not unsafe but clearly a wrong implementation len() will lead to a buggy behaviour.

You need to define "buggy". If buggy means undefined behavior or memory corruption because other parts of the code rely on `Buffer::len` result to do unsafe things, then the trait must be marked unsafe.

**Comment by fuzzypixelz** [comment]
@wyfo @Mallets I would like to review this before it gets merged :)

File: commons/zenoh-buffers/src/lib.rs (hunk 1 above)
Comment (wyfo) [reply commons/zenoh-buffers/src/lib.rs:147]: About unsafe trait, you may look at `ExactSizeIterator` [documentation](https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html):
> Note that this trait is a safe trait and as such does not and cannot guarantee that the returned length is correct. This means that unsafe code must not rely on the correctness of [Iterator::size_hint](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint). The unstable and unsafe [TrustedLen](https://doc.rust-lang.org/std/iter/trait.TrustedLen.html) trait gives this additional guarantee.

So, if you use `Buffer::len` in any unsafe precondition, you should mark `Buffer` unsafe, especially it is public.

Comment (wyfo) [reply commons/zenoh-buffers/src/lib.rs:147]: A proposal to make `with_slot` safe would be the following:
```rust
   fn with_slot<F>(&mut self, len: usize, write: F) -> Result<NonZeroUsize, DidntWrite>
    where
        F: FnOnce(&mut [u8]) -> &[u8],
    {
        if len > self.len() {
            return Err(DidntWrite);
        }
        let written = write(&mut self[..len]);
        assert_eq!(self.as_ptr(), written.as_ptr());
        let written_length = written.len();
        // SAFETY: Assertion ensures that `self` and `written` are the same slice, 
        // so `written.len()` is a valid length for `self`
        *self = unsafe { mem::take(self).get_unchecked_mut(written_length..) };
        NonZeroUsize::new(written_length).ok_or(DidntWrite)
    }
```
Here, the safety is ensured by an assertion, there is no need of `unsafe` precondition.
But because `write` callback should be inlined, the assertion should be elided. I have to test it.
Now, the only way to trigger memory corruption by calling `with_slot` would be to use `unsafe` code to forge a returned slice with the same base pointer and an overflowing length. However, the fault would be on the user `unsafe` code, not on `with_slot`, so it doesn't question `with_slot` safety.

Mallets (approved):

commons/zenoh-buffers/src/bbuf.rs
```diff
@@ -127,15 +127,15 @@ impl Writer for &mut BBuf {
...
+    unsafe fn with_slot<F>(&mut self, len: usize, write: F) -> Result<NonZeroUsize, DidntWrite>
     where
         F: FnOnce(&mut [u8]) -> usize,
     {
         if self.remaining() < len {
             return Err(DidntWrite);
         }
 
-        let written = f(self.as_writable_slice());
+        let written = write(self.as_writable_slice());
```
fuzzypixelz, line 138: I think `as_writable_slice` should be unsafe as well. It claims that the `Writer` ensures that `self.len` is less than the capacity.

The invariant was proved in `with_slot` before, and now it's the responsibility of the caller of `with_slot`. `unsafe` should not be used unless the current context can prove there is no UB. `as_writable_slice` cannot be safe if its invariant can be broken w/ safe code.

fuzzypixelz: The same applies to `as_mut_slice`.

commons/zenoh-buffers/src/bbuf.rs
```diff
@@ -127,15 +127,15 @@ impl Writer for &mut BBuf {
         self.capacity() - self.len()
     }
 
-    fn with_slot<F>(&mut self, len: usize, f: F) -> Result<NonZeroUsize, DidntWrite>
+    unsafe fn with_slot<F>(&mut self, len: usize, write: F) -> Result<NonZeroUsize, DidntWrite>
```
fuzzypixelz, line 130: The behavior doesn't exactly match the trait documentation. Wwe provide a buffer of _at least_ and not _exactly_ `len` bytes because `as_writable_slice` returns the entire available slice.

commons/zenoh-buffers/src/slice.rs
```diff
@@ -242,12 +201,10 @@ impl<'a> SiphonableReader for &'a [u8] {
...
     {
-        let res = writer.write(self).map_err(|_| DidntSiphon);
-        if let Ok(len) = res {
-            // SAFETY: len is returned from the writer, therefore it means
-            //         len amount of bytes have been written to the slice.
-            *self = crate::unsafe_slice!(self, len.get()..);
-        }
-        res
+        let len = writer.write(self).map_err(|_| DidntSiphon)?;
+        // SAFETY: len is returned from the writer, therefore it means
```
fuzzypixelz, line 205: We cannot rely on the correctness of the `Writer`. This should be unsafe as well. Otherwise `Writer` has to be an unsafe trait.

commons/zenoh-buffers/src/zbuf.rs
```diff
@@ -658,7 +657,7 @@ impl Writer for ZBufWriter<'_> {
         let s = crate::unsafe_slice_mut!(cache.spare_capacity_mut(), ..len);
         // SAFETY: converting MaybeUninit<u8> into [u8] is safe because we are going to write on it.
```
fuzzypixelz, line 658: This is backwards. We should write and to it and _then_ `MaybeUninit::assume_init`.

commons/zenoh-buffers/src/vec.rs
```diff
@@ -103,7 +103,7 @@ impl Writer for &mut Vec<u8> {
         let s = crate::unsafe_slice_mut!(self.spare_capacity_mut(), ..len);
         // SAFETY: converting MaybeUninit<u8> into [u8] is safe because we are going to write on it.
```
fuzzypixelz, line 104: This is backwards. We should write and to it and _then_ `MaybeUninit::assume_init`.

fuzzypixelz (commented):
Even though this pull request doesn't affect it, `zenoh_buffers::vec::uninit` is unsound. It returns a `Vec<u8>` without initializing the bytes. It claims that its call to `Vec::set_len` is safe because "we are setting the length equal to the allocated capacity" but the Safety section of `Vec::set_len` explicitly requires the bytes to be initilized.

This affects for example `BBuf` which passes`zenoh_buffers::vec::uninit()` to `.into_boxed_slice()`.

Many places in the code unsafely cast uninitilized values to initilized values and _then_ claim it's safe after writing to it. It should be the other way around. A prime example of this is the Safety comment on `[u8]::read_zslice` (which has a  a call `zenoh_buffers::vec::uninit` even though it's a safe call).

wyfo:
> Even though this pull request doesn't affect it, zenoh_buffers::vec::uninit is unsound.

We know that, it's a huge problem, but this is not the subject of this PR. The goal is to split the work into small PRs.

commons/zenoh-buffers/src/vec.rs
wyfo: Tackling the uninitialized memory issue not the subject of this PR.

commons/zenoh-buffers/src/slice.rs
wyfo: As long as the trait is public, yes. Otherwise, it's better, but not really mandatory to me.

commons/zenoh-buffers/src/zbuf.rs
wyfo: Tackling the uninitialized memory issue not the subject of this PR.

commons/zenoh-buffers/src/bbuf.rs
wyfo: True

commons/zenoh-buffers/src/bbuf.rs
wyfo: I think you're wrong there. It's the responsability of the caller of `with_slot`, but because `with_slot` is `unsafe`, it's fine.

There is nothing wrong to have the possibility of having invariant broken by *unsafe* code. For example, you can break  `Vec` invariant using `set_len`, and every safe methods of `Vec` will be corrupted; as `set_len` is `unsafe`, it's fine. 
It's the same here: if `self.len` is updated without checking the invariant, it must be done in `unsafe` method.

commons/zenoh-buffers/src/slice.rs
wyfo: I will rollback the change, as it has to be part of a bigger change (making `Writer` unsafe)

fuzzypixelz (approved):

PR #1191, opened 2024-06-25
fix: fix keyexpr canonization unsafe behavior

### LINKED ISSUE #651: [Bug] `memcpy` source and destination overlap in `zenoh_keyexpr`
### Describe the bug

Valgrind reports `memcpy` source and destination overlap errors in the `Canonizable` impl of `&mut str`. I think the following object is the problem:

```rs
pub struct Writer {
    pub ptr: *mut u8,
    pub len: usize,
}

impl Writer {
    pub fn write(&mut self, slice: &[u8]) {
        let len = slice.len();
        unsafe { ptr::copy(slice.as_ptr(), self.ptr.add(self.len), len) };
        self.len += len
    }
    pub fn write_byte(&mut self, byte: u8) {
        unsafe { *self.ptr.add(self.len) = byte };
        self.len += 1
    }
}
```

Writer is used in the `Canonizable::canonize` implementation of `&mut str` (and `String` by extension) to write from one slice of a keyexpr to another with possible overlap, which renders the use of `ptr::copy` and not `ptr::copy_nonoverlapping` in `Writer::write` necessary.

However, because of Rust's strict W^R rules, the compiler is allowed to assume that the `&mut self` and slice slice arguments of `Writer::write` point to non overlapping regions of memory. This is because if you take a mutable borrow of a slice in Rust, you can't have another immutable borrow to it (even if they're non-overlapping) using safe code. Thus, the compiler will replace the `memmove` with the `memcpy` after optimization passes.

Here's a simple Rust playground example which confirms this: https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=c8577ea64c6c1cc514442eb014a77c82
If you select Show ASM with the Debug optim level, you'll see that the `ptr::copy` compiles to a `memmove`. But when selecting the Release optim level, it becomes a `memcpy` and all hell breaks loose.

Also, `miri` complains about the exact same `ptr::copy` call, but I'm unfamiliar with their unsafe code rules (they have a model more general than the Rust compiler's):

```
error: Undefined Behavior: not granting access to tag <107636> because that would remove [SharedReadOnly for <109114>] which is strongly protected because it is an argument of call 48948
   --> commons/zenoh-keyexpr/src/key_expr/utils.rs:24:18
    |
24  |         unsafe { ptr::copy(slice.as_ptr(), self.ptr.add(self.len), len) };
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not granting access to tag <107636> because that would remove [SharedReadOnly for <109114>] which is strongly protected because it is an argument of call 48948
    |
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
    = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
help: <107636> was created by a SharedReadWrite retag at offsets [0x0..0x5]
```

### To reproduce

1. Change the rustup toolchain to nightly in `rust-toolchain.toml`
2. Run `cargo miri test keyexpr`

### System info

- Platform: macOS 13 Ventura
- CPU: Apple M1
- Zenoh SHA: b669489bc814a758741f09c671ecc3a0683697a0

wyfo:
Fixes #651

### DIFF
--- a/commons/zenoh-keyexpr/src/key_expr/canon.rs
+++ b/commons/zenoh-keyexpr/src/key_expr/canon.rs
@@ -23,6 +23,11 @@ fn canonize(bytes: &mut [u8]) -> usize {
     let mut double_wild = false;
     loop {
         match &bytes[index..] {
+            [b'*', b'*'] => {
+                bytes[written..written + 2].copy_from_slice(b"**");
+                written += 2;
+                return written;
+            }
             [b'*', b'*', b'/', ..] => {
                 double_wild = true;
                 index += 3;
@@ -47,7 +52,7 @@ fn canonize(bytes: &mut [u8]) -> usize {
                 index += 2;
             }
             _ => {
-                if double_wild {
+                if double_wild && &bytes[index..] != b"**" {
                     bytes[written..written + 3].copy_from_slice(b"**/");
                     written += 3;
                     double_wild = false;
@@ -134,6 +139,9 @@ fn canonizer() {
     let mut s = String::from("hello/**/**/bye");
     s.canonize();
     assert_eq!(s, "hello/**/bye");
+    let mut s = String::from("hello/**/**");
+    s.canonize();
+    assert_eq!(s, "hello/**");
 
     // Any $* chunk is replaced by a * chunk
     let mut s = String::from("hello/$*/bye");
--- a/commons/zenoh-keyexpr/src/key_expr/canon.rs
+++ b/commons/zenoh-keyexpr/src/key_expr/canon.rs
@@ -17,6 +17,7 @@ pub trait Canonizable {
     fn canonize(&mut self);
 }
 
+// Return the length of the canonized string
 fn canonize(bytes: &mut [u8]) -> usize {
     let mut index = 0;
     let mut written = 0;
@@ -96,8 +97,7 @@ impl Canonizable for &mut str {
         // and remaining garbage bytes are zeroed
         let bytes = unsafe { self.as_bytes_mut() };
         let length = canonize(bytes);
-        // SAFETY: canonized length is lesser than or equal to the original length
-        unsafe { bytes.get_unchecked_mut(length..) }.fill(b'\0');
+        bytes[length..].fill(b'\0');
     }
 }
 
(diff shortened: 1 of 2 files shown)


wyfo:
@Mallets @fuzzypixelz

Mallets (approved):

commons/zenoh-keyexpr/src/key_expr/canon.rs
```diff
@@ -12,114 +12,103 @@
...
 }
 
+impl Canonizable for &mut str {
+    fn canonize(&mut self) {
+        // SAFETY: canonize leave an UTF8 string within the returned length,
+        // and remaining garbage bytes are zeroed
+        let bytes = unsafe { self.as_bytes_mut() };
+        let length = canonize(bytes);
+        // SAFETY: canonized length is lesser than or equal to the original length
+        unsafe { bytes.get_unchecked_mut(length..) }.fill(b'\0');
```
wyfo, line 100: @Mallets I don't know how much performance of canonization are critical, but do you think I can remove this unsafe?

commons/zenoh-keyexpr/src/key_expr/canon.rs
```diff
@@ -12,114 +12,103 @@
...
-        let mut s = self.as_mut();
-        s.canonize();
-        let len = s.len();
-        self.truncate(len);
+        // SAFETY: canonize leave an UTF8 string within the returned length,
+        // and remaining garbage bytes are truncated
+        let bytes = unsafe { self.as_mut_vec() };
+        let length = canonize(bytes);
+        // SAFETY: canonized length is lesser than or equal to the original length
+        unsafe { bytes.set_len(length) };
```
wyfo, line 111: @Mallets I don't know how much performance of canonization are critical, but do you think I can remove this unsafe?

commons/zenoh-keyexpr/src/key_expr/canon.rs
Mallets: Yes, go ahead.

commons/zenoh-keyexpr/src/key_expr/canon.rs
Mallets: Yes, you can remove it. Go ahead.

wyfo:
I've at last replaced the non-critical unsafe code by a safe version. Sorry for the delay.

Mallets (approved):

PR #1192, opened 2024-06-26
Implement Error trait for ZBytes error types

Mallets:
See https://github.com/eclipse-zenoh/zenoh-python/pull/240

### DIFF
--- a/zenoh/src/api/bytes.rs
+++ b/zenoh/src/api/bytes.rs
@@ -345,6 +345,28 @@ where
     Deserialize(<T as TryFrom<ZBytes>>::Error),
 }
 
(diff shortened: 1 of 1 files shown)


Mallets:
@wyfo this PR is in response to your comment in https://github.com/eclipse-zenoh/zenoh-python/pull/240

milyin (approved):

wyfo:
I've missed #1174... (I'm watching all activities of the repository now, will not miss PR again).
I don't like these new error types, that make the API so much more complex. I think we just should use `Box<dyn Error>` everywhere.
Also, we track tuple member path, but not sequence path, so it's not very consistent to me. 

Anyway, I've nothing to say about this particular PR, LGTM.

PR #1193, opened 2024-06-26
feat: unify pub/sub and query/reply modules

wyfo:


wyfo:
@Mallets @kydos @milyin

wyfo:
I've let the `core` module as is for now. Maybe we should create a `qos` module to put QOS  related things (`Priority` and `CongestionControl` for now, but maybe more later), as they are common to pub/sub and query/reply.

wyfo:
About `Encoding`, I'm hesitating to put it in `core` with `Priority` and `CongestionControl`, but it may also be put in `bytes` module, as this is closely related. I think I prefer this solution now.

Mallets:
@wyfo it would be great if this PR comes with a battery of sister PRs to update all bindings/plugins/backends with the new module organization.

wyfo:
This is incredibly long to patch all backend/plugins, so I've written a Python script to modify the imports automatically 😅

wyfo:
@Mallets All the sister PRs have been opened.

Mallets (approved):

Mallets (approved):

PR #1196, opened 2024-06-26
Allow to enable/disable automatic batching from config

Mallets:


DEFAULT_CONFIG.json5
```diff
@@ -284,6 +284,8 @@
         /// Therefore, the maximum batch size is 2^16-1 (i.e. 65535).
         /// The default batch size value is the maximum batch size: 65535.
         batch_size: 65535,
+        /// Perform automatic batching of messages if they are smaller of the batch_size
+        automatic_batching: true,
```
JEnoch, line 288: `automatic_batching` can make one think a manual batching is possible if not automatic.
Why not just `batching: true` or `enable_batching: true` ?

Mallets: I'm fine with `batching`. I'll update the PR.

JEnoch (approved):

PR #899, opened 2024-04-04
fix(771): Adding mTLS support in QUIC

gabrik:
This PR contains

- [x] Some refactoring for shared method between TLS and QUIC transport
- [x] mTLS support for QUIC
- [x] tests

Closes #771

PR #1200, opened 2024-06-27
Update ACL config format to support AND/OR logic between subjects

oteffahi:
This PR reworks the ACL config to support boolean combinations of subjects. An examples of the new config format is the following:
```json
{
  "access_control":
    {
      "enabled": true,
      "default_permission": "deny",

      "rules":
      [
        {
          "id": "allow pub/sub ingress on test/demo",
          "permission": "allow",
          "flows": ["ingress"],
          "messages": [
            "put",
            "declare_subscriber"
          ],
          "key_exprs": [
            "test/demo"
          ]
        },
        {
          "id": "allow get/queryable test/demo ingress/egress",
          "permission": "allow",
          "flows": ["ingress", "egress"],
          "messages": [
            "get",
            "declare_queryable"
          ],
          "key_exprs": [
            "test/demo"
          ]
        },
      ],

      "subjects":
      [
        {
          "id": "client1 or client2 on domain1 through en0/en1",
          "interfaces": [
            "en0",
            "en1",
          ],
          "cert_common_names": [
            "domain1.local"
          ],
          "usernames": [
            "client1name",
            "client2name"
          ]
        },
        {
          "id": "domain2",
          "cert_common_names": [
            "domain2.local"
          ]
        },
        {
          "id": "all"
          // no fields defined = wildcard (all messages)
        }
      ],

      "policies":
      [
        {
          "rules": ["allow pub/sub ingress on test/demo"],
          "subjects": [
            "client1 or client2 on domain1 through en0/en1",
            "domain2"
          ]
        },
        {
          "rules": ["allow get/queryable test/demo ingress/egress"],
          "subjects": [
            "all",
          ]
        },
      ]
    }
  }
```

Within each subject, a cartesian product is performed to produce the `(interface, cert_common_name, username)` combinations. Each combination is a logical `AND` between its components, and different combinations within the same subject in the `subjects` list represent a logical `OR` between them.

Rules are declared seperately, and applied to these subject logical combinations in the `policies` list. Unique identifiers (`id` fields) are used to represent the subjects and rules in the `policies`.


oteffahi:
@Mallets This is ready for review.
Still need to write/update ACL tests, check some edge-cases, and update default_config.

commons/zenoh-config/src/lib.rs
```diff
@@ -104,26 +104,38 @@ pub struct DownsamplingItemConf {
...
-    pub subject: Subject,
+    pub subject_id: usize,
     pub key_expr: String,
     pub action: Action,
     pub permission: Permission,
     pub flow: InterceptorFlow,
 }
 
-#[derive(Serialize, Debug, Deserialize, Eq, PartialEq, Hash, Clone)]
+#[derive(Serialize, Debug, Deserialize, Eq, PartialEq, Hash, Clone, PartialOrd, Ord)]
```
OlivierHecart, line 138: `Subject` is apparently not used in the config and only used in interceptors. If so it should be moved there.

zenoh/src/net/routing/interceptor/access_control.rs
```diff
@@ -112,11 +109,25 @@ impl InterceptorFactoryTrait for AclEnforcer {
                     Ok(links) => {
                         for link in links {
                             for face in link.interfaces {
-                                let subject = &Subject::Interface(face.clone());
-                                if let Some(val) = self.enforcer.subject_map.get(subject) {
-                                    authn_ids.push(AuthSubject {
-                                        id: *val,
-                                        name: face,
+                                // combining each interface with remaining AuthId values to get existing combinations
+                                // NOTE: current ACL logic does not apply correctly when multiple interfaces are returned
```
Mallets, line 113: A `warning` should be printed in this case.

zenoh/src/net/routing/interceptor/authorization.rs
```diff
@@ -32,7 +35,81 @@ use zenoh_util::net::get_interface_names_by_addr;
...
+    }
+
+    pub(crate) fn build(self) -> SubjectMap {
+        SubjectMap {
+            inner: self.builder.build(),
+        }
+    }
+
+    /// Assumes subject contains at most one instance of each Subject variant
+    pub(crate) fn insert_or_get(&mut self, subject: Vec<Subject>) -> usize {
```
Mallets, line 101: It's not clear to me why a builder should have an `insert_or_get`. Usually a builder always overwrite existing values. E.g.:
```rust
let a = mybuilder()
    .foo("abc")
    .foo("def")
    .build();
    
assert_eq!(a.foo(), "def");
```

zenoh/src/net/routing/interceptor/authorization.rs
```diff
@@ -143,9 +220,13 @@ impl PolicyEnforcer {
...
         if self.acl_enabled {
-            if let Some(mut rules) = mut_acl_config.rules {
-                if rules.is_empty() {
-                    tracing::warn!("Access control rules are empty in config file");
+            if let (Some(mut rules), Some(mut subjects), Some(policy)) = (
+                mut_acl_config.rules,
+                mut_acl_config.subjects,
+                mut_acl_config.policy,
+            ) {
+                if rules.is_empty() || subjects.is_empty() || policy.is_empty() {
```
Mallets, line 228: Better be explicit with the logs and have one log per empty case.

zenoh/src/net/routing/interceptor/authorization.rs
oteffahi: This is required for the subject map because we want to avoid duplicating a subject combination. If the name `insert_or_get` is not fitting what should we call it?

zenoh/src/net/routing/interceptor/authorization.rs
```diff
@@ -156,59 +237,74 @@ impl PolicyEnforcer {
...
+                        if subject.id.trim().is_empty() {
+                            bail!("Found empty subject id in subjects list");
+                        }
+                        if subject.interfaces.is_none() {
+                            if subject.cert_common_names.is_none() && subject.usernames.is_none() {
+                                tracing::warn!(
+                                    "Subject '{}' is empty. Setting it to wildcard (all network interfaces)",
+                                    subject.id
+                                );
+                                subject.interfaces = Some(get_interface_names_by_addr(
```
Mallets, line 261: If `IPv4` is disabled in the system this operation will fail. Check first `IPv6` and then `IPv4`, return error when both fail.

zenoh/src/net/routing/interceptor/authorization.rs
Mallets: It's unclear to me then who is providing the `Vec<Subject>` as parameter that is ignored in case a subject is already present. What I want to be sure of is that a misusage of `insert_or_get` can lead to a security misconfiguration.

oteffahi: The `Vec<Subject>` is the "subject", or better say the subject combination. Each individual subject combination is mapped to an id in the SubjectMap (the combination containing at lease one subject).
The issue is that when parsing the ACL config, we perform the cartesian product of `interfaces`, `cert_common_names` and `usernames` within each configured subject, to derive the subject combinations from each individual subject object in config, which can yield subject combinations that intersect with another subject object in the config. Depending on the parsing order, some combinations may have already been mapped, and overwriting those mappings would require updating previously parsed subject references in the intermediate data structures.
With this context, could you please explain what security misconfiguration can arise?

Mallets: It's still not clear to me how the linear parsing of the configuration could lead lead to different parsing order. Would you mind clarifying this point?

oteffahi: Sorry I may have misspoken, I'll reiterate my point.

What I mean is that depending on the subjects declared in the config (in the `subjects` array), we may encounter during config parsing a subject combination that was previously put in the subject map and associated to another subject id (the unique string identifier provided in the config). The goal of `insert_or_get` is to avoid overwriting previously mapped subject combinations, in order to avoid updating the reference in the parsing data structure for previously parsed subject config (the data structure being a hashmap that maps subject string Ids to a `Vec<usize>` containing the "references" to the subject combinations).

On another note, I realize now that calling "subjects" both the combinations and the individual characteristics (interfaces, cert_common_names and usernames) is confusing.

zenoh/src/net/routing/interceptor/access_control.rs
```diff
@@ -112,11 +109,25 @@ impl InterceptorFactoryTrait for AclEnforcer {
...
+                                let face = Subject::Interface(face.clone());
+                                let mut subject_query = subjects.clone();
+                                subject_query.push(face);
+                                for subject_combination in
+                                    self.enforcer.subject_map.get(&subject_query)
+                                {
+                                    authn_ids.insert(AuthSubject {
+                                        id: *subject_combination.1,
+                                        name: format!(
```
fuzzypixelz, line 122: A quick "Search for references" using r-a shows that this field is only used in `tracing` events. While this path is not hot I believe that `name` ought to just be in the `Debug` or `Display` impl.

zenoh/src/net/routing/interceptor/authorization.rs
```diff
@@ -32,7 +35,81 @@ use zenoh_util::net::get_interface_names_by_addr;
...
+    inner: TrieMap<Subject, usize>,
+}
+
+impl SubjectMap {
+    pub(crate) fn builder() -> SubjectMapBuilder {
+        SubjectMapBuilder::new()
+    }
+
+    pub(crate) fn get(&self, subjects: &[Subject]) -> Vec<(Vec<Subject>, &usize)> {
+        let mut subjects = subjects.to_owned();
```
fuzzypixelz, line 50:
```suggestion
        let mut subjects = subjects.to_vec();
```

zenoh/src/net/routing/interceptor/authorization.rs
```diff
@@ -32,7 +35,81 @@ use zenoh_util::net::get_interface_names_by_addr;
...
+        subjects.sort_unstable();
+        let mut res: Vec<(Vec<Subject>, &usize)> = vec![];
+        for subject in subjects {
+            let map_query: Vec<(Vec<Subject>, &usize)> =
+                self.inner.predictive_search([subject]).collect();
+            for (subject, id) in map_query {
+                if subject.len() > 1 {
+                    for s in &subject {
+                        if subject.binary_search(s).is_err() {
```
fuzzypixelz, line 59: Isn't this always true?

zenoh/src/net/routing/interceptor/authorization.rs
```diff
@@ -235,105 +346,148 @@ impl PolicyEnforcer {
...
+                    "Subject id must be unique: id '{}' is repeated",
+                    config_subject.id
+                );
             }
-
-            for subject in config_rule.interfaces.as_ref().unwrap() {
-                if subject.trim().is_empty() {
-                    bail!("found an empty interface value in interfaces list");
+            // validate subject config fields
+            let interfaces = config_subject.interfaces.as_ref().unwrap();
```
fuzzypixelz, line 367: If the `unwrap` is not supposed to panic, I suggest using `expect` alongside the justification as to why. Same goes for the two `unwrap`s blow.

gabrik:
I have got a question, what if we do not know a priori the interfaces? can an empty list or setting them to none apply the configuration on all interfaces?

Mallets:
> Another question, I see that we can specify common names, so we can imagine that a user can create certificates for its clients with whathever common name, but seems that the router refuses to open the mTLS session in that case.
> 
> ```
> Can not accept TLS connection: received fatal alert: BadCertificate
> 2024-07-17T07:08:04.592149Z  WARN acc-0 ThreadId(36) zenoh_link_quic::unicast: QUIC acceptor failed: ConnectionClosed(ConnectionClose { error_code: Code::crypto(2a), frame_type: None, reason: b"invalid peer certificate: NotValidForName" }) at io/zenoh-links/zenoh-link-quic/src/unicast.rs:377.
> 2024-07-17T07:08:04.592192Z  WARN acc-0 ThreadId(36) zenoh_link_quic::unicast: QUIC acceptor failed: ConnectionClosed(ConnectionClose { error_code: Code::crypto(2a), frame_type: None, reason: b"invalid peer certificate: NotValidForName" }) at io/zenoh-links/zenoh-link-quic/src/unicast.rs:377. Hint: increase the system open file limit.
> ```

As far as I recall mTLS is not supported in QUIC

gabrik:
> > Another question, I see that we can specify common names, so we can imagine that a user can create certificates for its clients with whathever common name, but seems that the router refuses to open the mTLS session in that case.
> > ```
> > Can not accept TLS connection: received fatal alert: BadCertificate
> > 2024-07-17T07:08:04.592149Z  WARN acc-0 ThreadId(36) zenoh_link_quic::unicast: QUIC acceptor failed: ConnectionClosed(ConnectionClose { error_code: Code::crypto(2a), frame_type: None, reason: b"invalid peer certificate: NotValidForName" }) at io/zenoh-links/zenoh-link-quic/src/unicast.rs:377.
> > 2024-07-17T07:08:04.592192Z  WARN acc-0 ThreadId(36) zenoh_link_quic::unicast: QUIC acceptor failed: ConnectionClosed(ConnectionClose { error_code: Code::crypto(2a), frame_type: None, reason: b"invalid peer certificate: NotValidForName" }) at io/zenoh-links/zenoh-link-quic/src/unicast.rs:377. Hint: increase the system open file limit.
> > ```
> 
> As far as I recall mTLS is not supported in QUIC

It is: https://github.com/eclipse-zenoh/zenoh/pull/899

DEFAULT_CONFIG.json5
```diff
@@ -207,9 +229,43 @@
...
+  //       /// (interface="lo0" && cert_common_name="example2.zenoh.io") ||
+  //       /// (interface="en0" && cert_common_name="example2.zenoh.io")
+  //     },
+  //     {
+  //       "id": "subject3",
+  //       /// An empty subject combination is a wildcard (all interfaces)
+  //     },
+  //   ],
+  //   /// The policy list contains policy entries
+  //   "policy":
```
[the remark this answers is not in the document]
fuzzypixelz: This is more a nit-pick.

DEFAULT_CONFIG.json5
```diff
@@ -207,9 +229,43 @@
   //       /// Subjects can be usernames when using user/password authentication
   //       "usernames": [
   //         "zenoh-example"
-  //       ]
+  //       ],
+  //       /// This instance translates internally to this filter:
+  //       /// (interface="lo0" && cert_common_name="example.zenoh.io" && username="zenoh-example") ||
```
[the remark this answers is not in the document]
fuzzypixelz: Another nitpick.

zenoh/src/net/routing/interceptor/authorization.rs
oteffahi: @fuzzypixelz I don't see why it would be always false, could you elaborate?

DEFAULT_CONFIG.json5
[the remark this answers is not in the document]
oteffahi: A policy is a set of rules applied to subjects. Calling it "policies" in plural would be confusing in my opinion.

zenoh/src/net/routing/interceptor/authorization.rs
fuzzypixelz: Because `subject.contains(s)` as `s` is the iteration variable over `subjects`?

oteffahi: Ah there's a typo, it should've been `subjects.binary_search(s).is_err()`.
`subject` is a subject combination yielded from the predictive search within the trie, it needs to be checked if it contains subjects that are not present in the query, hence the search in `subjects` list.
The variable names are a bit confusing...

oteffahi:
@fuzzypixelz could you please explain why you rewrote all the subject combinations logic?
The previous logic worked and supported adding any new `Subject` type with very minor changes, and relied on a trie to avoid a sequential search within a list, which is the case of your newer implementation.

zenoh/src/net/routing/interceptor/authorization.rs
fuzzypixelz: Okay, this is likely the source of the bug I observed. I honestly couldn't be sure of the intent behind the code, so it was simpler for me to replace it for now. It's ultimately up to you to decide on an (efficient) data structure.

Mallets: I'd propose to focus on the correctness of this PR at this stage.
Configuration parsing could be further optimized but it should produce correct result first. 
Therefore I suggest to go ahead with the solution proposed by @fuzzypixelz if proved correctly at this stage.

fuzzypixelz:
> @fuzzypixelz could you please explain why you rewrote all the subject combinations logic? The previous logic worked and supported adding any new `Subject` type with very minor changes, and relied on a trie to avoid a sequential search within a list, which is the case of your newer implementation.

When my review was requested I noticed that `subject.binary_search(s).is_err()` which led to me find that:

1. If a subject specifies a “valid” interface and an “invalid” certification common name, then the subject is incorrectly authenticated with the “invalid” certification common name, even though it mismatches its actual common name. Same thing happens with “valid” certification common name and an “invalid” username or a “valid” interface and an “invalid” username.

2. If no subject property is defined (i.e. username, certificate common name or interface), then the result is a wildcard on all interfaces (this works because a subject always has an interface). This should arguably be a wildcard on all subject properties, not just interfaces.

3. If at least one, but not all subject properties are defined, the result is not a wildcard on the undefined subject properties. There should arguably be a wildcard on all missing subject properties to be more consistent.

I rewrote it to ensure that (1) doesn't occur, and ended up addressing (2) and (3) also to have a generic handling of wildcards. I believed that the root issue was the Trie data structure (it searches ordered subject properties by prefix).Perhaps the typo in the binary search call would've addressed this, but I'm not sure.

It was also not clear to me that it was necessarily faster (subject properties were sorted, one prefix search was performed for each subject property, possibly a binary search for each queried subject property and relative cache unfriendliness of the Trie).

Ultimately I had ensure the implementation is sound in a timely manner so I wrote the simplest code I could think of. Needless to say, it is not up to me to decide on the final implementation.

gabrik:
I just tested it a bit and now it works as expected, I just have one minor comment, the current implementation is always an `AND` in the subject, it would be interesting to configure the behavior between `AND` and `OR`?

DEFAULT_CONFIG.json5
```diff
@@ -204,20 +206,77 @@
   //       "key_exprs": [
   //         "test/demo"
   //       ],
+  //     },
+  //     {
+  //       "id": "rule2",
+  //       "actions": [
```
gabrik, line 212: I think `actions` is the wrong word here, as those are more message kinds.

DEFAULT_CONFIG.json5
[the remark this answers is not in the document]
gabrik: It is an array of `policy` , so `policies`

gabrik (changes requested):

gabrik:
Also, `queryable` and `get` actions are not clear, like a queryable must have the permissions to receive `get` and send `reply`, in the current implementation there is no such `reply` action.

I would expect rules like (pseudocode, all allows), from the PoV of a router
```
queryable: 
   - declare_queryable: ingress
   - get: egress
   - reply: ingress
getter:
  - declare_queryable: egress
  - get: ingress
  - reply: egress 
```

in current implementation we have

```
queryable:
 - declare_queryable: ingress
 - get: egress
 getter:
  - declare_queryable: egress
  - get: ingress
```

DEFAULT_CONFIG.json5
oteffahi: The naming is from ACL SOTA: subjects (interface, cert, username) perform actions (in the form of messages) on resources (key_exprs).

oteffahi:
> I just tested it a bit and now it works as expected, I just have one minor comment, the current implementation is always an `AND` in the subject, it would be interesting to configure the behavior between `AND` and `OR`?

@gabrik `OR` is expressed in the `policy` array by having one rule apply to multiple subjects that are constructed in a way to express a logical `OR`.

DEFAULT_CONFIG.json5
fuzzypixelz: But if you want to set a policy that prevents a node from _receiving_ a Subcriber Declaration that is configured as allowing **`declare_subscriber` actions** for egress flow. Even though the node receiving the declaration isn't doing anything. Same goes for `put` actions.

oteffahi:
>  It is an array of `policy` , so `policies`
In SOTA, the access control policy (singular) is the set of rules applied to subjects, so calling it `policies` would be confusing in that regard. The whole configuration creates the ACL policy.

oteffahi:
> Also, `queryable` and `get` actions are not clear, like a queryable must have the permissions to receive `get` and send `reply`, in the current implementation there is no such `reply` action.
> 
> I would expect rules like (pseudocode, all allows), from the PoV of a router
> 
> ```
> queryable: 
>    - declare_queryable: ingress
>    - get: egress
>    - reply: ingress
> getter:
>   - declare_queryable: egress
>   - get: ingress
>   - reply: egress 
> ```
> 
> in current implementation we have
> 
> ```
> queryable:
>  - declare_queryable: ingress
>  - get: egress
>  getter:
>   - declare_queryable: egress
>   - get: ingress
> ```

`Reply` actions were excluded from the initial ACL design because we assumed that allowing/denying a `get` means allowing/denying its associalted `Reply` even if it is not on the same `key_expr`. This can be changed in a future PR if deemed useful for certain applications. @OlivierHecart what do you think about this?

DEFAULT_CONFIG.json5
oteffahi: After further discussion within the team, the config's perspective will be changed from `actions` to `messages`. The naming will be updated in this PR.

zenoh/src/net/routing/interceptor/authorization.rs
oteffahi: Following @fuzzypixelz's implementation of wildcards, this interface scanning is no longer required.

gabrik:
> @gabrik `Reply` actions were excluded from the initial ACL design because we assumed that allowing/denying a `get` means allowing/denying its associated `Reply` even if it is not on the same `key_expr`

I'm against implicit things, because if that applies for `get` then also a `declare_subscribe` should imply the `put`

fuzzypixelz (approved):
I would rename "declare_subscriber" to "subscriber_declaration", "declare_queryable" to "queryable_declaration", "put" to "publication" and "get" to "query".

The idea is to use nouns instead of verbs now that we think of these as data and not actions.

This is more a matter of opinion so overall this LGTM 🚀

fuzzypixelz:
> > I would rename "declare_subscriber" to "subscriber_declaration", "declare_queryable" to "queryable_declaration", "put" to "publication" and "get" to "query".
> > The idea is to use nouns instead of verbs now that we think of these as data and not actions.
> > This is more a matter of opinion so overall this LGTM 🚀
> 
> If we move to `messages` than the messages are `declare_subscriber`, `declare_queryable`, `put`, and `get`. Those messages have the same name of the exposed API used by applications, making it quite easy to correlate `messages` to `API` for the user.

@Mallets If we're talking about the API names then let's not forget that `Publisher::put` returns a `Publication`. That a queryable receives `Query` objects (we don't sat it receives a "get"). That `Subscriber::undeclare()` returns a `SubscriberUndeclaration` (there is no `SubscriberDeclaration` type but that's more of an inconsistency in the API naming convention).

It is not enough for the message names to reflect the API of the subject impacted by `ingress` rules. Subjects impacted by egress rules also ought to be considered. 

If I was a user it would be immediately clear to me that a rule with on message `get` would impact my `Session:get` but I would very surprised to learn that it also impacts my `Queryable::recv_async`.

The message names cannot match the API calls of subjects on both ingress and egress sides. If they match either one, then the naming is misleading for the other. Thus the names ought to be abstracted over both.

Mallets:
> > > I would rename "declare_subscriber" to "subscriber_declaration", "declare_queryable" to "queryable_declaration", "put" to "publication" and "get" to "query".
> > > The idea is to use nouns instead of verbs now that we think of these as data and not actions.
> > > This is more a matter of opinion so overall this LGTM 🚀
> > 
> > 
> > If we move to `messages` than the messages are `declare_subscriber`, `declare_queryable`, `put`, and `get`. Those messages have the same name of the exposed API used by applications, making it quite easy to correlate `messages` to `API` for the user.
> 
> @Mallets If we're talking about the API names then let's not forget that `Publisher::put` returns a `Publication`. That a queryable receives `Query` objects (we don't say it receives a "get"). That `Subscriber::undeclare()` returns a `SubscriberUndeclaration` (there is no `SubscriberDeclaration` type but that's more of an inconsistency in the API naming convention).
> 
> It is not enough for the message names to reflect the API of the subject impacted by `ingress` rules. Subjects impacted by egress rules also ought to be considered.
> 
> If I was a user it would be immediately clear to me that a rule with on message `get` would impact my `Session:get` but I would very surprised to learn that it also impacts my `Queryable::recv_async`.
> 
> The message names cannot match the API calls of subjects on both ingress and egress sides. If they match either one, then the naming is misleading for the other. Thus the names ought to be abstracted over both.

The mapping has nothing to do with Rust types but with the Zenoh API and semantics. Here we are defining a config that can be loaded by a C/C++/Python/Java/Kotlin application. All those languages consistently define `declare_subscriber`, `declare_queryable`, etc. operations. So, IMHO restricting the scope to Rust type and driving the discussion around that would be a mistake.

Mallets (approved):

gabrik (approved):
with reserve to improve the semantic in the future.

i.e, split between `get` and `query`

oteffahi:
> with reserve to improve the semantic in the future.
> 
> i.e, split between `get` and `query`

You mean `get` and `reply`?

gabrik:
> > with reserve to improve the semantic in the future.
> > i.e, split between `get` and `query`
> 
> You mean `get` and `reply`?

Yup

fuzzypixelz:
> The mapping has nothing to do with Rust types but with the Zenoh API and semantics.

@Mallets My argument wasn't centered around the Rust API names, I was just giving examples. Consider the following ACL rule:
```json
{
  "id": "deny-get-egress",
  "flows": ["egress"],
  "messages": ["get"],
  "permission": "deny",
  "key_exprs": ["test/1"]
}
```

If `deny-get-egress` is applied to subject S in a router R for instance, then R would block  any queries matching `test/1` from reaching S.

What are the semantics of _denying_ "get" _messages_[^1] from reaching S? The word "get" refers to the Zenoh API operation `get` using your wording.

Here we are trying to configure whether or not _queries_ can reach S, the semantics are more clearly expressed in terms of data in this case, not API operations. Thus "operations" do not capture the semantics of `messages` in all cases (granted they do for `ingress` flows).

[^1]: If we're being pedantic here, there is not such thing as a "get" message, it is an operation.

Mallets (approved):

PR #1201, opened 2024-06-27
Add code coverage to CI

diogomatsubara:


diogomatsubara:
After some discussion with @Mallets we decided to take a different direction for now

PR #1207, opened 2024-07-01
Fix bug in keyexpr::includes leading to call get_unchecked on empty array UB

OlivierHecart:


wyfo:
I forgot to open the PR after our discussion 😅
LGTM

PR #1211, opened 2024-07-02
Cargo tarpaulin

diogomatsubara:


diogomatsubara:
After some discussion with @Mallets with decided to take a different direction for now

PR #1214, opened 2024-07-03
Use TCP MSS as TCP link MTU

Mallets:
This PR retrieves the configured TCP MSS (Maximum Segment Size) for TCP-based stream (i.e. TCP and TLS) and uses the MSS value as advertised link MTU. This allows to choose the Zenoh frame size more wisely based on the reported information.

PR #1215, opened 2024-07-04
Reduce the default batch_size

kydos:
Reduce the default batch size to 64512 bytes to ensure that a zenoh message always fits in 64K.


Mallets:
The proposed change only updates the `DEFAULT_CONFIG` reference file, which has no impact on any parameter configured in Zenoh. It's only the user-facing configuration reference.

See #1214 for additional discussion.

Mallets:
Superseded by #1214

PR #1219, opened 2024-07-05
fix(storage-manager): do not start when 'timestamping' is disabled

J-Loudet:
All storage must have a timestamp associated with a Sample.

As it is possible to publish without adding a timestamp, it means that a Zenoh node must add this timestamp "at some point". Up until now, the default configuration of a router ('timestamping' enabled) combined with the fact that only routers could load plugins (and, thus, storage) made it so that a timestamp was (by default) always added.

Recent changes in Zenoh — namely the fact that not only routers can load plugins and that peers and client have, by default, the 'timestamping' configuration disabled — invalidate these assumptions.

We should then enforce at runtime, that the 'timestamping' configuration is enabled when attempting to load the storage manager.

This commit adds this check by verifying that there is an HLC associated with the Zenoh Session — the HLC is only created if 'timestamping' is enabled (see `zenoh/zenoh/src/net/runtime/mod.rs::142`).

* plugins/zenoh-plugin-storage-manager/src/lib.rs: return an error if the storage manager is started while the configuration option 'timestamping' is disabled.
* plugins/zenoh-plugin-storage-manager/tests/operations.rs: updated the `config` used in the test to enable 'timestamping'.
* plugins/zenoh-plugin-storage-manager/tests/wildcard.rs: updated the `config` used in the test to enable 'timestamping'.

### DIFF
--- a/plugins/zenoh-plugin-storage-manager/src/lib.rs
+++ b/plugins/zenoh-plugin-storage-manager/src/lib.rs
@@ -31,6 +31,7 @@ use memory_backend::MemoryBackend;
 use storages_mgt::StorageMessage;
 use zenoh::{
     internal::{
+        bail,
         plugins::{Response, RunningPlugin, RunningPluginTrait, ZenohPlugin},
         runtime::Runtime,
         zlock, LibLoader,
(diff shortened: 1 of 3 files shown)


Charles-Schleich:
LGTM.
Related Issue on RocksDB:  https://github.com/eclipse-zenoh/zenoh-backend-rocksdb/issues/129

JEnoch (approved):

PR #1224, opened 2024-07-05
Configure `scouting/*/autoconnect/*` with a sequence

fuzzypixelz:
This is easier to read and less error prone. JSON already has a construct for sequences, we should use it.

### DIFF
--- a/commons/zenoh-protocol/src/core/whatami.rs
+++ b/commons/zenoh-protocol/src/core/whatami.rs
@@ -142,10 +142,12 @@ impl WhatAmIMatcher {
             Self::U8_R => WhatAmI::STR_R,
             Self::U8_P => WhatAmI::STR_P,
             Self::U8_C => WhatAmI::STR_C,
-            Self::U8_R_P => formatcp!("{}|{}", WhatAmI::STR_R, WhatAmI::STR_P),
-            Self::U8_R_C => formatcp!("{}|{}", WhatAmI::STR_R, WhatAmI::STR_C),
-            Self::U8_P_C => formatcp!("{}|{}", WhatAmI::STR_P, WhatAmI::STR_C),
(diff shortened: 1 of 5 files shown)


commons/zenoh-protocol/src/core/whatami.rs
```diff
@@ -142,10 +142,12 @@ impl WhatAmIMatcher {
...
             Self::U8_C => WhatAmI::STR_C,
-            Self::U8_R_P => formatcp!("{}|{}", WhatAmI::STR_R, WhatAmI::STR_P),
-            Self::U8_R_C => formatcp!("{}|{}", WhatAmI::STR_R, WhatAmI::STR_C),
-            Self::U8_P_C => formatcp!("{}|{}", WhatAmI::STR_P, WhatAmI::STR_C),
-            Self::U8_R_P_C => formatcp!("{}|{}|{}", WhatAmI::STR_R, WhatAmI::STR_P, WhatAmI::STR_C),
+            Self::U8_R_P => formatcp!("[{},{}]", WhatAmI::STR_R, WhatAmI::STR_P),
+            Self::U8_R_C => formatcp!("[{},{}]", WhatAmI::STR_R, WhatAmI::STR_C),
+            Self::U8_P_C => formatcp!("[{},{}]", WhatAmI::STR_P, WhatAmI::STR_C),
+            Self::U8_R_P_C => {
+                formatcp!("[{},{},{}]", WhatAmI::STR_R, WhatAmI::STR_P, WhatAmI::STR_C)
```
Mallets, line 149: This part has nothing to do with JSON. Moreover, the config file could be provided in YAML as well. Therefore, I believe it is wrong to enforce JSON syntax at this level.

fuzzypixelz: Fixed. I also restored the `FromStr` impl that I had deleted before

Mallets (approved):

PR #1229, opened 2024-07-08
fix(storage-manager): validate presence of timestamp

J-Loudet:
This commit introduces checks before accessing the `timestamp` associated with a Sample — instead of calling `unwrap()`.

In theory, a Sample should never arrive to a Storage without a Timestamp. In practice, we cannot guarantee this invariant with certainty (future modifications of the code base?). With these checks, the Storage will simply discard the Sample instead of panicking the entire storage manager.

* plugins/zenoh-plugin-storage-manager/src/replica/storage.rs: add checks when accessing the timestamp and remove `unwrap`.


Charles-Schleich (approved):
LGTM

JEnoch (approved):

PR #1231, opened 2024-07-08
fix: cover the zero capacity case of the handlers

### LINKED ISSUE #1230: [dev/1.0.0] Zero capacity handlers are not handled properly
### Describe the bug

Users may attempt to create a handler for a zero-capacity channel. For example, we support this function in zenoh-c v0.11.0. https://github.com/ZettaScaleLabs/zenoh-c/blob/a98d700e994598fe3d26f47c417d397b66902ead/src/closures/response_channel.rs?plain=1#L52.    

### To reproduce

Skip

### System info

- zenoh: 6df74c7

YuanYuYuan:
Closes #1230

### DIFF
--- a/zenoh/src/api/handlers/fifo.rs
+++ b/zenoh/src/api/handlers/fifo.rs
@@ -38,7 +38,11 @@ impl<T: Send + 'static> IntoHandler<'static, T> for FifoChannel {
     type Handler = flume::Receiver<T>;
 
     fn into_handler(self) -> (Callback<'static, T>, Self::Handler) {
-        flume::bounded(self.capacity).into_handler()
+        if self.capacity == 0 {
+            flume::unbounded().into_handler()
+        } else {
+            flume::bounded(self.capacity).into_handler()
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


zenoh/src/api/handlers/fifo.rs
```diff
@@ -38,7 +38,11 @@ impl<T: Send + 'static> IntoHandler<'static, T> for FifoChannel {
     type Handler = flume::Receiver<T>;
 
     fn into_handler(self) -> (Callback<'static, T>, Self::Handler) {
-        flume::bounded(self.capacity).into_handler()
+        if self.capacity == 0 {
```
Mallets, line 41: I don't like implicit behaviour with magic numbers that change the type of the internal channel.
Users can already pass an unbound channel with 
```rust
session
    .declare_subscriber("myke")
    .with(flume::unbounded())
    .await
    .unwrap();
```

For the above reason I don't see the need of such change.

wyfo:
`flume` supports bounded channel with a capacity of zero. It's called a "rendez-vous" channel. Even if I don't think any user would want to use that for their handlers, I agree with Luca that changing the semantic depending on the capacity being zero is not a Rust thing.

The "trick" we are using for C may be regarded as a specific C thing, but we can also choose to either expose an additional `zc_reply_fifo_new_unbounded`, or change the signature to accept `isize` with a negative value meaning unbounded.
And after consideration, knowing that a capacity of zero is supported as "rendez-vous", I think that `-1` would be better.

PR #1238, opened 2024-07-11
Make `rustls`, `rustls-webpki` and `webpki-roots` optional

fuzzypixelz:
In f36b228a18cb048e93751c2b58075d3d65e4b79c the `zenoh-link-commons` (used by `zenoh-link`) crate was made to depend on the aforementioned TLS crates. This is not necessary as the `tls` module is only used by `zenoh-link-tls` and `zenoh-link-quic`.

### DIFF
--- a/io/zenoh-link-commons/Cargo.toml
+++ b/io/zenoh-link-commons/Cargo.toml
@@ -26,14 +26,15 @@ version = { workspace = true }
 
 [features]
 compression = []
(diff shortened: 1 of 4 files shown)


torsteingrindvik (approved):
Checked `cargo tree`, looks good.
Also tried on our CI which had license issues before this PR, but now it's ok :tada:

JEnoch (approved):

Mallets (approved):

PR #1241, opened 2024-07-16
Null value opts for shm

yellowhatter:
Some null value optimizations to equalize SHM data structures for zenoh-c


commons/zenoh-shm/src/api/provider/types.rs
```diff
@@ -61,14 +59,18 @@ impl Default for AllocAlignment {
...
+        }
     }
 
     /// Get alignment in normal units (bytes)
     #[zenoh_macros::unstable_doc]
-    pub fn get_alignment_value(&self) -> usize {
-        1usize << self.pow
+    pub fn get_alignment_value(&self) -> NonZeroUsize {
+        // SAFETY: this is safe
+        unsafe { NonZeroUsize::new_unchecked(1usize << self.pow) }
```
Mallets, line 73: `AllocAlignment::new()` takes `pow: u8` that returns `Ok` if `pow < 64`.
`1usize << self.pow` may overflow in case `usize` is `32bit`.

commons/zenoh-shm/src/api/provider/types.rs
```diff
@@ -84,11 +86,13 @@ impl AllocAlignment {
...
     #[zenoh_macros::unstable_doc]
-    pub fn align_size(&self, size: usize) -> usize {
+    pub fn align_size(&self, size: NonZeroUsize) -> NonZeroUsize {
         let alignment = self.get_alignment_value();
-        match size % alignment {
+        match size.get() % alignment {
             0 => size,
-            remainder => size + (alignment - remainder),
+            remainder => unsafe {
+                NonZeroUsize::new_unchecked(size.get() + (alignment.get() - remainder))
```
Mallets, line 94: What's the guarantee of this `unsafe` block?

commons/zenoh-shm/src/api/provider/types.rs
```diff
@@ -61,14 +59,18 @@ impl Default for AllocAlignment {
 
 impl AllocAlignment {
     #[zenoh_macros::unstable_doc]
-    pub fn new(pow: u8) -> Self {
-        Self { pow }
+    pub fn new(pow: u8) -> Result<Self, ZLayoutError> {
+        match pow {
+            pow if pow < 64 => Ok(Self { pow }),
```
wyfo, line 64:
```suggestion
            pow if pow < usize::BITS as u8 => Ok(Self { pow }),
```

commons/zenoh-shm/src/api/provider/types.rs
```diff
@@ -61,14 +59,18 @@ impl Default for AllocAlignment {
 
 impl AllocAlignment {
     #[zenoh_macros::unstable_doc]
```
Mallets, line 61: Add doc on the function returning an error if `pow >= 64`.

commons/zenoh-shm/src/api/provider/types.rs
```diff
@@ -113,16 +117,16 @@ impl Display for MemoryLayout {
 impl MemoryLayout {
     /// Try to create a new memory layout
     #[zenoh_macros::unstable_doc]
-    pub fn new(size: usize, alignment: AllocAlignment) -> ZResult<Self> {
+    pub fn new(size: NonZeroUsize, alignment: AllocAlignment) -> Result<Self, ZLayoutError> {
```
Mallets, line 120: To improve ergonomics you could change the signature:
```rust
pub fn new<T>(size: T, alignment: AllocAlignment) -> Result<Self, ZLayoutError>
where
    T: TryInto<NonZeroUsize>
```

In this way user can still do:
```rust
let layout = MemoryLayout::new(100, AllocAlignment::default()).unwrap();
```
instead of:
```rust
let layout = MemoryLayout::new(100.try_into().unwrap(), AllocAlignment::default()).unwrap();
```

yellowhatter: There are two reasons why I propose to leave `Alignment` and `MemoryLayout` in it's current the form for a while:
1. I want to have a good benchmarking suite for SHM before doing any further opts. The integrated benchmarking, that is able to check big parts of our code, not just a small ones (that can be easily optimized by the compiler).
2. We need `NonZeroUsize's` Null Value Opt in some of our data structures that aggregate `MemoryLayout` for making some things very beautiful and effecient in zenoh-cpp. Placing `NonZeroUsize` in `MemoryLayout` is a good solution here.

commons/zenoh-shm/src/api/provider/types.rs
yellowhatter: Fixed through `usize::BITS` as @wyfo suggested

commons/zenoh-shm/src/api/provider/types.rs
yellowhatter: @Mallets this unsafe block is always safe:
1. remainder is always less than alignment
2. because of 1, the value of (alignment.get() - remainder) is always > 0
3. because of 2, we add nonzero size to nonzero (alignment.get() - remainder)
4. we make sure that there is no overflow condition in 3 by means of alignment limitation in `new` by limiting pow value

I will add a comment to the code

@wyfo unfortunately, the compiler is not able to infer this!

wyfo: Indeed, it's unfortunate, and I'm quite surprise that it's not able to even infer that `alignment.get() - remainder` is non-zero 🤔

commons/zenoh-shm/src/api/provider/shm_provider.rs
```diff
@@ -182,8 +183,14 @@ where
         // NOTE: Depending on internal implementation, provider's backend might relayout
         // the allocations for bigger alignment (ex. 4-byte aligned allocation to 8-bytes aligned)
 
+        // Obtain nonzero size
+        let nonzero_size = data
```
Mallets, line 187: I believe this block can be removed since the conversion is now done inside `MemoryLayout::new` which will return `ZLayoutError::IncorrectLayoutArgs` in case of *zero size*.

commons/zenoh-shm/src/api/provider/types.rs
yellowhatter: @wyfo  my understanding is:
1. Fist of all, `1usize << pow` is evaluated, and it's result is `usize`. An overflow might happen here, which will panic on Debug and be UB on Release (UB because it is processor-defined)
2. In `NonZeroUsize::new` compiler is able to guess that ON TARGET ARCHITECTURE any shift operation applied to `1usize` constant at step 1 will give nonzero result, even in case of overflow. It means that for TARGET ARCH a check may be elided, but fir other architectures this might be wrong and check assembly will be generated.

Mallets (approved):

PR #1250, opened 2024-07-22
Do not require mandatory "unstable" for "internal" feature

yellowhatter:


yellowhatter:
@DenisBiryukov91 @Mallets please review

"internal" should not require "unstable", otherwise we end up compiling and exporting to user a lot of "unstable" API in zenoh-c just for nothing

Mallets (approved):

PR #1256, opened 2024-07-24
SHM bugfix

yellowhatter:
Fix issues discovered through Raspberry PI evaluation:
- 32bit misaligned access
- Drop ZBuf in case of SHM error instead of dropping transport


yellowhatter:
@Mallets

Mallets (approved):

PR #1258, opened 2024-07-24
Fix #1239. Fix timestamp codec len calculation

### LINKED ISSUE #1239: Router modifies pico push timestamp ext size.
### Describe the bug

I'm trying to send a timestamp in my push through a router with my pub/sub on pico.

For some reason, the size field of my timestamp extension goes from 0x16 (22 bytes) when entering the router to 0x06 (6 bytes) when exiting the router towards by subscriber... the buffer has not changed size and the deserialization fails.

### To reproduce

Start a router
Start a publisher on pico sending a timestamp in the push message (not the put body), 
Start a subscriber.
With Wireshark, check that the incoming packet has the correct `0x42` (extension buffer) `0x16` (22 bytes size) and the ougoing one is now `0x42` `0x06`.

### System info

Ubuntu 22.04

Mallets:
Fix incorrect timestamp length calculation when timestamp is encoded in a message extension.

### DIFF
--- a/commons/zenoh-codec/src/core/timestamp.rs
+++ b/commons/zenoh-codec/src/core/timestamp.rs
@@ -23,7 +23,8 @@ use crate::{LCodec, RCodec, WCodec, Zenoh080};
 
 impl LCodec<&Timestamp> for Zenoh080 {
     fn w_len(self, x: &Timestamp) -> usize {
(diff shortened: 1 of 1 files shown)


Mallets:
Fix #1239 .

@jean-roland can you please review this PR?

PR #1259, opened 2024-07-24
Add publisher delete and queryable reply messages to ACL

### LINKED ISSUE #1255: Queryable reply keyexpr should be exposed to interceptors
### Describe the feature

[This TODO line](https://github.com/eclipse-zenoh/zenoh/blob/main/zenoh/src/net/routing/dispatcher/queries.rs#L810) should be implemented in order to expose the appropriate reply keyexpr to interceptors.

oteffahi:
Adds filtering of publisher delete and queryable reply messages to ACL.
Example of ACL rules with delete and reply messages:
```json
{
  "access_control":
    {
      "enabled": true,
      "default_permission": "deny",
      "rules":
      [
        {
          "id": "allow queryables",
          "permission": "allow",
          "flows": ["ingress", "egress"],
          "messages": [
            "query",
            "reply",
            "declare_queryable"
          ],
          "key_exprs": [
            "**"
          ]
        },
        {
          "id": "allow delete",
          "permission": "allow",
          "flows": ["ingress", "egress"],
          "messages": [
            "delete",
          ],
          "key_exprs": [
            "**"
          ]
        },
      ],
      // rest of ACL config
    }
}
```

### DIFF
--- a/zenoh/src/net/routing/dispatcher/queries.rs
+++ b/zenoh/src/net/routing/dispatcher/queries.rs
@@ -718,7 +718,7 @@ pub(crate) fn route_send_response(
                         ext_tstamp,
                         ext_respid,
                     },
-                    "".to_string(), // @TODO provide the proper key expression of the response for interceptors
+                    key_expr.to_string(),
                 ));
         }
         None => tracing::warn!(
--- a/zenoh/src/net/routing/dispatcher/queries.rs
+++ b/zenoh/src/net/routing/dispatcher/queries.rs
@@ -718,7 +718,7 @@ pub(crate) fn route_send_response(
                         ext_tstamp,
                         ext_respid,
                     },
-                    key_expr.to_string(),
+                    "".to_string(), // @TODO provide the proper key expression of the response for interceptors
                 ));
         }
         None => tracing::warn!(
--- a/zenoh/src/net/routing/interceptor/access_control.rs
+++ b/zenoh/src/net/routing/interceptor/access_control.rs
@@ -244,10 +244,8 @@ impl InterceptorTrait for IngressAclEnforcer {
                     return None;
                 }
             }
-            NetworkBody::Response(Response { wire_expr, .. }) => {
-                // @TODO: Remove wire_expr usage when issue #1255 is implemented
-                if self.action(AclMessage::Reply, "Reply (ingress)", wire_expr.as_str())
-                    == Permission::Deny
+            NetworkBody::Response(Response { .. }) => {
+                if self.action(AclMessage::Reply, "Reply (ingress)", key_expr?) == Permission::Deny
                 {
                     return None;
                 }
--- a/zenoh/src/net/routing/interceptor/access_control.rs
+++ b/zenoh/src/net/routing/interceptor/access_control.rs
@@ -235,6 +235,21 @@ impl InterceptorTrait for IngressAclEnforcer {
             .or_else(|| ctx.full_expr());
 
         match &ctx.msg.body {
+            NetworkBody::Request(Request {
+                payload: RequestBody::Query(_),
+                ..
+            }) => {
+                if self.action(AclMessage::Query, "Query (ingress)", key_expr?) == Permission::Deny
+                {
+                    return None;
+                }
+            }
+            NetworkBody::Response(Response { .. }) => {
+                if self.action(AclMessage::Reply, "Reply (ingress)", key_expr?) == Permission::Deny
+                {
+                    return None;
+                }
+            }
             NetworkBody::Push(Push {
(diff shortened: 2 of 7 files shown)


oteffahi:
Requesting reviews from @Mallets @OlivierHecart @fuzzypixelz
Thanks!

zenoh/src/net/routing/interceptor/access_control.rs
```diff
@@ -278,7 +278,49 @@ impl InterceptorTrait for IngressAclEnforcer {
...
-            _ => {}
+            NetworkBody::Response(Response { .. }) => {
+                if self.action(AclMessage::Reply, "Reply (ingress)", key_expr?) == Permission::Deny
+                {
+                    return None;
+                }
+            }
+            // Unfiltered Delete messages
+            NetworkBody::Push(Push {
+                payload: PushBody::Del(_),
```
Mallets, line 289: `Delete` should be handled in the same way a `Put` is handled. We should also add it as part of the config.

Mallets: Yes, it does 👍

zenoh/src/net/routing/dispatcher/queries.rs
```diff
@@ -718,7 +718,7 @@ pub(crate) fn route_send_response(
                         ext_tstamp,
                         ext_respid,
                     },
-                    "".to_string(), // @TODO provide the proper key expression of the response for interceptors
+                    key_expr.to_string(),
```
Mallets, line 721: This allocates everytime in the critical path. Why it should return a `String` and not a `keyexpr`?

zenoh/src/net/routing/interceptor/authorization.rs
```diff
@@ -184,25 +184,31 @@ impl PermissionPolicy {
...
 
 impl ActionPolicy {
     fn action(&self, action: AclMessage) -> &PermissionPolicy {
         match action {
             AclMessage::Query => &self.query,
             AclMessage::Put => &self.put,
+            AclMessage::Delete => &self.delete,
             AclMessage::DeclareSubscriber => &self.declare_subscriber,
             AclMessage::DeclareQueryable => &self.declare_queryable,
+            AclMessage::Reply => &self.reply,
```
Mallets, line 201: From a stylistics point of view let's group `Query` and `Reply` together while matching.
Moreover, let's match first the most common cases that are `Put`, `Query`, Reply.

zenoh/src/net/routing/dispatcher/queries.rs
oteffahi: It's because `RoutingContext<Msg>` expects a `String`. In other instances where this is called no additional allocation is made, but here it is the case...
For more detail see here: https://github.com/eclipse-zenoh/zenoh/blob/main/zenoh/src/net/routing/mod.rs

Mallets: Let's then return empty `String` for the time being until a proper solution is found. 
Does returning an empty `String` impact the behaviour?

oteffahi: Returning an empty string breaks the keyexpr matching (empty string does not match with any key expression, even `**`).
Also, rules are made on key expressions, so having an empty string on all replies would impact the behavior of ACL

Mallets: Ok, understood. Can you then verify the impact on performance?

oteffahi: I'll look into it 👍

oteffahi: Due to high impact on performance, this change will be reverted. We can use `Response{ wire_expr, ... }` in ACL code to only get the keyexpr when needed.

zenoh/src/net/routing/interceptor/access_control.rs
```diff
@@ -235,6 +235,23 @@ impl InterceptorTrait for IngressAclEnforcer {
...
+            }) => {
+                if self.action(AclMessage::Query, "Query (ingress)", key_expr?) == Permission::Deny
+                {
+                    return None;
+                }
+            }
+            NetworkBody::Response(Response { wire_expr, .. }) => {
+                // @TODO: Remove wire_expr usage when issue #1255 is implemented
+                if self.action(AclMessage::Reply, "Reply (ingress)", wire_expr.as_str())
+                    == Permission::Deny
```
oteffahi, line 250: This code, and its equivalent egress version, will need to be updated once #1255 is implemented.

Mallets (approved):

PR #1260, opened 2024-07-24
make examples SHM-agnostic

yellowhatter:
I adopt examples to illustrate that SHM operation is transparent at data reception side. For example, I removed `z_sub_shm` and instead `z_sub` will print out SHM datatype for SHM buffers if it is compiled with SHM support


yellowhatter:
@Mallets

Mallets:
I believe the proposed changes obfuscate the examples rather than clarifying them... Taking out some critical Zenoh code from the example and put it in a common module (`receive_sample`) and using it directly makes the example hard to read and follow.
Moreover, changing the output format has an impact on interoperability tests, which should be updated accordingly. 
Therefore, I'm rather against merging this PR,

yellowhatter:
> I believe the proposed changes obfuscate the examples rather than clarifying them... Taking out some critical Zenoh code from the example and put it in a common module (`receive_sample`) and using it directly makes the example hard to read and follow. Moreover, changing the output format has an impact on interoperability tests, which should be updated accordingly. Therefore, I'm rather against merging this PR,

The main goal of this PR is to illustrate that Zenoh works with SHM data in transparent manner and also to reduce the number of examples by removing excessive `z_sub_shm`.

So, for readability we can remove the common code block and put it into examples directly, but the output format change is necessary as it now indicates SHM type info, that is also a subject for interoperability tests, thus worth supporting there.

Mallets:
SHM is not available in all languages, so it's a no go for the time being to modify all bindings to support something that is available only in Rust, C, and C++. Let's keep `z_sub` and `z_pub` simple and let's add anything related to `shm` into `shm` example for the time being.

Although SHM is an extremely cool feature of Zenoh, not every user nor beginner will use it. So, my advise is not to make simply things more complex. Moreover, if we plan to change basic examples, the PR should come with sister PRs that update base interoperability tests.

Issue #1257, opened 2024-07-24
Porting async_std to tokio for all plugins


evshary:
### Describe the feature

This is the umbrella issue to track the progress of porting async_std to tokio for all plugins and backends

* Zenoh (examples, storage, REST...): https://github.com/eclipse-zenoh/zenoh/pull/1264
* DDS plugin: https://github.com/eclipse-zenoh/zenoh-plugin-dds/pull/280
* ROS 2 plugin: https://github.com/eclipse-zenoh/zenoh-plugin-ros2dds/pull/195
* ROS 1 plugin: https://github.com/eclipse-zenoh/zenoh-plugin-ros1/pull/118
* MQTT plugin: https://github.com/eclipse-zenoh/zenoh-plugin-mqtt/pull/154
* webserver plugin: https://github.com/eclipse-zenoh/zenoh-plugin-webserver/pull/128
* S3 backend: https://github.com/eclipse-zenoh/zenoh-backend-s3/pull/127
* Influxdb backend: https://github.com/eclipse-zenoh/zenoh-backend-influxdb/pull/169
* rockdb backend: https://github.com/eclipse-zenoh/zenoh-backend-rocksdb/pull/155
* filesystem backend: https://github.com/eclipse-zenoh/zenoh-backend-filesystem/pull/170

PR #1264, opened 2024-07-25
Use tokio to replace async_std for plugins and backends.

evshary:
Related to https://github.com/eclipse-zenoh/zenoh/issues/1257

### DIFF
--- a/commons/zenoh-core/src/macros.rs
+++ b/commons/zenoh-core/src/macros.rs
@@ -157,7 +157,7 @@ macro_rules! to_u64 {
 }
 
 // This macro allows to spawn the right amount of threads in the
-// tokio executor
+// async_std executor
 #[macro_export]
 macro_rules! zasync_executor_init {
     () => {
--- a/commons/zenoh-core/src/macros.rs
+++ b/commons/zenoh-core/src/macros.rs
@@ -157,7 +157,7 @@ macro_rules! to_u64 {
 }
 
 // This macro allows to spawn the right amount of threads in the
-// async_std executor
+// tokio executor
 #[macro_export]
 macro_rules! zasync_executor_init {
     () => {
--- a/plugins/zenoh-plugin-rest/src/lib.rs
+++ b/plugins/zenoh-plugin-rest/src/lib.rs
@@ -17,7 +17,7 @@
 //! This crate is intended for Zenoh's internal use.
 //!
 //! [Click here for Zenoh's documentation](../zenoh/index.html)
-use std::{borrow::Cow, convert::TryFrom, str::FromStr, sync::Arc};
+use std::{borrow::Cow, convert::TryFrom, str::FromStr, sync::Arc, time::Duration};
 
 use base64::Engine;
 use futures::StreamExt;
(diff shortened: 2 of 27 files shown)


Mallets:
@wyfo do you mind reviewing this PR?

YuanYuYuan:
BTW, if we decide to use another static Tokio runtime in the plugin and consequently remove the async-std dependency in zenoh-util, we can eliminate the outdated TimedEvent since it was used before Tokio-utils was available.

Mallets:
> BTW, if we decide to use another static Tokio runtime in the plugin and consequently remove the async-std dependency in zenoh-util, we can eliminate the outdated TimedEvent since it was used before Tokio-utils was available.

We need to double check first wether `TimedEvent` is used in any plugin.

YuanYuYuan:
A quick search over our eclipse project shows we use `TimedEvent` in the zenoh-plugin-dds, zenoh-plugin-ros2dds, and zenoh-backend-influxdb. And I agree that we should open another PR dedicated to cleaning this last async-std dependence in zenoh.

plugins/zenoh-plugin-storage-manager/tests/wildcard.rs
```diff
@@ -195,6 +193,7 @@ async fn test_wild_card_in_order() {
 
 #[test]
 fn wildcard_test() {
-    task::block_on(async { test_wild_card_in_order().await });
-    // task::block_on(async { test_wild_card_out_of_order() });
+    let rt = Runtime::new().unwrap();
+    rt.block_on(async { test_wild_card_in_order().await });
+    // rt.block_on(async { test_wild_card_out_of_order() });
```
Mallets, line 198: Dead code.
```suggestion
```

commons/zenoh-core/src/macros.rs
```diff
@@ -157,7 +157,7 @@ macro_rules! to_u64 {
 }
 
 // This macro allows to spawn the right amount of threads in the
-// async_std executor
+// tokio executor
```
Mallets, line 160: The `zasync_executor_init` is specific to `async-std` and it is needed for tests to run on the CI. If we are moving to *tokio* then `zasync_executor_init` shouldn't be needed anymore. This change should be reverted and any call to `zasync_executor_init` should be removed from all the tests ported to *tokio*.

Mallets:
A broader discussion should happen on wether every plugin has its own executor or a common executor can be used when statically linking the plugins.

Mallets:
Discussion happened on Discord and there is a common agreement to go for a single runtime in case of static linking. 

To deal with *static* vs *dynamic* linking and symbols re-export `S3` backend uses the [dynamic_plugin](https://github.com/eclipse-zenoh/zenoh-backend-s3/blob/0973237a4348528b2420d8729894618ed5807273/Cargo.toml#L37) feature to selectively activate ad-hoc support for each target. We should explore that approach and to eventually make it common to all plugins.

plugins/zenoh-plugin-rest/src/lib.rs
```diff
@@ -245,8 +257,9 @@ impl Plugin for RestPlugin {
 
         let conf: Config = serde_json::from_value(plugin_conf.clone())
             .map_err(|e| zerror!("Plugin `{}` configuration error: {}", name, e))?;
-        let task = async_std::task::spawn(run(runtime.clone(), conf.clone()));
-        let task = async_std::task::block_on(task.timeout(std::time::Duration::from_millis(1)));
+        let task = TOKIO_RUNTIME.spawn(run(runtime.clone(), conf.clone()));
```
Mallets, line 260: Here we should try to get the current runtime.

plugins/zenoh-plugin-rest/src/lib.rs
```diff
@@ -245,8 +257,9 @@ impl Plugin for RestPlugin {
...
+        let task = TOKIO_RUNTIME
```
Mallets, line 261: Here we should try to get the current runtime.

Mallets (approved):

Issue #1249, opened 2024-07-22
Unexpected Failover brokering


YuanYuYuan:
### Describe the bug

Failover brokering is activated unexpectedly in the following order

1. `./target/debug/zenohd -l tcp/127.0.0.1:7447` 
2. `./target/debug/examples/z_sub -e tcp/127.0.0.1:7447 -m peer`
3. `./target/debug/examples/z_pub -e tcp/127.0.0.1:7447 -m peer`
    
Note the below order doesn’t trigger.
    
1. `./target/debug/zenohd -l tcp/127.0.0.1:7447` 
2. `./target/debug/examples/z_pub -e tcp/127.0.0.1:7447 -m peer`
3. `./target/debug/examples/z_sub -e tcp/127.0.0.1:7447 -m peer`

### System info

- zenoh: b3e42cef38f6227a87a26725d0479bcf9ace69a9

PR #1272, opened 2024-07-25
Fix failover brokering bug reacting to linkstate changes

OlivierHecart:


YuanYuYuan:
I've confirmed that this PR resolves the issue #1249. Let's merge it or wait for the fix of CI?

Mallets:
> I've confirmed that this PR resolves the issue #1249. Let's merge it or wait for the fix of CI?

Great! Tests are passing so let's not block this PR because of new Clippy lints.

PR #1275, opened 2024-07-26
Don't duplicate token undeclarations in routers

### LINKED ISSUE #1274: Liveliness token undeclaration duplication in routers
## Background

(original issue: https://github.com/ZettaScaleLabs/rmw_zenoh/issues/2.)

After https://github.com/eclipse-zenoh/zenoh/pull/865 token undeclarations started to loopback in the internal router and external routers in some cases.

### Declare, Subscribe, Undeclare

The following diagrams show the result of these ordered operations: `declare_token(K)`, `declare_subscriber(K)`, `undeclare_token(K)` in a peer. Red arrow signify messages present after #865 but not before it.

#### Without router

This change should be maintained. An application is supposed to receive Liveliness updates to tokens it declares (PUTs have always been received).

#### With router

The routers (external & internal) should not duplicate liveliness token undeclarations.

#### Without router

No change of behavior is observed in this case.

#### With router

Again, the routers (external & internal) should not duplicate liveliness token undeclarations.

## Conclusion

I think it makes sense for us to loopback token undeclarations as a liveliness subscriber has been declared. But I don't think it makes sense to loopback twice (in the internal router and in external routers).

## System info

- ref: 9408f744c

fuzzypixelz:
Fixes #1274.

### DIFF
--- a/zenoh/src/net/routing/hat/router/token.rs
+++ b/zenoh/src/net/routing/hat/router/token.rs
@@ -411,7 +411,7 @@ fn propagate_forget_simple_token(
         // undeclaration, otherwise the undeclaration would be duplicated at the source Face. In
         // cases where we don't have access to a Face as we didnt't receive an undeclaration and we
         // default to true.
-        } else if dbg!(src_face.map_or(true, |src_face| src_face.id != face.id))
+        } else if src_face.map_or(true, |src_face| src_face.id != face.id)
             && face_hat!(face).remote_interests.values().any(|(r, o)| {
                 o.tokens() && r.as_ref().map(|r| r.matches(res)).unwrap_or(true) && !o.aggregate()
             })
--- a/zenoh/src/net/routing/hat/router/token.rs
+++ b/zenoh/src/net/routing/hat/router/token.rs
@@ -386,8 +386,13 @@ fn send_forget_sourced_token_to_net_clildren(
 fn propagate_forget_simple_token(
     tables: &mut Tables,
     res: &Arc<Resource>,
+    src_face: Option<&Arc<FaceState>>,
     send_declare: &mut SendDeclare,
 ) {
+    // FIXME(fuzzypixelz): This function should be split to accommodate for the divergent code paths
+    // (the ones coming from `HatTokenTrait::undeclare_token` vs the ones coming from
+    // `HatBaseTrait::closing` and `HatBaseTrait::handle_oam`). This way we can either pass
+    // `src_face` or not, rather than using an optional.
     for mut face in tables.faces.values().cloned() {
         if let Some(id) = face_hat_mut!(&mut face).local_tokens.remove(res) {
             send_declare(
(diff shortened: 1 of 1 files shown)


zenoh/src/net/routing/hat/router/token.rs
```diff
@@ -386,8 +386,14 @@ fn send_forget_sourced_token_to_net_clildren(
 fn propagate_forget_simple_token(
     tables: &mut Tables,
     res: &Arc<Resource>,
+    src_face: Option<&Arc<FaceState>>,
     send_declare: &mut SendDeclare,
 ) {
+    dbg!(src_face.as_ref().map(|f| f.to_string()));
+    // FIXME(fuzzypixelz): This function should be split to accommodate for the divergent code paths
```
OlivierHecart, line 393: Why should we do that ? Wouldn't this lead to a lot of code duplicate? What's wrong with an Option ?
(Note: if we do so, there are a lot of other functions that have an Optional src_face parameter that should also be changed)

zenoh/src/net/routing/hat/router/token.rs
```diff
@@ -406,9 +412,26 @@ fn propagate_forget_simple_token(
...
+        //   - `HatBaseTrait::handle_oam`
+        //   - `undeclare_client_token`: when a peer/client undeclares a token and no other
+        //     peer/client has declared it (or has undeclared it) then **this** router's propagation
+        //     of the token is undeclared. In this case we don't have a source Face (there is no
+        //     "self" face).
+        //
+        // In these cases we don't have access to a Face as we don't received an undeclaration and
+        // we default to true.
+        //
+        // TODO: Could we use ZIDs instead of Faces?
```
OlivierHecart, line 429: Why would ZIDs be better ? It seems to me that comparing faces ids is faster.

zenoh/src/net/routing/hat/router/token.rs
```diff
@@ -406,9 +412,26 @@ fn propagate_forget_simple_token(
                     res.expr(),
                 ),
             );
-        } else if face_hat!(face).remote_interests.values().any(|(r, o)| {
-            o.tokens() && r.as_ref().map(|r| r.matches(res)).unwrap_or(true) && !o.aggregate()
-        }) {
+        // NOTE(fuzzypixelz): We need to check that `face` is not the source Face of the token
+        // undeclaration, otherwise the undeclaration would be duplicated at the source Face.
+        //
+        // The code path that leads to `src_face` being `None` originate in:
```
OlivierHecart, line 418: The first two lines of comment are ok to me. 
But all the rest just to explain why src_face is an option and when it is None is overkill to me. If we go that far we'll be overwhelmed with comments. And one can easily search and find places where it is called with None and why it makes sense.

OlivierHecart (changes requested):

zenoh/src/net/routing/hat/router/token.rs
fuzzypixelz: It was because of:

https://github.com/eclipse-zenoh/zenoh/blob/f7486160c01514d601186cb234f6edf49c76de63/zenoh/src/net/routing/hat/router/token.rs#L683

I thought that the face would need to stay `None` yet we don't want to send a token to the client/peer that intitiated the undeclaration.

zenoh/src/net/routing/hat/router/token.rs
fuzzypixelz: My reasoning was that we always have ZIDs, even when there is no source Face.

zenoh/src/net/routing/hat/router/token.rs
fuzzypixelz: I wrote that to explain to myself, if anything :P

OlivierHecart (approved):

PR #1277, opened 2024-07-27
Update Cargo.toml

kydos:
Updated description to be aligned with what we use everywhere else

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,7 +70,7 @@ authors = [
 edition = "2021"
 license = "EPL-2.0 OR Apache-2.0"
 categories = ["network-programming"]
-description = "Zenoh: Zero Overhead Pub/sub, Store/Query and Compute."
+description = "Zenoh: The Zero Overhead Pub/Sub/Query Protocol."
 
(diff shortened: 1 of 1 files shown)


fuzzypixelz:
Should this also be the repository's description? It currently reads:

> zenoh unifies data in motion, data in-use, data at rest and computations. It carefully blends traditional pub/sub with geo-distributed storages, queries and computations, while retaining a level of time and space efficiency that is well beyond any of the mainstream stacks.

There is also the [Eclipse Project description](https://projects.eclipse.org/projects/iot.zenoh):

> Eclipse Zenoh provides is a stack  designed to (1) minimize network overhead, (2) support extremely constrained devices, (3) supports devices with low duty-cycle by allowing the negotiation of data exchange modes and schedules, (4) provide a rich set of abstraction for distributing, querying and storing data along the entire system, and (5) provide extremely low latency and high throughput.

Mallets:
I believe it's fine to accept this PR. We would need to eventually update the rest.

Mallets (approved):

PR #1278, opened 2024-07-29
Rework `plugins_loading/search_dirs` config option

### LINKED ISSUE #1236: `zenohd`'s directory is the last searched path for plugin loading
### Describe the bug

`/example/path/to/zenohd` searches for plugins in the following order:

1. `$PWD`
2. `$HOME/.zenoh/lib`
3. `/opt/homebrew/lib`
4. `/usr/local/lib`
5. `/usr/lib`
6. **`/example/path/to`**

### To reproduce

1. `touch libzenoh_plugin_rest.dylib`
2. `cargo run --bin zenohd`

### System info

- ref: ﻿﻿9e1c4a8680a04b34f09dd2776d07958c8aae6f92

fuzzypixelz:
Fixes #1236.

See:
1. https://github.com/ZettaScaleLabs/zenoh/blob/10df70f2a3b0104c083e83f950a1f95cd800cfa4/zenohd/src/main.rs#L146-L155
2. https://github.com/ZettaScaleLabs/zenoh/blob/b59a1645a8c1563ce1c660b025781c8b3565d28b/plugins/zenoh-backend-traits/src/config.rs#L35

### DIFF
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -199,7 +199,7 @@
   //       /// Id has to be unique within the rule set
   //       "id": "rule1",
   //       "messages": [
-  //         "put", "delete", "declare_subscriber", 
+  //         "put", "delete", "declare_subscriber",
   //         "query", "reply", "declare_queryable",
   //       ],
   //       "flows":["egress","ingress"],
@@ -211,7 +211,7 @@
   //     {
   //       "id": "rule2",
   //       "messages": [
-  //         "put", "delete", "declare_subscriber", 
+  //         "put", "delete", "declare_subscriber",
   //         "query", "reply", "declare_queryable",
   //       ],
   //       "flows":["ingress"],
@@ -462,11 +462,14 @@
   ///
   //
   //  plugins_loading: {
-  //    // Enable plugins loading.
+  //    /// Enable plugins loading.
   //    enabled: false,
   //    /// Directories where plugins configured by name should be looked for. Plugins configured by __path__ are not subject to lookup.
-  //    /// If `enabled: true` and `search_dirs` is not specified then `search_dirs` falls back to the default value: ".:~/.zenoh/lib:/opt/homebrew/lib:/usr/local/lib:/usr/lib"
-  //    search_dirs: [],
+  //    /// Directories are specified as object with fields `kind` and `value` is accepted.
+  //    ///   1. If `kind` is `current_exe_parent`, then the parent of the current executable's directory is searched and `value` should be `null`.
+  //    ///   2. If `kind` is `path`, then `value` is interpreted as a filesystem path. Simply supplying a string instead of a object is equivalent to this.
+  //    /// If `enabled: true` and `search_dirs` is not specified then `search_dirs` falls back to the default value:
+  //    search_dirs: [{ "kind": "current_exe_parent" }, ".", "~/.zenoh/lib", "/opt/homebrew/lib", "/usr/local/lib", "/usr/lib"],
   //  },
   //  /// Plugins are only loaded if `plugins_loading: { enabled: true }` and present in the configuration when starting.
   //  /// Once loaded, they may react to changes in the configuration made through the zenoh instance's adminspace.
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -467,6 +467,7 @@
   //    /// Directories where plugins configured by name should be looked for. Plugins configured by __path__ are not subject to lookup.
   //    /// Directories are specified as object with fields `kind` and `value` is accepted.
   //    ///   1. If `kind` is `current_exe_parent`, then the parent of the current executable's directory is searched and `value` should be `null`.
+  //    ///      In Bash notation, `{ "kind": "current_exe_parent" }` equals `$(dirname $(which zenohd))` while `"."` equals `$PWD`.
   //    ///   2. If `kind` is `path`, then `value` is interpreted as a filesystem path. Simply supplying a string instead of a object is equivalent to this.
   //    /// If `enabled: true` and `search_dirs` is not specified then `search_dirs` falls back to the default value:
   //    search_dirs: [{ "kind": "current_exe_parent" }, ".", "~/.zenoh/lib", "/opt/homebrew/lib", "/usr/local/lib", "/usr/lib"],
--- a/commons/zenoh-util/src/lib_search_dirs.rs
+++ b/commons/zenoh-util/src/lib_search_dirs.rs
@@ -1,3 +1,16 @@
+//
+// Copyright (c) 2024 ZettaScale Technology
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+// which is available at https://www.apache.org/licenses/LICENSE-2.0.
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+//
+// Contributors:
+//   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
+//
 use std::{env, error::Error, fmt::Display, path::PathBuf, str::FromStr};
 
 use serde::{
--- a/commons/zenoh-util/src/lib_search_dirs.rs
+++ b/commons/zenoh-util/src/lib_search_dirs.rs
@@ -18,8 +18,6 @@ use serde::{
     Deserialize, Serialize,
 };
 
-use crate::lib_loader::LIB_DEFAULT_SEARCH_PATHS;
-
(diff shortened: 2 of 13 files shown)


DEFAULT_CONFIG.json5
```diff
@@ -462,11 +462,14 @@
...
+  //    /// Enable plugins loading.
   //    enabled: false,
   //    /// Directories where plugins configured by name should be looked for. Plugins configured by __path__ are not subject to lookup.
-  //    /// If `enabled: true` and `search_dirs` is not specified then `search_dirs` falls back to the default value: ".:~/.zenoh/lib:/opt/homebrew/lib:/usr/local/lib:/usr/lib"
-  //    search_dirs: [],
+  //    /// Directories are specified as object with fields `kind` and `value` is accepted.
+  //    ///   1. If `kind` is `current_exe_parent`, then the parent of the current executable's directory is searched and `value` should be `null`.
+  //    ///   2. If `kind` is `path`, then `value` is interpreted as a filesystem path. Simply supplying a string instead of a object is equivalent to this.
+  //    /// If `enabled: true` and `search_dirs` is not specified then `search_dirs` falls back to the default value:
+  //    search_dirs: [{ "kind": "current_exe_parent" }, ".", "~/.zenoh/lib", "/opt/homebrew/lib", "/usr/local/lib", "/usr/lib"],
```
Mallets, line 472: What's hence the difference from `{ "kind": "current_exe_parent" }` and `"."`?

commons/zenoh-util/src/lib_search_dirs.rs
Mallets, line 1: Add license header.

DEFAULT_CONFIG.json5
fuzzypixelz: If I'm running from a Bash shell `zenohd` then:

- `{ "kind": "current_exe_parent" }` equals `$(which zenohd)`
- `"."` equals `$PWD`

As soon as `zenohd` is not in my working directory the two are different.

Mallets: Understood, then should we be simply using something like `__exec_folder__` instead of the whole `{ "kind": "current_exe_parent" }`. In other parts of the config we are using `__<x>__` to mark reserved/special config parameters.

Mallets: Technically it is a valid possibility in JSON. But how would you express it in YAML? There is some discussion on potentially having YAML as default config in Zenoh and I'd like to avoid JSON-specific way of constructing config fields which are not so ergonomic in YAML (or any other data format we may want to support).

fuzzypixelz:
```yaml
# config.yml
plugins_loading:
  enabled: true
  search_dirs:
    - { kind: current_exe_parent }
    - .
    # Example of a custom field, equivalent to '~/source/zenoh/target/debug'
    - { kind: path, value: ~/source/zenoh/target/debug }
    - ~/.zenoh/lib
    - /opt/homebrew/lib
    - /usr/local/lib
    - /usr/local
```

`yq -o=json < config.yml` yields:

```json
{
  "plugins_loading": {
    "enabled": true,
    "search_dirs": [
      {
        "kind": "current_exe_parent"
      },
      ".",
      {
        "kind": "path",
        "value": "~/source/zenoh/target/debug"
      },
      "~/.zenoh/lib",
      "/opt/homebrew/lib",
      "/usr/local/lib",
      "/usr/local"
    ]
  }
}
```

Even better, [YAML version 1.2 is a superset of JSON](https://en.wikipedia.org/wiki/YAML).

Mallets: Having something like `- { kind: path, value: ~/source/zenoh/target/debug }` looks pretty weird in YAML to be honest. I would expect to have something:
```yaml
# config.yml
plugins_loading:
  enabled: true
  search_dirs:
    - .
    - ~/foo
    # equivalent to
    - path: .
    - path: ~/foo
    # getting information from the environment
    - env: current_exec_folder
```

Which yields the following JSON: 
```json
{
    "plugins_loading": {
        "enabled": true,
        "search_dirs": [
            {
                "path": "."
            },
            {
                "path": "~/foo"
            },
            {
                "env": "current_exec_folder"
            },
            ".",
            "~/foo"
        ]
    }
}
```

fuzzypixelz: In any case, the user is free not to use them, at the cost of an extra line in the config.

However, the `{ k: v }` syntax or "inline blocks" have been there since YAML 1.0. [You find them 6 examples into the specification.](https://yaml.org/spec/1.0/#id2489726)

They're useful when an object is small enough to span one line:

```yaml
build:
  - { target: x86_64-unknown-linux-gnu, os: ubuntu-20.04, cross: false }
  - { target: x86_64-unknown-linux-musl, os: ubuntu-20.04, cross: true }
  - { target: arm-unknown-linux-gnueabi, os: ubuntu-20.04, cross: true }
  - { target: arm-unknown-linux-gnueabihf, os: ubuntu-20.04, cross: true }
  - { target: armv7-unknown-linux-gnueabihf, os: ubuntu-20.04, cross: true }
  - { target: aarch64-unknown-linux-gnu, os: ubuntu-20.04, cross: true }
  - { target: aarch64-unknown-linux-musl, os: ubuntu-20.04, cross: true }
  - { target: x86_64-apple-darwin, os: macos-12, cross: false }
  - { target: aarch64-apple-darwin, os: macos-12, cross: false }
  - { target: x86_64-pc-windows-msvc, os: windows-2019, cross: false }
  - { target: x86_64-pc-windows-gnu, os: windows-2019, cross: false }
```

In the above example, without inline blocks the section would be 3x times bigger, going from 11 lines to 33 (not to mention the extra indentation), for no good reason.

They are syntactically very similar to inline arrays (e.g. `[x, y, z]`). Although inline arrays are more common than inline blocks in small configurations.

Mallets: Good point. The very last thing I'd simply add the equivalence of paths in the comment in the config. I.e.:
```
{ "kind": "current_exe_parent" } equals basename $(which zenohd)
"." equals $PWD
```

commons/zenoh-util/src/lib_search_dirs.rs
```diff
@@ -0,0 +1,230 @@
...
+    fn into_iter(self) -> Self::IntoIter {
+        IntoIter {
+            iter: self.0.into_iter(),
+        }
+    }
+}
+
+impl Default for LibSearchDirs {
+    fn default() -> Self {
+        let de = &mut serde_json::Deserializer::from_str(&LIB_DEFAULT_SEARCH_PATHS);
```
Mallets, line 94: Here we are assuming `JSON` format which we shouldn't.

fuzzypixelz: Why shouldn't we assume JSON? If not JSON then something else will be assumed. The preview colon-separated list was also an assumption inspired by path separators in Unix (Windows uses semicolons).

Before it was a colon-separated list of strings. I changed it to JSON to re-use the `Deserialize` impl and to support the new `{ "kind": "...", "value": "..." }` syntax. We can switch it to YAML or TOML.

If it weren't for the fact that this is a compile-time configuration option, I could simply do away with this deserialization and use a `const`.

gabrik: I do not like to use `serde_json` specific things as it imposes a format.

The config per-se is a DSL, the representation must not matter so I would avoid any format-specific methods.

fuzzypixelz: > I do not like to use `serde_json` specific things as it imposes a format.
> 
> The config per-se is a DSL, the representation must not matter so I would avoid any format-specific methods.

`LIB_DEFAULT_SEARCH_PATHS` is a `const` `String` computed at compile time from an environment variable (see `zconfigurable`). It has nothing to do with Zenoh configuration.

fuzzypixelz: Here's the relevant snippet:

https://github.com/ZettaScaleLabs/zenoh/blob/f7eed49016c112c59304338d01a183a58bd5fca2/commons/zenoh-util/src/lib_loader.rs#L28-L42

gabrik: Nvm, I'm wrong, what I meant is that this environment variable HAS to be in JSON format, but do we really need it?

Can't just be a const like you mentioned in the first comment?

fuzzypixelz: > Nvm, I'm wrong, what I meant is that this environment variable HAS to be in JSON format, but do we really need it?
> 
> Can't just be a const like you mentioned in the first comment?

Yes, it can be `const` if no one needs to override it at build time...

gabrik: That's what I'm questioning it there a use-case when it is needed to override?

Mallets (approved):

PR #1281, opened 2024-07-30
Shm segment cleanup

yellowhatter:


commons/zenoh-shm/src/api/client_storage/mod.rs
```diff
@@ -31,17 +31,16 @@ use crate::{
...
-    pub static ref GLOBAL_CLIENT_STORAGE: Arc<ShmClientStorage> = Arc::new(
-        ShmClientStorage::builder()
-            .with_default_client_set()
-            .build()
-    );
-}
+#[dynamic(lazy, drop)]
+/// A global lazily-initialized SHM client storage.
+/// When initialized, contains default client set,
+/// see ShmClientStorage::with_default_client_set
```
Mallets, line 37: Nitpick here. If you want to reference the actual function/type in the doc you can use markdown-like syntax `[`with_default_client_set`](ShmClientStorage::with_default_client_set)`

commons/zenoh-shm/src/cleanup.rs
```diff
@@ -0,0 +1,48 @@
...
+use static_init::dynamic;
+
+/// A global cleanup, that is guaranteed to be dropped at normal program exit and that will
+/// execute all registered cleanup routines at this moment
+#[dynamic(lazy, drop)]
+pub(crate) static mut CLEANUP: Cleanup = Cleanup::new();
+
+/// An RAII object that calls all registered routines upon destruction
+pub(crate) struct Cleanup {
+    cleanups: lockfree::queue::Queue<Option<Box<dyn FnOnce() + Send>>>,
```
Mallets, line 24: It's not clear to me why the `Queue` needs to contain `Option<Box<...>>` and not simply `Box<...>` if  in the `Drop` implementation you say that `// SAFETY: this is safe as cleanup will never have None elements`.

commons/zenoh-shm/src/watchdog/validator.rs
```diff
@@ -14,16 +14,15 @@
...
 
 pub(super) type InvalidateCallback = Box<dyn Fn() + Send>;
 
-lazy_static! {
-    pub static ref GLOBAL_VALIDATOR: WatchdogValidator =
-        WatchdogValidator::new(Duration::from_millis(100));
-}
+#[dynamic(lazy, drop)]
+pub static mut GLOBAL_VALIDATOR: WatchdogValidator =
+    WatchdogValidator::new(Duration::from_millis(100));
```
Mallets, line 25: Not a comment on this PR but rather on future evolution. The duration the `WatchdogValidator` operates on should be configurable.

yellowhatter: Yes, I have this in plan. There are some tricky points with adding this which require careful design

commons/zenoh-shm/src/cleanup.rs
yellowhatter: lockfree data structure under the hood uses `core::mem::uninitialized` that states:
`// SAFETY: the caller must guarantee that an uninitialized value is valid for `T`.`
and using `Box<...>` I get runtime panic in `uninitialized` at this line:
`intrinsics::assert_mem_uninitialized_valid::<T>();`

commons/zenoh-shm/src/api/client_storage/mod.rs
yellowhatter: fixed

yellowhatter:
@Mallets all done, please see my replies

commons/zenoh-shm/src/cleanup.rs
Mallets: Then it's worth adding a comment that also help justifying the `unsafe` in the `Drop`.

commons/zenoh-shm/src/cleanup.rs
```diff
@@ -0,0 +1,48 @@
...
+    pub(crate) fn register_cleanup(&self, cleanup_fn: Box<dyn FnOnce() + Send>) {
+        self.cleanups.push(Some(cleanup_fn));
+    }
+}
+
+impl Drop for Cleanup {
+    fn drop(&mut self) {
+        while let Some(cleanup) = self.cleanups.pop() {
+            // SAFETY: this is safe as cleanup will never have None elements
+            unsafe {
```
Mallets, line 43: At a second look, I think that in any case the `unsafe` can be removed with a simple `if let Some(f) = cleanup { f() }`.

yellowhatter: I'm pretty much sure that rustc will not be able to elide branching instruction here, but this is not a hotpath, so let's remove it for the sake of safety and clean code :)

yellowhatter:
unicast_authenticator test failed because of some port collision, not related to this PR

Mallets (approved):

PR #1283, opened 2024-07-30
Fix failing Liveliness Subscriber Undeclaration

### LINKED ISSUE #1282: Failure to undeclare Liveliness Subscribers
### Describe the bug

https://github.com/eclipse-zenoh/zenoh/blob/502d3be20b78f2873ec069cbcd5fefd308ecf5c7/zenoh/src/api/session.rs#L1239-L1243

Only non-Liveliness Subscribers are checked.

### To reproduce

1. Declare a Liveliness Subscriber
2. Undeclare it.

### System info

- ref: 502d3be20b78f2873ec069cbcd5fefd308ecf5c7

fuzzypixelz:
Fixes #1282.

### DIFF
--- a/zenoh-ext/tests/liveliness.rs
+++ b/zenoh-ext/tests/liveliness.rs
@@ -81,18 +81,18 @@ async fn test_liveliness_querying_subscriber_clique() {
     assert_eq!(sample.kind(), SampleKind::Put);
     assert_eq!(sample.key_expr().as_str(), LIVELINESS_KEYEXPR_2);
 
-    assert!(token1.undeclare().await.is_ok());
+    token1.undeclare().await.unwrap();
     tokio::time::sleep(SLEEP).await;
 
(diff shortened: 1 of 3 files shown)


zenoh-ext/tests/liveliness.rs
```diff
@@ -179,12 +185,20 @@ async fn test_liveliness_querying_subscriber_brokered() {
     assert_eq!(sample.kind(), SampleKind::Put);
     assert_eq!(sample.key_expr().as_str(), LIVELINESS_KEYEXPR_2);
 
-    drop(token1);
+    assert!(token1.undeclare().await.is_ok());
```
Mallets, line 188: Doing an `unwrap()` is preferable in this case since the error will be printed out.
If the test fails in the CI then it's hard to debug what was the error that caused the failure.
Same comment applies to the rest of `assert!(x.is_ok())`

Mallets (approved):

PR #1286, opened 2024-07-31
Rename `close` to `undeclare` in `zenoh-ext`

### LINKED ISSUE #1285: `PublicationCache::close` and `FetchSubscriber::close` should be named `undeclare`
### Describe the feature

This is to align the naming convention with the `zenoh` crate.

fuzzypixelz:
Fixes #1285.

### DIFF
--- a/zenoh-ext/src/publication_cache.rs
+++ b/zenoh-ext/src/publication_cache.rs
@@ -256,9 +256,9 @@ impl<'a> PublicationCache<'a> {
         })
     }
 
-    /// Close this PublicationCache
+    /// Undeclare this [`PublicationCache`]`.
     #[inline]
-    pub fn close(self) -> impl Resolve<ZResult<()>> + 'a {
+    pub fn undeclare(self) -> impl Resolve<ZResult<()>> + 'a {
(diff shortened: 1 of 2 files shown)


Mallets (approved):
Before merging it is necessary to check wether these functions are used in `zenoh-c`.
If yes, then a sister PR should be open.

fuzzypixelz:
> Before merging it is necessary to check wether these functions are used in `zenoh-c`. If yes, then a sister PR should be open.

You're right, both are used. But we need this merged first before I can reference the merge commit in the lockfile.

Mallets:
> > Before merging it is necessary to check wether these functions are used in `zenoh-c`. If yes, then a sister PR should be open.
> 
> You're right, both are used. But we need this merged first before I can reference the merge commit in the lockfile.

The good way to proceed is to minimize the misalignment window between repos as much as possible. It would be good to have the CI running already on zenoh-c pointing to this branch. If the CI in zenoh-c is passing then we can quickly merge this PR and point zenoh-c's PR to `dev/1.0.0`.

fuzzypixelz:
@Mallets Ready to merge.

PR #1287, opened 2024-07-31
fix: invalid mime crash on REST plugin

gabrik:
Fixing a crash when receiving an HTTP GET with `_raw=true` and no data received from the Zenoh queryables.

This leads to an `""` (empty) Mime type that is not correctly converted.


Charles-Schleich:
LGTM.

Mallets (approved):

PR #1290, opened 2024-08-01
Zslice kind

wyfo:
Requires #1289 before being merged

### DIFF
--- a/commons/zenoh-buffers/src/zslice.rs
+++ b/commons/zenoh-buffers/src/zslice.rs
@@ -14,16 +14,15 @@
 use alloc::{boxed::Box, sync::Arc, vec::Vec};
 use core::{
     any::Any,
-    convert::AsRef,
-    fmt,
+    fmt, iter,
     num::NonZeroUsize,
-    ops::{Deref, Index, Range, RangeFrom, RangeFull, RangeInclusive, RangeTo, RangeToInclusive},
-    option,
+    ops::{Bound, Deref, RangeBounds},
 };
 
 use crate::{
     buffer::{Buffer, SplitBuffer},
     reader::{BacktrackableReader, DidntRead, HasReader, Reader},
+    writer::{BacktrackableWriter, DidntWrite, Writer},
 };
 
 /*************************************/
@@ -37,7 +36,7 @@ pub trait ZSliceBuffer: Any + Send + Sync + fmt::Debug {
 
 impl ZSliceBuffer for Vec<u8> {
     fn as_slice(&self) -> &[u8] {
-        self.as_ref()
+        self
     }
 
     fn as_any(&self) -> &dyn Any {
@@ -51,7 +50,7 @@ impl ZSliceBuffer for Vec<u8> {
 
 impl ZSliceBuffer for Box<[u8]> {
     fn as_slice(&self) -> &[u8] {
-        self.as_ref()
+        self
     }
 
     fn as_any(&self) -> &dyn Any {
@@ -65,7 +64,7 @@ impl ZSliceBuffer for Box<[u8]> {
 
 impl<const N: usize> ZSliceBuffer for [u8; N] {
     fn as_slice(&self) -> &[u8] {
-        self.as_ref()
+        self
     }
 
     fn as_any(&self) -> &dyn Any {
@@ -91,9 +90,9 @@ pub enum ZSliceKind {
 /// A clonable wrapper to a contiguous slice of bytes.
 #[derive(Clone)]
 pub struct ZSlice {
-    pub(crate) buf: Arc<dyn ZSliceBuffer>,
-    pub(crate) start: usize,
-    pub(crate) end: usize,
+    buf: Arc<dyn ZSliceBuffer>,
+    start: usize,
+    end: usize,
     #[cfg(feature = "shared-memory")]
     pub kind: ZSliceKind,
 }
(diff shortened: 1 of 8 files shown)


wyfo:
@Mallets @yellowhatter

wyfo:
I've run throughput example, and performances after this PR are better, as expected.

commons/zenoh-buffers/src/zslice.rs
```diff
@@ -211,65 +201,9 @@ impl AsRef<[u8]> for ZSlice {
     }
 }
 
-impl Index<usize> for ZSlice {
```
Mallets, line 214: Why removing all these impls?

commons/zenoh-buffers/src/zslice.rs
```diff
@@ -108,58 +97,52 @@ impl ZSlice {
...
+    ///
+    /// Buffer modification must not modify slice range.
     #[inline]
     #[must_use]
-    pub fn downcast_mut<T>(&mut self) -> Option<&mut T>
-    where
-        T: Any,
-    {
-        Arc::get_mut(&mut self.buf).and_then(|val| val.as_any_mut().downcast_mut::<T>())
+    pub unsafe fn downcast_mut<T: Any>(&mut self) -> Option<&mut T> {
```
Mallets, line 129: Why marking this function as `unsafe`? `Arc::get_mut` should already guarantee that reference count is 1 and it is safe to mutate.

commons/zenoh-buffers/src/zslice.rs
wyfo: (This change is brought by #1289)
Because they are not needed. Rust indexing syntax already includes an auto-deref. I know that some standard collections like `Vec` reimplement `Index` (by just manually doing the deref), but I suspect it may comes from a time where there was no auto-deref.
On the other hand, `bytes::Bytes` doesn't implement `Index`.

commons/zenoh-buffers/src/zslice.rs
wyfo: (This change is brought by https://github.com/eclipse-zenoh/zenoh/pull/1289 and should be discussed there)
The answer is in the safety comment. It may be safe to mutate the inner buffer inside the Arc, but `ZSlice` still have an invariant to maintain. If the function was safe, you would be able to truncate the vector, and the `end` would then be invalid. Thus, we need to ensure that the invariant is maintained, i.e. `start`/`end` remain valid for the modified slice. The simple precondition for that is to have the slice size unmodified, as it is the case for SHM buffers which are the only use case of this method.

commons/zenoh-codec/src/core/zbuf.rs
```diff
@@ -164,9 +165,8 @@ mod shm {
                                     reader.read_zslices(len, |s| zbuf.push_zslice(s))?;
                                 }
                                 SHM_PTR => {
-                                    let mut zslice: ZSlice = self.codec.read(&mut *reader)?;
-                                    zslice.kind = ZSliceKind::ShmPtr;
-                                    zbuf.push_zslice(zslice);
+                                    let vec: Vec<u8> = self.codec.read(&mut *reader)?;
```
Mallets, line 168: We should read a `ZSlice` here and not a `Vec<u8>` which always causes an allocation and a copy.
This is something that should be avoided as much as possible. 
In short, while this PR reduces the size on the stack of the `ZSlice`, it introduces an allocation for every `ShmBufInfoBuffer`.

commons/zenoh-shm/src/lib.rs
```diff
@@ -228,3 +228,20 @@ impl ZSliceBuffer for ShmBufInner {
         self
     }
 }
+
+#[derive(Debug)]
+pub struct ShmBufInfoBuffer(pub Vec<u8>);
```
Mallets, line 233: I believe `ShmBufInfoBuffer` can simply be a wrapper around a `ZSlice`?

wyfo: You cannot use a wrapper around ZSlice, otherwise you would not be able to use it in `ZBuf`.

commons/zenoh-codec/src/core/zbuf.rs
wyfo: Indeed, forget about that again... I'm not sure there is a simple way to handle that. In fact the issue is our two-phase deserialization, first into a `ZSlice` containing a `ShmBufInfo`, then into  a ZSlice containing a SHM buffer.
Why don't we do both at once?

wyfo:
As this PR is neither urgent nor important, it can be postponed or closed because my implementation isn't good indeed.

commons/zenoh-codec/src/core/zbuf.rs
Mallets: We can definitely define a specific codec (e.g. `ZenohZSliceSHM`) to directly decode a `ZSlice` containing a  `ShmBufInfoBuffer`. The code would look like something like:

```rust
{
   ...
   SHM_PTR => {
      let shm_codec = ZenohZSliceSHM::new();
      let zslice: ZSlice = shm_codec.read(&mut *reader)?;
      zbuf.push_zslice(zslice);
   }
}

impl<R> RCodec<ZSlice, &mut R> for ZenohZSliceSHM
where
    R: Reader,
{
   // Do everything all-in-one
}
```

PR #1291, opened 2024-08-02
reliability propagation

wyfo:


OlivierHecart:
It has been decided to remove reliability from subscriber side.

PR #1294, opened 2024-08-02
Hide `UndeclarableInner::undeclare_inner`

fuzzypixelz:
The aim of this pull request is to remove `UndeclarableInner::undeclare_inner` from the public API.

### DIFF
--- a/zenoh/src/api/session.rs
+++ b/zenoh/src/api/session.rs
@@ -449,17 +449,17 @@ impl fmt::Debug for SessionRef<'_> {
     }
 }
 
-pub(crate) trait UndeclarableInner<S, O, T = ZResult<()>>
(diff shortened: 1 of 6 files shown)


zenoh/src/api/session.rs
```diff
@@ -449,24 +449,40 @@ impl fmt::Debug for SessionRef<'_> {
     }
 }
 
-/// A trait implemented by types that can be undeclared.
-pub trait Undeclarable<S, O, T = ZResult<()>>
+pub(crate) trait UndeclarableInner<S, O, T = ZResult<()>>
```
Mallets, line 452: I would call it `UndeclarableSealed` which is make it clear that follows the *sealed trait* pattern.

Mallets (approved):

PR #1295, opened 2024-08-05
feat: don't undeclare objects on drop

wyfo:


wyfo:
It was not possible to remove the internal pool of the session in zenoh-python, because publishers still need an `Arc`, not a `Weak`.

wyfo:
@Mallets @kydos

wyfo:
Closed in favor of #1334

PR #1301, opened 2024-08-07
Add append method to ZByteWriter

Mallets:
This PR enables to append a `ZBytes` in the `ZBytesWriter` by taking ownership.
This allows to compose a `ZBytes` out of multiple `ZBytes` that may point to different memory regions.
Said in other terms, it allows to create a linear view on different memory regions without copy.
This is pretty useful when writing by removing additional copies.


Mallets:
@DenisBiryukov91 can you please review this PR?

DenisBiryukov91 (approved):

PR #1312, opened 2024-08-14
make SHM enabled by default in Config

yellowhatter:


yellowhatter:
@Mallets SHM now enabled in the Config
(CI failure is `transport_tcp_concurrent`  on Windows and not related to SHM)

Mallets:
I would add the description

>  // A probing procedure for shared memory is performed upon session opening. To enable `z_pub_shm` to operate
> // over shared memory (and to not fallback on network mode), shared memory needs to be enabled also on the
> // subscriber side. By doing so, the probing procedure will succeed and shared memory will operate as expected.

that is present in the examples to the `shared_memory` section in the `DEFAULT_CONFIG` to not loose the information.

yellowhatter:
> I would add the description
> 
> > // A probing procedure for shared memory is performed upon session opening. To enable `z_pub_shm` to operate
> > // over shared memory (and to not fallback on network mode), shared memory needs to be enabled also on the
> > // subscriber side. By doing so, the probing procedure will succeed and shared memory will operate as expected.
> 
> that is present in the examples to the `shared_memory` section in the `DEFAULT_CONFIG` to not lose the information.

done

zenoh/tests/shm.rs
```diff
@@ -128,6 +125,7 @@ async fn test_session_pubsub(peer01: &Session, peer02: &Session, reliability: Re
             .declare_subscriber(&key_expr)
             .callback(move |sample| {
                 assert_eq!(sample.payload().len(), size);
+                assert!(sample.payload().deserialize::<&zshm>().is_ok());
```
Mallets, line 128: In this particular case it won't make much difference but in the tests is better to use `unwrap()` since upon failure the actual error will be printed out.

yellowhatter: Agree, fixed

Mallets (approved):

Issue #1320, opened 2024-08-21
Liveliness token are mistakenly undeclared


YuanYuYuan:
### Describe the bug

A local liveliness token undeclaration will trigger the other token undeclaration in a chain reaction.

### To reproduce

```rust
use zenoh::{prelude::*, Config};
use zenoh::sample::SampleKind;

#[tokio::main]
async fn main() {
    // Initiate logging
    zenoh::try_init_log_from_env();

    println!("Opening session...");
    let session = zenoh::open(Config::default()).await.unwrap();
    let subscriber = session
        .liveliness()
        // .declare_subscriber("@ros2_lv/0/**")
        .declare_subscriber("**")
        .callback(|sample| {
            match sample.kind() {
                SampleKind::Put => println!(
                    ">> [LivelinessSubscriber] New alive token ('{}')",
                    sample.key_expr().as_str()
                ),
                SampleKind::Delete => println!(
                    ">> [LivelinessSubscriber] Dropped token ('{}')",
                    sample.key_expr().as_str()
                ),
            }
        })
        .await
        .unwrap();

    // Create a node
    println!("Create a node");
    let node_token = session.liveliness().declare_token(format!("node")).await.unwrap();

    // Create a publisher
    println!("Create a publisher");
    let pub_token = session.liveliness().declare_token(format!("pub")).await.unwrap();

    // Create subscriber
    println!("Create a subscriber");
    let sub_token = session.liveliness().declare_token(format!("sub")).await.unwrap();

    // Destroy a publisher
    println!("Destroy a publisher");
    pub_token.undeclare().await.unwrap();

    println!("Press CTRL-C to undeclare LivelinessToken and quit...");
    std::thread::park();

    sub_token.undeclare().await.unwrap();
    node_token.undeclare().await.unwrap();
}
```

```console
Opening session...
2024-08-21T14:44:26.656773Z  INFO main ThreadId(01) zenoh::net::runtime: Using ZID: 49758512d68759c06b8cce305e60b6ef
2024-08-21T14:44:26.658523Z  INFO main ThreadId(01) zenoh::net::runtime::orchestrator: Zenoh can be reached at: tcp/[fe80::694a:ae45:bf2b:33a0]:33891
2024-08-21T14:44:26.658551Z  INFO main ThreadId(01) zenoh::net::runtime::orchestrator: Zenoh can be reached at: tcp/192.168.1.140:33891
[...]

PR #1321, opened 2024-08-21
Fix liveliness bug

OlivierHecart:


YuanYuYuan:
Confirmed that the issue #1320 and the related bug in rmw_zenoh_cpp have been resolved by this PR.

PR #1324, opened 2024-08-26
Select links based on message Priority

fuzzypixelz:
The goal of this pull request is to enable users to choose the link on which a message is sent as a function of the message's priority.

### DIFF
--- a/io/zenoh-transport/src/unicast/establishment/accept.rs
+++ b/io/zenoh-transport/src/unicast/establishment/accept.rs
@@ -20,9 +20,9 @@ use zenoh_buffers::{reader::HasReader, writer::HasWriter, ZSlice};
 use zenoh_codec::{RCodec, WCodec, Zenoh080};
 use zenoh_core::{zasynclock, zcondfeat, zerror};
 use zenoh_crypto::{BlockCipher, PseudoRng};
-use zenoh_link::LinkUnicast;
+use zenoh_link::{EndPoint, LinkUnicast};
 use zenoh_protocol::{
-    core::{Field, Resolution, WhatAmI, ZenohIdProto},
+    core::{Field, Reliability, Resolution, WhatAmI, ZenohIdProto},
     transport::{
         batch_size,
         close::{self, Close},
@@ -55,7 +55,7 @@ pub(super) type AcceptError = (zenoh_result::Error, Option<u8>);
 struct StateTransport {
     batch_size: BatchSize,
     resolution: Resolution,
-    ext_qos: ext::qos::StateAccept,
+    ext_qos: ext::qos::QoS,
     #[cfg(feature = "transport_multilink")]
     ext_mlink: ext::multilink::StateAccept,
     #[cfg(feature = "shared-memory")]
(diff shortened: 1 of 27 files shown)


io/zenoh-transport/src/unicast/universal/tx.rs
```diff
@@ -11,51 +11,56 @@
...
 
-        // No best match found, take the first available link
-        if let Some(pl) = guard.iter().map(|tl| &tl.pipeline).next() {
-            zpush!(guard, pl, msg);
-        }
+        let Some(transport_link) = transport_links
+            .iter()
+            .find(|transport_link| {
+                transport_link.link.config.reliability == msg_reliability
+                    && transport_link.link.config.priority == Some(msg.priority())
```
Mallets, line 34: This is a strict comparison, I believe a better strategy would be finding the link with closest priority matching. 
Moreover, there are too many calls to `find` which iterates over the list of the links every time. Wouldn't be better to just iterate once?

io/zenoh-transport/src/unicast/establishment/accept.rs
```diff
@@ -637,16 +639,19 @@ impl<'a, 'b: 'a> AcceptFsm for &'a mut AcceptLink<'b> {
...
-        direction: TransportLinkUnicastDirection::Inbound,
+        direction,
         batch: BatchConfig {
             mtu,
             is_streamed,
             #[cfg(feature = "transport_compression")]
             is_compression: false,
         },
+        reliability: link_reliability(&link),
+        priority: link_priority(&link, direction)?,
```
Mallets, line 654: I believe this won't work since I'm not sure the metadata are correctly set/stored in the link.

io/zenoh-transport/src/unicast/establishment/mod.rs
```diff
@@ -116,3 +119,33 @@ pub(super) fn compute_sn(
...
+    }
+}
+
+pub(super) fn link_priority(
+    link: &LinkUnicast,
+    direction: TransportLinkUnicastDirection,
+) -> ZResult<Option<Priority>> {
+    const PRIORITY_METADATA_KEY: &str = "priority";
+
+    let link_locator = match direction {
```
Mallets, line 137: I believe setting the right priority would need both `src` and `dst` and find the min matching set. Moreover, I would assume that an extension to `Init/Open` should be added to communicate to the remote peer the actual `QoS` parameters should be associated with the link.

io/zenoh-transport/src/unicast/universal/tx.rs
fuzzypixelz: What do you mean by "closest priority matching"? If the message data specified priority 3 and we have two linked tagged with priorities 2 and 4, then which link do we choose?

Rounding down will mean that the links with lower priorities will get more of the messages for which there is no exact match. And vice-versa for rounding up.

I'll remove the extra loops.

io/zenoh-transport/src/unicast/establishment/mod.rs
fuzzypixelz: What do you mean by "the min matching set"? The minimum of the `src` and `dst` priorities? If so then can't both peers agree on the priority tag of a link without exchanging further data?

io/zenoh-transport/src/unicast/establishment/accept.rs
fuzzypixelz: I'm assuming this comment is referring to the next comment.

io/zenoh-transport/src/unicast/universal/tx.rs
Mallets: I'll go with rounding-up. It's less problematic impacting data with lower priority than higher priority.

io/zenoh-transport/src/unicast/establishment/mod.rs
Mallets: To agree on the priority an extension to `Init/Open` should be added.

fuzzypixelz: Priority ranges are now exchanged in `INIT` messages. Right now priorities are only applied if both sides agree on the exact range. I don't see a clear way unifying the two so I leave it for discussion:

https://github.com/fuzzypixelz/zenoh/blob/90b068f11293328da3dfffe59fa37fbed432d0ec/io/zenoh-transport/src/unicast/establishment/ext/qos.rs#L127-L135

io/zenoh-transport/src/unicast/establishment/accept.rs
fuzzypixelz: Priority ranges are not obtained from `INIT` messages using the `QoS` extension.

io/zenoh-transport/src/unicast/universal/tx.rs
fuzzypixelz: As we're using ranges now, I interpreted "closest priority matching" to mean belonging to the smallest range available. Please let me know what you think about that.

io/zenoh-transport/src/unicast/establishment/mod.rs
fuzzypixelz: Also, endpoints are now recovered from the the "new link" channel and not from the `.get_src()` and `.get_dst()` methods.

Issue #1292, opened 2024-08-02
zenohd default config error


Sobczaczek:
### Describe the bug

While running zenohd with default (DEFAULT_CONFIG.json5) config -> error:
INFO main ThreadId(01) zenohd: zenohd v1.0.0-alpha.5 built with rustc 1.75.0 (82e1608df 2023-12-21)
thread 'main' panicked at zenohd/src/main.rs:122:42:
called `Result::unwrap()` on an `Err` value: JSON error: invalid type: string "", expected a list of whatami variants ('router', 'peer', 'client') at commons/zenoh-config/src/lib.rs:748.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

### To reproduce

1. zenohd -c DEFAULT_CONFIG.json5 

### System info

- Ubuntu 24.04

anhaabaete:
Please, test it.

https://github.com/eclipse-zenoh/zenoh/compare/main...anhaabaete:zenoh:patch-1

it work to me.

milyin:
The approach used for fixing this seems to be wrong, we need to correct DEFAULT_CONFIG or provide default value for this parameter. See discussion in https://github.com/eclipse-zenoh/zenoh/pull/1326
Going to revert the fix

PR #1326, opened 2024-08-27
Merge main into dev/1.0.0

milyin:


zenohd/src/main.rs
```diff
@@ -120,7 +120,11 @@ fn config_from_args(args: &Args) -> Config {
         .config
         .as_ref()
         .map_or_else(Config::default, |conf_file| {
-            Config::from_file(conf_file).unwrap()
+            Config::from_file(conf_file).unwrap_or_else(|e| {
+                // if file load fail, wanning it, and load default config
+                tracing::warn!("Warn: File {} not found! {}", conf_file, e.to_string());
+                Config::default()
```
Mallets, line 126: If config file is provided as CLI arg and the file does not exist, I believe we should just return an error and exit. I don't think it's a good a idea to silently fallback to a default configuration here.

Mallets: Not only `DEFAULT_CONFIG.json5`, I believe zenohd should refuse to start if wrong parameters are passed and not to silently fallback. The user may not even see the traficing warn if zenohd is run in a container. But it's very apparent if zenohd does not start at all.

milyin: Reverted the fix, reopened issue https://github.com/eclipse-zenoh/zenoh/issues/1292

Mallets (approved):

PR #1334, opened 2024-08-28
feat!: bind callback subscriber/queryable to session lifetime

wyfo:
To determine if the entity is callback-only, the only elegant way I've found is the rule "handler is ZST means callback-only". Unless users starts writing fancy implementations, it should be correct 100% of the time.

Session entities now uses weak references, except publishers because it would impact performances. Weak references also solves the issue of mass undeclarations before closing the session (when the session is an `Arc`), except for publishers.

`Undeclarable` trait has been refactored a little bit to better match its use in the code.

### DIFF
--- a/zenoh/src/api/key_expr.rs
+++ b/zenoh/src/api/key_expr.rs
@@ -551,7 +551,7 @@ impl<'a> KeyExpr<'a> {
 
 impl<'a> UndeclarableSealed<&'a Session> for KeyExpr<'a> {
     type Undeclaration = KeyExprUndeclaration<'a>;
-    
+
     fn undeclare_inner(self, session: &'a Session) -> Self::Undeclaration {
         KeyExprUndeclaration {
             session,
--- a/zenoh/src/api/key_expr.rs
+++ b/zenoh/src/api/key_expr.rs
@@ -551,6 +551,7 @@ impl<'a> KeyExpr<'a> {
 
 impl<'a> UndeclarableSealed<&'a Session> for KeyExpr<'a> {
     type Undeclaration = KeyExprUndeclaration<'a>;
+    
     fn undeclare_inner(self, session: &'a Session) -> Self::Undeclaration {
         KeyExprUndeclaration {
             session,
--- a/zenoh/src/api/key_expr.rs
+++ b/zenoh/src/api/key_expr.rs
@@ -549,8 +549,9 @@ impl<'a> KeyExpr<'a> {
     }
 }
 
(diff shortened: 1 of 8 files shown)


wyfo:
Replaces #1295 
@Mallets

wyfo:
@kydos

wyfo:
Idk why the test failed... ¯\_(ツ)_/¯

zenoh/src/api/queryable.rs
```diff
@@ -534,43 +538,13 @@ impl fmt::Debug for QueryableState {
     }
 }
 
-/// An entity able to reply to queries through a callback.
```
Mallets, line 537: Even if the code has changed, I believe the doc is still relevant when pointing to the right type and with an appropriate description.

wyfo: What would you write here?

zenoh/src/api/liveliness.rs
```diff
@@ -578,11 +575,13 @@ where
...
-                subscriber: SubscriberInner {
-                    session,
+                inner: SubscriberInner {
+                    #[cfg(feature = "unstable")]
+                    session_id: session.zid(),
+                    session: session.into(),
                     state: sub_state,
                     kind: SubscriberKind::LivelinessSubscriber,
-                    undeclare_on_drop: true,
+                    undeclare_on_drop: size_of::<Handler::Handler>() > 0,
```
Mallets, line 584: I understand that a zero-sized handler is not dropped. Callback are zero-size and therefore the handler will not be dropped. Could you add a comment explaining this?

zenoh/src/api/queryable.rs
Mallets: Nevermind, after a second read I believe the content is covered in the new comments.

zenoh/src/api/key_expr.rs
```diff
@@ -549,8 +549,9 @@ impl<'a> KeyExpr<'a> {
     }
 }
 
-impl<'a> UndeclarableSealed<&'a Session, KeyExprUndeclaration<'a>> for KeyExpr<'a> {
-    fn undeclare_inner(self, session: &'a Session) -> KeyExprUndeclaration<'a> {
+impl<'a> UndeclarableSealed<&'a Session> for KeyExpr<'a> {
+    type Undeclaration = KeyExprUndeclaration<'a>;
```
Mallets, line 553:
```suggestion
    type Undeclaration = KeyExprUndeclaration<'a>;
    
```

wyfo:
Closed in favor of #1364

PR #1344, opened 2024-08-30
Allow transport defragmentation errors and fix dropping of old transport messages

### LINKED ISSUE #1343: Unresponsive UDP listener and deadlock closing session when using UDP link and a fragmentation error occurs
### Describe the bug

Using UDP link to do pub/sub with big payload sizes at high throughput, fragmentation is triggered and some fragments can be lost. When a fragment is lost this triggers a fragmentation error at subscriber (reader) side and closes the transport. The listener becomes unresponsive, and the reader session can't be closed following this, due to a deadlock between the session close and the listener.

### To reproduce

1- start a subscriber listening on `udp/127.0.0.1:7447` with logs enabled.
2- start a publisher which connects on `udp/127.0.0.1:7447` and publishes big messages (around 9500bytes or more) at high throughput.
3- after fragmentation error occurs, listener is unresponive and closing its session hangs forever.

### System info

MacBook Air, Mac OS, Apple Silicon M2
Zenoh 1.0.0-alpha.6

oteffahi:
With these changes, transport will not be closed following a defragmentation error (skipping a fragment SN, failure to defragment, defragmentation buffer full). These errors are logged to the `TRACE` level.

Also fixes a bug where old transport messages were still passed to application instead of being dropped.

Closes #1343


io/zenoh-transport/src/unicast/universal/rx.rs
```diff
@@ -97,19 +97,19 @@ impl TransportUnicastUniversal {
...
+                    self.trigger_callback(callback.as_ref(), msg)?;
+                }
+            } else {
+                tracing::debug!(
+                    "Transport: {}. No callback available, dropping messages: {:?}",
+                    self.config.zid,
+                    payload
+                );
             }
-        } else {
```
Mallets, line 107: I would keep the `else` statement but at `trace` level.

io/zenoh-transport/tests/unicast_defragmentation.rs
```diff
@@ -93,19 +93,8 @@ async fn run(endpoint: &EndPoint, channel: Channel, msg_size: usize) {
...
-    });
-
-    // Wait on the router manager that the transport has been closed
-    ztimeout!(async {
-        while !router_manager.get_transports_unicast().await.is_empty() {
-            tokio::time::sleep(SLEEP).await;
-        }
-    });
+    // wait a little bit for the message to be sent
+    tokio::time::sleep(SLEEP).await;
```
Mallets, line 97: Why this change?

oteffahi: The test was checkîng that the transport is closed after a defrag error. Since now there is no expected error, I added a sleep to wait for potential errors to happen after the message is sent. I assumed `schedule` does not effectively send the message and that we have to wait for the TX task to pick it up and send it.

io/zenoh-transport/src/unicast/universal/rx.rs
oteffahi: This is already the case within the `verify_sn` function. When it returns `Ok(false)` it also logs to `trace` that a message was dropped

io/zenoh-transport/src/unicast/universal/rx.rs
```diff
@@ -97,19 +97,19 @@ impl TransportUnicastUniversal {
             Reliability::BestEffort => zlock!(c.best_effort),
         };
 
-        self.verify_sn(sn, &mut guard)?;
-
-        let callback = zread!(self.callback).clone();
-        if let Some(callback) = callback.as_ref() {
-            for msg in payload.drain(..) {
-                self.trigger_callback(callback.as_ref(), msg)?;
+        if self.verify_sn(sn, &mut guard)? {
```
Mallets, line 100: In general I find a high level of nesting harder to read than multiple early returns.
E.g.:
```rust
if cond1 {
  if cond2 {
    // Do stuff
  } else {
    println!("Cond 2 not met"):
  }
} else {
  println!("Cond 1 not met");
}
```
*vs*
```rust
if !cond1! {
  println!("Cond 1 not met);
  return Err(());
}
if !cond2! {
  println!("Cond 2 not met);
  return Err(());
}
// Do stuff
```

io/zenoh-transport/src/unicast/universal/rx.rs
Mallets: Ok, see my most recent comment on nesting.

io/zenoh-transport/tests/unicast_defragmentation.rs
Mallets: Then does it make sense at all to keep this test?

oteffahi: It seems to be the only fragmentation test we have in the codebase. At least if something goes wrong with fragmentation this test will error out.

Mallets: Other tests trigger fragmentation, like https://github.com/eclipse-zenoh/zenoh/blob/983908b35b2a5bc8830b8eccec8025e62d69dc7f/io/zenoh-transport/tests/unicast_transport.rs#L234

oteffahi: Then I guess I'll remove the test since it doesn't serve a purpose anymore.

io/zenoh-transport/src/unicast/universal/rx.rs
```diff
@@ -97,8 +97,10 @@ impl TransportUnicastUniversal {
             Reliability::BestEffort => zlock!(c.best_effort),
         };
 
-        self.verify_sn(sn, &mut guard)?;
-
+        if !self.verify_sn(sn, &mut guard)? {
```
Mallets, line 100: Log the dropped message in trace.

io/zenoh-transport/src/unicast/universal/rx.rs
```diff
@@ -140,28 +143,33 @@ impl TransportUnicastUniversal {
...
```
Mallets, line 146: Log the dropped message in trace.

PR #1351, opened 2024-09-03
Fix storage error when keyexpr equals the configured `strip_prefix`

### LINKED ISSUE #1350: Storages don't accept a put on configured `strip_prefix`
### Describe the bug

When a storage is  configured on a key expression with `strip_prefix`, a `PUT` on a key expression equal to the prefix generates an error in the logs, and the message is not stored.

### To reproduce

Using the REST API and adminspace.
1. Run `RUST_LOG=DEBUG ./zenohd --adminspace-permissions rw --plugin-search-dir .`
2. Load a volume on any supported backend, and create a storage on `demo/example/**` with `strip_prefix=demo/example`.
3. `PUT` on `demo/example`: notice error log in `zenohd` logs. 
4. `GET` on `demo/example/**`: returns an empty reply, meaning the previous `PUT` message was not stored.

### System info

MacBook Air, Mac OS, Apple Silicon M2
Zenoh 1.0.0-beta.1

oteffahi:
Following the changes in #1325, issue #1350 was identified. This PR fixes said issue.

Closes #1350.

### DIFF
--- a/plugins/zenoh-plugin-storage-manager/src/lib.rs
+++ b/plugins/zenoh-plugin-storage-manager/src/lib.rs
@@ -443,12 +443,15 @@ pub fn strip_prefix(
                 );
             }
 
-            match key_expr.strip_prefix(prefix).as_slice() {
-                [stripped_key_expr] => {
-                    if stripped_key_expr.is_empty() {
-                        return Ok(None);
-                    }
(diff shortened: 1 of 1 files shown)


JEnoch (approved):
Good catch!

PR #1347, opened 2024-09-03
feat!: make session an arc-like object 

wyfo:
The refactoring is quite deep, so this is the first (dirty) iteration which passes the tests.

The branch is based on #1334, which should be merged first

PR #1364, opened 2024-09-06
feat!: make session an arc-like object

Mallets:
Superseds https://github.com/eclipse-zenoh/zenoh/pull/1347


zenoh/src/api/publisher.rs
```diff
@@ -990,15 +830,20 @@ impl<'a> MatchingListenerInner<'a> {
...
+
+    fn undeclare_inner(self, _: ()) -> Self::Undeclaration {
         MatchingListenerUndeclaration { subscriber: self }
     }
 }
 
 /// A listener that sends notifications when the [`MatchingStatus`] of a
 /// publisher changes.
 ///
+/// Matching litsteners run in background until the publisher is undeclared.
```
Mallets, line 844:
```suggestion
/// Matching listeners run in background until the publisher is undeclared.
```

zenoh/src/api/publisher.rs
```diff
@@ -990,15 +830,20 @@ impl<'a> MatchingListenerInner<'a> {
...
+/// They can be manually undeclared, but will not be undeclared on drop.
```
Mallets, line 845: Does `They` refer to both `matching listener` and `publisher`? What does `but will not be undeclared on drop` refer to here?

zenoh/src/api/session.rs
```diff
@@ -359,142 +367,120 @@ impl Resource {
...
+            .field("id", &self.runtime.zid())
+            .finish()
     }
-    fn info(&'s self) -> SessionInfo<'a> {
-        SessionInfo {
-            session: self.clone(),
-        }
+}
+
+/// A zenoh session.
```
Mallets, line 412: I'm just realising now that documentation on `Session` is practically inexistent (this is true also for the published doc: https://docs.rs/zenoh/1.0.0-beta.1/zenoh/struct.Session.html)

zenoh/src/api/session.rs
```diff
@@ -359,142 +367,120 @@ impl Resource {
...
+// NOTE: `UndeclarableInner` is only pub(crate) to hide the `undeclare_inner` method. So we don't
+// care about the `private_bounds` lint in this particular case.
+#[allow(private_bounds)]
+/// A trait implemented by types that can be undeclared.
+pub trait Undeclarable<S = ()>: UndeclarableSealed<S> {}
+
+impl<T, S> Undeclarable<S> for T where T: UndeclarableSealed<S> {}
+
+pub(crate) struct SessionInner {
+    weak_counter: Mutex<usize>,
```
Mallets, line 396: @wyfo  Why the `weak_counter` needs to be `Mutex<usize>` instead of `AtomicUsize`?

zenoh/src/api/session.rs
```diff
@@ -359,142 +367,120 @@ impl Resource {
...
         }
     }
 }
 
-pub(crate) trait UndeclarableSealed<S, O, T = ZResult<()>>
-where
-    O: Resolve<T> + Send,
-{
-    fn undeclare_inner(self, session: S) -> O;
+pub(crate) struct WeakSession(Arc<SessionInner>);
```
Mallets, line 447: Why `WeakSession` contains an `Arc` instead of `Weak`? I find it a beat confusing.

zenoh/src/api/session.rs
```diff
@@ -968,28 +984,50 @@ impl Session {
...
         })
     }
+}
+impl SessionInner {
+    fn close(&self) -> impl Resolve<ZResult<()>> + '_ {
+        ResolveFuture::new(async move {
+            let Some(primitives) = zwrite!(self.state).primitives.take() else {
+                return Ok(());
+            };
+            trace!("close()");
```
Mallets, line 1005: Improve log message to indicate which session is being closed..

zenoh/src/api/session.rs
wyfo: At first, I used Weak everywhere (as you can see in the commit history), because it was simpler for a first iteration. However, I also add to use Weak for the primitives, and it there was the rub...
When the session was dropped, primitive was supposed to be still usable for the closing, and using Weak here was causing chain reaction leading to panic in runtime.
I didn't wanted to investigate to much, because it seemed complex, and I already had the idea in my mind to replace Weak by a pseudo-weak reference but using Arc. I mean, Weak means worse performance, notably for Publisher and impl Primitive, so if we have a simple way to avoid it, we should do it. But here, I also had to do it to avoid the panic.
So I implemented this "pseudo-weak", using an additional counter in a mutex. Why? because using atomic seemed very complicated to me — what happens when pseudo-weak are cloned when the last Session is dropped?
So I went with this simple but powerful implementation, bringing the pseudo-weak semantic we need for session closing, and the guaranteeing the best performance for the critical paths.
The mutex only gives a tiny overload for clone/drop operations, but we are talking about session entities, so I assume we don't care at all here.

zenoh/src/api/session.rs
wyfo: See https://github.com/eclipse-zenoh/zenoh/pull/1364#discussion_r1746726921

zenoh/src/api/session.rs
wyfo: Still, your question highlight the lack of comments in the code to explain the approach. I will have to write them.

Mallets: It sounds reasonable, thanks!

zenoh/src/api/publisher.rs
wyfo: It's seems more like remains of my previous works. `MatchingListener` should follow the same semantic than `Subscribers`, i.e. undeclaration on drop if the handler is not a ZST. I will fix it.

zenoh/src/api/session.rs
```diff
@@ -503,104 +489,42 @@ impl Session {
...
-    ///     .await
-    ///     .unwrap();
-    /// tokio::task::spawn(async move {
-    ///     while let Ok(sample) = subscriber.recv_async().await {
-    ///         println!("Received: {:?}", sample);
-    ///     }
-    /// }).await;
-    /// # }
-    /// ```
-    pub fn into_arc(self) -> Arc<Self> {
```
Mallets, line 561: @wyfo I'd be tempted to not remove `into_arc` but to deprecate it saying it's no longer needed. In this way we minimise a bit the quantity of code breaking (at least for betas).

wyfo: I was hesitating too. But for any real exemple where the session is stored in a struct, the user will have to update the type, so removing `into_arc` will follow naturally.
Only one-file example just using inference will benefit keeping this method. That's why I've finally removed it.

PR #1367, opened 2024-09-06
feat(storage-manager): enforce History::Latest based on Timestamp

J-Loudet:
This commit forces all Storage that have the capability `History::Latest` to only keep the Latest value based on their Timestamp. This is achieved by (i) keeping track of the Timestamp associated to all key expressions (for which a publication was received) and (ii) discarding received publications that have an older Timestamp (compared to what is tracked for the same key expression).

The main motivation is the Replication (or Storage Alignment) feature: the algorithm we use rely on the assumption that a Storage with the capability `History::Latest` only keeps the latest publication based on the Timestamp. If that property is not upheld, no guarantee can be made on the Replication algorithm.

Hence, to avoid leaving this responsibility on Storage developers, it has been decided to enforce that behaviour at the Storage Manager level. This is what this commit does.

Additionally, to avoid having the `latest_updates` structure grow infinitely, it was included in the `GarbageCollectionEvent` structure to be garbage collected at regular interval. This garbage collection is controlled by the `GarbageCollectionConfig` which defaults to removing entries that are more than 24h old.

* plugins/zenoh-plugin-storage-manager/src/replica/storage.rs:
  - Add a new field `latest_updates` to the `StorageService` structure, which is used to keep track of the latest publication timestamp for all stripped key expression for which the Storage received a publication.
  - In the `process_sample` function, add a check to ensure that the Sample being processed is more recent than what is currently stored in the Storage. If the Sample is less recent, it is discarded.
  - If the `process_sample` function, after the Sample was successfully inserted, update the `latest_updates` structure with the more recent Timestamp. Note that the `StorageInsertionResult::Outdated` could still be used as an escape hatch for a Storage to filter Samples on something else than the Timestamp.
  - In the `GarbageCollectionEvent`, add the `last_updates` structure such that it is garbage collected.
  - Updated the `run` method of the `GarbageCollectionEvent` to retain only entries that are less than `GarbageCollectionConfig::lifespan` old.


J-Loudet:
@Mallets @JEnoch 
Relevant internal tests: https://github.com/ZettaScaleLabs/zenoh-tests/actions/runs/10735338846

JEnoch (approved):
LGTM.
I checked relevant backends: none was enforcing the timestamps.

PR #1373, opened 2024-09-06
feat(zenoh_id): exposing into `[u8; 16]` & `to_le_bytes()`

DariusIMP:
Allowing users to create a ZenohId from a slice, using TryFrom, and also allowing users to convert a ZenohId into `[u8; 16]`.

Related to https://github.com/eclipse-zenoh/zenoh-kotlin/pull/196.

### DIFF
--- a/commons/zenoh-config/src/wrappers.rs
+++ b/commons/zenoh-config/src/wrappers.rs
@@ -54,6 +54,15 @@ impl From<ZenohIdProto> for ZenohId {
     }
 }
 
+impl TryFrom<&[u8]> for ZenohId {
+    type Error = zenoh_result::Error;
+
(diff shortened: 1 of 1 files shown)


commons/zenoh-config/src/wrappers.rs
```diff
@@ -72,6 +81,13 @@ impl From<ZenohId> for OwnedKeyExpr {
     }
 }
 
+impl From<ZenohId> for [u8; ZenohIdProto::MAX_SIZE] {
```
Mallets, line 84: Instead of this `From` it would be better to expose `to_le_bytes` which provides stronger guarantees on the endianess.

PR #1383, opened 2024-09-09
feat!: make callback opaque

wyfo:
Use an enum hidden to use channel senders directly. It should improve performances, and may allow later to use async sender methods.

### DIFF
--- a/zenoh/src/api/handlers/callback.rs
+++ b/zenoh/src/api/handlers/callback.rs
@@ -13,7 +13,10 @@
 //
 
 //! Callback handler trait.
-use super::{Dyn, IntoHandler};
+
+use std::sync::Arc;
+
+use super::{IntoHandler, RingChannelSender};
 
 /// A function that can transform a [`FnMut`]`(T)` to
 /// a [`Fn`]`(T)` with the help of a [`Mutex`](std::sync::Mutex).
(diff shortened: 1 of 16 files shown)


zenoh/src/api/handlers/callback.rs
```diff
@@ -22,32 +25,88 @@ pub fn locked<T>(fnmut: impl FnMut(T)) -> impl Fn(T) {
...
+    }
+
+    #[inline]
+    pub fn call(&self, arg: &T)
+    where
+        T: Clone,
+    {
+        match &self.0 {
+            CallbackInner::Dyn(cb) => cb(arg),
+            CallbackInner::Flume(tx) => {
```
DenisBiryukov91, line 66: Are you sure that this kind of optimization is really justified given that flume/ring channels are just one of many that users can use in rust, and that many people are still going to just use function callbacks ? 
Also to make use of such an optimization we would need to complicate zenoh-c interface since right now it only accepts a "closure" with pointer to c-function (which is transformed into Arc<dyn>).

wyfo: Flume is not "just one of many that users can use in rust", it's the default one, and one of the few with a `IntoHandler` implementation (the other being `std:sync::mpsc`.
And the goal of this change is also to optimize callback use, see https://godbolt.org/z/G6vcT4b93

> Also to make use of such an optimization we would need to complicate zenoh-c interface since right now it only accepts a "closure" with pointer to c-function (which is transformed into Arc).

I don't see the how this PR changes anything here

DenisBiryukov91: > I don't see the how this PR changes anything here

It adds an extra overhead due to enum match (not sure whether it is significant though). Which means that you probably improve performance of flume channel in rust, but maybe also make it worse for the whole zenoh-c/cpp + produce slightly more complex code. What I mean is that I would really like to see the performance gain for channels / loss for callbacks if there is any.

wyfo: > It adds an extra overhead due to enum match (not sure whether it is significant though).

Branch prediction enters the room.

> Which means that you probably improve performance of flume channel in rust, but maybe also make it worse for the whole zenoh-c/cpp + produce slightly more complex code.


Still better than the previous "by value" callback, see godbolt. Why is the produced code more complex?

> This also means that if in several month we decide to have another "default" channel we would need to revisit this code.

And then, we will change the internal enum, we don't care, it's internal, and it's only a dozen lines to change.

What I mean is that I would really like to see the performance gain for channels / loss for callbacks if there is any.

Trust branch prediction, especially if you are using *only* callbacks.

DenisBiryukov91: > till better than the previous "by value" callback, see godbolt. Why is the produced code more complex?
> 
> >

I do not argue that passing by ref is faster than by value, this can still be achieved through only Arc<dyn>, without any enum matching. 
The produced code is more complex  since it adds more lines => higher maintenance cost (which is always better to be avoided if there is no any significant benefit)

wyfo: > But more than a performance optimization, the long-term vision is to be able to get flume sender directly in order to use its async method.

This is the goal behind making the type opaque. Now, using the enum is not a high maintenance code, all the code is contained in one module and is only a few lines.

wyfo:
Using a immutable reference as callback argument prevent getting a mutable access to a shared memory segment contained in payload. 
The solutions would be either to use a mutable reference, or to keep the argument by value. @kydos said he prefer keeping the current semantic, so let's go. I will the PR accordingly.

DenisBiryukov91:
> Using a immutable reference as callback argument prevent getting a mutable access to a shared memory segment contained in payload.

Access by value is still available through handlers if needed.

Mallets:
> Using a immutable reference as callback argument prevent getting a mutable access to a shared memory segment contained in payload. The solutions would be either to use a mutable reference, or to keep the argument by value. @kydos said he prefer keeping the current semantic, so let's go. I will the PR accordingly.

With current semantics you mean passing `Sample` by value?

wyfo:
Yes

wyfo:
Replaced by #1405

PR #1403, opened 2024-09-11
refactor: use builder method for undeclaration on drop

wyfo:
The previous decided solution for undeclaration on drop "broke" the `FetchingSubscriber`, and using the associated constant doesn't make it easy to repair it. 

After a quick discussion with @kydos, it seems simpler and better to just expose an `undeclare_on_drop` builder method, having `callback` being just a shortcut for `.with(my_callback).undeclare_on_drop(false). This way, there is no hidden behavior, and it should be easier for the user to understand. And more importantly, it's way easier for the extensions and bindings to use it.


Mallets:
I'm curious to understand more on how the solution based on associated constant for `Handler` broke `FetchingSubscriber`.

Mallets (approved):

PR #1404, opened 2024-09-12
Bug/liveliness duplicate

OlivierHecart:
This PR fixes several issues occurring when multiple Liveliness subs or liveliness subs and liveliness get with intersecting key expressions with or without history. The issues were mainly token duplicates or missed token undeclares.

### DIFF
--- a/zenoh/src/net/routing/hat/p2p_peer/token.rs
+++ b/zenoh/src/net/routing/hat/p2p_peer/token.rs
@@ -475,7 +475,7 @@ pub(crate) fn declare_token_interest(
                 for src_face in tables
                     .faces
                     .values()
-                    .filter(|f| mode.future() || f.whatami != WhatAmI::Router)
+                    .filter(|f| f.whatami != WhatAmI::Router)
                     .cloned()
                     .collect::<Vec<Arc<FaceState>>>()
                 {
@@ -508,7 +508,7 @@ pub(crate) fn declare_token_interest(
             for src_face in tables
                 .faces
                 .values()
-                .filter(|f| mode.future() || f.whatami != WhatAmI::Router)
+                .filter(|f| f.whatami != WhatAmI::Router)
                 .cloned()
                 .collect::<Vec<Arc<FaceState>>>()
             {
--- a/zenoh/src/net/routing/hat/p2p_peer/token.rs
+++ b/zenoh/src/net/routing/hat/p2p_peer/token.rs
@@ -421,6 +421,21 @@ pub(super) fn token_new_face(
     }
 }
 
+#[inline]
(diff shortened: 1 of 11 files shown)


Mallets:
Would it be possible to add tests to the CI to catch this bug in the future?

zenoh/src/net/routing/hat/p2p_peer/token.rs
```diff
@@ -466,18 +475,13 @@ pub(crate) fn declare_token_interest(
                 for src_face in tables
                     .faces
                     .values()
+                    .filter(|f| f.whatami != WhatAmI::Router)
```
fuzzypixelz, line 478: Since we send back remote tokens of P2P Peers to the Session, I believe (correct me if I'm wrong) that this can still result in duplicates during the interval where this remote P2P Peer's Initial Interest is not yet finalized:

https://github.com/eclipse-zenoh/zenoh/blob/ac36506326bceac48d6d428e93203c609ba629df/zenoh/src/net/routing/hat/p2p_peer/interests.rs#L137-L139

OlivierHecart: I guess it could. But now that 
1. Token ids are reused if resent to the API: https://github.com/eclipse-zenoh/zenoh/blob/872948828b9355d754d9a261812f8a70e67320e5/zenoh/src/net/routing/hat/p2p_peer/token.rs#L484
2. The API deduplicates tokens: https://github.com/eclipse-zenoh/zenoh/blob/872948828b9355d754d9a261812f8a70e67320e5/zenoh/src/api/session.rs#L2291
The API should deduplicate.

fuzzypixelz: Fair enough. Can we point this out in a comment somewhere?

OlivierHecart: Where ? 
Here https://github.com/eclipse-zenoh/zenoh/blob/872948828b9355d754d9a261812f8a70e67320e5/zenoh/src/api/session.rs#L2291
?

fuzzypixelz: I would put it next to the `f.whatami != WhatAmI::Router` change. But anywhere is fine by me, as long as it makes reference to all relevant code locations (Session de-duplication, Interest propagation from P2P Peers to P2P Peers with non-finalized Initial Interests, and Token propagation from P2P Peers to P2P Peers in general).

OlivierHecart: Let's add comments in another PR.

PR #1407, opened 2024-09-12
refactor: remove state from entities

wyfo:
Entities having having a reference to their state prevented to drop their callback when closing the session.


Mallets (approved):

Mallets:
@OlivierHecart the proposed changes look good to me but I'd like to have also your opinion.

OlivierHecart (approved):

PR #1412, opened 2024-09-13
fix: remove `Session::put` onehost publisher

wyfo:
This publisher caused error when it was dropped.


OlivierHecart (approved):

wyfo:
I've tested with the throughput benchmark, and I've seen no performance difference.

Mallets (approved):

PR #1413, opened 2024-09-13
Add `.private` to `.gitignore`

fuzzypixelz:
A folder to store intentionally untracked changes for a less cluttered git-status.


fuzzypixelz:
Nevermind, just use `.git/info/exclude`.

Issue #1417, opened 2024-09-13
Much of the `Config` API should not be public


fuzzypixelz:
### Describe the bug

N/A

### To reproduce

N/A

### System info

- ref: 53ad3138fb39d412952339b923d42121b66825d5

Mallets:
Closed by #1419

Issue #1304, opened 2024-08-09
Review of public `zenoh` items


fuzzypixelz:
### Describe the bug

Quickly skimming https://docs.rs/zenoh/1.0.0-alpha.6/zenoh/all.html, the presence of some items may be problematic

## Duplicate definitions

- https://docs.rs/zenoh/1.0.0-alpha.6/zenoh/type.Error.html vs https://docs.rs/zenoh/1.0.0-alpha.6/zenoh/prelude/type.ZError.html
- https://docs.rs/zenoh/1.0.0-alpha.6/zenoh/type.Result.html vs https://docs.rs/zenoh/1.0.0-alpha.6/zenoh/prelude/type.ZResult.html

## `zenoh-keyexpr::keyexpr_tree` re-exports

- https://docs.rs/zenoh/1.0.0-alpha.6/zenoh/key_expr/keyexpr_tree/trait.IKeyExprTree.html
- https://docs.rs/zenoh/1.0.0-alpha.6/zenoh/key_expr/keyexpr_tree/trait.IKeyExprTreeMut.html
- https://docs.rs/zenoh/1.0.0-alpha.6/zenoh/key_expr/keyexpr_tree/struct.KeBoxTree.html
- https://docs.rs/zenoh/1.0.0-alpha.6/zenoh/key_expr/keyexpr_tree/struct.KeyedSetProvider.html
- https://docs.rs/zenoh/1.0.0-alpha.6/zenoh/key_expr/keyexpr_tree/struct.NonWild.html
- https://docs.rs/zenoh/1.0.0-alpha.6/zenoh/key_expr/keyexpr_tree/struct.UnknownWildness.html

## `zenoh-protocol` re-exports

- https://docs.rs/zenoh/1.0.0-alpha.6/zenoh/query/type.ConsolidationMode.html
- https://docs.rs/zenoh/1.0.0-alpha.6/zenoh/query/type.QueryTarget.html

## Leaked implementation details

- https://docs.rs/zenoh/1.0.0-alpha.6/zenoh/sample/type.SourceSn.html
- https://docs.rs/zenoh/1.0.0-alpha.6/zenoh/session/type.EntityId.html

## Inconsistency in naming constants

[...]

PR #1419, opened 2024-09-13
Hide `zenoh_config` internals

fuzzypixelz:
Addresses #1417 and #1304 with the following changes:

- Replace `zenoh::Config` with `zenoh::config::Config` leaving `zenoh_config::Config` unchanged.
- Add a new `zenoh_config::Config` with the following public items:
  - `Default`, `Debug`, `Clone` and `Default` impls
  - `Deref<Target = InternalConfig>` and `DerefMut<Target = InternalConfig>` impls with `cfg(feature = "unstable_config")`]
  - `from_env`, `from_file`, `insert_json5`, `get` and `remove`
- Reduce `zenoh::config` to the following public items:
  - `Config`
  - `InsertionError`
  - `LookupError`
  - `LookupGuard`
  - `WhatAmI` (for `zenoh::scout`)
  - `WhatAmIMatcher` (for `zenoh::scout`)
  - `Notifier` (for `zenoh::Session::config`)
  - `Notification` (for `zenoh::config::Notifier`)
- Move `Notifier` to the zenoh crate with the following API:
  - `new` (**not sure about this one**)
  - `subscribe`
  - `notify`
  - `lock`
  - `remove`
  - `insert_json5`
  - `get`

Older code using the `zenoh_config::Config` API will still compile provided that items such as `ModeDependentValue` are imported from `zenoh_config`.

<hr>

I believe the `Config` API changes is ready to be propagated to bindings. Here are CI runs to know what's going wrong:

- cc @wyfo https://github.com/eclipse-zenoh/zenoh-python/actions/runs/10886444693/job/30245908828
- cc @sashacmc https://github.com/eclipse-zenoh/zenoh-c/actions/runs/10886423431/job/30245904074
- cc @DariusIMP https://github.com/eclipse-zenoh/zenoh-java/actions/runs/10899780569
- cc @DariusIMP https://github.com/eclipse-zenoh/zenoh-kotlin/actions/runs/10886495298

### DIFF
--- a/zenoh/src/api/config.rs
+++ b/zenoh/src/api/config.rs
@@ -31,18 +31,17 @@ impl Config {
     }
 
     pub fn insert_json5(&mut self, key: &str, value: &str) -> Result<(), InsertionError> {
-        <zenoh_config::Config as validated_struct::ValidatedMap>::insert_json5(
-            &mut self.0,
-            key,
-            value,
-        )
-        .map_err(InsertionError)
+        self.0.insert_json5(key, value).map_err(InsertionError)
     }
 
     #[zenoh_macros::unstable]
     pub fn get<'a>(&'a self, key: &str) -> Result<&'a dyn Any, LookupError> {
-        <zenoh_config::Config as validated_struct::ValidatedMap>::get(&self.0, key)
-            .map_err(LookupError)
+        self.0.get(key).map_err(LookupError)
+    }
+
+    #[zenoh_macros::unstable]
+    pub fn get_json(&self, key: &str) -> Result<String, LookupError> {
+        self.0.get_json(key).map_err(LookupError)
     }
 
     #[zenoh_macros::unstable]
--- a/zenoh/src/api/config.rs
+++ b/zenoh/src/api/config.rs
@@ -12,7 +12,9 @@
 //   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
 //
 use std::{
+    any::Any,
     env, fmt,
+    ops::Deref,
     path::Path,
     sync::{Arc, Mutex, MutexGuard},
 };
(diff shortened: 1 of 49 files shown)


zenoh/src/api/config.rs
```diff
@@ -0,0 +1,224 @@
...
+    pub fn from_file<P: AsRef<Path>>(path: P) -> ZResult<Self> {
+        Ok(Config(zenoh_config::Config::from_file(path)?))
+    }
+
+    pub fn insert_json5(&mut self, key: &str, value: &str) -> Result<(), InsertionError> {
+        self.0.insert_json5(key, value).map_err(InsertionError)
+    }
+
+    #[zenoh_macros::unstable]
+    pub fn get<'a>(&'a self, key: &str) -> Result<&'a dyn Any, LookupError> {
```
Mallets, line 38: Is this function really needed? I would remove it.

zenoh/src/api/config.rs
```diff
@@ -0,0 +1,224 @@
...
+impl Config {
+    pub fn from_env() -> ZResult<Self> {
+        Ok(Config(zenoh_config::Config::from_env()?))
+    }
+
+    pub fn from_file<P: AsRef<Path>>(path: P) -> ZResult<Self> {
+        Ok(Config(zenoh_config::Config::from_file(path)?))
+    }
+
+    pub fn insert_json5(&mut self, key: &str, value: &str) -> Result<(), InsertionError> {
```
Mallets, line 33: I would add as well:
```rust
pub fn from_json5(value: &str) -> ZResult<Config>;
```

zenoh/src/api/config.rs
```diff
@@ -0,0 +1,224 @@
...
+        self.0.get(key).map_err(LookupError)
+    }
+
+    #[zenoh_macros::unstable]
+    pub fn get_json(&self, key: &str) -> Result<String, LookupError> {
+        self.0.get_json(key).map_err(LookupError)
+    }
+
+    #[zenoh_macros::unstable]
+    pub fn remove<K: AsRef<str>>(&mut self, key: K) -> ZResult<()> {
```
Mallets, line 48: Is this function really needed? If not I would remove it.

zenoh/src/api/config.rs
fuzzypixelz: It's only used as an example in the `Session::config` rustdoc:

https://github.com/eclipse-zenoh/zenoh/blob/fb0f4ead6283df9f74b79a9d5e06cc3227a34256/zenoh/src/api/session.rs#L636-L644

I don't have anything against removing it, the above is the only reason why I kept it.

zenoh/src/api/config.rs
fuzzypixelz: The `Notifier` version is used by the adminspace:

https://github.com/eclipse-zenoh/zenoh/blob/fb0f4ead6283df9f74b79a9d5e06cc3227a34256/zenoh/src/net/runtime/adminspace.rs#L417-L430

We can make it `pub(crate)`...

zenoh/src/api/config.rs
fuzzypixelz: See 78d8780d2c083820f9c7be91281ceb02c30ff616.

zenoh/src/api/config.rs
fuzzypixelz: Deleted.

zenoh/src/api/config.rs
fuzzypixelz: Deleted.

Mallets (approved):

PR #1422, opened 2024-09-16
feat: improving pluging feature

gabrik:
Adding a feature for a more fine graned control on plugins loading


Mallets:
To handle optional dependency we may invert the feature logic.
You could add a `static_plugins_loading` feature that enables `async-std` dependency.

PR #1424, opened 2024-09-16
Update depdendencies

fuzzypixelz:
Not much to say here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,40 +3,19 @@
 version = 3
 
 [[package]]
-name = "CoreFoundation-sys"
-version = "0.1.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d0e9889e6db118d49d88d84728d0e964d973a5680befb5f85f55141beea5c20b"
-dependencies = [
- "libc",
- "mach",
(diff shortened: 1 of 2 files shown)


Mallets:
@fuzzypixelz can we make sure the change is reflected in the `Cargo.toml` as well?

fuzzypixelz:
@Mallets The `Cargo.toml` specify version ranges, this pull request bumps dependencies within those ranges.

For example tokio has been bumped from `1.39.3` to `1.40.0` because in `Cargo.toml` we got `tokio = "1.39.3"` and that means any `1.x.y` version where `x >= 39` and `y >= 3`.

Same goes for tokio-util which went from `0.7.11` to `0.7.12`.

Updating the `Cargo.toml` is only necessary if we go through semver-incompatible changes, which is beyond the scope of this pull request.

Mallets:
It should be in the scope of the next `-rc` for Zenoh. Even though `cargo` does what you described, it would be good to have the version bumped and verify that Zenoh depends on the very latest crates.

fuzzypixelz:
@Mallets Done.

Mallets (approved):

PR #1425, opened 2024-09-16
Improve API docs

Mallets:
Remove certain type aliases that leaked internal crates directly in the docs.


zenoh/src/api/query.rs
```diff
@@ -25,6 +25,10 @@ use zenoh_keyexpr::OwnedKeyExpr;
 #[cfg(feature = "unstable")]
 use zenoh_protocol::core::ZenohIdProto;
 use zenoh_protocol::core::{CongestionControl, Parameters};
+/// The [`Queryable`](crate::query::Queryable)s that should be target of a [`get`](Session::get).
```
fuzzypixelz, line 28: So we've decided against wrapping this.

Mallets: It's an `enum` so it can't be really wrapped. It can only be redefined.

fuzzypixelz: It's still possible to redefine all the variants. If that's too much churn then I suggest we add `#[non_exhaustive]` to leave some degree of flexibility.

Mallets: `#[non_exhaustive]` has some counter indications and we decided in the past to not use it.
I'm not such worried about having this `enum` and to potentially separate the types in the future. 

E.g., the same principle applies to https://docs.rs/zenoh/1.0.0-beta.2/zenoh/qos/enum.CongestionControl.html or https://docs.rs/zenoh/1.0.0-beta.2/zenoh/qos/enum.Priority.html.

fuzzypixelz: Alright, I guess we can re-define it anyway in case we want to change the underlying zenoh-protocol definition.

fuzzypixelz (approved):

PR #1430, opened 2024-09-17
Fix docs

Mallets:
Fix docs

### DIFF
--- a/zenoh/src/api/queryable.rs
+++ b/zenoh/src/api/queryable.rs
@@ -147,7 +147,7 @@ impl Query {
         }
     }
 
-    /// Sends a reply to this Query.
+    /// Sends a [`crate::sample::Sample`] of kind [`crate::sample::SampleKind::Put`] as a reply to this Query.
     ///
     /// By default, queries only accept replies whose key expression intersects with the query's.
     /// Unless the query has enabled disjoint replies (you can check this through [`Query::accepts_replies`]),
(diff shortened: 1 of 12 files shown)


zenoh/src/api/queryable.rs
```diff
@@ -132,7 +132,7 @@ impl Query {
         self.attachment.as_mut()
     }
 
-    /// Sends a reply in the form of [`Sample`] to this Query.
+    /// Send a reply in the form of [`Sample`] to this Query.
```
fuzzypixelz, line 135:
```suggestion
    /// Sends a reply in the form of [`Sample`] to this Query.
```

I would stick with the third person pronoun as is done in `std` and most Rust crates.

Here's an example: https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.push

zenoh/src/api/queryable.rs
```diff
@@ -147,7 +147,7 @@ impl Query {
         }
     }
 
-    /// Sends a reply to this Query.
+    /// Send a put as a reply to this Query.
```
fuzzypixelz, line 150:
```suggestion
    /// Sends a [`crate::api::sample::Sample`] of kind [`crate::api::sample::SampleKind::Put`] as a reply to this Query.
```

zenoh/src/api/queryable.rs
```diff
@@ -191,7 +190,7 @@ impl Query {
         }
     }
 
-    /// Sends a delete reply to this Query.
+    /// Send a delete as a reply to this Query.
```
fuzzypixelz, line 193:
```suggestion
    /// Sends a [`crate::sample::Sample`] of kind [`crate::sample::SampleKind::Delete`] as a reply to this Query.
```

fuzzypixelz (changes requested):
I would stick with the third person pronoun as is done in `std` and most Rust crates.

Here's an example: https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.push

Mallets:
@fuzzypixelz  Suggestions applied!

fuzzypixelz (approved):

PR #1436, opened 2024-09-18
Fix config rexport

Mallets:
Rexport some config type under unstable_config feature.
Some type was erroneously removed by https://github.com/eclipse-zenoh/zenoh/pull/1419.

### DIFF
--- a/zenoh/src/lib.rs
+++ b/zenoh/src/lib.rs
@@ -352,10 +352,6 @@ pub mod time {
 /// Configuration to pass to [`open`] and [`scout`] functions and associated constants
 pub mod config {
     pub use zenoh_config::{EndPoint, Locator, WhatAmI, WhatAmIMatcher, ZenohId};
-    #[zenoh_macros::unstable_config]
-    pub use zenoh_config::{
-        ModeDependentValue, PermissionsConf, PluginsConfig, SecretValue, ValidatedMap,
-    };
 
     pub use crate::api::config::Config;
     #[zenoh_macros::unstable]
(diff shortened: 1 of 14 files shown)


zenoh/src/lib.rs
```diff
@@ -351,7 +351,11 @@ pub mod time {
 
 /// Configuration to pass to [`open`] and [`scout`] functions and associated constants
 pub mod config {
-    pub use zenoh_config::{WhatAmI, WhatAmIMatcher};
+    pub use zenoh_config::{EndPoint, Locator, WhatAmI, WhatAmIMatcher, ZenohId};
```
wyfo, line 354: why putting back `EndPoint` and `Locator`?

fuzzypixelz: They can't really be used from the stable API, having them there is superfluous I would say

Although users will write Endpoints in their configuration and they would benefit from the docs, it's still odd having a public item that once constructed can't be used in any way.

zenoh/src/lib.rs
```diff
@@ -351,7 +351,11 @@ pub mod time {
...
+    #[zenoh_macros::unstable_config]
+    pub use zenoh_config::{
+        ModeDependentValue, PermissionsConf, PluginsConfig, SecretValue, ValidatedMap,
```
fuzzypixelz, line 357: How was this chosen? It seems like an odd selection from what's available.

zenoh/src/lib.rs
Mallets: `Locator` is part of the `Hello` and scouting API. And they are stable.

zenoh/src/lib.rs
Mallets: I've simply reverted the changes.

PR #1445, opened 2024-09-18
Alternative SHM segment cleanup

yellowhatter:


commons/zenoh-shm/src/api/cleanup.rs
```diff
@@ -0,0 +1,35 @@
...
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+//
+// Contributors:
+//   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
+//
+
+use crate::cleanup::CLEANUP;
+
+/// Make forced cleanup
```
fuzzypixelz, line 17:
```suggestion
/// Make forced cleanup.
///
```

commons/zenoh-shm/src/posix_shm/cleanup.rs
```diff
@@ -0,0 +1,134 @@
...
+//
+// Contributors:
+//   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
+//
+
+pub(crate) use platform::cleanup_orphaned_segments;
+
+#[cfg(not(unix))]
+mod platform {
+    pub(crate) fn cleanup_orphaned_segments() {}
```
fuzzypixelz, line 19: Maybe we can add a `tracing::warn` in case `cleanup_orphaned_segments` does nothing?

yellowhatter: Well, no, because what I want is to abstract our users from the platform-specific details of our implementation (that even might change in the future!)

PR #1454, opened 2024-09-19
refactor(storage-manager): implement cache for `History::Latest`

J-Loudet:
The changes introduced in #1367 were actually redundant (and incomplete): the method `is_latest` was checking that a received Sample was more recent than what is currently stored.

This commit removes the redundant checks and implements a correct caching strategy by leveraging both approaches. The changes consist mostly in:
1. Changing the outer test by continuing early if the Sample is detected as deleted. This allows lowering the overall indentation of the `process_sample` method by one level.
2. Modifying the behaviour of the `is_latest` method to check the cache first and only query the Storage if there is a cache miss.

To help make potential concurrency issues visible, the method `is_latest` was renamed to `guard_cache_if_latest` as the lock on the Cache should only be released when the Sample has been processed *and* the Cache updated.

Additionally, the cache is filled at initialisation.

The `read_cost` configuration parameter was removed as it was not used and made obsolete with this change.

* plugins/zenoh-backend-example/src/lib.rs:
* plugins/zenoh-backend-traits/src/lib.rs:
* plugins/zenoh-plugin-storage-manager/src/memory_backend/mod.rs: removed the obsolete `read_cost` parameter.

* plugins/zenoh-plugin-storage-manager/src/storages_mgt/mod.rs:
  - Populate the Cache at initialisation.
  - Updated the call to `StorageService::start` to also pass the Cache.

* plugins/zenoh-plugin-storage-manager/src/storages_mgt/service.rs
  - Updated the call to `StorageService::start` to also pass the Cache.
  - "Reversed" the logic when checking if a Sample is deleted: if it is continue to the next one. This allows reducing the indentation level.
  - Keep the guard over the cache if a Sample is more recent and only release it once it has been updated (after the Sample has been successfully processed).

### DIFF
--- a/plugins/zenoh-plugin-storage-manager/src/storages_mgt/service.rs
+++ b/plugins/zenoh-plugin-storage-manager/src/storages_mgt/service.rs
@@ -22,7 +22,7 @@ use std::{
 use async_trait::async_trait;
 use flume::Receiver;
 use futures::select;
-use tokio::sync::{Mutex, RwLock};
+use tokio::sync::{Mutex, MutexGuard, RwLock};
 use zenoh::{
     internal::{
         buffers::{SplitBuffer, ZBuf},
@@ -75,9 +75,8 @@ impl StorageService {
         storage: Arc<Mutex<Box<dyn zenoh_backend_traits::Storage>>>,
         capability: Capability,
         rx: Receiver<StorageMessage>,
+        latest_updates: Arc<Mutex<HashMap<Option<OwnedKeyExpr>, Timestamp>>>,
     ) {
-        // @TODO: optimization: if read_cost is high for the storage, initialize a cache for the
-        // latest value
         let mut storage_service = StorageService {
             session,
             key_expr: config.key_expr,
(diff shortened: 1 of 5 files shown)


plugins/zenoh-plugin-storage-manager/src/storages_mgt/service.rs
```diff
@@ -236,127 +235,118 @@ impl StorageService {
...
-                    }
-                };
+            let mut storage = self.storage.lock().await;
+            let storage_result = match sample.kind() {
+                SampleKind::Put => {
+                    storage
+                        .put(
+                            stripped_key.clone(),
+                            Value::new(
+                                sample_to_store.payload().clone(),
```
Charles-Schleich, line 320:
```suggestion
                                sample_to_store.payload(),
```

Charles-Schleich:
@J-Loudet, Cache implementation looks good.

J-Loudet:
Relevant internal tests: https://github.com/ZettaScaleLabs/zenoh-tests/actions/runs/10944755270

PR #1462, opened 2024-09-20
Polish the public API

fuzzypixelz:
n/a


Mallets (approved):

wyfo:
Why do we want a breaking change that can be avoided simply?

PR #1467, opened 2024-09-20
feat!: improve serialization efficiency

wyfo:
Fixed-size objects are now written directly on the writer, instead of being wrapped first into a ZBytes.
Slice like objects are also written on the writer, but obviously prefixed by their size, like ZBytes.

### DIFF
--- a/ci/valgrind-check/src/pub_sub/bin/z_pub_sub.rs
+++ b/ci/valgrind-check/src/pub_sub/bin/z_pub_sub.rs
@@ -38,7 +38,7 @@ async fn main() {
                 sample.key_expr().as_str(),
                 sample
                     .payload()
-                    .deserialize::<String>()
+                    .try_deserialize::<String>()
                     .unwrap_or_else(|e| format!("{}", e))
             );
         })
--- a/zenoh/src/api/bytes.rs
+++ b/zenoh/src/api/bytes.rs
@@ -13,16 +13,12 @@
 //
 
 //! ZBytes primitives.
-use std::{
-    borrow::Cow, collections::HashMap, convert::Infallible, fmt::Debug, marker::PhantomData,
-    str::Utf8Error, string::FromUtf8Error, sync::Arc,
-};
+use std::{borrow::Cow, convert::Infallible, fmt::Debug, marker::PhantomData, mem::size_of};
 
 use uhlc::Timestamp;
-use unwrap_infallible::UnwrapInfallible;
 use zenoh_buffers::{
     buffer::{Buffer, SplitBuffer},
-    reader::{HasReader, Reader},
+    reader::{DidntRead, HasReader, Reader},
     writer::HasWriter,
     ZBuf, ZBufReader, ZBufWriter, ZSlice, ZSliceBuffer,
 };
@@ -31,12 +27,9 @@ use zenoh_protocol::{
     core::{Encoding as EncodingProto, Parameters},
     zenoh::ext::AttachmentType,
 };
-#[cfg(feature = "shared-memory")]
+#[cfg(all(feature = "shared-memory", feature = "unstable"))]
 use zenoh_shm::{
-    api::buffer::{
-        zshm::{zshm, ZShm},
-        zshmmut::{zshmmut, ZShmMut},
-    },
+    api::buffer::{zshm::zshm, zshmmut::zshmmut},
     ShmBufInner,
 };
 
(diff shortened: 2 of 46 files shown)


zenoh/src/api/bytes.rs
```diff
@@ -819,2538 +712,831 @@ impl std::fmt::Display for ZDeserializeError {
...
+            fn deserialize(_zbytes: &'_ ZBytes) -> Result<Self, ZDeserializeError> {
+                unimplemented!()
+            }
+            fn try_from_zbytes(_zbytes: ZBytes) -> Result<Self, ZDeserializeError> {
+                unimplemented!()
+            }
+        }
+    )*};
 }
+impl_num!(i8, i16, i32, i64, i128, isize, u8, u16, u32, u64, u128, usize, f32, f64);
```
DenisBiryukov91, line 1077: I'm not sure implementation for isize/usize makes sense anymore since their sizes are platform dependent (we could always use vle, but the issue is that in some languages they are not independent types and are defined in terms of other integers).

ci/valgrind-check/src/pub_sub/bin/z_pub_sub.rs
```diff
@@ -38,7 +38,7 @@ async fn main() {
                 sample.key_expr().as_str(),
                 sample
                     .payload()
-                    .deserialize::<String>()
+                    .try_deserialize::<String>()
```
Mallets, line 41: Why changing the name? There is already quite a bit of material out there, not mentioning all bindings, that uses `deserialize`. I don't think this change is needed.

zenoh/src/api/bytes.rs
```diff
@@ -819,2538 +712,831 @@ impl std::fmt::Display for ZDeserializeError {
...
 
-impl From<&Cow<'_, [u8]>> for ZBytes {
-    fn from(t: &Cow<'_, [u8]>) -> Self {
-        ZSerde.serialize(t)
-    }
+impl SerializedSize for bool {
```
DenisBiryukov91, line 1079: Are we confident that the size of bool will be the same for each platform ?

zenoh/src/api/bytes.rs
```diff
@@ -456,69 +439,59 @@ impl ZBytes {
...
     /// let buf: Vec<u8> = vec![0, 1, 2, 3];
     /// let bytes = ZBytes::from(buf.clone());
-    /// let deser: Vec<u8> = bytes.into();
+    /// let deser: Vec<u8> = bytes.deserialize();
     /// assert_eq!(buf.as_slice(), deser.as_slice());
     /// ```
     ///
     /// If you want to be sure that no copy is performed at all, then you should use [`ZBytes::slices`].
     /// Please note that in this case data may not be contiguous in memory and it is the responsibility of the user to properly parse the raw slices.
-    pub fn into<'a, T>(&'a self) -> T
```
Mallets, line 465: I don't think we need to introduce this API change.
See my previous comment.

zenoh/src/api/bytes.rs
```diff
@@ -819,2538 +712,831 @@ impl std::fmt::Display for ZDeserializeError {
...
-
-// [u8; N]
-impl<const N: usize> Serialize<[u8; N]> for ZSerde {
-    type Output = ZBytes;
-
-    fn serialize(self, t: [u8; N]) -> Self::Output {
-        ZBytes::new(t)
+impl Deserialize<'_> for Vec<u8> {
+    fn read(_reader: &mut ZBytesReader) -> Result<Self, ZDeserializeError> {
+        unimplemented!()
```
Mallets, line 887: It should be implemented.

zenoh/src/api/bytes.rs
```diff
@@ -86,48 +79,59 @@ impl From<OptionZBytes> for Option<ZBytes> {
...
+}
 
-    /// The implementer should take care of deserializing a [`ZBytes`] into the type `T`.
-    /// If [`Encoding`] metadata is required in a given system, the caller should take care of checking the proprer
-    /// [`Encoding`] value (e.g. [`crate::sample::Sample::encoding`]) to select the right type `T` to deserialize into.
-    fn deserialize(self, t: Self::Input<'_>) -> Result<T, Self::Error>;
+pub trait Deserialize<'a>: Sized + SerializedSize {
+    fn read(reader: &mut ZBytesReader) -> Result<Self, ZDeserializeError>;
+    fn deserialize(zbytes: &'a ZBytes) -> Result<Self, ZDeserializeError>;
+    fn try_from_zbytes(zbytes: ZBytes) -> Result<Self, ZDeserializeError>;
```
Mallets, line 115: What's the purpose of `try_from_zbytes` as part of the trait? It seems not needed to me.

ci/valgrind-check/src/pub_sub/bin/z_pub_sub.rs
wyfo: everything has been changed in the last iteration with @DenisBiryukov91 and @kydos, see the coming PR. This one will be closed.

zenoh/src/api/bytes.rs
wyfo: No it doesn't, because it should never be called, as the the semantic was:
- call `read` if the size is fixed (not the case for `Vec<T>`)
- call `deserialize`/`try_from_zbytes` if the size is variable

zenoh/src/api/bytes.rs
wyfo: Even if everything will be dropped so we don't care anymore, here was the reason:
`deserialize` couldn't be call in the context of an iterator when you deserialize first a `ZBytes`, which would then have an anonymous lifetime. So deserializing an iterator of `Cow<'a, [u8]>` was not possible as `'a` would be anonymous in the implementation.
That's why two different methods were needed, one for `ZBytes::deserialize`, and the other one for the iterator.

PR #1468, opened 2024-09-24
feat(storage-manager): replication initial alignment

J-Loudet:
These commits change the way a replicated Storage starts: if it is empty
and configured to be replicated, it will attempt to align with an active
and compatible (i.e. same configuration) Replica before anything.

The previous behaviour made a query on the key expression of the
Storage. Although, it could, in some cases, actually perform the same
initial alignment, it could not guarantee to only query a Storage that
was configured to be replicated.

⚠️  See individual commits for more details, they are meant to be reviewed separately.
⚠️ This PR is meant to be rebased (_not_ squashed).


Charles-Schleich (approved):
LGTM !

PR #1472, opened 2024-09-25
feat!: move serialization into zenoh-ext

wyfo:
@kydos @DenisBiryukov91

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5130,6 +5130,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "which"
+version = "6.0.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b4ee928febd44d98f2f459a4a79bd4d928591333a494a10a868418ac1b39cf1f"
+dependencies = [
+ "either",
+ "home",
+ "rustix 0.38.37",
+ "winsafe",
+]
+
+[[package]]
 name = "win-sys"
 version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5409,6 +5421,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "winsafe"
+version = "0.0.19"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d135d17ab770252ad95e9a872d365cf3090e3be864a34ab46f48555993efc904"
+
+[[package]]
 name = "x509-parser"
 version = "0.16.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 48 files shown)


milyin:
I like the proposed solution. And maybe it can be made even simpler. It seems that we don't need to put this to zenoh-ext. One of our goals is to explicitly detach serialization from `ZBytes`. But serialization in any case will be present in all bindings. Moving it to zenoh-ext doesn't change a lot.
Instead the traits `ZBytesExt`, `ZBytesWriterExt` which attaches serialization to `ZBytes` can be removed without losing functionality. We already have all necessary functionality in `Serialize` and `Deserialize` traits and may allow user to use them directly.

My proposal is to define `Serialize` in zenoh and make it something like this:
```rust
pub trait Serialize {
    fn serialize(&self, writer: &mut ZBytesWriter);
    fn serialize_iter<I: IntoIterator<Item = &Self>>(iter: I, writer: &mut ZBytesWriter) where I::IntoIter: ExactSizeIterator {
        let iter = iter.into_iter();
        self.write_vle(iter.len() as u64);
        for t in iter {
            t.serialize(self);
        }
   }
   fn into_zbytes(self) -> ZBytes;
   fn into_zbytes_iter< ......
}
```

The idea is that `ZBytes` is not aware at all about serialization. Instead each type can implement the `Serialize` and `Deserialize` traits and use the methods directly:

```
    let mut bytes = ZBytes::new();
    let mut writer = bytes.writer();
    let i1 = 1234_u32;
    let i2 = String::from("test");
    let i3 = vec![1, 2, 3, 4];
    i1.serialize(&mut writer);
    i2.serialize(&mut writer);
    i3.serialize(&mut writer);
    // deserialization
    let mut reader = bytes.reader();
    let o1 = u32::deserialize(&mut reader).unwrap();
    let o2 = String::deserialize(&mut reader).unwrap();
    let o3 = Vec::<u8>::deserialize(&mut reader).unwrap();
```

wyfo:
There are two issues with your proposal:
- `serialize_iter` shouldn't be a trait method, because otherwise it would be meant to be overridden, and that's not the case;
- my previous draft was in fact using method on the (de)serialized type instead of the writer/reader, but @DenisBiryukov91 was against it.

milyin:
> There are two issues with your proposal:
> 
>     * `serialize_iter` shouldn't be a trait method, because otherwise it would be meant to be overridden, and that's not the case;
Ok, then it should be real method, no problem

>     * my previous draft was in fact using method on the (de)serialized type instead of the writer/reader, but @DenisBiryukov91 was against it.

@DenisBiryukov91 mentioned that the code looks ugly from his point of view. But I think the full separation of serialization and writer worths it. The visible fact that there is no connection from writer to serializer will really help user get the idea, where is the base functionality and where is the replaceable helper

wyfo:
Superseded by #1473

PR #1475, opened 2024-09-26
do not use Linux Ephemeral ports in tests

yellowhatter:
Sometimes this causes tests to fail because of port collision

### DIFF
--- a/zenoh/tests/authentication.rs
+++ b/zenoh/tests/authentication.rs
@@ -41,10 +41,10 @@ mod test {
         create_new_files(TESTFILES_PATH.to_path_buf())
             .await
             .unwrap();
-        test_pub_sub_deny_then_allow_usrpswd(37447).await;
-        test_pub_sub_allow_then_deny_usrpswd(37447).await;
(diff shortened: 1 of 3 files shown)


oteffahi:
@yellowhatter could you please also update the port numbers in these tests: https://github.com/eclipse-zenoh/zenoh/blob/main/zenoh/tests/authentication.rs#L44-L81
Please respect the unicity of ports as it is, quic and tls do not close their ports in a consistent time, and the different tests run concurrently
Thanks 😄

yellowhatter:
> @yellowhatter could you please also update the port numbers in these tests: https://github.com/eclipse-zenoh/zenoh/blob/main/zenoh/tests/authentication.rs#L44-L81 Please respect the unicity of ports as it is, quic and tls do not close their ports in a consistent time, and the different tests run concurrently Thanks 😄

Yep, my text search wasn't good enough to find also these ports ;)

yellowhatter:
@oteffahi done

oteffahi (approved):
LGTM.

PR #1477, opened 2024-09-26
Fix peers start conditions bug

OlivierHecart:
connect_peers returns false if not connected to avoid wrong start_condititions notif

### DIFF
--- a/zenoh/src/net/runtime/orchestrator.rs
+++ b/zenoh/src/net/runtime/orchestrator.rs
@@ -911,7 +911,7 @@ impl Runtime {
         }
     }
 
-    /// Returns `true` if a new Transport instance exists between us and `zid`.
+    /// Returns `true` if a new Transport instance is established between us and `zid` or had already been established.
     #[must_use]
     async fn connect(&self, zid: &ZenohIdProto, scouted_locators: &[Locator]) -> bool {
         if !self.insert_pending_connection(*zid).await {
@@ -1024,7 +1024,7 @@ impl Runtime {
         }
     }
 
-    /// Returns `true` if a new Transport instance exists between us and `zid`.
+    /// Returns `true` if a new Transport instance is established between us and `zid` or had already been established.
     pub async fn connect_peer(&self, zid: &ZenohIdProto, locators: &[Locator]) -> bool {
         let manager = self.manager();
         if zid != &manager.zid() {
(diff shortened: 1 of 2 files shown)


oteffahi (approved):
Bug seems fixed.
LGTM.

zenoh/src/net/runtime/orchestrator.rs
```diff
@@ -1024,7 +1024,7 @@ impl Runtime {
         }
     }
 
-    pub async fn connect_peer(&self, zid: &ZenohIdProto, locators: &[Locator]) {
+    pub async fn connect_peer(&self, zid: &ZenohIdProto, locators: &[Locator]) -> bool {
```
fuzzypixelz, line 1027: Correct me if I'm wrong. `Runtime::connect` returns true if it has opened a new Transport with `zid`. This seems to do the opposite.

oteffahi: If I understand correctly, `connect_peer` only returns `false` if `Runtime::connect` returns `false`, and `Runtime::connect` returns `true` if it succeeds in opening a new transport with `zid`.
The only confusing bit to me is that if a connection is pending with `zid`, then both functions will return `false`.

fuzzypixelz: Nevermind, it returns true if a Transport is opened with `zid`.

zenoh/src/net/runtime/orchestrator.rs
```diff
@@ -911,7 +911,7 @@ impl Runtime {
         }
     }
 
-    /// Returns `true` if a new Transport instance has been opened with `zid`.
+    /// Returns `true` if a new Transport instance exists between us and `zid`.
```
oteffahi, line 914: This is misleading, as the `connect` function doesn't only check if a transport exists, but also tries to open one if it does not exist.

oteffahi (changes requested):

zenoh/src/net/runtime/orchestrator.rs
oteffahi:
```suggestion
    /// Opens a new transport with `zid`.
    /// Returns `true` if successful, or if a Transport instance already exists with `zid`.
```

oteffahi (approved):

PR #1486, opened 2024-09-30
Routing refactoring I

fuzzypixelz:


zenoh/src/api/session.rs
```diff
@@ -2145,7 +2141,7 @@ impl SessionInner {
     ) -> ZResult<()> {
         tracing::trace!("liveliness.get({}, {:?})", key_expr, timeout);
         let mut state = zwrite!(self.state);
-        let id = state.liveliness_qid_counter.fetch_add(1, Ordering::SeqCst);
+        let id = self.runtime.next_id();
```
OlivierHecart, line 2144: This is meant to be used for long lived entities, not for one shot things like liveliness queries. 
It is better if each session use it's own sequencing as before, ids will be exhausted less quickly.

OlivierHecart (changes requested):

PR #1487, opened 2024-09-30
feat: add `SessionClosed` error

wyfo:


zenoh/src/lib.rs
```diff
@@ -193,7 +193,7 @@ pub mod session {
         builders::publisher::{SessionDeleteBuilder, SessionPutBuilder},
         info::{PeersZenohIdBuilder, RoutersZenohIdBuilder, SessionInfo, ZenohIdBuilder},
         query::SessionGetBuilder,
-        session::{open, OpenBuilder, Session, Undeclarable},
+        session::{open, OpenBuilder, Session, SessionClosed, Undeclarable},
```
milyin, line 196: This should be named `SessionCloseError` I suppose, by analogy with `ReplyError`, `ZAllocError`

milyin (changes requested):

zenoh/src/lib.rs
wyfo: Done

milyin (approved):

PR #1499, opened 2024-10-02
feat(storage-manager): replication supports wildcard updates

J-Loudet:
This PR adds support for Wildcard updates (be it deletion or addition) when replicating storage.

⚠️  Commits are meant to be reviewed individually (once the PR is no longer a draft).
⚠️ This PR is meant to be rebased _not squashed_.

### DIFF
--- a/plugins/zenoh-plugin-storage-manager/src/replication/core/aligner_query.rs
+++ b/plugins/zenoh-plugin-storage-manager/src/replication/core/aligner_query.rs
@@ -107,8 +107,8 @@ impl Replication {
                         .intervals
                         .get(&interval_idx)
                     {
-                        interval.sub_intervals.values().for_each(|sub_interval| {
-                            events_to_send.extend(sub_interval.events.values().map(Into::into));
+                        interval.sub_intervals().for_each(|sub_interval| {
+                            events_to_send.extend(sub_interval.events().map(Into::into));
                         });
                     }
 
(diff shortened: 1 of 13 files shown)


plugins/zenoh-plugin-storage-manager/src/replication/core/aligner_query.rs
```diff
@@ -272,60 +269,74 @@ impl Replication {
...
+                    let mut storage = self.storage_service.storage.lock().await;
+                    match storage
+                        .get(event_to_retrieve.stripped_key.clone(), "")
+                        .await
+                    {
+                        Ok(stored_data) => stored_data,
+                        Err(e) => {
+                            tracing::error!(
+                                "Failed to retrieve data associated to key < {:?} >: {e:?}",
+                                event_to_retrieve.key_expr()
```
oteffahi, line 296: The log currently looks something like
```
2024-10-08T14:58:30.624185Z ERROR tokio-runtime-worker ThreadId(02) zenoh_plugin_storage_manager::replication::core::aligner_query: Failed to retrieve data associated to key < Some("wildcard/delete/a/1") >: "Key Some(\"wildcard/delete/a/1\") is not present"
```

Maybe we should unwrap if possible, for a cleaner error message.

oteffahi (approved):
Remove unnecessary logs. Aside from that it LGTM.

PR #1504, opened 2024-10-03
Encapsulate `flume::Receiver`

fuzzypixelz:
The entirety of `flume::Receiver`'s API was exposed to anyone who declared subscribers with the default handler (which at least one person, myself). This is arguably not good:

1. `flume` could break semver (it seems to be a cool crate but mistakes can happen and you can never be too careful)
2. Even if `flume` always respects semver, we can't update it from 0.11 to a semver-incompatible version without break the _`zenoh`_ API.
3. We can't change `FifoChannel::Handler` if a new faster channel comes around.

### DIFF
--- a/zenoh/src/api/handlers/fifo.rs
+++ b/zenoh/src/api/handlers/fifo.rs
@@ -90,10 +90,8 @@ impl<T> FifoChannelHandler<T> {
 
     /// Wait for an incoming value from the channel associated with this receiver, returning an
     /// error if all senders have been dropped or the timeout has expired.
-    pub fn recv_timeout(&self, dur: Duration) -> ZResult<T> {
-        self.0
-            .recv_deadline(Instant::now().checked_add(dur).unwrap())
-            .map_err(Into::into)
+    pub fn recv_timeout(&self, duration: Duration) -> ZResult<T> {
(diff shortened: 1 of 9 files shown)


gabrik (approved):

gabrik:
The CI is not happy yet :D

zenoh/src/api/handlers/fifo.rs
```diff
@@ -41,11 +49,284 @@ impl Default for FifoChannel {
...
+                if let Err(error) = sender.send(t) {
+                    tracing::error!(%error)
+                }
+            })),
+            FifoChannelHandler(receiver),
+        )
+    }
+}
+
+impl<T> FifoChannelHandler<T> {
```
OlivierHecart, line 72: Shouldn't we mark all those functions #[inline] ?

fuzzypixelz: I think that Rust (or LLVM) functions rarely _should_ be `#[inline]`. To quote the [reference](https://doc.rust-lang.org/reference/attributes/codegen.html#the-inline-attribute):

> Note: The rustc compiler automatically inlines functions based on internal heuristics. Incorrectly inlining functions can make the program slower, so this attribute should be used with care.

PR #1505, opened 2024-10-03
refactor(storage-manager): search replication log only once

J-Loudet:
This PR refactors the Aligner part of the Replication to (i) make the code easier to read and (ii) slightly optimize the alignment process by only searching the Replication Log once.

As a bonus, an `assert!` was added when the storage manager is compiled in Debug to help us guarantee that the Replication Log only contains a single Event per key expression.

⚠️ See individual commits for details.


Charles-Schleich (approved):
Changes look good ! @J-Loudet

PR #1509, opened 2024-10-04
Feat/lookup optimization

Charles-Schleich:
Small Optimization to avoid searching through full BTreeMap for event

### DIFF
--- a/plugins/zenoh-plugin-storage-manager/src/replication/log.rs
+++ b/plugins/zenoh-plugin-storage-manager/src/replication/log.rs
@@ -215,14 +215,41 @@ impl LogLatest {
     }
 
     /// Lookup the provided key expression and, if found, return its associated [Event].
-    pub fn lookup(&self, stripped_key: &Option<OwnedKeyExpr>) -> Option<&Event> {
+    pub fn search_more_recent_event(
+        &self,
+        stripped_key: &Option<OwnedKeyExpr>,
+        timestamp: &Timestamp,
+    ) -> Option<&Event> {
         if !self.bloom_filter_event.check(stripped_key) {
             return None;
         }
 
-        for interval in self.intervals.values().rev() {
-            if let Some(event) = interval.lookup(stripped_key) {
-                return Some(event);
+        // NOTE: If the call to `get_time_classification` returns an error it either means that the
+        // Timestamp is far in the future or the clock of the Host is in a weird state. In both
+        // cases, there is nothing we can do.
+        let Ok((start_interval_idx, start_sub_interval_idx)) =
+            self.configuration.get_time_classification(timestamp)
+        else {
+            return None;
+        };
+
+        for (interval_idx, interval) in self
+            .intervals
+            .iter()
+            .filter(|(&idx, _)| idx >= start_interval_idx)
+        {
+            let search_result = if interval_idx == &start_interval_idx {
+                interval.search_more_recent_event(
+                    stripped_key,
+                    timestamp,
+                    Some(start_sub_interval_idx),
+                )
+            } else {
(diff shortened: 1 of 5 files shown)


plugins/zenoh-plugin-storage-manager/src/replication/log.rs
```diff
@@ -215,11 +215,26 @@ impl LogLatest {
...
+        if let Ok((interval, sub_idx)) = self.configuration.get_time_classification(timestamp) {
+            if let Some(event) = self
+                .intervals
+                .get(&interval)
+                .and_then(|interval| interval.get(&sub_idx))
+                .and_then(|sub_interval| sub_interval.get(stripped_key))
+            {
+                return Some(event);
+            };
+        };
```
J-Loudet, line 236: The objective of this optimization is to replace the `for` loop below this one with a loop that starts looking at `interval` (and within that _specific_ `interval` starting at `sub_idx`) and keeps on visiting greater intervals if it does not find any Event for the provided `stripped_key`.

With an example: if we have the following Replication Log,
```
- Interval(0)
   - SubInterval(0)
   - SubInterval(1)
   - SubInterval(2)

- Interval(1)
   - SubInterval(0)
   - SubInterval(1)
   - SubInterval(2)

- Interval(2)
   - SubInterval(0)
   - SubInterval(1)
   - SubInterval(2)
```

And the call to `self.configuration.get_time_classification(timestamp)` returns `(Interval(1), SubInterval(1))`. Then the lookup method should check if there is an Event with the same stripped key expression `stripped_key` but a greater timestamp in:

```
- Interval(1)
   - SubInterval(1)
   - SubInterval(2)

- Interval(2)
   - SubInterval(0)
   - SubInterval(1)
   - SubInterval(2)
```

The optimization skips the Interval and SubInterval in which, because of how the Replication Log is constructed, we know for a fact that all the Events have a lower timestamp.

Possibly the name of the function should be changed to something other than `lookup` as we are only interested in checking if there is an Event with the same stripped key expression _and a greater timestamp_.

J-Loudet (changes requested):

plugins/zenoh-plugin-storage-manager/src/replication/log.rs
Charles-Schleich: Alright, I believe that I better understand the implementation and have updated function to iterate forward from `interval_idx, sub_idx` returned from `configuration.get_time_classification`

In the case that `get_time_classification` returns an error, i am ignoring the error and returning None.

plugins/zenoh-plugin-storage-manager/src/replication/log.rs
```diff
@@ -224,22 +224,28 @@ impl LogLatest {
...
-            if let Some(event) = interval.lookup(stripped_key) {
-                return Some(event);
+            // Search forward all intervals, including all Sub Intervals
+            interval_idx.0 = interval_idx.0 + 1;
+            while let Some(interval) = self.intervals.get(&interval_idx) {
+                if let Some(event) = interval.lookup(stripped_key) {
+                    return Some(event);
+                }
+                interval_idx.0 = interval_idx.0 + 1;
             }
```
J-Loudet: We have no guarantee that the intervals and sub-intervals are continuous, I think an iterator coupled with a filter is a viable approach.

For instance we might have:
```
- Interval(0)
   - SubInterval(0)
   - SubInterval(2)
- Interval(2)
   - SubInterval(1)
- Interval(4)
   - SubInterval(0)
   - SubInterval(2)
```

J-Loudet (changes requested):
Could you also squash your last commits? (from 48f0e6a to f3a4805)
That would make it easier to follow.

J-Loudet:
Done in #1566

PR #1510, opened 2024-10-04
deadline adoption for multi-fragment message

### LINKED ISSUE #1494: QoS: message fragments dropped
### Describe the bug

During investigation of unixsocket failing on macos13, I found that our batching algo allows dropping single fragments of big message if transmitted with CC==Drop.

Due to default small batch count, if transmitting big messages with sizes >> batch size, there can be a situation when all transmitted messages will have at least one dropped fragment, which makes it impossible to defragment them and leaves RX with no valuable data at all. That exactly what happened with our macos13 runner.

It is completely useless to transmit message if at least one of it's fragments is dropped, so in order to fix our macos13 test and have more optimal CC operation we need to fix our CC algo to:
- drop only the whole message, not one of it's fragments
- block if congestion happens in the middle of sending a multi-fragment message

### To reproduce

-

### System info

-

yellowhatter:
[closes #1494 ]

### DIFF
--- a/io/zenoh-transport/src/common/pipeline.rs
+++ b/io/zenoh-transport/src/common/pipeline.rs
@@ -127,6 +127,40 @@ impl StageInMutex {
     }
 }
 
+struct Deadline {
+    deadline: Option<Option<Instant>>,
+    wait_time: Duration,
+}
+
+impl Deadline {
+    fn new(wait_time: Duration) -> Self {
+        let deadline = Self::calc_deadline(wait_time);
+        Self {
+            deadline,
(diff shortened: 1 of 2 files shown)


yellowhatter:
[closes #1494]

yellowhatter:
Briefly, improvements:
- fix tests regression for multi-fragment message problem
- fix perf regression for some platforms caused by too frequent time query
- improve batching code readability (enum instead of cascaded options)

fuzzypixelz (approved):

Issue #1507, opened 2024-10-03
rework list of predefined encodings


milyin:
### Describe the release item

There are several predefined encodings which became ambigious after serialization rework in https://github.com/eclipse-zenoh/zenoh/issues/1474. The main reason of ambiguity is that the prefix `ZENOH_` for encoding doesn't say anything is serialization used or not.

E.g. is `ZENOH_STRING` means that `ZBytes::to_string()` was used or string was serialized into `ZBytes`. The difference is important: in second case the string is additionally prefixed by length.
Same for `ZENOH_BYTES` value: is it `ZBytes::to_bytes()` or serialization of vector of `u8`. Difference is the same: just `to_bytes()` doesn't add any prefix

The proposal is to 
- keep encodings `ZENOH_STRING` and `ZENOH_BYTES`, they should correspond to `ZBytes::to_string()`, `ZBytes::to_bytes()`
- add type `ZENOH_SERIALIZED`. it's suffix can be used to store serializing schema
- remove types `ZENOH_INT8`, `ZENOH_INT16`, etc

The fixes should be also propagated to bindings
- [ ] https://github.com/eclipse-zenoh/zenoh/pull/1520
- [ ] zenoh-c
- [ ] zenoh-pico
- [ ] zenoh-cpp
- [ ] zenoh-python
- [ ] kotlin
- [ ] typescript

milyin:
@wyfo @kydos if this that you finally decided about this issue?

PR #1512, opened 2024-10-04
encoding list updated after moving serialization to zenoh-ext

milyin:
fix for https://github.com/eclipse-zenoh/zenoh/issues/1507

PR #1513, opened 2024-10-04
fix: remove specific encoding constants related to serialization

wyfo:
fixes https://github.com/eclipse-zenoh/zenoh/issues/1507

### DIFF
--- a/zenoh/src/api/encoding.rs
+++ b/zenoh/src/api/encoding.rs
@@ -14,18 +14,14 @@
 use std::{borrow::Cow, convert::Infallible, fmt, str::FromStr};
 
 use phf::phf_map;
-use zenoh_buffers::{ZBuf, ZSlice};
+use zenoh_buffers::ZSlice;
 use zenoh_protocol::core::EncodingId;
-#[cfg(feature = "shared-memory")]
-use zenoh_shm::api::buffer::{zshm::ZShm, zshmmut::ZShmMut};
-
-use super::bytes::ZBytes;
 
 /// Default encoding values used by Zenoh.
 ///
 /// An encoding has a similar role to *Content-type* in HTTP: it indicates, when present, how data should be interpreted by the application.
 ///
-/// Please note the Zenoh protocol does not impose any encoding value nor it operates on it.
+/// Please note the Zenoh protocol does not impose any encoding value, nor it operates on it.
 /// It can be seen as some optional metadata that is carried over by Zenoh in such a way the application may perform different operations depending on the encoding value.
 ///
 /// A set of associated constants are provided to cover the most common encodings for user convenience.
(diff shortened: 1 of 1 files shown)


milyin (changes requested):
Looks good for me. I did similar PR (https://github.com/eclipse-zenoh/zenoh/pull/1512), I think they can be combined in the following way:
- explicitly say that `ZENOH_BYTES` and `ZENOH_STRING` corresponds to `to_bytes()` and `try_to_string()` methods
- reference zenoh-ext in `ZENOH_SERIALIZATION`
- remove `ZENOH_ERROR` - it doesn't make sense at all. The `ReplyError` type contains encoding field whcih is set to data packet, this leaves no place to special encoding for `ReplyError`
- (questionable) avoid full renumbering of encoding ids and keep the range 0..15 for possible future `ZENOH_*` encodings

wyfo:
> explicitly say that ZENOH_BYTES and ZENOH_STRING corresponds to to_bytes() and try_to_string() methods

Encoding is neither mandatory nor automatic, so I don't see why we would assign an encoding to a method of `ZBytes`. You can encode a string without specifying an encoding, and use `try_to_string` on the other side.
Also, if you encode some JSON data, you will use either `to_bytes` or `try_to_string` to retrieve the bytes and read your JSON, so the methods are not really related to the encoding.

> reference zenoh-ext in ZENOH_SERIALIZATION
Will do

> remove ZENOH_ERROR - it doesn't make sense at all. The ReplyError type contains encoding field whcih is set to data packet, this leaves no place to special encoding for ReplyError

If you say so

> (questionable) avoid full renumbering of encoding ids and keep the range 0..15 for possible future ZENOH_* encodings

Questionable indeed. Honestly, I want to remove all `ZENOH_XXX` encoding, so I don't want to support that idea.

milyin:
> > explicitly say that ZENOH_BYTES and ZENOH_STRING corresponds to to_bytes() and try_to_string() methods
> 
> Encoding is neither mandatory nor automatic, so I don't see why we would assign an encoding to a method of `ZBytes`. You can encode a string without specifying an encoding, and use `try_to_string` on the other side. Also, if you encode some JSON data, you will use either `to_bytes` or `try_to_string` to retrieve the bytes and read your JSON, so the methods are not really related to the encoding.

The `ZENOH_` encodings existed for a long time and I see no reason to remove them now. They makes sense in zenoh infrastructure and thay means that data can be interpleted using zenoh's tools. Now it can be some unknown blob "ZENOH_BYTES", an UTF-8 string "ZENOH_STRING", is it a zenoh serialized data "ZENOH_SERIALIZED".  So it makes sens to keep them and reference to corresponding zenoh's functions

> 
> > (questionable) avoid full renumbering of encoding ids and keep the range 0..15 for possible future ZENOH_* encodings
> 
> Questionable indeed. Honestly, I want to remove all `ZENOH_XXX` encoding, so I don't want to support that idea.

I'm against removing `ZENOH_XXX` and if we keeping them, it makes sense to keep range for them

milyin:
@wyfo : discussed it on morning meeting with @kydos . We need to keep ZENOH_ prefixes. I'd also would recommend to keep the gap 0..15 for ZENOH_ prefixes - this gap occured naturally and it can be useful in the future

milyin:
replaced with https://github.com/eclipse-zenoh/zenoh/pull/1512

PR #1514, opened 2024-10-04
Add timeout for TLS handshakes

fuzzypixelz:
After establishing a TCP connections, the TLS link waits for the TLS handshake to complete in `rustls`. If the TLS handshake never takes place, then the TLS link waits _forever_ and no one else can establish a Zenoh session. It is thus possible to block a router with a TLS on `tls/127.0.0.1:7447` listener from servicing any incoming connections with the following:

```rust
fn main() {
    let _stream = TcpStream::connect("127.0.0.1:7447")?;
    thread::park();
}
```

### DIFF
--- a/io/zenoh-links/zenoh-link-tls/src/lib.rs
+++ b/io/zenoh-links/zenoh-link-tls/src/lib.rs
@@ -81,8 +81,6 @@ zconfigurable! {
     // Amount of time in microseconds to throttle the accept loop upon an error.
     // Default set to 100 ms.
     static ref TLS_ACCEPT_THROTTLE_TIME: u64 = 100_000;
-    /// The time duration in milliseconds to wait for the TLS handshake to complete.
-    static ref TLS_HANDSHAKE_TIMEOUT_MS: u64 = 10_000;
 }
 
 pub mod config {
@@ -110,4 +108,8 @@ pub mod config {
 
     pub const TLS_VERIFY_NAME_ON_CONNECT: &str = "verify_name_on_connect";
     pub const TLS_VERIFY_NAME_ON_CONNECT_DEFAULT: bool = true;
+
+    /// The time duration in milliseconds to wait for the TLS handshake to complete.
+    pub const TLS_HANDSHAKE_TIMEOUT_MS: &str = "tls_handshake_timeout_ms";
+    pub const TLS_HANDSHAKE_TIMEOUT_MS_DEFAULT: u64 = 10_000;
 }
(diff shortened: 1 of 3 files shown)


io/zenoh-links/zenoh-link-tls/src/unicast.rs
```diff
@@ -436,9 +437,18 @@ async fn accept_task(
                         };
 
                         // Accept the TLS connection
-                        let tls_stream = match acceptor.accept(tcp_stream).await {
-                            Ok(stream) => TlsStream::Server(stream),
+                        let tls_stream = match tokio::time::timeout(
+                            dbg!(Duration::from_millis(*TLS_HANDSHAKE_TIMEOUT_MS)),
```
gabrik, line 441: remove the `dbg!`

io/zenoh-links/zenoh-link-tls/src/lib.rs
```diff
@@ -81,6 +81,8 @@ zconfigurable! {
     // Amount of time in microseconds to throttle the accept loop upon an error.
     // Default set to 100 ms.
     static ref TLS_ACCEPT_THROTTLE_TIME: u64 = 100_000;
+    /// The time duration in milliseconds to wait for the TLS handshake to complete.
+    static ref TLS_HANDSHAKE_TIMEOUT_MS: u64 = 10_000;
```
gabrik, line 85: should this be configurable?

gabrik (changes requested):

gabrik:
I'm thinking if we should also spawn a task on accept, or have a task pool where we sent the accepted socket before waiting for the the TLS handshake to complete.

fuzzypixelz:
> I'm thinking if we should also spawn a task on accept, or have a task pool where we sent the accepted socket before waiting for the the TLS handshake to complete.

Yes, this is effectively what the [`tcp-listener`](https://docs.rs/tls-listener/latest/tls_listener/) crate does.

We should be careful as `tcp-listener` has had [`CVE-2024-28854`](https://www.cve.org/CVERecord?id=CVE-2024-28854) where it was vulnerable to a slow-loris DoS attack.

gabrik:
I think we need to add a key in order to make the timeout configurable, see: https://github.com/eclipse-zenoh/zenoh/blob/tls-handshake-timeout/io/zenoh-links/zenoh-link-tls/src/lib.rs#L88

fuzzypixelz:
> I think we need to add a key in order to make the timeout configurable, see: https://github.com/eclipse-zenoh/zenoh/blob/tls-handshake-timeout/io/zenoh-links/zenoh-link-tls/src/lib.rs#L88

Those keys are for runtime configuration, but `TLS_HANDSHAKE_TIMEOUT_MS` is a compile-time constant. Maybe we should make this runtime-configurable.

gabrik:
> Those keys are for runtime configuration, but `TLS_HANDSHAKE_TIMEOUT_MS` is a compile-time constant. Maybe we should make this runtime-configurable.

I think it must be runtime-configurable as the timeout depends the deployment

gabrik (approved):

oteffahi (approved):

PR #1516, opened 2024-10-07
Fix undesired propagation of interests to routers from peers

OlivierHecart:


zenoh/src/net/routing/hat/p2p_peer/interests.rs
```diff
@@ -147,7 +147,8 @@ impl HatInterestTrait for HatCode {
         };
         for dst_face in tables.faces.values_mut().filter(|f| {
             f.whatami == WhatAmI::Router
-                || (options.tokens()
+                || (face.whatami == WhatAmI::Client
```
fuzzypixelz, line 150: I suggest writing this as:

```rust
match dst_face.whatami {
    WhatAmI::Router => true,
    WhatAmI::Client => false,
    WhatAmI::Peer => options.tokens() 
        && src_face.whatami == WhatAmI::Client
        && mode == InterestMode::Current 
        && !initial_interest(f).map(|i| i.finalized).unwrap_or(true))
}
```

So that the condition reads more like a case by case analysis.

fuzzypixelz, line 150: Under which circumstances do routers send peer-to-peer peers interest? A simple grep in `zenoh/src/net/routing/hat` shows only results in `client` and `p2p_peer`.

imstevenpmwork: Just peaking over here but maybe:
```rust
WhatAmI::Peer => {
        options.tokens() 
            && src_face.whatami == WhatAmI::Client
            && mode == InterestMode::Current
            && !initial_interest(f).map_or(true, |i| i.finalized)
    }
```
Instead?

PR #1517, opened 2024-10-07
zenoh::config::default() restored

milyin:
it was removed on cleaning up type leaks on inport zenoh_config::*, restored now


fuzzypixelz (changes requested):
@milyin @kydos I don't think it should be restored, users can simply write `zenoh::Config::default`, which exactly as many characters as `zenoh::config::default`.

(Assming the argument for adding this back in is ease of typing).

wyfo:
I think it was removed on purpose, and it's indeed redundant with `zenoh::Config::default()`, so it may bring confusion.

PR #1520, opened 2024-10-08
Encoding and timestamp serialization remove from zenoh-ext

milyin:
fixes https://github.com/eclipse-zenoh/zenoh/issues/1507 also


zenoh/src/api/encoding.rs
```diff
@@ -582,21 +466,8 @@ impl Encoding {
...
-        8u16 => "zenoh/uint32",
-        9u16 => "zenoh/uint64",
-        10u16 => "zenoh/uint128",
-        11u16 => "zenoh/float32",
-        12u16 => "zenoh/float64",
-        13u16 => "zenoh/bool",
-        14u16 => "zenoh/string",
-        15u16 => "zenoh/error",
+        1u16 => "zenoh/string",
+        15u16 => "zenoh/serialized",
```
wyfo, line 470: You really want to "reserve" encoding slots?

zenoh-ext/src/serialization.rs
```diff
@@ -608,12 +545,11 @@ mod tests {
     fn timestamp_serialization() {
```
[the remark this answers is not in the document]
milyin: removed. Made it to make sure that we still can serialize timestamp and as an example, but yes, this is excessive

zenoh/src/api/encoding.rs
```diff
@@ -86,147 +82,35 @@ impl Encoding {
...
-        id: 13,
-        schema: None,
-    });
     /// A UTF-8 string.
     ///
     /// Constant alias for string: `"zenoh/string"`.
     ///
-    /// Usually used for types: `String`, `&str`, `Cow<str>`, `char`.
+    /// This encoding supposes that the payload was created with [`ZBytes::from::<String>`](crate::bytes::ZBytes::from) or similar
+    /// (`&str`, `Cow<str>`, `char`) and it's data can be acquired with [`ZBytes::try_to_string()`](crate::bytes::ZBytes::try_to_string) without an error.
```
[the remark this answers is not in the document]
milyin: removed `char`, added similar comment to bytes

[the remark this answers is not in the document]
milyin: fixed

zenoh/src/api/encoding.rs
milyin: I think this makes sense. Later we can even restore primitive types encoding to these slots, with names like `zenoh/serialized;u8`, `zenoh/serialized;i16`, etc - when the schema for zenoh encoding is elaborated (this is @kydos 's idea)

wyfo: Then you will surely need more than 2..16 range, so why not reserving more? Why do we care having zenoh encodings first? (except for `zenoh/bytes` which must obviously be 0)

milyin: Mostly because it allows us not to change existing numeration. Yes, it's internal and as we know these codes are not used anywhere, but as it costs nothing to us, better not to break compatibility when we can.
But this is really minor question. If you are strongly against this gap, let's remove it, it's cheaper than discuss it too long.

milyin: @DenisBiryukov91 is also against the gap and this makes zenoh-pico a bit more complex. Then let's renumber them, fixing

milyin: renumbered

PR #1544, opened 2024-10-17
fallible serialization

milyin:
- `z_try_serialize` function added
- trait `TrySerialize` added. The trait `Serialize` remains for infallible serialization
   - For basic types and containers which allows infallible serialization this trait is implemeted as `TrySerialize<Error = Infallible>`
   - The `Serialize` trait is implemeted automatically for all types which implements `TrySerialize<Error = Infallible>`
   - It's possible to implement `Serialize` explicitly for types which already implements `TrySerialize<Error = ZSerializeError>` or any other error. This is necessary for e.g. `HashMap`
 - fallible serialization of `CString` added: in demonstration purpose and may be for future use in zenoh-c

### DIFF
--- a/zenoh-ext/src/serialization.rs
+++ b/zenoh-ext/src/serialization.rs
@@ -36,15 +36,11 @@ impl fmt::Display for ZDeserializeError {
 impl std::error::Error for ZDeserializeError {}
 
 #[derive(Debug)]
-pub struct ZSerializeError(String);
+
+pub struct ZSerializeError;
 impl fmt::Display for ZSerializeError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "serialization error: {}", self.0)
-    }
-}
-impl From<String> for ZSerializeError {
-    fn from(s: String) -> Self {
-        Self(s)
+        write!(f, "serialization error")
     }
 }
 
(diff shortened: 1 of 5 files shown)


zenoh-ext/src/serialization.rs
```diff
@@ -33,18 +35,58 @@ impl fmt::Display for ZDeserializeError {
 }
 impl std::error::Error for ZDeserializeError {}
 
-fn default_serialize_n<T: Serialize>(slice: &[T], serializer: &mut ZSerializer) {
+#[derive(Debug)]
+pub struct ZSerializeError(String);
```
wyfo, line 39: Why not using a ZST as it is the case for deserialization?

wyfo: Why not a `Box<dyn Error + Send + Sync>`?

zenoh-ext/src/serialization.rs
```diff
@@ -174,13 +245,39 @@ impl ZSerializer {
...
     }
 
+    /// Serialize the given iterator into a [`ZSerializer`].
+    /// Returns an error if serialization of an element fails.
+    ///
+    /// Sequence serialized with this method may be deserialized with [`ZDeserializer::deserialize_iter`].
+    /// See [Zenoh serialization format RFC][1].
+    ///
+    /// [1]: https://github.com/eclipse-zenoh/roadmap/blob/main/rfcs/ALL/Serialization.md#sequences
+    pub fn try_serialize_iter<T: TrySerialize, I: IntoIterator<Item = T>>(
```
wyfo, line 266: Why no `try_serialize` method?

zenoh-ext/src/serialization.rs
milyin: What is a ZST ? About Box - yes, agree, fixing

milyin: made it ZST for now

zenoh-ext/src/serialization.rs
milyin: added

Mallets:
@milyin At the end we decided to not include this PR in 1.0.0. 
Therefore I believe it can be safely closed as no longer relevant.

PR #1548, opened 2024-10-18
Don't close multicast transport on OS error 55

OlivierHecart:
When sending high throughput through multicast transport the following could occur:
```
TX task failed: Write error on UDP link 192.168.0.1:51381 => 224.0.0.1:17511: No buffer space available (os error 55) at io/zenoh-links/zenoh-link-udp/src/multicast.rs:88.
Closing multicast transport on udp/224.0.0.1:17511
```
Now Zenoh logs and ignores the error.

### DIFF
--- a/io/zenoh-links/zenoh-link-udp/src/multicast.rs
+++ b/io/zenoh-links/zenoh-link-udp/src/multicast.rs
@@ -81,14 +81,23 @@ impl LinkMulticastTrait for LinkMulticastUdp {
     }
 
     async fn write(&self, buffer: &[u8]) -> ZResult<usize> {
-        self.unicast_socket
+        match self
+            .unicast_socket
             .send_to(buffer, self.multicast_addr)
             .await
-            .map_err(|e| {
-                let e = zerror!("Write error on UDP link {}: {}", self, e);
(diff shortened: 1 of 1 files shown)


io/zenoh-links/zenoh-link-udp/src/multicast.rs
```diff
@@ -81,14 +81,24 @@ impl LinkMulticastTrait for LinkMulticastUdp {
...
             .await
-            .map_err(|e| {
-                let e = zerror!("Write error on UDP link {}: {}", self, e);
-                tracing::trace!("{}", e);
-                e.into()
-            })
+        {
+            Ok(size) => Ok(size),
+            std::io::Result::Err(e) => {
+                if let Some(55) = e.raw_os_error() {
```
Mallets, line 91: The error code handling should be behind a platform-specific feature gate.
The code `55` means `no buffer space available` on *mac os x* (which version?) but it not guaranteed the same code is used by other platforms.

Mallets (changes requested):
In general it LGTM. Please see my comment on platform-specific codes.

Mallets (approved):

PR #1551, opened 2024-10-19
Fix for Err: Too many open files

Ricardicus:
I am using zenoh in a project (called [chat-pgp](https://github.com/Ricardicus/chat-pgp)), where I open sessions with a zenoh configuration read from file.

When launching multiple processes this way, I run into this error:

`
Message:  called Result::unwrap() on an Err value: Too many open files (os error 24) at /Users/rickard/.cargo/registry/src/index.crates.io-6f17d22bba15001f/zenoh-config-1.0.0-alpha.6/src/lib.rs:747.`

Looking at the code, it seems like the file, opened using the standard library, isn't closed.
Just to be sure I added this, and now the problem is gone for me. Not sure why the file descriptors aren't closed
in the original code, but the problem is solved for me like this.

### DIFF
--- a/commons/zenoh-config/src/lib.rs
+++ b/commons/zenoh-config/src/lib.rs
@@ -776,32 +776,35 @@ impl Config {
     }
 
     fn _from_file(path: &Path) -> ZResult<Config> {
-        match std::fs::File::open(path) {
-            Ok(mut f) => {
-                let mut content = String::new();
-                if let Err(e) = f.read_to_string(&mut content) {
-                    bail!(e)
-                }
-                match path
-                    .extension()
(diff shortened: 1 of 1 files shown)


Ricardicus:
Nah, this fix still didn't fix it.
I will open this as an issue instead.

The way I know to reproduce the error is by using my program.

Run these three at the same time:
./target/debug/relay -z zenoh/config_localhost.json5
./target/debug/client -z zenoh/config_localhost.json5
./target/debug/client -z zenoh/config_localhost.json5

Do the "init 1" initialize a chat session.. then the crash occurs..

PR #1553, opened 2024-10-21
P2p peers don't send OAM messages when gossip disabled

OlivierHecart:


fuzzypixelz:
Closed in favor of https://github.com/eclipse-zenoh/zenoh/pull/1557.

PR #1554, opened 2024-10-21
Fix KeepAlive transmission interval

Mallets:
Fix KeepAlive regression introduced during Tokio porting. Send a KeepAlive only in the case no messages have been sent in a KeepAlive period.


YuanYuYuan:
Thanks @Mallets. I misunderstood the keep_alive as a periodic event and thus messed up the logic. This LGTM.

PR #1555, opened 2024-10-21
No more MIN_FREE_CHUNK_SIZE limitation

yellowhatter:
Do not use MIN_FREE_CHUNK_SIZE limitation - the provider's alignment will do the trick!


YuanYuYuan:
LGTM

PR #1557, opened 2024-10-21
Fix bug in return_conditions declares leading to wait for the full delay

OlivierHecart:


fuzzypixelz (commented):
This is rather tangential but I have a question about a change in #1535:

https://github.com/eclipse-zenoh/zenoh/blob/3f97bfc566a3a0f025c74f9205fce0deb790d2fd/zenoh/src/net/routing/hat/p2p_peer/gossip.rs#L454-L460

After scouting a router that didn't gossip about any other nodes, we only return in `open` if `open/return_conditions/declares` is true. Even though the router sends no initial interest declarations.

I tested with one router and one peer configured with `open/return_conditions/declares:false` and `open/return_conditions/connect_scouted:true`. The peer doesn't return from `open` before the scouting delay. Although AFAICT no declarations are expected from the router anyway.

fuzzypixelz (approved):

PR #1564, opened 2024-10-25
Close TLS and Quic links on certificate expiration

oteffahi:
Quic and TLS links currently do not close when remote certificate expires. This PR makes instances monitor the expiration of the remote certificate chain, closing the link at the moment of expiration.

This is configurable via the `transport.link.tls.close_link_on_expiration` key in the config file. It is set by default to `false` to maintain the old behavior as default.

### DIFF
--- a/io/zenoh-link-commons/src/tls.rs
+++ b/io/zenoh-link-commons/src/tls.rs
@@ -135,6 +135,12 @@ pub mod expiration {
         const MAX_EXPIRATION_SLEEP_DURATION: tokio::time::Duration =
             tokio::time::Duration::from_secs(600);
 
+        tracing::trace!(
+            "Expiration task started for link {} => {}",
+            expiration_info.src_locator,
+            expiration_info.dst_locator,
+        );
+
         loop {
             let now = OffsetDateTime::now_utc();
             if expiration_info.expiration_time <= now {
--- a/io/zenoh-link-commons/src/tls.rs
+++ b/io/zenoh-link-commons/src/tls.rs
@@ -146,13 +146,9 @@ pub mod expiration {
         expiration_time: OffsetDateTime,
         token: CancellationToken,
     ) {
-        // NOTE: should expose or tune sleep duration
-        const MAX_EXPIRATION_SLEEP_DURATION: tokio::time::Duration =
-            tokio::time::Duration::from_secs(600);
-
-        loop {
-            let now = OffsetDateTime::now_utc();
-            if expiration_time <= now {
+        tokio::select! {
+            _ = token.cancelled() => {},
+            _ = sleep_until_date(expiration_time) => {
                 // close link
                 if let Some(link) = link.upgrade() {
                     tracing::warn!(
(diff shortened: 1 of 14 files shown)


oteffahi:
Commit [efb6cb931fa34ab9ab2df5d4912390684c539432](https://github.com/eclipse-zenoh/zenoh/pull/1564/commits/efb6cb931fa34ab9ab2df5d4912390684c539432) changes the approach to one expiration task per link instance, as the previous implementation had a design flaw where `Weak` references to closed links wouldn't be cleaned up if said links were closed by the transport (i.e not closed by expiration task), leading to potential memory exhaustion attacks.

io/zenoh-link-commons/src/tls.rs
```diff
@@ -86,3 +86,121 @@ impl WebPkiVerifierAnyServerName {
...
+                            expiration_info.src_locator,
+                            expiration_info.dst_locator,
+                            e
+                        )
+                    }
+                }
+                break;
+            }
+            // next sleep duration is the minimum between MAX_EXPIRATION_SLEEP_DURATION and the duration till next expiration
+            // this mitigates the unsoundness of using `tokio::time::sleep_until` with long durations
```
OlivierHecart, line 166: Why is it bad to use `tokio::time::sleep_until` with long durations ?

OlivierHecart: ok

OlivierHecart (approved):

io/zenoh-link-commons/src/tls.rs
```diff
@@ -86,3 +86,104 @@ impl WebPkiVerifierAnyServerName {
...
+    }
+
+    // Cleanup expiration task when manager is dropped
+    impl Drop for LinkCertExpirationManager {
+        fn drop(&mut self) {
+            if let Some(handle) = self.handle.take() {
+                self.token.cancel();
+                zenoh_runtime::ZRuntime::Acceptor.block_in_place(async {
+                    let _ = handle.await;
+                })
```
oteffahi, line 136: @OlivierHecart @Mallets is this worth the cost, or should we just call `self.token.cancel()` without polling the handle?
During a review with @wyfo we reached the conclusion that the cost of `block_in_place` outweights the benefit here.

io/zenoh-links/zenoh-link-quic/src/unicast.rs
```diff
@@ -416,19 +453,50 @@ async fn accept_task(
                             }
                         };
                         let dst_addr = quic_conn.remote_address();
+                        let dst_locator = Locator::new(QUIC_LOCATOR_PREFIX, dst_addr.to_string(), "")?;
                         // Get Quic auth identifier
                         let auth_id = get_cert_common_name(&quic_conn)?;
 
+                        // Get certificate chain expiration
+                        let maybe_expiration_time = get_cert_chain_expiration(&quic_conn)?;
```
wyfo, line 461: use something like 
```
tls_close_link_on_expiration
    .then(|| get_cert_chain_expiration(&quic_conn))
    .transpose()?
    .flatten()
// OR
if tls_close_link_on_expiration {
    get_cert_chain_expiration(&quic_conn)?
} else {
    None
}
```

io/zenoh-link-commons/src/tls.rs
```diff
@@ -86,3 +86,104 @@ impl WebPkiVerifierAnyServerName {
...
+            }
+        }
+    }
+
+    async fn expiration_task(
+        link: Weak<dyn LinkUnicastTrait>,
+        src_addr: SocketAddr,
+        dst_addr: SocketAddr,
+        link_type: String,
```
wyfo, line 145: If we need a `String` later, then we can change the type. For now, `&'static str` is enough.

oteffahi:
Following commits rework the implementation to dodge the performance issues of `block_in_place` for expiration_task cleanup , and also addresses another issue that arises from the expiration task and transport potentially calling `link.close().await` concurrently.

OlivierHecart (approved):

PR #1566, opened 2024-10-28
refactor(storage-manager): replication log lookup

J-Loudet:
This commit optimises the method `lookup` to only lookup for newer (and
thus rename it to `lookup_newer`) Event.

As it is only looking for a newer Event, this method skips the Intervals
and SubIntervals that, by construction of the Replication Log, only
contain Events with lower timestamps.

### DIFF
--- a/plugins/zenoh-plugin-storage-manager/src/replication/classification.rs
+++ b/plugins/zenoh-plugin-storage-manager/src/replication/classification.rs
@@ -126,8 +126,8 @@ impl Interval {
     }
 
     /// Returns an iterator over the [SubInterval]s contained in this `Interval`.
-    pub(crate) fn sub_intervals(&self) -> impl Iterator<Item = &SubInterval> {
-        self.sub_intervals.values()
+    pub(crate) fn sub_intervals(&self) -> impl Iterator<Item = (&SubIntervalIdx, &SubInterval)> {
+        self.sub_intervals.iter()
     }
 
     /// Returns, if one exists, a reference over the [SubInterval] matching the provided
--- a/plugins/zenoh-plugin-storage-manager/src/replication/classification.rs
+++ b/plugins/zenoh-plugin-storage-manager/src/replication/classification.rs
@@ -39,6 +39,21 @@ pub(crate) enum EventRemoval {
     RemovedOlder(Event),
 }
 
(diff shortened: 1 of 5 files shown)


plugins/zenoh-plugin-storage-manager/src/replication/classification.rs
```diff
@@ -397,8 +401,25 @@ impl SubInterval {
         EventRemoval::NotFound
     }
 
-    fn lookup(&self, event_to_lookup: &EventMetadata) -> Option<&Event> {
-        self.events.get(&event_to_lookup.log_key())
+    /// Looks up the key expression of the provided `event_to_lookup` in this [SubInterval] and,
+    /// depending on its timestamp and an [Event] has been found, returns the [EventLookup].
```
oteffahi, line 405: Nitpick
```suggestion
    /// depending on its timestamp and if an [Event] has been found, returns the [EventLookup].
```

plugins/zenoh-plugin-storage-manager/src/replication/classification.rs
```diff
@@ -397,8 +401,25 @@ impl SubInterval {
...
+    /// `Older` is returned.
+    ///
+    /// If this SubInterval contains no Event with the same key expression, `NotFound` is returned.
+    pub(crate) fn lookup(&self, event_to_lookup: &EventMetadata) -> EventLookup {
+        match self.events.get(&event_to_lookup.log_key()) {
+            Some(event) => {
+                if event.timestamp >= event_to_lookup.timestamp {
+                    EventLookup::NewerOrIdentical(event)
+                } else {
+                    EventLookup::Older
```
oteffahi, line 418: Are we sure it will never be useful to get the event in case of an `EventLookup::Older`?
IMO semantically the event exists, we should return it and it should be up to the calling logic to use it or not (unless this has a significant impact on performance).

oteffahi (changes requested):
The logic makes sense to me. I also ran the intergration tests and it looks good.

Found two nitpicks to be addressed 😅

plugins/zenoh-plugin-storage-manager/src/replication/classification.rs
J-Loudet: Fixed 👍

oteffahi (approved):
LGTM

PR #1569, opened 2024-10-29
Able to check Tokio Runtime is alive inside is_closed.

evshary:
The draft PR to deal with the issue
https://github.com/ZettaScaleLabs/rmw_zenoh/issues/26

Here is what we do:
1. Provide a clear panic log to tell developers not to use Zenoh API inside atexit.
2. Check Tokio Runtime availability inside `is_closed` API. This can help developers avoid calling Zenoh API in the wrong place.

### DIFF
--- a/commons/zenoh-runtime/src/lib.rs
+++ b/commons/zenoh-runtime/src/lib.rs
@@ -129,12 +129,23 @@ impl ZRuntime {
     where
         F: Future<Output = R>,
     {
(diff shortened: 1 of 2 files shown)


evshary:
It would be better to check it on rmw_zenoh. Let's replace it with https://github.com/ZettaScaleLabs/rmw_zenoh/pull/37

PR #1571, opened 2024-10-29
add empty gravestone values for Sample, Reply, ReplyError, Query and Hello

DenisBiryukov91:
add empty gravestone values for Sample, Reply, ReplyError, Query and Hello - required to support move operations in C++

### DIFF
--- a/commons/zenoh-config/src/wrappers.rs
+++ b/commons/zenoh-config/src/wrappers.rs
@@ -119,6 +119,17 @@ impl Hello {
     pub fn whatami(&self) -> WhatAmI {
         self.0.whatami
     }
+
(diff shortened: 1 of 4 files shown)


DenisBiryukov91:
@Mallets. Yes it is still valid, although the related task to provide functionality to move out of loaned types in zenoh-c and make std::move in c++ work more closely to how users would expect, has quite low priority.

milyin (approved):
The approach for Query can be changed (better to have real empty state, but this will require too much rework of other code), but this variant is ok.
The goal of this update is to provide "dummy" states for zenoh-c "take from loaned mut" operation - see PR https://github.com/eclipse-zenoh/zenoh-c/pull/718

Issue #1562, opened 2024-10-24
ACL unfiltered messages


nduchaux:
### Describe the bug

Hi all,

ACL seems buggy.
I am playing with my own program and the 2 example from zenoh: z_sub_liveliness and z_sub.
Here is my program:
```rust
let _liveness_token = session.liveliness().declare_token("alive").await.unwrap();
session.put("test", "test").await.unwrap();
let subscriber = session.declare_subscriber("**").await.unwrap();
```
Without ACL, here is the result from z_sub_liveliness:
```
>> [LivelinessSubscriber] New alive token ('alive')
>> [LivelinessSubscriber] Dropped token ('alive')
```
and from z_sub:
```
>> [Subscriber] Received PUT ('test': 'test')
```
So far so good.
Now I set an ACL on my program (but not on the examples):
```js
  access_control: {
         "enabled": true,
         "default_permission": "deny",
         "rules": [],
         "subjects": [],
         "policies": [],
  }
```
So basically, deny all from everywhere and anybody.

Here is the result from z_sub_liveliness:
```
>> [LivelinessSubscriber] New alive token ('alive')
2024-10-24T15:42:01.796836Z ERROR rx-3 ThreadId(25) zenoh::net::routing::dispatcher::pubsub: Face{3, 905d53f9087a70a728cced371077c1ac} Undeclare unknown subscriber 2
>> [LivelinessSubscriber] Dropped token ('alive')
```
The liveliness token is still received (not good, I wanted to deny all).
The subscrition to "**" is also received (or maybe only the unsubscribe) somehow but partly ignored (thus the error on the unsubscribe)

Here is the result from z_sub:
```
2024-10-24T15:42:01.796831Z ERROR rx-0 ThreadId(22) zenoh::net::routing::dispatcher::pubsub: Face{3, 905d53f9087a70a728cced371077c1ac} Undeclare unknown subscriber 2
```
So the sample is not received (good!).
But same error with the unsubscribe.

Zenoh v1.0.0

### To reproduce

see bug description

### System info

Windows
Zenoh v1.0.0

oteffahi:
You configured ACL to deny all, but that does not mean all Zenoh messages are affected, only the ones supported by ACL are. In this case, liveliness tokens are not filtered in ACL in 1.0.0, which explains why the linveliness subscriber received it.

The subscription is not received, but the unsubscribe is, which is indeed a bug.

PR #1573, opened 2024-10-30
Fix(ACL): Apply queryable and subscriber declaration filters on their respective undeclarations

oteffahi:
Addresses the unfiltered Undeclare messages reported in https://github.com/eclipse-zenoh/zenoh/issues/1562

### DIFF
--- a/zenoh/src/net/routing/dispatcher/pubsub.rs
+++ b/zenoh/src/net/routing/dispatcher/pubsub.rs
@@ -172,7 +172,8 @@ pub(crate) fn undeclare_subscription(
         Resource::clean(&mut res);
         drop(wtables);
     } else {
-        tracing::trace!("{} Undeclare unknown subscriber {}", face, id);
+        // NOTE: This is expected behavior if subscriber declarations are denied with ingress ACL interceptor.
+        tracing::debug!("{} Undeclare unknown subscriber {}", face, id);
     }
 }
 
--- a/zenoh/src/net/routing/dispatcher/pubsub.rs
+++ b/zenoh/src/net/routing/dispatcher/pubsub.rs
@@ -124,7 +124,6 @@ pub(crate) fn undeclare_subscription(
     node_id: NodeId,
     send_declare: &mut SendDeclare,
 ) {
-    tracing::trace!("Undeclare subscription {}", face);
     let res = if expr.is_empty() {
         None
     } else {
@@ -133,7 +132,7 @@ pub(crate) fn undeclare_subscription(
             Some(prefix) => match Resource::get_resource(prefix, expr.suffix.as_ref()) {
                 Some(res) => Some(res),
                 None => {
-                    tracing::trace!(
+                    tracing::error!(
                         "{} Undeclare unknown subscriber {}{}!",
                         face,
                         prefix.expr(),
@@ -143,7 +142,7 @@ pub(crate) fn undeclare_subscription(
                 }
             },
             None => {
-                tracing::trace!(
+                tracing::error!(
                     "{} Undeclare subscriber with unknown scope {}",
                     face,
                     expr.scope
--- a/zenoh/src/net/routing/dispatcher/pubsub.rs
+++ b/zenoh/src/net/routing/dispatcher/pubsub.rs
@@ -124,7 +124,7 @@ pub(crate) fn undeclare_subscription(
     node_id: NodeId,
     send_declare: &mut SendDeclare,
 ) {
-    tracing::debug!("Undeclare subscription {}", face);
+    tracing::trace!("Undeclare subscription {}", face);
     let res = if expr.is_empty() {
         None
     } else {
@@ -133,7 +133,7 @@ pub(crate) fn undeclare_subscription(
             Some(prefix) => match Resource::get_resource(prefix, expr.suffix.as_ref()) {
                 Some(res) => Some(res),
                 None => {
-                    tracing::error!(
+                    tracing::trace!(
                         "{} Undeclare unknown subscriber {}{}!",
                         face,
                         prefix.expr(),
@@ -143,7 +143,7 @@ pub(crate) fn undeclare_subscription(
                 }
             },
             None => {
-                tracing::error!(
+                tracing::trace!(
                     "{} Undeclare subscriber with unknown scope {}",
                     face,
                     expr.scope
@@ -173,7 +173,7 @@ pub(crate) fn undeclare_subscription(
         Resource::clean(&mut res);
         drop(wtables);
     } else {
-        tracing::error!("{} Undeclare unknown subscriber {}", face, id);
+        tracing::trace!("{} Undeclare unknown subscriber {}", face, id);
     }
 }
 
--- a/zenoh/src/net/routing/dispatcher/queries.rs
+++ b/zenoh/src/net/routing/dispatcher/queries.rs
@@ -186,7 +186,8 @@ pub(crate) fn undeclare_queryable(
         Resource::clean(&mut res);
         drop(wtables);
     } else {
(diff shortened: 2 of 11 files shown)


zenoh/src/net/routing/interceptor/access_control.rs
```diff
@@ -429,14 +473,6 @@ impl InterceptorTrait for EgressAclEnforcer {
...
-            | NetworkBody::Declare(Declare {
-                body: DeclareBody::UndeclareQueryable(_),
-                ..
-            })
-            | NetworkBody::Declare(Declare {
-                body: DeclareBody::UndeclareSubscriber(_),
-                ..
             }) => {}
             // Unfiltered remaining message types
             NetworkBody::Interest(_) | NetworkBody::OAM(_) | NetworkBody::ResponseFinal(_) => {}
```
fuzzypixelz, line 478: If `Request` and `Response` are handled, then `ResponseFinal` should be handled too.

oteffahi: `ResponseFinal` Cannot and should not be filtered:
- It cannot be filtered because it does not have a keyexpr
- It should not be filtered because it should always be allowed, even if all responses are denied the `ResponseFinal` is allowed through to mark the end of the query.

oteffahi:
The current routing and interceptor logic is relying on the `ext_wire_expr` field to fill the keyexpr field in the `RoutingContext`. This means that a manually crafted `UndeclareSubscriber` or `UndeclareQueryable` message can feed inconsistent keyexpr and subscriber/queyable id to the interceptor, effectively undeclaring the sub/qbl even if the keyexpr of the declaration does not match the `ext_wire_expr` data.

@Mallets should we merge this now or delay it to the next patch to fix this?

zenoh/src/net/routing/interceptor/access_control.rs
fuzzypixelz: > It cannot be filtered because it does not have a keyexpr

If at least one `Response` keyexpr is denied, then all `ResponseFinal` would be denied.

> It should not be filtered because it should always be allowed, even if all responses are denied the ResponseFinal is allowed through to mark the end of the query.

Queries are cleaned up after a timeout, the query sender doesn't need `ResponseFinal` to finalize them. If I understand correctly, not filtering the `ResponseFinal` would return an empty reply rather than error. If I misconfigure my ACL rules w.r.t queries, I would expect my queries to return errors and not succeed with empty result sets.

oteffahi: > If I misconfigure my ACL rules w.r.t queries, I would expect my queries to return errors and not succeed with empty result sets.

ACL should give the illusion that the denied queryable replies do not exist in the network, which is the current behavior of "succeed with empty result sets" (i.e. a set that only contains the allowed replies). An error when replies are denied would leak the existence of the replies and the respective ACL config that is denying them in the network.

fuzzypixelz (approved):

zenoh/src/net/routing/dispatcher/pubsub.rs
```diff
@@ -124,7 +124,7 @@ pub(crate) fn undeclare_subscription(
     node_id: NodeId,
     send_declare: &mut SendDeclare,
 ) {
-    tracing::debug!("Undeclare subscription {}", face);
+    tracing::trace!("Undeclare subscription {}", face);
```
OlivierHecart, line 127: I would just remove this trace. It's not coherent with the rest of the code and contains too few info. It's probably a leftover from a debug session.

zenoh/src/net/routing/dispatcher/pubsub.rs
```diff
@@ -133,7 +133,7 @@ pub(crate) fn undeclare_subscription(
             Some(prefix) => match Resource::get_resource(prefix, expr.suffix.as_ref()) {
                 Some(res) => Some(res),
                 None => {
-                    tracing::error!(
+                    tracing::trace!(
```
OlivierHecart, line 136: This one should not occur with proper ACL (expo is not empty). Please revert to error.

zenoh/src/net/routing/dispatcher/pubsub.rs
```diff
@@ -143,7 +143,7 @@ pub(crate) fn undeclare_subscription(
                 }
             },
             None => {
-                tracing::error!(
+                tracing::trace!(
```
OlivierHecart, line 146: This one should not occur with proper ACL (expo is not empty). Please revert to error.

zenoh/src/net/routing/dispatcher/pubsub.rs
```diff
@@ -173,7 +173,7 @@ pub(crate) fn undeclare_subscription(
         Resource::clean(&mut res);
         drop(wtables);
     } else {
-        tracing::error!("{} Undeclare unknown subscriber {}", face, id);
+        tracing::trace!("{} Undeclare unknown subscriber {}", face, id);
```
OlivierHecart, line 176: Should we set to debug rather than trace ?

zenoh/src/net/routing/dispatcher/queries.rs
```diff
@@ -146,7 +146,7 @@ pub(crate) fn undeclare_queryable(
             Some(prefix) => match Resource::get_resource(prefix, expr.suffix.as_ref()) {
                 Some(res) => Some(res),
                 None => {
-                    tracing::error!(
+                    tracing::trace!(
```
OlivierHecart, line 149: This one should not occur with proper ACL (expo is not empty). Please revert to error.

zenoh/src/net/routing/dispatcher/queries.rs
```diff
@@ -156,7 +156,7 @@ pub(crate) fn undeclare_queryable(
                 }
             },
             None => {
-                tracing::error!(
+                tracing::trace!(
```
OlivierHecart, line 159: This one should not occur with proper ACL (expo is not empty). Please revert to error.

zenoh/src/net/routing/dispatcher/queries.rs
```diff
@@ -186,7 +186,7 @@ pub(crate) fn undeclare_queryable(
         Resource::clean(&mut res);
         drop(wtables);
     } else {
-        tracing::error!("{} Undeclare unknown queryable {}", face, id);
+        tracing::trace!("{} Undeclare unknown queryable {}", face, id);
```
OlivierHecart, line 189: Should we set to debug rather than trace ?

OlivierHecart (changes requested):

zenoh/src/net/routing/dispatcher/pubsub.rs
oteffahi: I think setting it to Debug would send people in the wrong direction when debugging, since it's expected behavior when ACL is enabled, but the message sounds like something went wrong.

zenoh/src/net/routing/dispatcher/queries.rs
oteffahi: I think it should be left at Trace for the same reason I gave for comment on `pubsub`

OlivierHecart: Fine. 
Maybe add a comment here stating that this is expected in case of ACL.

zenoh/src/net/routing/dispatcher/pubsub.rs
OlivierHecart: Fine.
Maybe add a comment here stating that this is expected in case of ACL.

OlivierHecart (approved):

OlivierHecart (approved):

PR #1576, opened 2024-10-31
atexit-safe close builder draft

yellowhatter:


io/zenoh-transport/src/manager.rs
```diff
@@ -477,10 +477,12 @@ impl TransportManager {
     }
 
     pub async fn close(&self) {
+        self.close_with_timeout(Duration::from_secs(10)).await;
+    }
+
+    pub async fn close_with_timeout(&self, timeout: Duration) {
```
Mallets, line 483: It's an `async` function, is this really needed? All `async` functions return a `Future` that can be `await`ed with a timeout `xxx.close().timeout(_).await`.

zenoh/src/api/builders/close.rs
```diff
@@ -0,0 +1,182 @@
...
+        self
+    }
+
+    /// Set the backoff for close operation
+    ///
+    /// # Arguments
+    ///
+    /// * `backoff` - The backoff behavior for close operation
+    ///
+    pub fn backoff(mut self, backoff: CloseBackoff) -> Self {
```
Mallets, line 76: What's the purpose of a `backoff` in a close operation? A backoff is usually employed as a mechanism to minimise the probability of collisions/contention.

zenoh/src/api/builders/close.rs
```diff
@@ -0,0 +1,182 @@
...
+use async_trait::async_trait;
+use tokio::{select, time::sleep};
+use tracing::warn;
+use zenoh_core::{Resolvable, Wait};
+use zenoh_result::ZResult;
+use zenoh_runtime::ZRuntime;
+use zenoh_task::TaskController;
+
+/// A behavior selector for close operations used if initial timeout was reached
+pub enum CloseBackoff {
```
Mallets, line 30: I think the term `backoff` is misused here. It seems to be a normal timeout.

io/zenoh-transport/src/manager.rs
yellowhatter: Inside I pass this `timeout` to `self.task_controller.terminate_all_async` .

zenoh/src/api/builders/close.rs
yellowhatter: Well, this is a policy describing what to do if the initial timeout was not enough, that's why I used `backoff` term. The idea is that user might have some requirements to have typical finalization delay that will become an initial timeout and some absolute maximum delay that can be described through `CloseBackoff::Wait(_)`

io/zenoh-transport/src/manager.rs
Mallets: Yes, but looking at [terminate_all_async implementation](https://github.com/ZettaScaleLabs/zenoh/blob/e73a89d3bab5d3960159ddfc0ef315407d2fdf7e/commons/zenoh-task/src/lib.rs#L117) I'm not sure it needs to have a timeout either...

zenoh/src/api/builders/close.rs
Mallets: If the timeout was not enough, then why bother? If the timeout was too short, then it needs to be increased. I don't see really the use case nor the need for such added complexity in the API.

yellowhatter: And at the moment of switching in between the initial and backoff timeouts a warning message will be printed out to the log. Currently logging here is very minimal, but in the future I want to extend it.

zenoh/src/api/builders/close.rs
yellowhatter: please check the comment above

yellowhatter: https://github.com/eclipse-zenoh/zenoh/pull/1576/files#r1830677025

io/zenoh-transport/src/manager.rs
yellowhatter: Oh, right, I missed that
I will refactor this

zenoh/src/api/builders/close.rs
Mallets: Again, why the use should set 2 timeouts? The use case and the need is unclear.

yellowhatter: To give user warning in case if the typical timeout failed - that can be used to early catch finalization delay problems etc - but on the other hand to give the ability to have extended timeout for proper finalization completion to avoid having semi-finalized `Session`.

Mallets: I don't think it's a real need nor we have a real use case that requires that. 
Can you provide one? Otherwise I believe a simple timeout will suffice.

yellowhatter: Started describing the use-case, but then I understood that everything of this can be implemented in the user code. I will fall back to the simple timeout then :)

yellowhatter:
Ci failed because of this: https://github.com/eclipse-zenoh/zenoh/pull/1586

for `TaskController::terminate_all_async` change, repo search shows that there is no other projects using it through "internal" feature

YuanYuYuan:
Hi @yellowhatter , here are my comments.

- These functions have no timeout, unlike their previous versions:
  - `TransportManager::close`
  - `TerminatableTask::terminate`
- Consider splitting the ephemeral port removal into a separate PR.
- Introducing two additional traits to create a concrete `SessionCloseBuilder` might be over engineering.
Instead, let's focus on configuring the close behavior through a well-structured `SessionCloseBuilder::new(config)`, just like we do in the other builders.
This could include setting options like switching between a timeout-based blocking mode and a non-blocking mode.

yellowhatter:
Hi, @YuanYuYuan 

> * These functions have no timeout, unlike their previous versions:
>   
>   * `TransportManager::close`
>   * `TerminatableTask::terminate`
This is intentional. They used internal future timeout, and now I rely on top-level future timeout that can be modified by the user.

yellowhatter:
> Without entering the internal details of this particular PR, I think this kind of API addition should be coordinated with @milyin for all the bindings (C, C++, Python, etc.). Adding a timeout in the close builder/option is something we may want to expose as well in all other languages.

Sure. This change doesn't break the current bindings, and we will coordinate on it's support everywhere after this one get merged

YuanYuYuan:
Yes. We could first discuss the use cases to see what options we want to expose. For instance, making the zenoh session terminate itself in a manner of nonblocking way if it has already been exit stage.

zenoh/src/api/builders/close.rs
```diff
@@ -0,0 +1,103 @@
...
+            closee: closeable.get_closee(),
+            timeout: Duration::from_secs(3600),
+        }
+    }
+
+    /// Set the timeout for close operation
+    ///
+    /// # Arguments
+    ///
+    /// * `timeout` - The timeout value for close operation
```
milyin, line 50: Documentation in zenoh-c is more detailed than here, in rust. E.g. in C default timeout is mentioned

milyin (changes requested):
Looks good, but test for new functionality is needed

zenoh/src/api/builders/close.rs
```diff
@@ -0,0 +1,103 @@
...
+    pub fn timeout(mut self, timeout: Duration) -> Self {
+        self.timeout = timeout;
+        self
+    }
+
+    /// Run Close operation concurrently
+    pub fn concurrently(
+        self,
+    ) -> tokio::task::JoinHandle<<CloseBuilder<TCloseable> as Resolvable>::To> {
+        ZRuntime::Application.spawn(self.into_future())
```
Mallets, line 61: The task shouldn't be spawn during the *building* phase but only in the builder finalization.

zenoh/src/api/builders/close.rs
```diff
@@ -0,0 +1,103 @@
...
+    ///
+    /// * `timeout` - The timeout value for close operation
+    ///
+    pub fn timeout(mut self, timeout: Duration) -> Self {
+        self.timeout = timeout;
+        self
+    }
+
+    /// Run Close operation concurrently
+    pub fn concurrently(
```
Mallets, line 58: Should it be simply call *background*?

yellowhatter: Maybe it's even better: `in_background`?

yellowhatter: Made with `in_`, correct me if you are for background :)

zenoh/src/api/builders/close.rs
yellowhatter: fixed

yellowhatter:
`three_node_combination_multicast` test is failing on windows, as I can see this is a known flaky one

zenoh/src/api/builders/close.rs
```diff
@@ -0,0 +1,151 @@
...
+            .into_future(),
+        )
+    }
+}
+
+#[cfg(all(feature = "unstable", feature = "internal"))]
+/// A builder for close operations running in background
+// NOTE: `Closeable` is only pub(crate) because it is zenoh-internal trait, so we don't
+// care about the `private_bounds` lint in this particular case.
+#[zenoh_macros::unstable_doc]
```
Mallets, line 103: For internal data it should actually be `[doc(hidden)]`.

zenoh/src/api/builders/close.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    // NOTE: yes, we need to return a future that returns JoinHandle
+    #[allow(clippy::async_yields_async)]
+    fn into_future(self) -> Self::IntoFuture {
+        Box::pin(async move { ZRuntime::Application.spawn(self.inner) }.into_future())
+    }
+}
+
+#[async_trait]
+pub(crate) trait Closee: Send + Sync + 'static {
```
Mallets, line 144: Very weird name. What does `Closee` mean?

zenoh/src/api/builders/close.rs
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    #[cfg(all(feature = "unstable", feature = "internal"))]
+    /// Set the timeout for close operation
+    ///
+    /// # Arguments
+    ///
+    /// * `timeout` - The timeout value for close operation (10s by default)
+    ///
+    #[zenoh_macros::unstable_doc]
```
Mallets, line 53: It's internal, not unstable. So no doc.

zenoh/src/api/builders/close.rs
```diff
@@ -0,0 +1,151 @@
...
+    pub fn timeout(mut self, timeout: Duration) -> Self {
+        self.timeout = timeout;
+        self
+    }
+
+    #[cfg(all(feature = "unstable", feature = "internal"))]
+    /// Run Close operation concurrently
+    #[zenoh_macros::unstable_doc]
```
Mallets, line 61: It's internal, not unstable. So no doc.

zenoh/src/api/builders/close.rs
yellowhatter: > Very weird name. What does `Closee` mean?

https://dictionary.cambridge.org/dictionary/english/ee

The quick search showed there is no 'closee' word, however the '-ee' added to 'close' is smth that perfectly reflects the trait's purpose.

Mallets: Ok, still it's not clear what the trait implementer is expected to do and who should implement it.

zenoh/src/api/builders/close.rs
yellowhatter: fixed

zenoh/src/api/builders/close.rs
yellowhatter: fxed

zenoh/src/api/builders/close.rs
yellowhatter: fixed

zenoh/src/api/builders/close.rs
Mallets: Mmmm not convinced either. Anyway, I don't think it's a blocker. LGTM.

yellowhatter:
> Looks good, but test for new functionality is needed

added tests

milyin (approved):

PR #1578, opened 2024-11-01
fix: break fragment chain when dropping fragmented message

wyfo:


io/zenoh-transport/src/common/pipeline.rs
```diff
@@ -322,7 +322,9 @@ impl StageIn {
         let mut reader = self.fragbuf.reader();
         while reader.can_read() {
             // Get the current serialization batch
-            batch = zgetbatch_rets!(tch.sn.set(sn).unwrap());
+            // If deadline is reached, sequence number is incremented in order to break the
+            // fragment chain already sent.
+            batch = zgetbatch_rets!(tch.sn.set(fragment.sn + 1).unwrap());
```
Mallets, line 327: Sequence numbers are handled as circular number space. `+1` is not a valid way of incrementing the SN. The proper increment function should be used instead that takes care of all modulo operations and wrap around.

wyfo: Indeed. At first, I used `tch.sn.get()`, but I found that a little bit ugly, hence this buggy `+1`. It's fixed.

Mallets (approved):

PR #1579, opened 2024-11-05
Provide a clearer panic log while missing TLS

evshary:
Although we decided not to provide a function to detect missing TLS [here](https://github.com/eclipse-zenoh/zenoh/pull/1569), it's still worth providing a clearer panic to users.
For example, https://github.com/eclipse-zenoh/roadmap/discussions/160

### DIFF
--- a/commons/zenoh-runtime/src/lib.rs
+++ b/commons/zenoh-runtime/src/lib.rs
@@ -137,7 +137,7 @@ impl ZRuntime {
             }
             Err(e) => {
                 if e.is_thread_local_destroyed() {
-                    panic!("The TLS inside Tokio is destroyed. You might call Zenoh API in the exiting process, which isn't supported now.");
+                    panic!("The Thread Local Storage inside Tokio is destroyed. This might happen when Zenoh API is called at process exit, e.g. in the atexit handler. Calling the Zenoh API at process exit is not supported and should be avoided.");
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


commons/zenoh-runtime/src/lib.rs
```diff
@@ -129,9 +129,16 @@ impl ZRuntime {
...
-                panic!("Zenoh runtime doesn't support Tokio's current thread scheduler. Please use multi thread scheduler instead, e.g. a multi thread scheduler with one worker thread: `#[tokio::main(flavor = \"multi_thread\", worker_threads = 1)]`");
+        match Handle::try_current() {
+            Ok(handle) => {
+                if handle.runtime_flavor() == RuntimeFlavor::CurrentThread {
+                    panic!("Zenoh runtime doesn't support Tokio's current thread scheduler. Please use multi thread scheduler instead, e.g. a multi thread scheduler with one worker thread: `#[tokio::main(flavor = \"multi_thread\", worker_threads = 1)]`");
+                }
+            }
+            Err(e) => {
+                if e.is_thread_local_destroyed() {
+                    panic!("The TLS inside Tokio is destroyed. You might call Zenoh API in the exiting process, which isn't supported now.");
```
Mallets, line 140:
```suggestion
                    panic!("The Thread Local Storage inside Tokio is destroyed. This might happen when Zenoh API is called at process exit, e.g. in the atexit handler. Calling the Zenoh API at process exit is not supported and should be avoided.");
```

Mallets (approved):

PR #1585, opened 2024-11-06
fix: do not drop callbacks with session lock held

wyfo:


Mallets (approved):

zenoh/src/api/session.rs
```diff
@@ -1092,16 +1092,22 @@ impl SessionInner {
...
-            state.subscribers.clear();
-            state.liveliness_subscribers.clear();
-            state.local_resources.clear();
-            state.remote_resources.clear();
+            let _queryables = std::mem::take(&mut state.queryables);
+            let _subscribers = std::mem::take(&mut state.subscribers);
+            let _liveliness_subscribers = std::mem::take(&mut state.liveliness_subscribers);
+            let _local_resources = std::mem::take(&mut state.local_resources);
+            let _remote_resources = std::mem::take(&mut state.remote_resources);
+            drop(state);
```
Mallets, line 1104: Wouldn't be enough to drop the state at the end of the function?
Does the `unstable` feature need to drop first the state and then reacquiring it straight after?

wyfo: Fixed, it's simpler and clearer now.

Mallets (approved):

PR #1586, opened 2024-11-06
fix small bug in backoff update

yellowhatter:


yellowhatter:
@Mallets I fixed it by forcing operand evaluation order that makes it impossible to have negative values. Maybe this is the fastest approach (I will run zenoh benchmarks for this)

Mallets (approved):

yellowhatter:
additional slight improvement: 
2024:11:08-17:10:06 throughput aggregated

PR #1590, opened 2024-11-12
Document callback drop on subscriber undeclaration

fuzzypixelz:


wyfo:
If I'm not mistaken, the callback is not dropped during the undeclaration, and that is an issue regarding C/C++ users' point of view and "unsafe" use.
In fact, some read task may have cloned the callback before the undeclaration, and therefore use it after. We would like to "improve" this behavior and wait the last callback to be dropped, notably to match C/C++ users' expectations, but for now, I will not touch the documentation.

OlivierHecart (approved):

OlivierHecart:
I confirm that in Zenoh v1.3.4 the callback is indeed dropped during subscriber undeclaration.

Issue #1031, opened 2024-05-15
Zenoh in Docker does not load plugins from /root/.zenoh/lib


philrhc:
### Describe the bug

Zenoh router does not load plugins from /root/.zenoh/lib

- Followed the steps here: https://zenoh.io/docs/getting-started/quick-test/#adding-plugins-and-backends-to-the-container
- Attached shell shows the .so files in `/root/.zenoh/lib`
- See the log line `2024-05-15T14:30:14.863318Z DEBUG main ThreadId(01) zenoh_util::std_only::lib_loader: Search for library libzenoh_plugin_rest.so to load in ["/", "/root/.zenoh/lib", "/usr/local/lib", "/usr/lib", "/"]`
- See the below response:
```
curl 'http://localhost:8000/@/router/local'
[
{ "key": "@/router/79887a32dad6ef34f82f4ed312390b69", "value": {"locators":["tcp/172.29.0.2:7447"],"metadata":null,"plugins":{"rest":{"path":"/libzenoh_plugin_rest.so"}},"sessions":[],"version":"v0.11.0-rc.3 built with rustc 1.72.0 (5680fa18f 2023-08-23)","zid":"79887a32dad6ef34f82f4ed312390b69"}, "encoding": "application/json", "time": "None" }
]
```

### To reproduce

Created a minimum repo w/ the docker compose but should be fairly easy to test w/ a docker compose file
https://github.com/philrhc/zenoh_docker_plugins

### System info

OS: Ubuntu 22.04.4 LTS
Arch: x86_64
CPU: 12th Gen Intel(R) Core(TM) i7-1265U
version:v0.11.0-rc.3 built with rustc 1.72.0 (5680fa18f 2023-08-23)
zid:79887a32dad6ef34f82f4ed312390b69

philrhc:
ok, closing because i don't think this is a bug, i think it is a fault in the documentation. I'll continue investigating...

philrhc:
if anyone else has the same problem, this should be helpful:
https://github.com/xvanov/advrob-project/blob/zenoh-router/desktop/compose.yaml

alex-quilt:
This seems to still be an issue with the documentation, which I believe they're referring to in https://github.com/eclipse-zenoh/zenoh/pull/1593. For example, the compose.yaml in https://zenoh.io/docs/getting-started/quick-test/ should be:
```
services:
  zenoh:
    image: eclipse/zenoh
    restart: unless-stopped
    ports:
[...]

PR #1593, opened 2024-11-13
chore: enable rest plugin only if in config or if --rest-http-port is set

gabrik:
Currently the rest plugin is enabled by default for `zenohd`, which means that its a required dependency when running a simple zenoh router without configuration.

This PR changes the behaviour by enabling the rest plugin only if the `--rest-http-port` is passed.


OlivierHecart:
Should be:
```rust
    // apply '--rest-http-port' to config only if explicitly set (overwriting config)
    if args.rest_http_port.is_some() || args.config.is_none() {
        match args.rest_http_port.as_deref() {
            Some(value) => {
                if !value.eq_ignore_ascii_case("none") {
                    config
                        .insert_json5("plugins/rest/http_port", &format!(r#""{value}""#))
                        .unwrap();
                    config
                        .insert_json5("plugins/rest/__required__", "true")
                        .unwrap();
                }
            }
            None => {
                config
                    .insert_json5("plugins/rest/http_port", "8000")
                    .unwrap();
            }
        }
    }
```

gabrik:
shouldn't the `|| args.config.is_none()` be removed to load only if the argument passed?

gabrik:
> > shouldn't the `|| args.config.is_none()` be removed to load only if the argument passed?
> 
> Well the point is that we want rest to start on port 8000 by default if nothing is specified.

the point of my PR is not not have it unless `--rest-http-port` is specified

OlivierHecart:
> > > shouldn't the `|| args.config.is_none()` be removed to load only if the argument passed?
> > 
> > 
> > Well the point is that we want rest to start on port 8000 by default if nothing is specified.
> 
> the point of my PR is not not have it unless `--rest-http-port` is specified

I believed the point of the PR was to not fail if rest was not available but still start it on port 8000 by default if available.

Mallets:
I agree with this PR and not having unnecessary services running if not specified. 
Moreover, with current status it's impossible to compile and run `zenohd` without forcing the compilation of the `rest plugin`, which is quite annoying tbh.
The PR should also include updates to the README to explicitly set the `--http-port` parameter when needed.

Mallets:
Documentation to be updated: https://zenoh.io/docs/manual/configuration/ 
Dockerfile to be updated: https://github.com/eclipse-zenoh/ci/blob/7cbe301bce9b932eea7ffea2b463890ab712f97e/publish-crates-docker/Dockerfile#L14

gabrik:
> * [ ]  Documentation to be updated: https://zenoh.io/docs/manual/configuration/ [Explicit mention zenohd --http-rest-port in documentation atolab/zenoh-web#88](https://github.com/atolab/zenoh-web/pull/88)
> 
> 
> Update: ~Dockerfile to be updated: https://github.com/eclipse-zenoh/ci/blob/7cbe301bce9b932eea7ffea2b463890ab712f97e/publish-crates-docker/Dockerfile#L14~ Current Dockerfile already passes all CLI arguments to the inner `zenohd`. So no new changes are expected.
> 
> @gabrik this PR is complete from PoV. Anything else?

Looks good to me.

JEnoch:
We have examples of REST API usage in backends READMEs that shall be updated with the correct `zenohd` command line for REST plugin activation:
* [ ] https://github.com/eclipse-zenoh/zenoh-backend-influxdb?tab=readme-ov-file#setup-at-runtime-via-curl-commands-on-the-admin-space
* [ ] https://github.com/eclipse-zenoh/zenoh-backend-rocksdb?tab=readme-ov-file#setup-at-runtime-via-curl-commands-on-the-admin-space
* [ ] https://github.com/eclipse-zenoh/zenoh-backend-filesystem?tab=readme-ov-file#setup-at-runtime-via-curl-commands-on-the-admin-space

Mallets (approved):

alex-quilt:
From https://github.com/eclipse-zenoh/zenoh/issues/1031, https://zenoh.io/docs/getting-started/quick-test/ should also be updated since the instructions currently don't work

PR #1597, opened 2024-11-15
feat: add start/stop extension markers for fragment chain

wyfo:
A fragment chain must begin with a start marker, otherwise the whole chain will be dropped.

When a fragmented message is dropped by congestion control, a final fragment with stop marker must be emitted. Because no batch is available, an ephemeral batch, i.e. not recycled in the pipeline after, is created.


wyfo:
@Mallets I've moved `ephemeral` flag out of `BatchConfig`, see the second commit.

Mallets:
I think for backward compatibility we still need to address two main points:

1. Zenoh-Pico should be updated accordingly and implement the start and stop markers
2. To avoid protocol breaking, the RX side should be able to handle also the case where no start marker is sent. This could be done lazily.

Finally, I wonder if we could add the information about the total message length in the start marker. The total length could be eventually used to preallocate a single buffer to copy the message on. I understand this would require changing quite a bit the deserialisation code. As well, it's still unclear to me the real impact on performance.

wyfo:
> To avoid protocol breaking, the RX side should be able to handle also the case where no start marker is sent. This could be done lazily.

It's handled by dropping the fragment. I don't see a way to provide backward compatibility here, that was my main concern about this change.

> Finally, I wonder if we could add the information about the total message length in the start marker. The total length could be eventually used to preallocate a single buffer to copy the message on.

I don't think it's a good idea regarding routing, because the message would be refragmented right after. However, on the client side, it would allow to avoid `ZBytes` fragmentation, which may be a big plus. But if you want to avoid copy, the deserialization code will indeed need a deep refactor.

wyfo:
Still, adding the size on start extension doesn't cost anything and is still a nice information to have when debugging, so why not adding it.

Mallets:
> It's handled by dropping the fragment. I don't see a way to provide backward compatibility here, that was my main concern about this change.

It's about improving the behaviour for future Zenoh versions (especially in patch release) without requiring dropping old versions (i.e. that would require bumping the protocol version). A viable option is to start by default in *no start marker mode* and as soon as *start marker* is detected we switch mode. Additionally, we could handle this in a `Init/Open` extension.

Mallets:
> Still, adding the size on start extension doesn't cost anything and is still a nice information to have when debugging, so why not adding it.

It costs bytes on the network. If we don't have a use for it, let's not send it.

wyfo:
> A viable option is to start by default in no start marker mode and as soon as start marker is detected we switch mode. Alternatevely, we could handle this in a Init/Open extension.

I prefer an `Init` extension here, otherwise, if you loose the first fragment, it's already too late. Do I implement it?

Mallets:
> > A viable option is to start by default in no start marker mode and as soon as start marker is detected we switch mode. Alternatevely, we could handle this in a Init/Open extension.
> 
> I prefer an `Init` extension here, otherwise, if you loose the first fragment, it's already too late. Do I implement it?

After a second thought I also prefer the `Init/Open` approach. Go for it :)

commons/zenoh-protocol/src/transport/fragment.rs
```diff
@@ -75,14 +75,26 @@ pub struct Fragment {
...
+    };
 
     pub type QoS = zextz64!(0x1, true);
     pub type QoSType = crate::transport::ext::QoSType<{ QoS::ID }>;
+
+    /// # Start extension
+    /// Mark the first fragment of a fragmented message
+    pub type Start = zextunit!(0x2, false);
+    /// # Stop extension
+    /// Indicate that the remaining fragments has been dropped
```
Mallets, line 96:
```suggestion
    /// Indicate that the following fragments have been dropped and the previous fragments should be discarded
```

commons/zenoh-protocol/src/transport/join.rs
```diff
@@ -125,6 +129,13 @@ pub mod ext {
     /// # Shm extension
     /// Used to advertise shared memory capabilities
     pub type Shm = zextzbuf!(0x2, true);
+
+    /// # Patch extension
+    /// Used to negotiate the patch version of the protocol
+    /// if not present (or 0), then protocol as released with 1.0.0
+    /// if >= 1, then fragmentation start/stop marker
+    pub type Patch = zextz64!(0x3, false);
```
Mallets, line 137: Even if not necessary, it may be good to use the same extension ID of the `Init/Open`?

```suggestion
    pub type Patch = zextz64!(0x7, false);
```

commons/zenoh-protocol/src/transport/open.rs
```diff
@@ -124,6 +121,13 @@ pub mod ext {
     /// # Compression extension
     /// Used to negotiate the use of compression on the link
     pub type Compression = zextunit!(0x6, false);
+
+    /// # Patch extension
+    /// Used to negotiate the patch version of the protocol
+    /// if not present (or 0), then protocol as released with 1.0.0
+    /// if >= 1, then fragmentation start/stop marker
+    pub type Patch = zextz64!(0x7, false);
```
Mallets, line 129: I think the patch extension is not needed in the `Open` message. Having it simply in the `Init` will suffice.

io/zenoh-transport/src/unicast/establishment/ext/patch.rs
```diff
@@ -0,0 +1,196 @@
...
+        Ok(())
+    }
+
+    type SendOpenSynIn = &'a StateOpen;
+    type SendOpenSynOut = open::ext::PatchType;
+    async fn send_open_syn(
+        self,
+        _state: Self::SendOpenSynIn,
+    ) -> Result<Self::SendOpenSynOut, Self::Error> {
+        Ok(open::ext::PatchType::DEFAULT)
```
Mallets, line 88: I think there is no need to send back the default patch here, especially if what is returned is the `DEFAULT` value.
Also if the rule is to agree on the minimum between the two versions, the `InitSyn/InitAck` exchange is enough and there is no need for the extension in the `Open`.

io/zenoh-transport/src/unicast/establishment/ext/patch.rs
```diff
@@ -0,0 +1,196 @@
...
+        Ok(())
+    }
+
+    type SendInitAckIn = &'a StateAccept;
+    type SendInitAckOut = init::ext::PatchType;
+    async fn send_init_ack(
+        self,
+        _state: Self::SendInitAckIn,
+    ) -> Result<Self::SendInitAckOut, Self::Error> {
+        Ok(init::ext::PatchType::CURRENT)
```
Mallets, line 176: Here the agreed minimum patch version could be directly sent.

io/zenoh-transport/src/unicast/establishment/accept.rs
```diff
@@ -545,6 +563,13 @@ impl<'a, 'b: 'a> AcceptFsm for &'a mut AcceptLink<'b> {
             .await
             .map_err(|e| (e, Some(close::reason::GENERIC)))?;
 
+        // Extension Patch
+        #[cfg(feature = "transport_compression")]
```
Mallets, line 567: See my comment on the need of the `ext_patch` in the `Open` message.
Besides that, the `cfg` directive is for compression and not for the patch.

io/zenoh-transport/src/unicast/establishment/open.rs
```diff
@@ -419,6 +429,13 @@ impl<'a, 'b: 'a> OpenFsm for &'a mut OpenLink<'b> {
             None
         );
 
+        // Extension Patch
+        let ext_patch = self
```
Mallets, line 433: See my previous comment on the need of `ext_patch` in the `Open` message.

io/zenoh-transport/src/unicast/universal/link.rs
```diff
@@ -191,8 +191,10 @@ async fn tx_task(
                             stats.inc_tx_bytes(batch.len() as usize);
                         }
 
-                        // Reinsert the batch into the queue
-                        pipeline.refill(batch, priority);
+                        if !batch.is_ephemeral() {
```
Mallets, line 194: I would move the check inside the `refill` function for better encapsulation and avoid checking wether a batch is ephemeral every time it is used.

io/zenoh-transport/src/unicast/universal/rx.rs
```diff
@@ -147,7 +149,15 @@ impl TransportUnicastUniversal {
             // Drop invalid message and continue
             return Ok(());
         }
+        if self.config.patch.has_fragmentation_start_stop() && ext_stop.is_some() {
```
Mallets, line 152: We should clear the defragmentation buffer in this case.

io/zenoh-transport/src/unicast/establishment/open.rs
Mallets, line 433: See my comment on the need of `ext_patch` in the `Open` message.

io/zenoh-transport/src/unicast/universal/link.rs
Mallets, line 194: See my comment about moving the check inside the `refill` function.

wyfo:
@Mallets I came to the conclusion that both ends don't need to agree on a patch version, i.e. selecting the minimum of both. In fact, each end can just keep the other's patch version, and let the code do the rest.
For example, if A has `patch = 1` and B has `patch = 0`, after init syn/ack exchange,  `PatchType::has_fragmentation_start_stop` will return `false` on A, and on B side, there will be no `PatchType`, so both "agree" to not use fragmentation markers.
On the other hand, if B has `patch = 2` this time, then `PatchType::has_fragmentation_start_stop` will return `true` on both sides, and an hypothetical `PatchType::has_new_feature` will return `false` on B, so again, there is a de facto "agreement" on the patch version.

Mallets:
> @Mallets I came to the conclusion that both ends don't need to agree on a patch version, i.e. selecting the minimum of both. In fact, each end can just keep the other's patch version, and let the code do the rest. For example, if A has `patch = 1` and B has `patch = 0`, after init syn/ack exchange, `PatchType::has_fragmentation_start_stop` will return `false` on A, and on B side, there will be no `PatchType`, so both "agree" to not use fragmentation markers. On the other hand, if B has `patch = 2` this time, then `PatchType::has_fragmentation_start_stop` will return `true` on both sides, and an hypothetical `PatchType::has_new_feature` will return `false` on B, so again, there is a de facto "agreement" on the patch version.

I agree with the reasoning. However, the `Init/Open` role is to also negotiate parameters and setting the minimum value in the message will make clear which version is chosen also when looking at Wireshark. Does it make sense?

io/zenoh-transport/src/unicast/establishment/ext/patch.rs
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    type RecvInitAckIn = (&'a mut StateOpen, init::ext::PatchType);
+    type RecvInitAckOut = ();
+    async fn recv_init_ack(
+        self,
+        input: Self::RecvInitAckIn,
+    ) -> Result<Self::RecvInitAckOut, Self::Error> {
+        let (state, other_ext) = input;
+        state.patch = other_ext;
```
Mallets, line 78: `other_ext` should be verified to be `<=` of `self`. In other words, we should not accept as a reply any value larger than what has been sent.

io/zenoh-transport/src/unicast/establishment/ext/patch.rs
```diff
@@ -0,0 +1,196 @@
...
+        Ok(())
+    }
+
+    type SendOpenSynIn = &'a StateOpen;
+    type SendOpenSynOut = ();
+    async fn send_open_syn(
+        self,
+        _state: Self::SendOpenSynIn,
+    ) -> Result<Self::SendOpenSynOut, Self::Error> {
+        unimplemented!()
```
Mallets, line 88: Add a comment that this is not `unimplemented` but rather no extension exists for that. 
Or should we simply return `Ok`?

io/zenoh-transport/src/unicast/establishment/ext/patch.rs
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    type RecvOpenAckIn = (&'a mut StateOpen, ());
+    type RecvOpenAckOut = ();
+    async fn recv_open_ack(
+        self,
+        _state: Self::RecvOpenAckIn,
+    ) -> Result<Self::RecvOpenAckOut, Self::Error> {
+        unimplemented!()
```
Mallets, line 97: Add a comment that this is not `unimplemented` but rather no extension exists for that. 
Or should we simply return `Ok`?

io/zenoh-transport/src/unicast/establishment/ext/patch.rs
```diff
@@ -0,0 +1,196 @@
...
+        Ok(())
+    }
+
+    type SendInitAckIn = &'a StateAccept;
+    type SendInitAckOut = init::ext::PatchType;
+    async fn send_init_ack(
+        self,
+        _state: Self::SendInitAckIn,
+    ) -> Result<Self::SendInitAckOut, Self::Error> {
+        Ok(init::ext::PatchType::CURRENT)
```
Mallets, line 176: See my comment on sending the minimum version as a way to be explicit rather than implicit.

io/zenoh-transport/src/unicast/establishment/ext/patch.rs
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    type RecvOpenSynIn = (&'a mut StateAccept, ());
+    type RecvOpenSynOut = ();
+    async fn recv_open_syn(
+        self,
+        _state: Self::RecvOpenSynIn,
+    ) -> Result<Self::RecvOpenSynOut, Self::Error> {
+        unimplemented!()
```
Mallets, line 185: Add a comment that this is not `unimplemented` but rather no extension exists for that. 
Or should we simply return `Ok`?

wyfo:
> I agree with the reasoning. However, the Init/Open role is to also negotiate parameters and setting the minimum value in the message will make clear which version is chosen also when looking at Wireshark. Also, in the Cookie the agreed patch version should stored, so it's more about being explicit rather than implicit. Does it make sense?

It does. I will roll back the minimum check. Then I will have to find a proper error code in zenoh-pico...

wyfo:
@Mallets Should we store minimum patch version in `TransportMulticastPeer`, to be "consistent" with the unicast workflow?

Mallets:
This PR LGTM. However, I prefer to wait for having https://github.com/eclipse-zenoh/zenoh-pico/pull/813 tested and approved before merging.

wyfo:
The lint error came from the recently 1.83 release. I will fix it while resolving the conflict. By the way, zeno-pico PR has been approved.

io/zenoh-transport/src/common/pipeline.rs
```diff
@@ -863,8 +880,10 @@ impl TransmissionPipelineConsumer {
     }
 
     pub(crate) fn refill(&mut self, batch: WBatch, priority: Priority) {
-        self.stage_out[priority as usize].refill(batch);
-        self.status.set_congested(priority, false);
+        if !batch.is_ephemeral() {
+            self.stage_out[priority as usize].refill(batch);
+            self.status.set_congested(priority, false);
```
wyfo, line 885: @Mallets The merge conflict was on these lines. Just check I took the right decision about calling `set_congested` only if the batch is not ephemeral.

Mallets: That's the right decision.

Mallets (approved):

PR #1601, opened 2024-11-20
Fix version macro & config deserialisation.

Jezza:
Just a couple of small fixes.

Nothing severe.

There's two changes here:

1) re-export `git_version`, this lets us use it in the macro, meaning downstream crates don't have to pull in a whole new crate just for the macro.

2) Fix number reading via `ModeDependentValue`.
I don't honestly know _how_ I was hitting this path, I suspect it has something to do with the arbitrary precision feature, but I'll admit that I didn't look too deeply into it.
This change just accepts `u64`s as well, and then casts it to a signed number.

I can split it into two PRs if needed, but I thought I'd just cut down on the noise.


Mallets:
It seems the code format check is failing, make sure to run:

```sh
cargo fmt -- --config "unstable_features=true,imports_granularity=Crate,group_imports=StdExternalCrate"
```

Regarding point 2., wouldn't be that some integer number is interpreted directly as `u64` instead of `i64` or `f64`?
E.g., in the config you may declare `field_f64: 4` and `4` is somehow interpreted as `u64`?

I'm just curious, besides that the PR LGTM and as soon as the CI is happy I'll merge it.

Jezza:
Thanks!

I pushed the formatter changes (Just a single newline).

Re: point 2.

I dived a bit deeper into it, and it seems that when `serde_json` parses a "0" when `arbitrary_precision` is disabled, it emits a U64.
image

Which after some thought, I now see why that differs.
Zenoh doesn't use `serde_json`, but `json5`.

PR #1603, opened 2024-11-21
implement non-linear wait before drop(close)

yellowhatter:


yellowhatter:
On my HW this doesn't make any visible performance change.

DEFAULT_CONFIG.json5
```diff
@@ -412,12 +412,26 @@
             drop: {
               /// The maximum time in microseconds to wait for an available batch before dropping a droppable message if still no batch is available.
               wait_before_drop: 1000,
+              /// The maximum deadline limit for multi-fragment messages.
+              /// When sending multi-fragment message, for each consecutive fragment the deadline from
+              /// the start point of message transmission is calculated as 
+              /// wait_before_drop * 2 ^ min(N, ttl) where N is a fragment number.
+              /// Thus, this parameter allows setting a limitation for maximum time for multi-fragment
+              /// message transmission as wait_before_drop * 2 ^ ttl
+              ttl: 8,
```
Mallets, line 421: I'd expose the configuration option differently without the need of leaking internal details about exponential backoff. 
What about calling it something like `max_wait_before_drop_fragments`?

yellowhatter: So, you suggest just representing the max limitation in milliseconds, completely hiding the inner calculations? Yes, it makes sense!

DEFAULT_CONFIG.json5
```diff
@@ -413,25 +413,15 @@
...
               /// if still no batch is available.
               wait_before_close: 5000000,
               /// The maximum deadline limit for multi-fragment messages.
-              /// When sending multi-fragment message, for each consecutive fragment the deadline from
-              /// the start point of message transmission is calculated as 
-              /// wait_before_close * 2 ^ min(N, ttl) where N is a fragment number.
-              /// Thus, this parameter allows setting a limitation for maximum time for multi-fragment
-              /// message transmission as wait_before_close * 2 ^ ttl
-              ttl: 2,
+              max_wait_before_close_fragments: 15000000,
```
Mallets, line 424: I don't think it makes sense to have a special treatment for the `wait_before_close`. And 15 seconds seems a bit too much anyway.

DEFAULT_CONFIG.json5
```diff
@@ -413,25 +413,15 @@
               /// The maximum time in microseconds to wait for an available batch before dropping a droppable message if still no batch is available.
               wait_before_drop: 1000,
               /// The maximum deadline limit for multi-fragment messages.
-              /// When sending multi-fragment message, for each consecutive fragment the deadline from
-              /// the start point of message transmission is calculated as 
-              /// wait_before_drop * 2 ^ min(N, ttl) where N is a fragment number.
-              /// Thus, this parameter allows setting a limitation for maximum time for multi-fragment
-              /// message transmission as wait_before_drop * 2 ^ ttl
-              ttl: 8,
+              max_wait_before_drop_fragments: 255000,
```
Mallets, line 416: 255 milliseconds seems a bit too much to me. What about a less aggressive 50 ms?

commons/zenoh-config/src/lib.rs
```diff
@@ -446,25 +446,15 @@ validated_struct::validator! {
                                 /// if still no batch is available.
                                 wait_before_drop: i64,
                                 /// The maximum deadline limit for multi-fragment messages.
-                                /// When sending multi-fragment message, for each consecutive fragment the deadline from
-                                /// the start point of message transmission is calculated as
-                                /// wait_before_drop * 2 ^ min(N, ttl) where N is a fragment number.
-                                /// Thus, this parameter allows setting a limitation for maximum time for multi-fragment
-                                /// message transmission as wait_before_drop * 2 ^ ttl
-                                ttl: usize,
+                                max_wait_before_drop_fragments: i64,
```
Mallets, line 449: Are negative values allowed? It's a bit weird to me. Same thing for `wait_before_drop` and `wait_before_close`.

yellowhatter:
The perf is ok:
2024:11:27-16:48:25 throughput aggregated

DEFAULT_CONFIG.json5
yellowhatter: This is just the same as previous `ttl: 8`
Ok, let it be 50

commons/zenoh-config/src/lib.rs
yellowhatter: Negative values are interpreted as infinity:
https://github.com/eclipse-zenoh/zenoh/blob/075f2ad30f4143104bf917246f1f90f100770c3b/io/zenoh-transport/src/manager.rs#L44

And that makes sense for some cases

DEFAULT_CONFIG.json5
yellowhatter: Well, this is CC::Block functionality, which has basic wait period of 5 seconds. I guess even 60 seconds will be ok here.

commons/zenoh-config/src/lib.rs
yellowhatter: > an awful, implicit, and non-documented way of configuring infinite duration

Agree!

https://github.com/eclipse-zenoh/zenoh/issues/1614

DEFAULT_CONFIG.json5
Mallets: Do we really want to treat this case differently? To me the `wait_before_close` is more a *I didn't manage to send even 1 message*. So, as long as every fragment is sent within *5s* we are good and there is no need to set yet another option.

yellowhatter: This is the same question as the one that was the root cause to have this PR. What if I succeeded sending 100 fragments spending 1 second on each one and then spent 6 seconds on sending 101th? Should I treat this as the condition to fail and throw to the garbage 100 seconds of occupied link?
I think the same rule as for Drop is applied here: the more progress we have, the more expensive the cancellation will be.

Mallets: Well, here is not about just dropping the message but closing the session. It's a safeguard about unresponsive readers and I don't think it depends on wether we are fragmenting or not.

Mallets: Moreover, in which way `wait_before_close` is different from the `session lease`?

yellowhatter:
Updated with old functionality for `Block`.
The perf is good:
2024:11:28-11:09:01 throughput aggregated

DEFAULT_CONFIG.json5
yellowhatter: Right! I removed non-linear part for Block

Mallets (approved):

PR #1606, opened 2024-11-22
fix: precise query id rollover semantic

wyfo:
Allow managing rollover and uniqueness, as well as being faster to access.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5467,6 +5467,7 @@ dependencies = [
  "rustc_version 0.4.1",
  "serde",
  "serde_json",
+ "slab",
  "socket2 0.5.7",
  "tokio",
  "tokio-util",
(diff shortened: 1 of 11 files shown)


wyfo:
I've realized that just making the query id available right after a timeout is not appropriate, because some response may come after. A better solution in case of timeout would be to add another delay, for example 1min in which the query id cannot be reused. And to handle the case where a lot of queries time out, those pending ids could be be inserted in a dedicated set, to be reused only in case of available id shortage.
@Mallets What do you think about it?

Mallets:
I think the proposed change introduces quite a bit of complexity. We could simply rollover on `QueryId` assuming the query is no longer alive, which is a realistic assumption given the number of queries required to be performed in parallel.

wyfo:
I think it's a valid assumption, but it quite defeats the point of having varint encoding for request id.

Mallets:
There might be some room for optimization indeed but I don't believe it's critical.

wyfo:
I've updated the PR taking in account the recent discussion.

Mallets (approved):

PR #1607, opened 2024-11-22
Add trace log while dropping the messages.

evshary:
It will be convenient to debug if there is a trace log while dropping the messages

### DIFF
--- a/io/zenoh-transport/src/common/pipeline.rs
+++ b/io/zenoh-transport/src/common/pipeline.rs
@@ -235,6 +235,10 @@ impl StageIn {
                                     // Still no available batch.
                                     // Restore the sequence number and drop the message
                                     $($restore_sn)?
+                                    tracing::trace!(
+                                        "Zenoh message dropped because it's over the deadline {:?}",
+                                        deadline.lazy_deadline.wait_time
+                                    );
                                     return false;
                                 }
                                 c_guard = self.mutex.current();
--- a/io/zenoh-transport/src/common/pipeline.rs
+++ b/io/zenoh-transport/src/common/pipeline.rs
@@ -236,8 +236,8 @@ impl StageIn {
                                     // Restore the sequence number and drop the message
                                     $($restore_sn)?
                                     tracing::trace!(
-                                        "Zenoh message dropped because it's over the deadline {:?}",
-                                        deadline.lazy_deadline.wait_time
+                                        "Zenoh message dropped because it's over the deadline {:?}: {:?}",
+                                        deadline.lazy_deadline.wait_time, msg
                                     );
                                     return false;
                                 }


io/zenoh-transport/src/common/pipeline.rs
```diff
@@ -235,6 +235,10 @@ impl StageIn {
                                     // Still no available batch.
                                     // Restore the sequence number and drop the message
                                     $($restore_sn)?
+                                    tracing::trace!(
+                                        "Zenoh message dropped because it's over the deadline {:?}",
```
Mallets, line 239: Print as well `msg` so there is the information which message has been dropped.

evshary: Sounds good to me

io/zenoh-transport/src/common/pipeline.rs
```diff
@@ -235,6 +235,10 @@ impl StageIn {
                                     // Still no available batch.
                                     // Restore the sequence number and drop the message
                                     $($restore_sn)?
+                                    tracing::trace!(
```
Mallets, line 238: Sorry for catching it only now, but in case `$restore_sn` returns early, the log will not be printed but the message will still be dropped. What about moving the log before `$restore_sn`? Please double check what is actually executed in the `$restore_sn`.

evshary: Sure. I will double check that, although the return value is bool and I don't think it can be returned earlier

Mallets: You are right, then I'm a bit confused about the `$($restore_sn)?` syntax. What is `?`? :)

evshary: This means the code is optional. It will be added if the macro has the argument `$restore_sn`
For example, https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=07557c8c874e8944528c3f030c7f96dc

In our case, `$($restore_sn)?` will become `tch.sn.set(sn).unwrap()` according to https://github.com/eclipse-zenoh/zenoh/blob/dacb2a5a448287af035a3f6d35cfdbfc06c19a08/io/zenoh-transport/src/common/pipeline.rs#L295

And will become empty when it comes to https://github.com/eclipse-zenoh/zenoh/blob/dacb2a5a448287af035a3f6d35cfdbfc06c19a08/io/zenoh-transport/src/common/pipeline.rs#L265

BTW, I just learned that we can easily get expanded macro with rust-analyzer in vscode
https://stackoverflow.com/a/71454821/22130710
Learn something new each day :laughing:

Mallets: Cool!

Mallets (approved):

PR #1609, opened 2024-11-25
fix: fix z_sub_thr termination and improve perf

wyfo:
Using `Thread::unpark` `std::process::exit` allows session to be closed, and callback final print to be executed.
Callback lock accounted for 2.5% of z_sub_thr perf. Using atomic and fine-grained lock remove the performance impact (even if it doesn't really change the global outcome).

### DIFF
--- a/examples/examples/z_sub_thr.rs
+++ b/examples/examples/z_sub_thr.rs
@@ -1,3 +1,7 @@
+use std::sync::{
+    atomic::{AtomicU64, Ordering},
+    Mutex,
+};
 //
 // Copyright (c) 2023 ZettaScale Technology
 //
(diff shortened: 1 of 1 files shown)


Mallets:
Any change on the example should be aligned with all Zenoh bindings.
All examples should provide the same way of providing measures.

fuzzypixelz (approved):

wyfo:
The throughput examples are not really synchronized across bindings.
Python is the only one working as expected, i.e. printing a recap at the end. The other ones declare a drop callback which is not executed, their behavior is buggy and misleading.
Also Rust is the only one to use a mutex (with `callback_mut`). C and C++ example doesn't do that (Python don't need obviously), so their is also an example alignment issue there, as well as a more serious issue of presenting example with "buggy code" of unsynchronized callback (which can be called multiple times concurrently).

That's why I think that both C and C++ example should be fixed and aligned on this more correct new Rust version.

Mallets:
Then we need a review of the examples. Again to reiterate on the point, any change in examples needs to be coordinated across all bindings. Modifying this example right now would also bring a difference w.r.t. historical measurements.

wyfo:
> Modifying this example right now would also bring a difference w.r.t. historical measurements.

Ok, so it's better to close this PR anyway.

PR #1613, opened 2024-11-26
Add QoS configuration overwrites for publication messages

oteffahi:
Enables users to overwrite QoS parameters for publication messages through config file. These overwrites will apply over Zenoh API configurations.

Overwrites are defined per key expression, which will apply the config on publications with key expressions included (as in the relationship of inclusion between key expressions) in the configured key expressions.

The new config format is the following. Note that config fields are `Option`s, users don't have to overwrite everything.
```json5
{
  qos: {
    publication: [
      {
        key_exprs: ["demo/**", "example/key"],
        config: {
          congestion_control: "block",
          priority: "data_high",
          express: true,
          // Note: reliability and destination require Zenoh to be compiled with the "unstable" feature
          reliability: "best_effort",
          allowed_destination: "remote",
        },
      },
    ],
  }
}
```


DEFAULT_CONFIG.json5
```diff
@@ -184,6 +184,22 @@
     },
   },
 
+  //  /// Overwrite default builders for specific key expressions
```
OlivierHecart, line 187: I'd rather have :
```
publishers: {
    default_builders: [
        {
             key_exprs: ["demo/**", "example/key"],
             config: {
                    congestion_control: "Block",
                    encoding: "zenoh/example",
                    priority: "data_high",
                    express: true,
                    reliability: "best_effort",
                    allowed_destination: "remote",
             },
         },
     ],
}
```

zenoh/src/api/session.rs
```diff
@@ -826,16 +836,45 @@ impl Session {
         TryIntoKeyExpr: TryInto<KeyExpr<'b>>,
         <TryIntoKeyExpr as TryInto<KeyExpr<'b>>>::Error: Into<zenoh_result::Error>,
     {
+        let maybe_key_expr = key_expr.try_into().map_err(Into::into);
+        let mut builder_overwrites = PublisherBuilderOptionsConf::default();
```
OlivierHecart, line 840: Could we move all that in `PublisherBuilder` impl in `builders/publishers.rs` ?

commons/zenoh-config/src/builders.rs
```diff
@@ -0,0 +1,127 @@
...
+#[serde(remote = "Self")]
+pub struct PublisherBuildersConf(pub(crate) Vec<PublisherBuildersInnerConf>);
+
+impl<'de> Deserialize<'de> for PublisherBuildersConf {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let builders = PublisherBuildersConf::deserialize(deserializer)?;
+        // check for invariant: each key_expr should be unique
```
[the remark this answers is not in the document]
oteffahi: Having duplicate key expressions shouldn't be allowed, the config would possibly have different builder values for the same keyexpr, how would we decide which values to apply in that case?

We already have an issue with the fact that one publisher key expression can be included in more than one key expression in the config, but this is inherent to the use of key expressions. We shouldn't add another layer of unsoundness by allowing conflicts on the same configured keyexpr.

DEFAULT_CONFIG.json5
oteffahi: Agreed, it seperates the keyexprs from the config, which is a lot more readable.

gabrik:
@RemiBarthe :eyes:

oteffahi:
Converting to Draft to apply the following changes:
- Config should overwrite default builder and even API calls.
- Config should also apply on `session.put` and overwrite its API calls.

zenoh/src/api/builders/publisher.rs
```diff
@@ -264,6 +267,19 @@ impl IntoFuture for PublicationBuilder<PublisherBuilder<'_, '_>, PublicationBuil
     }
 }
 
+/// Marks which [`PublisherBuilder`] QoS configurations were overwritten by Zenoh config.
+/// Associated [`PublisherBuilder`] methods calls will not modify overwritten configurations
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct PublicationOverwrittenQoS {
```
OlivierHecart, line 273: Maybe I'm missing something but this looks overcomplicated to me. 
Instead of overwriting values at builder creation and mark some values as overwritten with those booleans, why don't we simply overwrite the values at builder resolution ?

oteffahi: Overwriting at builder resolution would require doing the lookup in `PublisherBuilder` resolution and in `SessionInner.resolve_put` since a `session.put` operation uses a `PublisherBuilder` to store and update the configuration. Doing the lookup in `resolve_put` means doing it for every publication made by a publisher, which would heavily impact the performance.

oteffahi: To avoid doing the lookup in `resolve_put` we'd have to do it in every implementation of `Wait for PublicationBuilder` (currently 4)

oteffahi: Done.

commons/zenoh-config/src/builders.rs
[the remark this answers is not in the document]
oteffahi: Removed it as per our last discussion.

OlivierHecart (approved):

PR #1615, opened 2024-11-27
Fix message storm when using multiple peers with multicast

OlivierHecart:
The first commit avoids the declare message loop that occurred when using multiple peers with multicast.

Then this PR prevents any routing from/to multicast groups from the same process (router) as this can easily lead to messages loop.

### DIFF
--- a/zenoh/src/net/routing/hat/client/mod.rs
+++ b/zenoh/src/net/routing/hat/client/mod.rs
@@ -275,8 +275,8 @@ impl HatBaseTrait for HatCode {
         _expr: &mut RoutingExpr,
     ) -> bool {
         src_face.id != out_face.id
-            && (out_face.mcast_group.is_none()
-                || (src_face.whatami == WhatAmI::Client && src_face.mcast_group.is_none()))
+            && out_face.mcast_group.is_none()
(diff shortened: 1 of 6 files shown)


zenoh/src/net/routing/hat/client/mod.rs
```diff
@@ -275,10 +275,8 @@ impl HatBaseTrait for HatCode {
         _expr: &mut RoutingExpr,
     ) -> bool {
         src_face.id != out_face.id
-            && match (src_face.mcast_group.as_ref(), out_face.mcast_group.as_ref()) {
-                (Some(l), Some(r)) => l != r,
-                _ => true,
-            }
+            && out_face.mcast_group.is_none()
```
wyfo, line 278: why this change?

sashacmc (approved):

zenoh/src/net/routing/hat/client/mod.rs
Mallets: It's to avoid sending data to a multicast group when in client mode.
This is also to align with zenoh-pico where joining a multicast group can only be done when in peer mode.

Mallets (approved):

PR #1618, opened 2024-11-29
Fix clippy warnings

OlivierHecart:


commons/zenoh-protocol/src/transport/open.rs
```diff
@@ -11,69 +11,77 @@
 // Contributors:
 //   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
 //
+
+//! # Open message
```
evshary, line 15: Just out of curiosity, do you move the comments before `use mod` intentionally?

evshary (approved):
Give my +1 although I'm not the committer :laughing:

PR #1619, opened 2024-11-29
Fix bug declaring/undeclaring tokens with the same key

OlivierHecart:
There was 2 possible paths here: 
1. Change `declare_liveliness_inner` to manage twin tokens (similar to subscribers).
2. Change `undeclare_liveliness_inner` to manage twin tokens independently.

Since the plan to have a routing less client is dead, the twin management in `Session` does not make much sense any more as it is also performed in routing. It just adds complexity for few or inexistent benefit.
So I opted for path 2.

### DIFF
--- a/zenoh/src/api/session.rs
+++ b/zenoh/src/api/session.rs
@@ -71,7 +71,7 @@ use zenoh_task::TaskController;
 use crate::api::selector::ZenohParameters;
 #[cfg(feature = "unstable")]
 use crate::api::{
-    liveliness::{Liveliness, LivelinessTokenState},
+    liveliness::Liveliness,
     publisher::Publisher,
     publisher::{MatchingListenerState, MatchingStatus},
     query::LivelinessQueryState,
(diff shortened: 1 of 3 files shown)


zenoh/src/api/session.rs
```diff
@@ -1555,10 +1555,7 @@ impl SessionInner {
         let mut state = zwrite!(self.state);
         tracing::trace!("declare_liveliness({:?})", key_expr);
         let id = self.runtime.next_id();
-        let tok_state = Arc::new(LivelinessTokenState {
-            id,
-            key_expr: key_expr.clone().into_owned(),
-        });
+        let tok_state = Arc::new(LivelinessTokenState { id });
 
         state.tokens.insert(tok_state.id, tok_state.clone());
```
fuzzypixelz, line 1560: Now `state.tokens` is essentially a map form token ids to token ids; it ought to simply be a set of token ids.

wyfo: Actually, I'm not even sure we need an set.

OlivierHecart: Well, it in fact became completely useless. I completely removed it.

fuzzypixelz (approved):

PR #1623, opened 2024-12-02
Change default config listen endpoints when transport_tcp feature is disabled

OlivierHecart:
When `transport_tcp` feature is enabled, default listen endpoints are (as before) : 
```
{
    router: ["tcp/[::]:7447"],
    peer: ["tcp/[::]:0"],
    client: [],
}
```
When `transport_tcp` feature is disabled, default listen endpoints are (now) : 
```
{
    router: [],
    peer: [],
    client: [],
}
```

This avoids having this error: `Unsupported protocol: tcp.` when running a router or peer build with `transport_tcp` feature disabled and the default config.


YuanYuYuan:
LGTM. Thanks @OlivierHecart

PR #1633, opened 2024-12-06
SHM channels

yellowhatter:


yellowhatter:
We decided to usie io_uring transport instead of SHM channels

PR #1635, opened 2024-12-06
Prioritize endpoint config over configuration file

### LINKED ISSUE #1634: Endpoint config should not be overwritten by Zenoh config file
### Describe the bug

Some link configurations can either be passed by endpoint parameters or via the zenoh config file. Currently the config file is given priority over endpoint parameters, which shouldn't be the case. Config file should be a fallback if no config is given on the specific endpoint.

### To reproduce

1. Create a zenoh config file with TLS config.
2. Set `transport/link/tls/enable_mtls` to true.
3. Set `listen/endpoints` to `["tls/[::]:7447#enable_mtls=false"]`
4. Start a peer/router with the config file, connect to it with a client via tls and notice how mtls is enabled.

### System info

Zenoh 1.0.4

oteffahi:
Closes #1634

### DIFF
--- a/commons/zenoh-protocol/src/core/endpoint.rs
+++ b/commons/zenoh-protocol/src/core/endpoint.rs
@@ -385,9 +385,8 @@ impl ConfigMut<'_> {
             self.0.address(),
             self.0.metadata(),
             parameters::from_iter(parameters::sort(parameters::join(
-                // Endpoint config overwrites incoming config
-                iter.map(|(k, v)| (k.borrow(), v.borrow())),
                 self.0.config().iter(),
+                iter.map(|(k, v)| (k.borrow(), v.borrow())),
             ))),
         )?;
 
--- a/commons/zenoh-protocol/src/core/endpoint.rs
+++ b/commons/zenoh-protocol/src/core/endpoint.rs
@@ -385,8 +385,9 @@ impl ConfigMut<'_> {
             self.0.address(),
             self.0.metadata(),
             parameters::from_iter(parameters::sort(parameters::join(
-                self.0.config().iter(),
+                // Endpoint config overwrites incoming config
                 iter.map(|(k, v)| (k.borrow(), v.borrow())),
+                self.0.config().iter(),
             ))),
         )?;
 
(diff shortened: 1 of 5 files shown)


commons/zenoh-protocol/src/core/endpoint.rs
```diff
@@ -385,8 +385,9 @@ impl ConfigMut<'_> {
             self.0.address(),
             self.0.metadata(),
             parameters::from_iter(parameters::sort(parameters::join(
-                self.0.config().iter(),
+                // Endpoint config overwrites incoming config
                 iter.map(|(k, v)| (k.borrow(), v.borrow())),
+                self.0.config().iter(),
```
Mallets, line 390: I don't think this is the right change to solve the issue. 
IMHO this changes is wrong w.r.t. the expected behaviour of this function where the iter in this is overwritten by the config. 

The change should be done in the transport like here: https://github.com/eclipse-zenoh/zenoh/blob/233240622f548c18fe41450144010ad1b7e1fe17/io/zenoh-transport/src/unicast/manager.rs#L386.

oteffahi: This function is currently only used for merging config file with endpoint config. Would renaming it to indicate its purpose solve the issue then?

io/zenoh-transport/zenoh-test-unix-socket-0.sock.lock
Mallets, line 1: Remove this file

io/zenoh-transport/zenoh-test-unix-socket-1.sock.lock
Mallets, line 1: Remove this file

oteffahi: oops 😅

io/zenoh-transport/zenoh-test-unix-socket-0.sock.lock
oteffahi: oops 😅

Mallets (approved):

zenoh/tests/tcp_buffers.rs
Mallets, line 1: This test shouldn't be part of `zenoh/tests` but rather in `zenoh-links` itself or `zenoh-transport`.

Mallets: All link-specific tests are part of `zenoh-transport`. We can move the test in a different PR.

PR #1642, opened 2024-12-09
DEFAULT_CONFIG.json5: add missing default config for autoconnect.client

JEnoch:
In `DEFAULT_CONFIG.json5` both `scouting.multicast.autoconnect` and `scouting.gossip.autoconnect` had the default values correctly set for `router` and `peer` modes. But the default value for `client` mode was missing.
It's actually set to `["router", "peer"]` [here](https://github.com/eclipse-zenoh/zenoh/blob/233240622f548c18fe41450144010ad1b7e1fe17/commons/zenoh-config/src/defaults.rs#L84).

I also tried to clarify the description of `autoconnect` config which seems not clear to the user on why there are different settings for `router`, `peer` and `client` in a config file which applies to only 1 of those cases at startup.

### DIFF
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -128,8 +128,8 @@
       /// The time-to-live on multicast scouting packets
       ttl: 1,
       /// Which type of Zenoh instances to automatically establish sessions with upon discovery on UDP multicast.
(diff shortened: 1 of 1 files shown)


Mallets:
`downsampling_by_keyexpr` test is failing but is not related to this PR.

PR #1649, opened 2024-12-10
Fix deadlock in subscriber undeclaration

oteffahi:
Closes https://github.com/eclipse-zenoh/zenoh/issues/1650

### DIFF
--- a/zenoh/src/api/session.rs
+++ b/zenoh/src/api/session.rs
@@ -1584,8 +1584,6 @@ impl SessionInner {
                                     false,
                                 )
                             }
-                        } else {
-                            drop(state);
                         }
                     } else {
                         drop(state);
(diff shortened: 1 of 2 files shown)


wyfo (approved):

zenoh/src/api/session.rs
```diff
@@ -1574,19 +1574,31 @@ impl SessionInner {
...
+                                let state = zread!(self.state);
+                                self.update_matching_status(
+                                    &state,
+                                    &sub_state.key_expr,
+                                    MatchingStatusType::Subscribers,
+                                    false,
+                                )
+                            }
+                        } else {
+                            drop(state);
```
wyfo: This `drop` is not necessary per se, as the code is at the end of the function, but I think it's a good thing to keep it, as it would avoid further error

OlivierHecart (approved):

zenoh/src/api/session.rs
oteffahi: Removed it, I found it ugly 😅

PR #1661, opened 2024-12-11
Fix bug leading links to be closed just after establishment with multiple peers

OlivierHecart:
Apply a hardcoded timeout of 10 seconds to connection attempts.


Mallets:
Superseded by #1663

PR #1662, opened 2024-12-12
Add CI for checking markdown format.

evshary:
The README format is sometimes not well-formatted. Add CI to make sure this can be checked automatically.
For vscode developer, installing `markdownlint` plugin can provide the lint warning.


evshary:
@Mallets do you think it's worth adding the markdown CI check here? If yes, I can apply it to all other repos

Mallets:
I think it's good to have the check enabled in the CI.

PR #1664, opened 2024-12-12
Fix stats bugs

OlivierHecart:
When routing, the `tx_z_put_pl_bytes` of the source session was increased instead fo the `tx_z_put_pl_bytes` of the destination sessions.


gabrik (approved):
LGTM

gabrik:
The test is flaky, as this PR does not touch downsampling.

PR #1676, opened 2024-12-17
Stop publication cache task on disconnected channels

fuzzypixelz:
While debugging https://github.com/ZettaScaleLabs/rmw_zenoh/issues/44, GDB backtraces showed that the `Net` runtime was apparently blocked executing this publication cache task; applying this patch solved the issue.

Even though the blocking behavior I observed might've been a consequence of different underlying issue, I think this patch should be applied anyway. This is because:
1. If I understand correctly, the `tokio::select!` macro doesn't execute poll each branch on every iteration, but rather uses a random number generator to choose the branch.
2. Errors should be handled ;)


Mallets:
*Advanced Pub/Sub* is now available in `main`.
@fuzzypixelz is this PR still relevant?

fuzzypixelz:
> _Advanced Pub/Sub_ is now available in `main`. @fuzzypixelz is this PR still relevant?

I think so, I don't think that the publication cache code has changed.

OlivierHecart:
Publication cacheis deprecated.

PR #1680, opened 2024-12-19
Add expression caching to routing Resource tree

OlivierHecart:
The full key expression string is stored in each `Resource` of the routing Resource tree so that `Resource::expr` does not need to perform recursive string concatenation to rebuild the full key expression string. This consumes more memory but saves a lot of CPU.


OlivierHecart:
Waiting https://github.com/eclipse-zenoh/zenoh/pull/1678

zenoh/src/net/routing/dispatcher/resource.rs
```diff
@@ -345,7 +349,7 @@ impl Resource {
 
     #[cfg(test)]
     pub fn print_tree(from: &Arc<Resource>) -> String {
-        let mut result = from.expr();
+        let mut result = from.expr().to_string();
```
Mallets, line 352: I agree with the change of having `expr() -> &str`. I'm not sure `expr_to_string() -> String` is really needed since, like here, you can easily do `expr().to_string()`.

OlivierHecart: I agree, `expr_to_string()` is a pure, non very useful, syntaxic shortcut. The main reason fo it was easier query replace and easier discrimination of places we need an owned string from places where we don't.
It can be removed.

Mallets (approved):

PR #1686, opened 2025-01-02
[ Tracing] Add instrumentation support

Jezza:
When spawning futures, we lose the original span information.  
This should fix that.  
I tried to capture the obvious spots, but I probably didn't get them all.

### DIFF
--- a/commons/zenoh-runtime/src/lib.rs
+++ b/commons/zenoh-runtime/src/lib.rs
@@ -34,6 +34,7 @@ use std::{
 use lazy_static::lazy_static;
 use serde::Deserialize;
 use tokio::runtime::{Handle, Runtime, RuntimeFlavor};
+use tokio::task::JoinHandle;
 use zenoh_macros::{GenericRuntimeParam, RegisterParam};
 use zenoh_result::ZResult as Result;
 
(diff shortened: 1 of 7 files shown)


Jezza:
Fails on formatting.  
I'll update it tomorrow.

commons/zenoh-runtime/src/lib.rs
```diff
@@ -141,6 +158,9 @@ impl ZRuntime {
                 }
             }
         }
+
+        let f = tracing::Instrument::instrument(f, tracing::Span::current());
```
Mallets, line 162: For performance reasons I'd put all these tracing calls behind a feature flag that is disabled by default.

Jezza: Makes sense.

I'll add it in.

commons/zenoh-runtime/src/lib.rs
```diff
@@ -125,6 +128,20 @@ pub enum ZRuntime {
 }
 
 impl ZRuntime {
+    pub fn handle(&self) -> &Handle {
+        ZRUNTIME_POOL.get(self)
+    }
+
+    pub fn spawn<F>(&self, future: F) -> JoinHandle<F::Output>
```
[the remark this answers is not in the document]
Jezza: The reason behind `spawn` was because a lot of futures were being spawned without an opportunity to add the instrumentation.  
With the feature gate now, this just does nothing, but just delegate to the `Handle`.

`handle` itself was just because I would prefer the handle to be ignored.  
I'd prefer if devs just spawned tasks via `spawn` instead of derefing to the handle, then using that.

That way we can forward the instrumentation.

I've removed it for now, but at some point I'd like to remove the Deref impl.

Mallets (approved):

PR #1689, opened 2025-01-06
fix(storage-manager): incorrectly revive key after deletion

J-Loudet:
This commit fixes a bug where a key would be incorrectly "revived" in a scenario where a deletion is followed by a Wildcard Put.

Assuming we have two Replicas, R1 and R2, the problematic scenario was:

0. Replicas R1 and R2 are disconnected.
1. `z_put "a" = 1 @t1` on Replica R1.
2. `z_delete "a" @t2` on Replica R2.
3. `z_put "**" = 42 @t3` on Replica R2.
4. Connect Replicas R1 and R2.

After aligning, before this fix, Replicas R1 and R2 would have `"a" = 42 @t3` in their storage whereas there should have been no key associated to `"a"` because of the delete at `@t2`.

This commit fixes this bug.

* plugins/zenoh-plugin-storage-manager/src/replication/core/aligner_reply.rs: instead of eagerly applying a Wildcard Update that should not be applied, do nothing. The comments were updated to provide more context.

### DIFF
--- a/plugins/zenoh-plugin-storage-manager/src/replication/core/aligner_reply.rs
+++ b/plugins/zenoh-plugin-storage-manager/src/replication/core/aligner_reply.rs
@@ -505,18 +505,22 @@ impl Replication {
                     return false;
                 }
             }
-            // The timestamp of the Wildcard Update is greater than that of the Event, we apply the
-            // Wildcard Update.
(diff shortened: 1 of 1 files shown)


oteffahi (approved):
Tested and validated internally.
LGTM.

Issue #1657, opened 2024-12-11
Add Advanced Pub/Sub Heartbeat


OlivierHecart:
### Describe the feature

Update the AdvancedPublisher so that it can optionally periodically send it's last published sample sequence number on the meta key expression (`@adv/pub/<zid>/<eid>/_/@/user/key`).
Update the AdvancedSubscriber to optionally subscribe to the periodic sequence number above and retrieve missed sample if detected.

PR #1701, opened 2025-01-09
Add advanced pub/sub last sample miss detection

oteffahi:
Closes https://github.com/eclipse-zenoh/zenoh/issues/1657

### DIFF
--- a/zenoh-ext/src/advanced_subscriber.rs
+++ b/zenoh-ext/src/advanced_subscriber.rs
@@ -989,9 +989,13 @@ impl<Handler> AdvancedSubscriber<Handler> {
                     let mut lock = zlock!(statesref);
                     let states = &mut *lock;
                     let entry = states.sequenced_states.entry(source_id);
-                    if matches!(&entry, Entry::Vacant(_)) && states.global_pending_queries > 0 {
-                        tracing::debug!("Skipping heartbeat on '{}' from publisher that is currently being pulled by global query", heartbeat_keyexpr);
-                        return;
+                    if matches!(&entry, Entry::Vacant(_)) {
+                        // NOTE: API does not allow both heartbeat and periodic_queries
+                        spawn_periodoic_queries!(states, source_id, statesref.clone());
+                        if states.global_pending_queries > 0 {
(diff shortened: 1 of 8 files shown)


zenoh-ext/src/advanced_subscriber.rs
```diff
@@ -921,6 +940,102 @@ impl<Handler> AdvancedSubscriber<Handler> {
...
+                        pending_queries: 0,
+                        pending_samples: BTreeMap::new(),
+                    });
+                    // TODO: add state to avoid sending multiple queries for the same heartbeat if its periodicity is higher than the query response time
+
+                    // check that it's not an old sn or a pending sample's sn
+                    if (state.last_delivered.is_none()
+                        || state.last_delivered.is_some_and(|sn| heartbeat_sn > sn))
+                        && !state.pending_samples.contains_key(&heartbeat_sn)
+                    {
```
oteffahi, line 996: This TODO could be covered by not sending a query unless there are no pending queries. Does that sound right @OlivierHecart ?

oteffahi: Doing so seems to risk that (in some potential high-load cases) heartbeats would get denied by other querying mechanisms and never result in a heartbeat query...

zenoh-ext/src/advanced_subscriber.rs
```diff
@@ -921,6 +940,102 @@ impl<Handler> AdvancedSubscriber<Handler> {
...
+                            heartbeat_keyexpr
+                        );
+                        return;
+                    };
+
+                    let mut lock = zlock!(statesref);
+                    let states = &mut *lock;
+                    let entry = states.sequenced_states.entry(source_id);
+                    if matches!(&entry, Entry::Vacant(_)) && states.global_pending_queries > 0 {
+                        tracing::debug!("Skipping heartbeat on '{}' from publisher that is currently being pulled by liveliness task", heartbeat_keyexpr);
```
OlivierHecart, line 980: Please update wrong log msg ("by global query")

OlivierHecart (approved):

PR #1708, opened 2025-01-13
Fix TLS endpoint port propagation in gossip

Mallets:
Properly store the TLS listener locator and endpoint as returned by the OS upon socket opening.

### DIFF
--- a/io/zenoh-links/zenoh-link-tls/src/unicast.rs
+++ b/io/zenoh-links/zenoh-link-tls/src/unicast.rs
@@ -426,7 +426,12 @@ impl LinkManagerUnicastTrait for LinkManagerUnicastTls {
             format!("{host}:{local_port}"),
             endpoint.metadata(),
         )?;
-
+        let endpoint = EndPoint::new(
+            locator.protocol(),
+            locator.address(),
+            locator.metadata(),
+            endpoint.config(),
+        )?;
         self.listeners
             .add_listener(endpoint, local_addr, task, token)
             .await?;
--- a/io/zenoh-links/zenoh-link-tls/src/unicast.rs
+++ b/io/zenoh-links/zenoh-link-tls/src/unicast.rs
@@ -11,7 +11,14 @@
 // Contributors:
 //   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
 //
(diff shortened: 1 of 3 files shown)


io/zenoh-links/zenoh-link-tls/src/unicast.rs
```diff
@@ -426,7 +426,12 @@ impl LinkManagerUnicastTrait for LinkManagerUnicastTls {
             format!("{host}:{local_port}"),
             endpoint.metadata(),
         )?;
-
+        let endpoint = EndPoint::new(
+            locator.protocol(),
+            locator.address(),
```
oteffahi, line 431: In QUIC link this is `local_addr.to_string`. Not sure if it can be different from `locator.address()` in this case.

oteffahi (changes requested):

io/zenoh-links/zenoh-link-tls/src/unicast.rs
Mallets: `local_addr.to_string()` allocates and then the creation of the `EndPoint` creation allocates a second time. `locator.address()` returns `&str` so it allocates one time less.

oteffahi: I was more concerned about if the contents of the two variables diverge. Indeed there is a case where they diverge, it seems that `locator.address()` contains the unresolved hostname, while `local_addr` contains the resolved one.

For example when listening on `localhost:0`:
- locator.address(): "localhost:49999"
- local_addr.to_string(): "[::1]:49999"

oteffahi: Looks like multicast scouting fails to connect when the propagated locator is `tls/[::1]:49999`, so it's correct to use `locator.address()`. Will check if QUIC link also has this issue

oteffahi: After some testing, I can confirm there is a similar issue on QUIC.

Mallets: Just added

oteffahi (approved):
LGTM.

oteffahi (approved):

PR #1712, opened 2025-01-14
Use client multicast autoconnect config

oteffahi:
Clients were hardcoded to autoconnect to routers through multicast scouting. This PR makes clients use the autoconnect config, and updates default config to autoconnect to routers to maintain the current client behavior.

While on the subject, this PR removes all mention of clients in gossip config (since they do not participate to gossip).


Mallets:
Failing CI unrelated to this PR

PR #1599, opened 2024-11-20
POC for session-local optimization

yellowhatter:
POC for optimization for session-local  communication.

For session-local 8 byte payload:
7M msg/s -> 9M msg/s

(tested with z_local_pub_sub_thr example)

PR #1713, opened 2025-01-14
refactor: lazily compute the push body

wyfo:
It allows to optimize the local workflow, as it remove all the memmove caused by the body computation if there is no remote route.

Credit to @yellowhatter for the original PR #1599

### DIFF
--- a/zenoh/src/net/routing/dispatcher/pubsub.rs
+++ b/zenoh/src/net/routing/dispatcher/pubsub.rs
@@ -410,10 +410,8 @@ pub fn route_data(
                             drop(tables);
                             #[cfg(feature = "stats")]
                             if !admin {
-                                inc_stats!(face, rx, user, payload);
                                 inc_stats!(outface, tx, user, payload);
                             } else {
-                                inc_stats!(face, rx, admin, payload);
                                 inc_stats!(outface, tx, admin, payload);
                             }
 
--- a/zenoh/src/net/routing/dispatcher/pubsub.rs
+++ b/zenoh/src/net/routing/dispatcher/pubsub.rs
@@ -353,6 +353,9 @@ macro_rules! inc_stats {
     };
 }
 
+// having all the arguments instead of an intermediate struct seems to enable a better inlining
+// see https://github.com/eclipse-zenoh/zenoh/pull/1713#issuecomment-2590130026
+#[allow(clippy::too_many_arguments)]
 pub fn route_data(
     tables_ref: &Arc<TablesLock>,
     face: &FaceState,
--- a/zenoh/src/net/routing/dispatcher/pubsub.rs
+++ b/zenoh/src/net/routing/dispatcher/pubsub.rs
@@ -19,10 +19,7 @@ use std::sync::Arc;
 use zenoh_core::zread;
 use zenoh_protocol::{
     core::{key_expr::keyexpr, Reliability, WireExpr},
-    network::{
-        declare::{ext, SubscriberId},
-        Push,
-    },
+    network::{declare::SubscriberId, push::ext, Push},
     zenoh::PushBody,
 };
 use zenoh_sync::get_mut_unchecked;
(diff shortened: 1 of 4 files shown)


wyfo:
The local throughput example found in #1599 goes from 11,5Mmsg/s to 13,5Mmsg/s.
The remote throughput example isn't impacted as expected.

However, there is a subtle and interesting change in the generated assembly that may interest you @Mallets and @yellowhatter 
Let's look at the flamegraphs
BEFORE
before
AFTER
after
As you can see, with the PR, `route_data` is now fully inlined. We can also see that the `memmove` part is quite diminished.

We could expect looking at the graphs that the new code should be faster, and in fact it is. It is not visible in the remote throughput example, because, again, this one is fully limited by the subscriber, so any publisher improvement is invisible. But if we use a dummy subscriber (I do it for my tests, just a `z_sub_thr` but with the `Push` message handling removed in the transport code), we can see the publisher going from 6,5Mmsg/s to 7,3Mmsg/s.
So this change is completely worth for both local and remote workflows!

Now, how can we explain this? My quick theory is that by not using an intermediate `Push` struct, the compiler is better able to track the value for the arguments of `route_data`, and has a better interest in inlining it. But obviously, I'm not sure of anything.

wyfo:
The open questions are:
- should we do the same for the requests? I think yes, maybe in another PR?
- should we change the signature of `Primitives` methods, at least `send_push` and `send_requests`? I'm not sure. For the `Demux`, the current signature is more natural, and maybe for other implementors too. I'm fine having this lazy implementation exposed only for `Face` to be used in `Session`.

zenoh/src/net/routing/dispatcher/pubsub.rs
```diff
@@ -391,39 +388,47 @@ macro_rules! inc_stats {
...
-                msg.wire_expr.suffix.as_ref()
+                wire_expr.suffix.as_ref()
             );
-            let mut expr = RoutingExpr::new(&prefix, msg.wire_expr.suffix.as_ref());
+            let mut expr = RoutingExpr::new(&prefix, wire_expr.suffix.as_ref());
 
             #[cfg(feature = "stats")]
             let admin = expr.full_expr().starts_with("@/");
             #[cfg(feature = "stats")]
+            let mut payload = payload();
```
wyfo, line 415: Should we count the payload in the stats even if it is not routed? This is the case in the current implementation, so I keep it as is, it may not be relevant after all, may it?

yellowhatter: In such cases I always say myself: "zenoh aims to be location-agnostic". So we should have the same stats for intra-Session ant inter-Session pubsub

wyfo: Currently, we don't have dedicated stats for local subscriber. But, because the remote code was always executed, we had the stats incremented, but only for `rx`, not `tx`. As I said above, I don't think this is the correct behavior. And I don't think either we should increment `rx` stats from a `put`; this increment is maybe an error coming from the fact that `route_data` is used by both `Demux` and `put`.

wyfo: Maybe the `rx` stats increment should be controlled by a boolean parameter, and only be activated by the `Demux` call

zenoh/src/net/routing/dispatcher/pubsub.rs
```diff
@@ -388,42 +385,53 @@ macro_rules! inc_stats {
     };
 }
 
+// having all the arguments instead of an intermediate struct seems to enable a better inlining
+// see https://github.com/eclipse-zenoh/zenoh/pull/1713#issuecomment-2590130026
+#[allow(clippy::too_many_arguments)]
```
Mallets, line 390: Even it seems to provide advantages in terms of performance, these kind of changes hinder code maintainability and future extensibility. In the past we had more granular function signatures, like the one proposed in this PR, and any change/addition in the protocol required heavy backtracking and manual extension of countless functions.

So, in general I'm against this change. @OlivierHecart ?

wyfo: The thing is, we need to be more lazy for better performance. And, I'm not touching the `Primitives`/`EPrimitives` traits in this PR. The lazy form is only for publication. Also, because the `Push` passed in argument was already destructurized and each, you still had some backtracking of protocol changes. I don't think this PR adds much.

yellowhatter: @Mallets here I want to say +1 to @wyfo . And the performance gain we have here is really a value.

yellowhatter (approved):

zenoh/src/net/routing/dispatcher/pubsub.rs
OlivierHecart: Couldn't `send_push_lazy` and `route_data` take a `impl FnOnce() -> Push` instead of `impl FnOnce() -> PushBody` ? Wouldn't this solve the issue ?

wyfo: > Couldn't send_push_lazy and route_data take a impl FnOnce() -> Push instead of impl FnOnce() -> PushBody ? Wouldn't this solve the issue ?

No, because you need the `WireExpr` and the `ext_nodeid` to compute before lazily compute the body or not.

OlivierHecart (approved):

Mallets (approved):

Issue #1563, opened 2024-10-24
Dependency on `lockfree` an unmaintained and unsound crate


VorpalBlade:
### Describe the bug

https://lib.rs/crates/zenoh-shm depends on https://lib.rs/crates/lockfree which is unmaintained and also unsound (using std::mem::uninitialized). I have tried reaching out to the author of the crate on their gitlab and plan to file an rustsec advisory when the 90 days are up (14th November).

As your reverse dependency is one of the more downloaded ones I thought I would give you a heads up so that you can switch to something else before then.

### To reproduce

N/A

### System info

N/A

PR #1714, opened 2025-01-14
SHM: optimize metadata

### LINKED ISSUE #1703: SHM: storages optimization
### Describe the feature

Tagged to #1697 
We have potential to optimize watchdog and header storages by joining them. That will complexify the code, but reduce memory, SHM buffer allocation, wire and SHM attachment operation costs.

yellowhatter:
closes #1703 

Combined SHM metadata in single storage that is now identified with single descriptor.

By-the-way fixed undiscovered edge case bug with buffer validity check in the receiver code.

Improves SHM performance by 35% (1.7M -> 2.3M) that moves SHM efficiency threshold to somewhat less than 512 bytes.

Breaks backward compatibility for SHM communication. Non-SHM is not affected.


yellowhatter:
Put ProtocolID and ChunkDescriptor into SHM metadata: less wire overhead and less internal structure sizes for SHM 

Gains 2% more performance for SHM, but this is not the only thing: it opens a way to implement publish-before-allocated-and-written concept, that should dramatically increase SHM performance and improve latency!

commons/zenoh-shm/src/metadata/allocated_descriptor.rs
```diff
@@ -0,0 +1,58 @@
+//
+// Copyright (c) 2023 ZettaScale Technology
```
Mallets, line 2: Make sure it's 2025 for newly-created files.

commons/zenoh-shm/src/metadata/segment.rs
```diff
@@ -0,0 +1,69 @@
...
+
+use super::descriptor::{MetadataIndex, MetadataSegmentID};
+use crate::{header::chunk_header::ChunkHeaderType, posix_shm::struct_in_shm::StructInSHM};
+
+const METADATA_SEGMENT_PREFIX: &str = "metadata";
+
+#[stabby::stabby]
+pub struct Metadata<const S: usize> {
+    headers: [ChunkHeaderType; S],
+    watchdogs: [AtomicU64; S], // todo: replace with (S + 63) / 64 when Rust supports it
```
Mallets, line 27: I'm curious to know more about the *todo* comment here. What's the intended behaviour and current limitations?

commons/zenoh-shm/src/metadata/segment.rs
```diff
@@ -0,0 +1,69 @@
...
+        let watchdog_index = index / 64;
+        let watchdog_mask_index = index % 64;
+        (
+            self.headers.as_ptr().offset(index as isize),
+            self.watchdogs.as_ptr().offset(watchdog_index as isize),
+            1u64 << watchdog_mask_index,
+        )
+    }
+
+    pub fn count(&self) -> usize {
```
Mallets, line 50: If `count` is const, couldn't be an associated `const` instead of a function?

Mallets (approved):

Mallets:
LGTM. I'll just wait to merge it because of the `1.1.2` release coming out tomorrow. I'd rather avoid introducing a breaking change in a patch release.

yellowhatter:
> LGTM. I'll just wait to merge it because of the `1.1.2` release coming out tomorrow. I'd rather avoid introducing a breaking change in a patch release.

Sure, that's intended as I didn't put "release" label :)

commons/zenoh-shm/src/metadata/segment.rs
yellowhatter: Watchdogs are packed as bits of `u64`. This means that for `S` headers we need `S/64`-sized `u64` array. However, Rust doesn't support const arithmetic for const generic parameter like this:
`watchdogs: [AtomicU64; (S + 63) / 64]`

Mallets: Thanks for the explanation

commons/zenoh-shm/src/metadata/segment.rs
yellowhatter: Changed in a bit different way: `pub const fn`
The reason is that associated constant cannot be called through object instance, only as `Type::constant`, so I need to carry full object type (and it is compilcated) everywhere
`const fn` is intended to work in the way we need

commons/zenoh-shm/src/metadata/allocated_descriptor.rs
yellowhatter: Done

Mallets:
@yellowhatter It's well understood that SHM is still marked as unstable, meaning that breaking changes may be introduced.
However, although I've already approved this PR, I was thinking wether we need to deal with this breaking change in a backward-compatible manner.
In Zenoh we have a `Patch` extension that identifies the actual protocol patch during the protocol handshake. Would that help, or be of any use in this case?

yellowhatter:
> @yellowhatter It's well understood that SHM is still marked as unstable, meaning that breaking changes may be introduced. However, although I've already approved this PR, I was thinking wether we need to deal with this breaking change in a backward-compatible manner. In Zenoh we have a `Patch` extension that identifies the actual protocol patch during the protocol handshake. Would that help, or be of any use in this case?

The idea is clear. I think we can add smth similar to `Patch` into existing `SHM` extension that is used to negotiate on SHM usage and supported data protocols at transport level. There will be no "backward compatibility", Zenoh will just fallback to non-SHM mode ;)

Mallets:
> @Mallets The idea is clear. I think we can add smth similar to `Patch` into existing `SHM` extension that is used to negotiate on SHM usage and supported data protocols at transport level. There will be no "backward compatibility", Zenoh will just fallback to non-SHM mode ;)

I think it makes sense. This would allow us to better control the introduction of additional changes in the future.

yellowhatter:
> > @Mallets The idea is clear. I think we can add smth similar to `Patch` into existing `SHM` extension that is used to negotiate on SHM usage and supported data protocols at transport level. There will be no "backward compatibility", Zenoh will just fallback to non-SHM mode ;)
> 
> I think it makes sense. This would allow us to better control the introduction of additional changes in the future.

I have recently added versioning support. There is no additional OAM extension - I made version negotiation in SHM probing segment. That is more flexible and saves bytes on wire

commons/zenoh-shm/src/metadata/descriptor.rs
```diff
@@ -0,0 +1,120 @@
...
+
+#[derive(Clone)]
+pub struct OwnedMetadataDescriptor {
+    pub(crate) segment: Arc<MetadataSegment>,
+    header: *const ChunkHeaderType,
+    watchdog_atomic: *const AtomicU64,
+    watchdog_mask: u64,
+}
+
+unsafe impl Send for OwnedMetadataDescriptor {}
```
Mallets, line 46: What are the guarantees of these `unsafe impls` for `Send` and `Sync` traits?

commons/zenoh-shm/src/metadata/descriptor.rs
```diff
@@ -0,0 +1,120 @@
...
+            segment,
+            header,
+            watchdog_atomic,
+            watchdog_mask,
+        }
+    }
+
+    #[inline(always)]
+    pub fn header(&self) -> &ChunkHeaderType {
+        unsafe { &(*self.header) }
```
Mallets, line 66: Add `# SAFETY`  comment

commons/zenoh-shm/src/metadata/descriptor.rs
```diff
@@ -0,0 +1,120 @@
...
+
+    pub fn confirm(&self) {
+        unsafe {
+            (*self.watchdog_atomic)
+                .fetch_or(self.watchdog_mask, std::sync::atomic::Ordering::SeqCst);
+        };
+    }
+
+    pub(crate) fn validate(&self) -> u64 {
+        unsafe {
```
Mallets, line 77: Add `# SAFETY`  comment.

commons/zenoh-shm/src/metadata/descriptor.rs
```diff
@@ -0,0 +1,120 @@
...
+        }
+    }
+
+    #[inline(always)]
+    pub fn header(&self) -> &ChunkHeaderType {
+        unsafe { &(*self.header) }
+    }
+
+    pub fn confirm(&self) {
+        unsafe {
```
Mallets, line 70: Add `# SAFETY`  comment.

commons/zenoh-shm/src/metadata/descriptor.rs
```diff
@@ -0,0 +1,120 @@
...
+}
+
+// The ordering strategy is important. See storage implementation for details
+impl Ord for OwnedMetadataDescriptor {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        match self.watchdog_atomic.cmp(&other.watchdog_atomic) {
+            core::cmp::Ordering::Equal => {}
+            ord => return ord,
+        }
+        self.watchdog_mask.cmp(&other.watchdog_mask)
```
Mallets, line 97: It's a very stylistic comment but this check could be moved to the line 94 as part of the `Equal` branch body.

commons/zenoh-shm/src/metadata/storage.rs
```diff
@@ -0,0 +1,84 @@
...
+        Ok(AllocatedMetadataDescriptor::new(descriptor))
+    }
+
+    pub fn reclaim(&self, descriptor: OwnedMetadataDescriptor) {
+        // header deallocated - increment it's generation to invalidate any existing references
+        descriptor
+            .header()
+            .generation
+            .fetch_add(1, std::sync::atomic::Ordering::SeqCst);
+        let mut guard = self.available.lock().unwrap();
```
Mallets, line 79: In `zenoh-core` there is the `zlock!` macro, which allows to quickly instrument any lock in the whole Zenoh codebase if needed.

commons/zenoh-shm/tests/metadata.rs
```diff
@@ -12,43 +12,134 @@
...
+
+fn metadata_link_failure_fn() -> impl Fn(usize, usize) -> ZResult<()> + Clone + Send + Sync + 'static
+{
+    |_task_index: usize, _iteration: usize| {
+        let allocated_metadata = GLOBAL_METADATA_STORAGE.read().allocate()?;
+        let descr = MetadataDescriptor::from(allocated_metadata.deref());
+        drop(allocated_metadata);
+
+        // Some comments on this behaviour...
+        // Even though the allocated_metadata is dropped, it's SHM segment still exists in GLOBAL_METADATA_STORAGE,
```
Mallets, line 75:
```suggestion
        // Even though the allocated_metadata is dropped, its SHM segment still exists in GLOBAL_METADATA_STORAGE,
```

commons/zenoh-shm/tests/metadata.rs
```diff
@@ -12,43 +12,134 @@
...
+        // so there is no way to detect that metadata is "deallocated" and the code below succeeds. The invalidation
+        // functionality is implemented on higher level by means of generation mechanism and protects from both metadata
```
Mallets, line 77:
```suggestion
        // functionality is implemented on higher level by means of a generation mechanism that protects from both metadata
```

io/zenoh-transport/src/unicast/establishment/ext/shm.rs
```diff
@@ -63,11 +69,45 @@ impl AuthSegment {
 
     pub fn open(id: AuthSegmentID) -> ZResult<Self> {
         let array = ArrayInSHM::open(id, AUTH_SEGMENT_PREFIX)?;
+
+        // validate minimal array length
+        if array.elem_count() < ID_START_INDEX {
+            bail!("SHM auth sgemnt too small, maybe other side is using incompatible version?")
```
Mallets, line 75:
```suggestion
            bail!("SHM auth segment is too small, maybe the other side is using an incompatible SHM version?")
```

io/zenoh-transport/src/unicast/establishment/ext/shm.rs
```diff
@@ -63,11 +69,45 @@ impl AuthSegment {
...
-        unsafe { *self.array.elem(CHALLENGE_INDEX) }
+        // challenge field is inverted to prevent SHM probing between new versioned
+        // SHM implementation and the old one
+        unsafe { !(*self.array.elem(CHALLENGE_INDEX)) }
+    }
+
+    pub fn validate_challenge(&self, expected_challenge: AuthChallenge, s: &str) -> bool {
+        let challnge_in_shm = self.challenge();
+        if challnge_in_shm != expected_challenge {
+            tracing::trace!(
```
Mallets, line 90: Should it be promoted to `debug!` level?

io/zenoh-transport/src/unicast/establishment/ext/shm.rs
```diff
@@ -63,11 +69,45 @@ impl AuthSegment {
...
+                s,
+                expected_challenge,
+                challnge_in_shm
+            );
+            return false;
+        }
+
+        let version_in_shm = unsafe { *self.array.elem(VERSION_INDEX) };
+        if version_in_shm != SHM_VERSION {
+            tracing::trace!(
```
Mallets, line 101: Should it be promoted to `debug!` level?

commons/zenoh-shm/src/lib.rs
```diff
@@ -122,28 +110,39 @@ impl Eq for ShmBufInner {}
...
+
     pub fn len(&self) -> NonZeroUsize {
         self.info.data_len
     }
 
     fn is_valid(&self) -> bool {
-        self.header.header().generation.load(Ordering::SeqCst) == self.info.generation
+        let header = self.metadata.owned.header();
+
+        header.generation.load(Ordering::SeqCst) == self.info.generation
```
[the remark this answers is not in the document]
yellowhatter: That makes sense. I will fix it, thank you!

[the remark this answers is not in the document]
yellowhatter: fixed

commons/zenoh-shm/src/metadata/descriptor.rs
yellowhatter: I removed these `unsafe impls`  and made `OwnedMetadataDescriptor` to be `Send` and `Sync` fairly

commons/zenoh-shm/src/metadata/descriptor.rs
yellowhatter: Not needed anymore, removed this `unsafe`

commons/zenoh-shm/src/metadata/descriptor.rs
yellowhatter: Not needed anymore, removed this `unsafe`

commons/zenoh-shm/src/metadata/descriptor.rs
yellowhatter: Not needed anymore, removed this `unsafe`

commons/zenoh-shm/src/metadata/descriptor.rs
yellowhatter: Style fixed

For the masks and `watchdog_atomic`: watchdogs are packed as bits of u64. Mask defines the exact bit within this u64. Descriptors with the same `watchdog_atomic` but different masks are different descriptors. I'm thinking of storing mask as u8 carrying the exact bit position, but it is a matter of further micro optimization

io/zenoh-transport/src/unicast/establishment/ext/shm.rs
yellowhatter: Made debug

Mallets:
@yellowhatter it seems the latest changes make some tests fail in the CI.

yellowhatter:
> @yellowhatter it seems the latest changes make some tests fail in the CI.

yes, I'm working on it

yellowhatter:
[closes #1563]

Mallets (approved):

PR #1717, opened 2025-01-15
Update routes pre computation strategy to pre compute less routes

OlivierHecart:
Routes pre computation strategy was a bit brutal and was computing too many unnecessary routes leading to high CPU consumption during system setup.
The routes pre computation strategy was updated to only compute 
- data routes for known keys that have no wildcard and have subscribers
- query routes for known keys that have no wildcard and have queriables

In a 30 talkers/30 listeners zenoh_rmw test the stabilisation time was reduced from 35 seconds to 7 seconds.

The p2p throughput and latency tests have been run on a MacBook Air M2. No observed regressions.

### DIFF
--- a/zenoh/src/net/routing/dispatcher/pubsub.rs
+++ b/zenoh/src/net/routing/dispatcher/pubsub.rs
@@ -224,7 +224,7 @@ pub(crate) fn compute_data_routes(tables: &Tables, expr: &mut RoutingExpr) -> Da
 }
 
 pub(crate) fn update_data_routes(tables: &Tables, res: &mut Arc<Resource>) {
-    if res.context.is_some() {
+    if res.context.is_some() && !res.expr().contains('*') && res.has_subs() {
         let mut res_mut = res.clone();
         let res_mut = get_mut_unchecked(&mut res_mut);
         compute_data_routes_(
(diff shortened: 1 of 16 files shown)


fuzzypixelz (approved):

zenoh/src/net/routing/dispatcher/pubsub.rs
```diff
@@ -177,61 +180,24 @@ pub(crate) fn undeclare_subscription(
...
-    compute_data_routes_(tables, &mut routes, expr);
+    tables
+        .hat_code
+        .compute_data_routes(tables, &mut routes, expr);
     routes
 }
 
 pub(crate) fn update_data_routes(tables: &Tables, res: &mut Arc<Resource>) {
-    if res.context.is_some() {
+    if res.context.is_some() && !res.expr().contains('*') && res.has_subs() {
```
Mallets, line 192: `expr()` returns `&str`, correct?
If `expr()` would return `keyexpr` you could use https://github.com/eclipse-zenoh/zenoh/blob/f90587b580a09b4ea66bb0f796a55699278c49b9/commons/zenoh-keyexpr/src/key_expr/borrowed.rs#L133

and rely on the key-expr logic instead of using `contains()` everywhere. It's more a matter of concern encapsulation than anything else.

OlivierHecart: I fully agree. The problem is that: 
- the resource tree cannot store keyexprs for now without a big refactor. 
- doing a try_into keyexpr each time accessing the expr when not needed has a significant cost.

So until we refactor the resource tree, it's far more efficient to do it like this.

PR #1720, opened 2025-01-15
Add getrandom crate feature=js for KeyExpr

Charles-Schleich:
Used to enable compiling key expressions to WASM for zenoh-ts

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -102,6 +102,7 @@ form_urlencoded = "1.2.1"
 futures = "0.3.30"
 futures-util = { version = "0.3.30", default-features = false } # Default features are disabled due to some crates' requirements
 git-version = "0.3.9"
+getrandom = { version = "0.2" }
 hashbrown = "0.14"
 hex = { version = "0.4.3", default-features = false } # Default features are disabled due to usage in no_std crates
 hmac = { version = "0.12.1", features = ["std"] }
--- a/commons/zenoh-keyexpr/Cargo.toml
+++ b/commons/zenoh-keyexpr/Cargo.toml
@@ -55,5 +55,6 @@ rand = { workspace = true, features = ["default"] }
 name = "keyexpr_tree"
 harness = false
 
+# NOTE: for the above reason, we need to explicitly ignore getrandom in the CI because it's an indirect dependency which is not used directly by zenoh.
 [package.metadata.cargo-machete]
 ignored = ["getrandom"]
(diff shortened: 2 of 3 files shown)


commons/zenoh-keyexpr/Cargo.toml
```diff
@@ -36,6 +36,7 @@ schemars = { workspace = true, optional = true }
 serde = { workspace = true, features = ["alloc"] }
 token-cell = { workspace = true }
 zenoh-result = { workspace = true }
+getrandom = { workspace = true, features = ["js"] }
```
Mallets, line 39: Wouldn't be better having a `js` feature in `zenoh-keyexpr` that optionally enables `js` on `getrandom` instead of always enabling it?

Cargo.toml
```diff
@@ -102,6 +102,7 @@ form_urlencoded = "1.2.1"
 futures = "0.3.30"
 futures-util = { version = "0.3.30", default-features = false } # Default features are disabled due to some crates' requirements
 git-version = "0.3.9"
+getrandom = { version = "0.2" }
```
Mallets, line 105: `getrandom` is added as dependency but no change in this PR seems to use it.
What's the role of this dependency alone?

commons/zenoh-keyexpr/Cargo.toml
Charles-Schleich: I agree with adding a feature, 
It was mentioned here https://docs.rs/getrandom/latest/getrandom/#webassembly-support
That libraries should not introduce their own `js` features just to enable getrandom’s `js` feature.

regardless i think its cleaner if we do add it.

Cargo.toml
Charles-Schleich: `rand` uses `getrandom` under the hood, and get random has the library feature that we need to enable to get the keyexpr package to compile to wasm successfully. 
Ive added an ignore for cargo machete which should fix the CI check

commons/zenoh-keyexpr/Cargo.toml
```diff
@@ -50,3 +54,6 @@ rand = { workspace = true, features = ["default"] }
 [[bench]]
 name = "keyexpr_tree"
 harness = false
+
+[package.metadata.cargo-machete]
```
Mallets, line 58:
```suggestion
# For the above reason, we need to explicitly ignore getrandom in the CI because it's an indirect dependency which is not used directly by zenoh.
[package.metadata.cargo-machete]
```

Mallets (approved):

PR #1722, opened 2025-01-15
Update DEFAULT_CONFIG.json5

Mallets:
Clarify DEFAULT_CONFIG documentation in transport section

### DIFF
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -354,11 +354,11 @@
       open_timeout: 10000,
       /// Timeout in milliseconds when accepting a link
       accept_timeout: 10000,
-      /// Maximum number of zenoh session in pending state while accepting
+      /// Maximum number of links in pending state while performing the handshake for accepting it
       accept_pending: 100,
-      /// Maximum number of sessions that can be simultaneously alive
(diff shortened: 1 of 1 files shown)


DEFAULT_CONFIG.json5
```diff
@@ -354,11 +354,11 @@
       open_timeout: 10000,
       /// Timeout in milliseconds when accepting a link
       accept_timeout: 10000,
-      /// Maximum number of zenoh session in pending state while accepting
+      /// Maximum number of links in pending state while performing the handshake for accepting it
       accept_pending: 100,
-      /// Maximum number of sessions that can be simultaneously alive
+      /// Maximum number of transports that can be simultaneously alive for a single zenoh sessions
```
evshary, line 359: I thought it was the maximum number of sessions for a single transport based on the config name `max_sessions`. Is it possible for us to change the name? For example, `max_per_session` or something else.

evshary (approved):
LGTM, but the config name is confusing.
I think we can deal with this in other PR.

DEFAULT_CONFIG.json5
Mallets: That would introduce a breaking change in the config I would rather avoid to introduce.

PR #1726, opened 2025-01-16
Expose `PluginManager` via the `Session`.

Jezza:
Statically defining plugins provides a lot of convenience (Don't have to worry about dynamic libs) and performance.


Jezza:
[An alternative was found.](https://github.com/eclipse-zenoh/zenoh-plugin-ros2dds/blob/82ec5f4d7aa7ddd07efbe007c33e9640b4d5b74e/zenoh-bridge-ros2dds/src/main.rs#L87)

PR #1731, opened 2025-01-20
Extend priority range syntax

fuzzypixelz:
This makes it possible to write , for example, `tcp/localhost:0?prio=1` instead of `tcp/localhost:0?prio=1-1`.


evshary (approved):
LGTM

Mallets (approved):

PR #1732, opened 2025-01-20
Test ports again

yellowhatter:
fix colliding port
fix ephmeral ports


zenoh-ext/tests/advanced.rs
oteffahi, line 1: For sake of consistency, let's also update the port for the remaining test to `27056`
https://github.com/eclipse-zenoh/zenoh/blob/ade92b9cd8533a4c7608ce1a3fb56023146c18a4/zenoh-ext/tests/advanced.rs#L727

PR #1736, opened 2025-01-22
refactor: avoid temporary allocation in `Resource::get_best_key`

wyfo:
The code has also be refactored to better express the intent.

Supersede #1730


wyfo:
@OlivierHecart  Now I've a transformed `Resource::get_best_key` into a method, do you want me to replace all `Resource::get_best_key(...)` in the code by a method call?

PR #1737, opened 2025-01-23
fix: revert #1627 and add congestion shortcut

wyfo:
#1627 introduced a congestion shortcut mechanism, but this one was not correctly synchronized. There was indeed situations experienced by users in which congested flag was set and never reset, which implies a drop of all successive messages (the publisher becomes kind of dead).

The congestion flag is in fact set after the deadline of a message is reached, while it is unset when batches were refilled, only with relaxed atomic operations. Also, after the deadline is reached, there is no further check of the queue.

The most obvious synchronization flow here is that between the instant where the thread is waken up because the deadline has been reached, and the one where the congested flag is set, it is possible that the tx task is unblocked and all the batches are sent and refilled. The congested flags would been set after, so there would be no batch to refill to unset it back. This flow seems hard to achieve when there are many batches in the queue, but it is still theoretically possible. And when fragmentation chain is dropped in the middle, pushing the ephemeral batch takes more time before setting the congested flag; this is precisely the case where the bug was observed by the way, so it may indicate the described flow is the reason behind, but it's not sure.

The proposed fix reverts #1627, and use a simpler and correctly synchronized shortcut: a congested flag is added behind the mutex of `StageIn`, it is set when the pipeline is congested and unset if the message is written (for both network and transport); if the congested flag is set, the deadline is not awaited. This shortcut is not "as short" as the previous, but it is again a lot simpler.

### DIFF
--- a/io/zenoh-transport/src/common/pipeline.rs
+++ b/io/zenoh-transport/src/common/pipeline.rs
@@ -274,6 +274,7 @@ impl StageIn {
                                     LOCAL_EPOCH.elapsed().as_micros() as MicroSeconds,
                                     Ordering::Relaxed,
                                 );
+                                self.congested = false;
                                 break batch;
                             }
                             None => {
@@ -303,14 +304,14 @@ impl StageIn {
                 if !self.batching || $msg.is_express() {
                     // Move out existing batch
                     self.s_out.move_batch($batch);
+                    return Ok(true);
                 } else {
                     let bytes = $batch.len();
                     *c_guard = Some($batch);
                     drop(c_guard);
                     self.s_out.notify(bytes);
+                    return Ok(true);
                 }
-                self.congested = false;
-                return Ok(true);
             }};
         }
 
@@ -419,7 +420,6 @@ impl StageIn {
 
         // Clean the fragbuf
         self.fragbuf.clear();
-        self.congested = false;
         Ok(true)
     }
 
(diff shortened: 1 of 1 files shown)


io/zenoh-transport/src/common/pipeline.rs
```diff
@@ -795,22 +764,14 @@ impl TransmissionPipelineProducer {
 
         // If message is droppable, compute a deadline after which the sample could be dropped
         let (wait_time, max_wait_time) = if msg.is_droppable() {
-            // Checked if we are blocked on the priority queue and we drop directly the message
-            if self.status.is_congested(priority) {
```
Mallets, line 799: By removing this check, the code will always block in the `zlock!(self.stage_in[idx])` at line `773` when congestion happens. I think the general idea is to avoid getting the lock if we are already in the congested scenario.

wyfo: It will not block for long, because if the pipeline is congested, no batch will be available and `push_network_message` will return directly.
Also, having the shortcut before the lock implies that each other messages waiting for the lock will have to wait the deadline while we know that the pipeline is congested.
And last, this is easier to synchronize properly, and touch the least possible amount of code.

Mallets: You will block for long because if the previous message is reliable, it will wait for a new batch. So it won't be possible to do an early drop in case of best effort message.

wyfo: That's a valid argument.

oteffahi:
Superseded by https://github.com/eclipse-zenoh/zenoh/pull/1737

PR #1739, opened 2025-01-23
refactor: use HashSet instead of Vec for network links

wyfo:
Links are used a lot for `contains` operation.

### DIFF
--- a/zenoh/src/net/routing/hat/linkstate_peer/network.rs
+++ b/zenoh/src/net/routing/hat/linkstate_peer/network.rs
@@ -1,3 +1,4 @@
+use std::collections::HashSet;
 //
 // Copyright (c) 2023 ZettaScale Technology
 //
@@ -52,7 +53,7 @@ pub(super) struct Node {
     pub(super) whatami: Option<WhatAmI>,
     pub(super) locators: Option<Vec<Locator>>,
     pub(super) sn: u64,
-    pub(super) links: Vec<ZenohIdProto>,
+    pub(super) links: HashSet<ZenohIdProto>,
 }
 
 impl std::fmt::Debug for Node {
@@ -149,7 +150,7 @@ impl Network {
             whatami: Some(runtime.whatami()),
             locators: None,
             sn: 1,
-            links: vec![],
+            links: HashSet::new(),
         });
         Network {
             name,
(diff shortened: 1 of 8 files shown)


zenoh/src/net/routing/hat/linkstate_peer/network.rs
```diff
@@ -1,3 +1,4 @@
+use std::collections::HashSet;
```
Mallets, line 1: Move the `use` after the header.

zenoh/src/net/routing/hat/p2p_peer/gossip.rs
```diff
@@ -1,3 +1,4 @@
+use std::collections::HashSet;
```
Mallets, line 1: Move the `use` after the header.

zenoh/src/net/routing/hat/linkstate_peer/network.rs
```diff
@@ -752,7 +753,7 @@ impl Network {
                 tracing::trace!("Update edge (link) {} {}", self.graph[self.idx].zid, zid);
                 self.update_edge(self.idx, idx);
             }
-            self.graph[self.idx].links.push(zid);
+            self.graph[self.idx].links.insert(zid);
```
Mallets, line 756: Here there is a semantic change where before the same `ZenohId` could be present multiple times in the `Vec<ZenohIdProto>` while now it can be present only once.
Maybe is it worth adding a debug log if `insert(zid)` returns `Some`? It would allow to detect if something else is going wrong. What do you think?

zenoh/src/net/routing/hat/router/network.rs
```diff
@@ -1,3 +1,4 @@
+use std::collections::HashSet;
```
Mallets, line 1: Move `use` after the header.

zenoh/src/net/routing/hat/router/network.rs
```diff
@@ -756,7 +757,7 @@ impl Network {
                 tracing::trace!("Update edge (link) {} {}", self.graph[self.idx].zid, zid);
                 self.update_edge(self.idx, idx);
             }
-            self.graph[self.idx].links.push(zid);
+            self.graph[self.idx].links.insert(zid);
```
Mallets, line 760: Same comment as above on adding a debug log message is `insert(zid)` returns `Some`?

Mallets (changes requested):

OlivierHecart:
No observed performance improvement.

Issue #1738, opened 2025-01-23
Publisher can get stuck dropping all messages with CongestionControl::Drop


oteffahi:
### Describe the bug

This bug was reported by a user [on the official Zenoh Discord server](https://discord.com/channels/914168414178779197/1319564474168119398). Their use-case has 4 publishers put 5Mb messages at 30Hz each, and one session declares a subscriber for each topic.
In this case, one or multiple publishers eventually get stuck in a dropping loop, where all PUT messages are dropped by congestion control and the subscriber does not receive any sample on the respective topics.

### To reproduce

1. Create 4 sessions that declare one publisher each, 30Hz, 5Mb per message.
2. Create one session that declares 4 subscribers, one for each topic, and recv on each subscriber with a 5 second timeout.
3. Depending on the machine's specs, you may need to artifically increase its CPU load.
4. Eventually one or multiple publishers will drop all of their messages, and their respective subscribers will timeout every 5 seconds.

### System info

- Ubuntu 24.04
- Tested on both x86 and ARM CPUs.
- Zenoh 1.1.1

wyfo:
Fixed by #1740

PR #1740, opened 2025-01-24
fix: fix pipeline congestion shortcut synchronization

wyfo:
#1627 introduced a congestion shortcut mechanism, but this one was not correctly synchronized. There was indeed situations experienced by users in which congested flag was set and never reset, which implies a drop of all successive messages (the publisher becomes kind of dead).

The congestion flag is in fact set after the deadline of a message is reached, while it is unset when batches were refilled, only with relaxed atomic operations. Also, after the deadline is reached, there is no further check of the queue.

The most obvious synchronization flow here is that between the instant where the thread is waken up because the deadline has been reached, and the one where the congested flag is set, it is possible that the tx task is unblocked and all the batches are sent and refilled. The congested flags would been set after, so there would be no batch to refill to unset it back. This flow seems hard to achieve when there are many batches in the queue, but it is still theoretically possible. And when fragmentation chain is dropped in the middle, pushing the ephemeral batch takes more time before setting the congested flag; this is precisely the case where the bug was observed by the way, so it may indicate the described flow is the reason behind, but it's not sure.

The proposed fix adds a additional synchronization step after setting the congested flag: we retry to push the message, this time with an already expired deadline. If the batches were refilled, the message should be able to be sent and the congestion flag will be reset.

If batches were in fact refilled, but the message was fragmented and still ends by being dropped, it still means batches have been pushed and will be refilled later, still unsetting the congested flag.

### DIFF
--- a/io/zenoh-transport/src/common/pipeline.rs
+++ b/io/zenoh-transport/src/common/pipeline.rs
@@ -254,7 +254,7 @@ struct StageIn {
 impl StageIn {
     fn push_network_message(
         &mut self,
-        msg: &mut NetworkMessage,
+        msg: &NetworkMessage,
         priority: Priority,
         deadline: &mut Deadline,
     ) -> Result<bool, TransportClosed> {
@@ -315,7 +315,7 @@ impl StageIn {
         // Get the current serialization batch.
         let mut batch = zgetbatch_rets!();
         // Attempt the serialization on the current batch
-        let e = match batch.encode(&*msg) {
+        let e = match batch.encode(msg) {
             Ok(_) => zretok!(batch, msg),
             Err(e) => e,
         };
@@ -335,7 +335,7 @@ impl StageIn {
 
         if let BatchError::NewFrame = e {
             // Attempt a serialization with a new frame
-            if batch.encode((&*msg, &frame)).is_ok() {
+            if batch.encode((msg, &frame)).is_ok() {
                 zretok!(batch, msg);
             }
         }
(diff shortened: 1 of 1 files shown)


io/zenoh-transport/src/common/pipeline.rs
```diff
@@ -806,9 +806,31 @@ impl TransmissionPipelineProducer {
         let mut deadline = Deadline::new(wait_time, max_wait_time);
         // Lock the channel. We are the only one that will be writing on it.
         let mut queue = zlock!(self.stage_in[idx]);
-        let sent = queue.push_network_message(&mut msg, priority, &mut deadline)?;
+        // Checked again for congestion in case in happens when blocking on the mutex.
```
oteffahi, line 809:
```suggestion
        // Check again for congestion in case it happens when blocking on the mutex
```

oteffahi (approved):
The bug reported in https://github.com/eclipse-zenoh/zenoh/issues/1738 is solved by the edge-case explained in the comments of this PR.
LGTM.

io/zenoh-transport/src/common/pipeline.rs
```diff
@@ -806,9 +806,31 @@ impl TransmissionPipelineProducer {
         let mut deadline = Deadline::new(wait_time, max_wait_time);
         // Lock the channel. We are the only one that will be writing on it.
         let mut queue = zlock!(self.stage_in[idx]);
-        let sent = queue.push_network_message(&mut msg, priority, &mut deadline)?;
+        // Check again for congestion in case it happens when blocking on the mutex.
+        if self.status.is_congested(priority) {
```
wyfo, line 810: @Mallets are you ok with this addition?

Mallets: Yes

PR #1742, opened 2025-01-27
feat: avoid duplicate autoconnection attempt

wyfo:
Gossip autoconnection could result in two nodes attempting to connect to each other concurrently.
To fix this issue, only one node initiate the connection now. The selection criteria is arbitrary but deterministic, currently the greater zid of both.


zenoh/src/net/routing/hat/linkstate_peer/network.rs
```diff
@@ -510,7 +510,10 @@ impl Network {
                             );
                         }
 
-                        if !self.autoconnect.is_empty() && self.autoconnect.matches(whatami) {
+                        if !self.autoconnect.is_empty()
+                            && self.autoconnect.matches(whatami)
+                            && select_autoconnect_node(self.graph[self.idx].zid, zid)
+                        {
```
oteffahi, line 516: This is a more complex problem that what it seems. Consider the following:
- Node1 is a peer, it is setup to autoconnect to routers and peers.
- Node2 is a router, it is setup to autoconnect only to routers.
- Node2 has a higher ZID, and therefore is the one returned by `select_autoconnect_node`.
In thise case, Node1 will not attempt to connect because it is not the selected node, and Node2 will not autoconnect because it does not have peers in its autoconnect config, and the connection will never be established...

oteffahi (changes requested):

zenoh/src/net/routing/hat/linkstate_peer/network.rs
oteffahi: Unless we add autoconnect config to the exchanged gossip information, the other solutions may involve closing one of the two links during or after a transport is opened.

commons/zenoh-config/src/lib.rs
```diff
@@ -184,6 +185,17 @@ pub enum Permission {
...
+/// Strategy for autoconnection, mainly to avoid nodes connecting to each other redundantly.
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, Eq, Hash, PartialEq)]
+#[serde(rename_all = "kebab-case")]
+pub enum AutoConnectStrategy {
+    /// A node will attempt to connect to another one only if its own zid is greater than the
+    /// other one. If both nodes use this strategy, only one will attempt the connection.
+    /// This strategy may not be suited if one of the node is not reachable by the other one,
+    /// for example because of a private IP.
+    GreaterZid,
+}
```
oteffahi, line 197: I would add a variant that signifies no autoconnect, this way we can have it as default value instead of simply a `None` of the wrapping `Option`.
@OlivierHecart what do you think?

DEFAULT_CONFIG.json5
```diff
@@ -161,6 +169,8 @@
       /// or different values for router or peer mode (e.g. autoconnect: { router: [], peer: ["router", "peer"] }).
       /// Each value is a list of: "peer" and/or "router".
       autoconnect: { router: [], peer: ["router", "peer"] },
+      /// Strategy for autoconnection, mainly to avoid nodes connecting to each other redundantly.
+      autoconnect_strategy: "always",
```
oteffahi, line 173:
```suggestion
      /// Strategy for autoconnection, mainly to avoid nodes connecting to each other redundantly.
      /// Accepts either:
      /// - "always": always attempt to autoconnect, may result in redundant connection which will then be closed.
      /// - "greater-zid": attempt to connect to another node only if its own zid is greater than the other's.
      ///   If both nodes use this strategy, only one will attempt the connection.
      ///   This strategy may not be suited if one of the nodes is not reachable by the other one, for example
      ///   because of a private IP.
      autoconnect_strategy: "always",
```

DEFAULT_CONFIG.json5
```diff
@@ -138,6 +138,14 @@
...
       /// Each value is a list of: "peer", "router" and/or "client".
       autoconnect: { router: [], peer: ["router", "peer"], client: ["router"] },
+      /// Strategy for autoconnection, mainly to avoid nodes connecting to each other redundantly.
+      /// Accepts either:
+      /// - "always": always attempt to autoconnect, may result in redundant connection which will be then be closed.
+      /// - "greater-zid": attempt to connect to another node only if its own zid is greater than the other one.
+      ///   If both nodes use this strategy, only one will attempt the connection.
+      ///   This strategy may not be suited if one of the node is not reachable by the other one, for example
+      ///   because of a private IP.
+      autoconnect_strategy: "always",
```
oteffahi, line 148:
```suggestion
      /// Strategy for autoconnection, mainly to avoid nodes connecting to each other redundantly.
      /// Accepts either:
      /// - "always": always attempt to autoconnect, may result in redundant connection which will then be closed.
      /// - "greater-zid": attempt to connect to another node only if its own zid is greater than the other's.
      ///   If both nodes use this strategy, only one will attempt the connection.
      ///   This strategy may not be suited if one of the nodes is not reachable by the other one, for example
      ///   because of a private IP.
      autoconnect_strategy: "always",
```

oteffahi (changes requested):
Typos and missing comments

commons/zenoh-config/src/lib.rs
wyfo: Done

DEFAULT_CONFIG.json5
```diff
@@ -138,6 +138,17 @@
...
+      /// Possible options are:
+      /// - "always": always attempt to autoconnect, may result in redundant connection which will then be closed.
+      /// - "greater-zid": attempt to connect to another node only if its own zid is greater than the other's.
+      ///   If both nodes use this strategy, only one will attempt the connection.
+      ///   This strategy may not be suited if one of the nodes is not reachable by the other one, for example
+      ///   because of a private IP.
+      /// Accepts a single value (e.g. autoconnect: "always") which applies whatever node would be auto-connected to,
+      /// or different values for router, peer, or client depending on the type of node detected
+      /// (e.g. autoconnect_strategy : { router:  "always", peer: "greater-zid", client: "always" })
+      autoconnect_strategy: { router:  "always", peer: "always", client: "always" },
```
oteffahi, line 151: Client cannot be an autoconnect target. Please update the comments accordingly, and add a check on the config to disallow clients here.
```suggestion
      /// Accepts a single value (e.g. autoconnect: "always") which applies whatever node would be auto-connected to,
      /// or different values for router and/or peer depending on the type of node detected
      /// (e.g. autoconnect_strategy : { router:  "always", peer: "greater-zid" })
      autoconnect_strategy: { router:  "always", peer: "always" },
```

DEFAULT_CONFIG.json5
```diff
@@ -161,6 +172,17 @@
...
```
oteffahi, line 185: See other comment regarding clients and autoconnect.
```suggestion
      /// Accepts a single value (e.g. autoconnect: "always") which applies whatever node would be auto-connected to,
      /// or different values for router and/or peer depending on the type of node detected
      /// (e.g. autoconnect_strategy : { router:  "always", peer: "greater-zid" })
      autoconnect_strategy: { router:  "always", peer: "always" },
```

DEFAULT_CONFIG.json5
```diff
@@ -138,6 +138,19 @@
...
+      /// - "greater-zid": attempt to connect to another node only if its own zid is greater than the other's.
+      ///   If both nodes use this strategy, only one will attempt the connection.
+      ///   This strategy may not be suited if one of the nodes is not reachable by the other one, for example
+      ///   because of a private IP.
+      /// Accepts a single value (e.g. autoconnect: "always") which applies whatever node would be auto-connected to,
+      /// or different values for router and/or peer depending on the type of node detected
+      /// (e.g. autoconnect_strategy : { "to-router":  "always", "to-peer": "greater-zid" }),
+      /// or different values for router or peer mode
+      /// (e.g. autoconnect_strategy : { peer: { "to-router":  "always", "to-peer": "greater-zid" } }).
+      autoconnect_strategy: { peer: { "to-router": "always", "to-peer": "always" } },
```
oteffahi, line 153: Need to add the default config for router

DEFAULT_CONFIG.json5
```diff
@@ -161,6 +174,19 @@
...
```
oteffahi, line 189: Need to add the default config for router

commons/zenoh-protocol/src/network/push.rs
```diff
@@ -47,6 +47,23 @@ pub struct Push {
...
+}
+
+impl WithIdxImpl for WireExpr<'static> {
+    type WithIdx = WithIdx<Self>;
+}
+
+pub struct PushWithIdx {
+    pub wire_expr: WithIdx<WireExpr<'static>>,
+}
+
```
oteffahi, line 66: Maybe I'm missing something, I don't see where this is used for this feature 🤔

oteffahi (changes requested):

commons/zenoh-protocol/src/network/push.rs
wyfo: Oups

DEFAULT_CONFIG.json5
wyfo: There is no default autoconnection for router, I could put the one for peer though

oteffahi: Whatever the current behavior is should be the default, I assume it's "always" for all cases

commons/zenoh-config/src/mode_dependent.rs
```diff
@@ -349,3 +352,177 @@ impl<T> ModeDependent<T> for Option<ModeDependentValue<T>> {
...
+                )))
+            }
+
+            fn visit_map<M>(self, map: M) -> Result<Self::Value, M::Error>
+            where
+                M: MapAccess<'de>,
+            {
+                let value =
+                    serde_json::Value::deserialize(de::value::MapAccessDeserializer::new(map))?;
+                dbg!(&value);
```
Mallets, line 515: Remove the `dbg!` directives that print unnecessary data.

Mallets (approved):

PR #1743, opened 2025-01-28
fix: handle linkstate decoding error

wyfo:
Panicking here was causing a major security breach concerning every non-encrypted transport, especially UDP ones. An attacker could simply forge a invalid OAM linkstate message to make the transport panic. Regarding UDP multicast network, a single message could make the whole network instantly unresponsive.


Mallets:
The problem is also present in case of encrypted transports. Besides that LGTM.

Mallets (approved):

PR #1751, opened 2025-02-03
Change default memory allocation policy and size of the queue

Mallets:
The default configuration now:

1.  allocates fewer batches 
2. adopts lazy memory allocations

overall reducing memory consumption out-of-the-box.

### DIFF
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -440,14 +440,14 @@
           /// then amount of memory being allocated for each queue is SIZE_XXX * LINK_MTU.
           /// If qos is false, then only the DATA priority will be allocated.
           size: {
-            control: 1,
-            real_time: 1,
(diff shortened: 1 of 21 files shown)


DEFAULT_CONFIG.json5
```diff
@@ -440,14 +440,14 @@
...
-            interactive_low: 1,
+            control: 2,
+            real_time: 2,
+            interactive_high: 2,
+            interactive_low: 2,
             data_high: 2,
-            data: 4,
-            data_low: 4,
-            background: 4,
+            data: 2,
```
wyfo, line 448: I think this would impact maximum throughput by a lot.

Mallets: In my tests I have observed no difference whatsoever... do you observe the same?

wyfo (approved):

PR #1753, opened 2025-02-03
AdvancedSub starts delivering samples as soon as sample_depth samples are received

OlivierHecart:
With this change, in case of SessionLocal pubsub when history_depth=1, the publisher threads synchronously calls the subscriber callback. This removes the race condition in rmw_zenoh test_client test.

### DIFF
--- a/zenoh-ext/src/advanced_subscriber.rs
+++ b/zenoh-ext/src/advanced_subscriber.rs
@@ -561,19 +561,15 @@ fn handle_sample(states: &mut State, sample: Sample) -> bool {
             pending_samples: BTreeMap::new(),
         });
         if state.last_delivered.map(|t| t < *timestamp).unwrap_or(true) {
-            if states.global_pending_queries == 0 && state.pending_queries == 0 {
+            if (states.global_pending_queries == 0 && state.pending_queries == 0)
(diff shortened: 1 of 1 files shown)


zenoh-ext/src/advanced_subscriber.rs
```diff
@@ -502,8 +503,25 @@ fn handle_sample(states: &mut State, sample: Sample) -> bool {
...
-            state.pending_samples.insert(source_sn, sample);
+        if state.last_delivered.is_none() && states.global_pending_queries != 0 {
+            // Avoid going through the Map if history_depth == 1
+            if states.history_depth == 1 {
+                state.last_delivered = Some(source_sn);
+                states.callback.call(sample);
+            } else {
+                state.pending_samples.insert(source_sn, sample);
+                if state.pending_samples.len() >= states.history_depth {
+                    if let Some((mut last_seq_num, sample)) = state.pending_samples.pop_first() {
```
DenisBiryukov91, line 514: the lines 515 - 521 are the same as in lines 546 - 553, could they be (at least partially moved) to a separate function ?

DenisBiryukov91 (approved):

PR #1764, opened 2025-02-07
Migrate zenohd to sync API

Mallets:
This PR migrates zenohd API usage from *async* to *sync*. By doing so, the unnecessary *main* tokio runtime is not spawn. As a result fewer threads are spawn and fewer resources are used by `zenohd`.

On my MacBook Pro M3, an idle `zenohd` now uses `7 MB` of RAM instead of the previous `28 MB`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5836,13 +5836,11 @@ name = "zenohd"
 version = "1.2.1"
 dependencies = [
  "clap",
- "futures",
  "git-version",
  "json5",
  "lazy_static",
  "rand 0.8.5",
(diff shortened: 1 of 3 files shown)


kydos:
Did you measure a release or debug build? I have completely different numbers on my MacBook  (see two routers running below).  FYI, this is for a **zenohd** compiled with default configuration.
image

Mallets:
I've measured it in release and using `instruments` which consider the full process memory, including vm stack.

OlivierHecart (approved):

PR #1768, opened 2025-02-12
feat: allocate rx buffer

wyfo:


DEFAULT_CONFIG.json5
```diff
@@ -514,12 +514,6 @@
       },
       /// Configure the zenoh RX parameters of a link
       rx: {
-        /// Receiving buffer size in bytes for each link
-        /// The default the rx_buffer_size value is the same as the default batch size: 65535.
-        /// For very high throughput scenarios, the rx_buffer_size can be increased to accommodate
-        /// more in-flight data. This is particularly relevant when dealing with large messages.
-        /// E.g. for 16MiB rx_buffer_size set the value to: 16777216.
-        buffer_size: 65535,
```
evshary, line 522: Should we keep the config and mark it not available instead of removing it?
I'm not sure how much we want to maintain compatibility.

wyfo: Don't know honestly. But this config is misleading, it's not a question of buffer size but of a number of batches, and the only case where more than one batch will be used is when a payload is cloned in a subscriber (e.g. a channel subscriber), so it's not the case for router for example.
So user may be tempted to increase the "buffer size" for a router with big traffic, but it will just allocate more unused memory.

PR #1769, opened 2025-02-12
Rework the pipeline: batch allocation + deterministic backoff

wyfo:
See more details in https://github.com/eclipse-zenoh/zenoh/pull/1769#issuecomment-2682179710

### DIFF
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -471,14 +471,14 @@
           /// then amount of memory being allocated for each queue is SIZE_XXX * LINK_MTU.
           /// If qos is false, then only the DATA priority will be allocated.
           size: {
-            control: 2,
-            real_time: 2,
-            interactive_high: 2,
-            interactive_low: 2,
+            control: 1,
+            real_time: 1,
+            interactive_high: 1,
+            interactive_low: 1,
             data_high: 2,
-            data: 2,
-            data_low: 2,
-            background: 2,
+            data: 4,
+            data_low: 4,
+            background: 4,
           },
           /// Congestion occurs when the queue is empty (no available batch).
           congestion_control: {
@@ -510,7 +510,7 @@
             /// Mode for memory allocation of batches in the priority queues. 
             /// - "init": batches are allocated at queue initialization time. 
             /// - "lazy": batches are allocated when needed up to the maximum number of batches configured in the size configuration parameter.
-            mode: "lazy",
+            mode: "init",
           },
         },
       },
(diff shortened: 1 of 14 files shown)


io/zenoh-transport/src/common/pipeline.rs
```diff
@@ -68,269 +62,376 @@ impl fmt::Display for TransportClosed {
...
+    batch_config: BatchConfig,
 }
 
-impl LazyDeadline {
-    fn new(wait_time: WaitTime) -> Self {
-        Self {
-            deadline: None,
-            wait_time,
+impl StageIn {
+    const SMALL_BATCH_SIZE: BatchSize = 1 << 11;
```
wyfo, line 307: This constant is of course arbitrary. It can be exposed in configuration if needed.

DEFAULT_CONFIG.json5
```diff
@@ -471,14 +471,14 @@
...
-            interactive_low: 2,
+            control: 1,
+            real_time: 1,
+            interactive_high: 1,
+            interactive_low: 1,
             data_high: 2,
-            data: 2,
-            data_low: 2,
-            background: 2,
+            data: 4,
```
wyfo, line 479: I've reverted the previous commit, so the numbers are back to the original ones. Throughputs of big messages was indeed very impacted by a smaller batch number, 130kmsg/s -> 50kmsg/s for 70kB messages.
@Mallets Do you think we should modify the numbers? for example increasing the hight priority to 2? In fact this PR does lazy allocation, and only keep one buffer in memory, so we could put any numbers, it would not change memory consumption at all, it just a matter of throughput/congestion.

DEFAULT_CONFIG.json5
```diff
@@ -471,14 +471,14 @@
...
+            data_low: 4,
```
OlivierHecart, line 480: The numbers are not coherent with `commons/zenoh-config/src/defaults.rs`.

wyfo: You're right, good catch !

OlivierHecart:
This PR implies too high latency for sporadic messages bursts (larger than a batch).

wyfo:
> This PR implies too high latency for sporadic messages bursts (larger than a batch).

What do you mean? That's exactly what we have right now. Try to send 1000msg in a row, then send one message and mesure its latency, you will get more than 1ms; even with `--cfg='transport/link/tx/queue/batching/time_limit:0'`, you will get more than 1ms most of the time.

With this PR, latency will also be more than 1ms, but `--cfg='transport/link/tx/queue/batching/time_limit:0'` will give less than 1ms 100% of the times.
At least, with this PR, I give the choice to the user to choose between latency and throughput, while the current code give you random jitter and bad latency.

PR #1780, opened 2025-02-18
move nested part of advanced pub-sub key expressions to the beginning to allow advanced pub-sub work with namespace prefixes

DenisBiryukov91:
move nested part of advanced pub-sub key expressions to the beginning


OlivierHecart:
Merged as part of: https://github.com/eclipse-zenoh/zenoh/pull/1792

PR #1781, opened 2025-02-18
Add policy id to ACL config

oteffahi:
Adds ACL policy id for consistency with ACL rules/subjects, and with downsampling interceptor.


gabrik:
@RemiBarthe  :eyes:

gabrik (approved):
LGTM

PR #1783, opened 2025-02-19
internal Query::empty doesn't require self

milyin:
The purpose of this "empty" method is to provide "take" operation for zenoh-c by replacing valid query with dummy one.
This variant of "empty" is safer: it reduces risks that destruction of `QueryInner` may cause some side-effects.

This update also adds internal `Keyexpr::dummy()` method for creating dummy key expression when `Option<KeyExpr>` is not convenient.

### DIFF
--- a/zenoh/src/api/query.rs
+++ b/zenoh/src/api/query.rs
@@ -158,7 +158,7 @@ impl Reply {
 
     /// Constructs an uninitialized empty Reply.
     #[zenoh_macros::internal]
-    pub unsafe fn empty() -> Self {
+    pub fn empty() -> Self {
         Reply {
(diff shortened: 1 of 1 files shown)


zenoh/src/api/sample.rs
```diff
@@ -434,7 +434,7 @@ impl Sample {
     #[zenoh_macros::internal]
     pub unsafe fn empty() -> Self {
```
DenisBiryukov91, line 435: with dummy keyexpr this function is no longer unsafe. I think same goes for Reply::empty().

DenisBiryukov91 (approved):

PR #1790, opened 2025-02-24
fix: change macos runners to use macos-latest

diogomatsubara:
this makes it easier to target self-hosted runners

This change was introduced in https://github.com/eclipse-zenoh/zenoh/pull/935 and since we don't use macos-12 anymore, it should be fine to revert back to `macos-latest`


YuanYuYuan:
LGTM

PR #1797, opened 2025-02-26
Add support for query/reply messages to Downsampling

oteffahi:
Adds query/reply message filters to Downsampling interceptor. Query/reply message types are respectively named `"query"` and `"reply"`, with the addition of `"push"` which is the previously supported Publication message type.
Downsampling config with different message types will apply its rules on all these message types as one flow of messages. Seperate Downsampling configurations with one type each will apply the rules per message type.

Also changes Downsampling config to accept a list of flows instead of a single one, to limit the need to duplicate a downsampling conf for query/reply messages which go in opposite flows.

Note that this PR breaks compatibility with previous format of Downsampling configuration.


oteffahi:
Clarification regarding my commit message for [963eb84](https://github.com/eclipse-zenoh/zenoh/pull/1797/commits/963eb845314515f36bf2a7dd012682c0d6651afa): two rules on the same keyexpr would have the last rule in the list overwrite the first, which further supports why message types need to be moved out of the rules list.

gabrik:
@RemiBarthe :eyes:

DEFAULT_CONFIG.json5
```diff
@@ -271,8 +271,18 @@
   //      /// Optional list of network interfaces messages will be processed on, the rest will be passed as is.
   //      /// If absent, the rules will be applied to all interfaces, in case of an empty list it means that they will not be applied to any.
   //      interfaces: [ "wlan0" ],
-  //      /// Data flow messages will be processed on. ("egress" or "ingress")
-  //      flow: "egress",
+  //      /// Optional list of data flows messages will be processed on ("egress" and/or "ingress").
+  //      /// If absent, the rules will be applied to both flows.
+  //      flow: ["ingress", "egress"],
```
OlivierHecart, line 276: Wouldn't it be simpler and less error prone to have something like:
```
flow: {
  ingress: true,
  egress: true,
}
```
?

oteffahi: We can make this change on all affected interceptors in a another PR. However I don't think it's worth the breaking change for the other interceptors (namely ACL).

PR #1800, opened 2025-02-27
fix: add cargo-machete ignore rule for base64ct

diogomatsubara:
This should fix the CI failures.

### DIFF
--- a/ci/valgrind-check/Cargo.toml
+++ b/ci/valgrind-check/Cargo.toml
@@ -23,6 +23,7 @@ description = "Internal crate for zenoh."
 
 [dependencies]
 base64ct = "~1.6.0"
+zerofrom = "~0.1.5"
 tokio = { version = "1.35.1", features = ["rt-multi-thread", "time", "io-std"] }
 zenoh = { path = "../../zenoh/" }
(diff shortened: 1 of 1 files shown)


ci/valgrind-check/Cargo.toml
```diff
@@ -23,6 +23,7 @@ description = "Internal crate for zenoh."
 
 [dependencies]
 base64ct = "~1.6.0"
+zerofrom = "~0.1.5"
```
JEnoch, line 26: Using `~0.1.5` doesn't pin the version. See https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#tilde-requirements
To pin to an exact version you must use `=0.1.5`.
The same applies to `base64ct` dependency.

JEnoch (changes requested):

ci/valgrind-check/Cargo.toml
diogomatsubara: Thanks, my bad! Fixed in 6b5b0cd

JEnoch (approved):

JEnoch (approved):

PR #1801, opened 2025-02-27
Update stabby to `36.2.2`

fuzzypixelz:
The new version fixes nightly builds (see https://github.com/ZettaScaleLabs/stabby/pull/100).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3847,9 +3847,9 @@ dependencies = [
 
 [[package]]
 name = "stabby"
-version = "36.1.1"
+version = "36.2.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "311d6bcf0070c462ff626122ec2246f42bd2acd44b28908eedbfd07d500c7d99"
+checksum = "89b7e94eaf470c2e76b5f15fb2fb49714471a36cc512df5ee231e62e82ec79f8"
 dependencies = [
  "rustversion",
  "stabby-abi",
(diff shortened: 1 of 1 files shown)


Cargo.lock
```diff
@@ -4516,7 +4517,7 @@ version = "1.6.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "97fee6b57c6a41524a810daee9286c02d7752c4253064d0b05472833a438f675"
 dependencies = [
- "cfg-if 1.0.0",
+ "cfg-if 0.1.10",
```
Mallets, line 4520: This is a weird downgrade...

fuzzypixelz: [The Cargo manifest of `twox-hash` specifies `">= 0.1, < 2`.](https://github.com/shepmaster/twox-hash/blob/79168f770e0e870a11c5bb69ec0382547ef04790/Cargo.toml#L18)

fuzzypixelz: I'll see if I can remove this change, good catch!

fuzzypixelz: So we have a dependency (blake3) that [depends on `cfg-if 0.1.10`](https://github.com/BLAKE3-team/BLAKE3/blob/12642538340b3d6dad36e566217604267f94fe68/Cargo.toml#L78). We also have both `1.0.0` and `0.1.10` in the dependency graph.

I tried running cargo-update with Rust 1.75 and I still got the same result. I don't understand why Cargo randomly chooses to update this dependency, but it's consistent with the requirements of `twox-hash 1.6.3` so I don't see an issue; they explicitly allow both `0.1.x` and `1.x.y`.

PR #1806, opened 2025-02-28
Fix/correct misleading qos overwrite warning

### LINKED ISSUE #1805: Misleading warning for qos overwrites starting with `**/`
### Describe the bug

When writing a QoS overwrite with a key expression starting with `**/`, `PublisherBuilder::apply_qos_overwrites` emits the following warning:

```
WARN main ThreadId(01) zenoh::api::builders::publisher: Publisher declared on `demo/example/zenoh-rs-pub` which is included by multiple key_exprs in qos config. Using qos config for `**/demo/**`
```

This is misleading, as after inspection, the only other "overwrite" is actually an empty overwrite for the `**` key. I may be wrong but I believe this is due to the tree nature of the `publisher_qos_tree`.

A way to only emit this warning when relevant would be to filter on nodes with a non empty QoS ovewrite. I can make a PR for that.

P.S.: I have noticed when there are actually multiple overwrite, the one applied can vary between launches.

### To reproduce

Start the z_pub example with the following configuration:

```json5
{
  qos: {
    publication: [
      {
        key_exprs: ["**/demo/**"],
        config: {
          priority: "data_high",
        },
      },
    ],
  },
}
```

### System info

- Platform: Archlinux
- Zenoh 2311334e

Hugal31:
Refactor `PublisherBuilder::apply_qos_overwrites` and only shows the warning if multiple nodes **with overwrites** matches the publisher.

Improve the warning to show all matching key expressions.

Fixes #1805


zenoh/src/api/builders/publisher.rs
```diff
@@ -383,20 +383,17 @@ impl PublisherBuilder<'_, '_> {
...
-                    if nodes_including.len() > 1 {
-                        tracing::warn!(
-                            "Publisher declared on `{}` which is included by multiple key_exprs in qos config. Using qos config for `{}`",
-                            key_expr,
-                            node.keyexpr(),
-                        );
-                    }
-                    break;
+            if let Some(node) = nodes_including.first() {
+                qos_overwrites = node.weight().unwrap().clone();
```
oteffahi, line 389: Please replace the `unwrap` with `expect("node weight should not be None")`

oteffahi (changes requested):
Please address nitpick comment. Other than that it LGTM.

oteffahi (approved):

PR #1808, opened 2025-03-03
feat: add code coverage report

diogomatsubara:
A few caveats
- I had to `--skip` a few tests that were hanging during CI runs under `llvm-cov` (e.g. `router_linkstate`, `three_node_combination`, `three_node_combination_multicast`). Those tests pass during normal unit tests runs, not sure why they hang when executed under llvm-cov
- the new workflow adds a matrix job for both nightly and stable. Enabling nightly allows us to run the llvm-cov with the `--doctests` option
- added `--ignore-run-fail` so we upload the code coverage report even if the tests fail. We still have flaky tests that causes failures, which end up cancelling the report upload step.
- I still need to configure the code cov token for zenoh but that will require a separate PR to `eclipse-zenho/.eclipsefdn`. The token is used for reports on protected branches, like zenoh main.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -246,7 +246,7 @@ jobs:
             uses: taiki-e/install-action@cargo-llvm-cov
 
           - name: Generate code coverage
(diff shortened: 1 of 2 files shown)


YuanYuYuan:
LGTM.

PR #1811, opened 2025-03-06
lower_latency feature

yellowhatter:
Add lower_latency feature that lowers latency by pushing RX executor.
https://discord.com/channels/914168414178779197/1311983497929560155/1342039248987095131

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -169,11 +169,11 @@ jobs:
 
       - name: Run tests with SHM + lowlatency
         if: ${{ matrix.os == 'macos-latest' || matrix.os == 'windows-latest' }}
-        run: cargo nextest run -F test -F lower_latency -F shared-memory -F unstable -F internal_config -E 'not (test(test_default_features))' --exclude zenoh-examples --exclude zenoh-plugin-example --workspace
+        run: cargo nextest run -F test -F fix_tokio_latency -F shared-memory -F unstable -F internal_config -E 'not (test(test_default_features))' --exclude zenoh-examples --exclude zenoh-plugin-example --workspace
 
       - name: Run tests with SHM + lowlatency + unixpipe
         if: ${{ matrix.os == 'ubuntu-latest' }}
-        run: cargo nextest run -F test -F lower_latency -F shared-memory -F unstable -F internal_config -F transport_unixpipe -E 'not (test(test_default_features))' --exclude zenoh-examples --exclude zenoh-plugin-example --workspace
+        run: cargo nextest run -F test -F fix_tokio_latency -F shared-memory -F unstable -F internal_config -F transport_unixpipe -E 'not (test(test_default_features))' --exclude zenoh-examples --exclude zenoh-plugin-example --workspace
 
       - name: Check for feature leaks
         if: ${{ matrix.os == 'ubuntu-latest' }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -167,13 +167,13 @@ jobs:
       - name: Run tests
         run: cargo nextest run -F test -F internal_config --exclude zenoh-examples --exclude zenoh-plugin-example --workspace
 
-      - name: Run tests with SHM
+      - name: Run tests with SHM + lowlatency
         if: ${{ matrix.os == 'macos-latest' || matrix.os == 'windows-latest' }}
-        run: cargo nextest run -F test -F shared-memory -F unstable -F internal_config -E 'not (test(test_default_features))' --exclude zenoh-examples --exclude zenoh-plugin-example --workspace
+        run: cargo nextest run -F test -F lower_latency -F shared-memory -F unstable -F internal_config -E 'not (test(test_default_features))' --exclude zenoh-examples --exclude zenoh-plugin-example --workspace
 
-      - name: Run tests with SHM + unixpipe
+      - name: Run tests with SHM + lowlatency + unixpipe
         if: ${{ matrix.os == 'ubuntu-latest' }}
-        run: cargo nextest run -F test -F shared-memory -F unstable -F internal_config -F transport_unixpipe -E 'not (test(test_default_features))' --exclude zenoh-examples --exclude zenoh-plugin-example --workspace
+        run: cargo nextest run -F test -F lower_latency -F shared-memory -F unstable -F internal_config -F transport_unixpipe -E 'not (test(test_default_features))' --exclude zenoh-examples --exclude zenoh-plugin-example --workspace
 
       - name: Check for feature leaks
         if: ${{ matrix.os == 'ubuntu-latest' }}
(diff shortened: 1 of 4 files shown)


commons/zenoh-runtime/src/lib.rs
```diff
@@ -79,6 +79,16 @@ impl RuntimeParam {
                 format!("{}-{}", zrt, id)
             })
             .build()?;
+        #[cfg(feature = "lower_latency")]
+        if zrt == ZRuntime::RX {
+            tracing::debug!("Spawning RX runtime polling task...");
+            #[allow(clippy::let_underscore_future)]
+            let _ = rt.spawn(async {
+                loop {
+                    tokio::time::sleep(tokio::time::Duration::from_micros(100)).await;
```
Mallets, line 88: Tokio does not have a time resolution smaller than `1 ms`.
As a consequence, this translates in a bursty spin loop.

yellowhatter: I know that tokio has some internal limitations on sleep resolution. However, for this code it sleeps for some (unmeasured) time and doesn't burn the CPU. The CPU consumption is around 0.6% on my laptop

commons/zenoh-runtime/Cargo.toml
```diff
@@ -14,6 +14,7 @@ description = { workspace = true }
 
 [features]
 tracing-instrument = []
+lower_latency = []
```
evshary, line 17: Just out of curiosity. Is there any downside enabling the feature?
If not, why not keep it as default?

commons/zenoh-runtime/src/lib.rs
Mallets: I understand the rationale behind that looping in backround keeps the thread warm.
But what's the gain in latency introduced with this PR?

yellowhatter: Ah yes, I should have added more context....

Before:
image

After:
image

commons/zenoh-runtime/Cargo.toml
yellowhatter: Well, it produces additional spurious wakeups. This slightly lowers the throughput (1-2%), increases power consumption on low-intensive workload setups.

commons/zenoh-runtime/src/lib.rs
yellowhatter: Hmm, I just realized that "before" was tested with SHM, so it has good extra bonus for latency on big buffers.

yellowhatter: Here is another test, with no SHM:
latency_mean

yellowhatter: Ape tests results slightly differs from run to run. But the trend is visible

wyfo: > Tokio does not have a time resolution smaller than `1 ms`. As a consequence, this translates in a bursty spin loop.

Tokio will round the sleep duration: https://github.com/tokio-rs/tokio/blob/e076d21f679a35ae2697165d46d111285d09e3b4/tokio/src/runtime/time/source.rs#L17

.github/workflows/ci.yml
```diff
@@ -167,13 +167,13 @@ jobs:
       - name: Run tests
         run: cargo nextest run -F test -F internal_config --exclude zenoh-examples --exclude zenoh-plugin-example --workspace
 
-      - name: Run tests with SHM
+      - name: Run tests with SHM + lowlatency
```
evshary, line 170: Why do we only test it with SHM? I think it can be used w/o SHM?

commons/zenoh-runtime/Cargo.toml
evshary: I see. FMPOV, if latency is more important to us, I prefer making it default. Anyway, both ways are fine to me.

Mallets, line 17: I would pick a better name, I think `lower_latency` is not expressive enough and it could create confusion with the [lowlatency](https://github.com/eclipse-zenoh/zenoh/blob/ee9fc7ed237f3cd493dc3e3f5d38f9812a7ebdd8/DEFAULT_CONFIG.json5#L403) config option.

commons/zenoh-runtime/src/lib.rs
```diff
@@ -79,6 +79,16 @@ impl RuntimeParam {
                 format!("{}-{}", zrt, id)
             })
             .build()?;
+        #[cfg(feature = "lower_latency")]
```
evshary, line 82: Since this is kind of a workaround, we can add some comments (why we do this).
My understanding is that the runtime does not wake up so frequently for some unknown reasons. Perhaps this is the limitation in tokio?

commons/zenoh-runtime/Cargo.toml
yellowhatter: We can make it default at anytime, we can put it ON in rmw.... Currently I want more discussions and tests.

.github/workflows/ci.yml
yellowhatter: Just in order not to run too much jobs.

commons/zenoh-runtime/src/lib.rs
yellowhatter: Agree.
Neighboring TX task works well at the same publication frequency. So, the problem is particular, there is no global limitation.

yellowhatter: added comment

commons/zenoh-runtime/Cargo.toml
yellowhatter: `fix_tokio_latency` ?

Mallets: It's a good proposal for me.

yellowhatter: Renamed

yellowhatter:
@OlivierHecart it seems that 100us sleeps give the best latency on my setup (Ubuntu 24.04 + AMD Ryzen 9). 200 us is slightly worse and 1ms is totally useless.

Mallets (approved):

evshary (approved):
LGTM

OlivierHecart:
I can confirm there is a significant impact on CPU consumption. Here are some results with 160 ros2 nodes:
image
image

yellowhatter:
> I can confirm there is a significant impact on CPU consumption. Here are some results with 160 ros2 nodes: image image

That's at least 160 000 tokio task wakeups per second - for sure the cpu consumption goes crazy.

commons/zenoh-runtime/Cargo.toml
wyfo: This feature name sound a bit pedantic to me. The message we send here is that tokio is bad, and we are able to fix it, while at the same time, we are using several runtime, which may be a start of explanation about why we have bad latency.

Issue #1810, opened 2025-03-04
Add option to bind TCP client ports to a given socket_address


OlivierHecart:
### Describe the feature

Add a `bind` Endpoint config option. For example:
`tcp/192.168.0.10:7447#bind=192.168.0.5:0`

When this option is set to a TCP `connect/endpoint` Zenoh should call the system `bind()` function with the given socket address on the created TCP socket before connecting to the remote address.

PR #1822, opened 2025-03-11
[ZEN-526] Add Bind config option 

Charles-Schleich:
This PR aims to address issue https://github.com/eclipse-zenoh/zenoh/issues/1810 

Add a bind Endpoint config option. For example:
`tcp/192.168.0.10:7447#bind=192.168.0.5:0`
- [x] TCP 
- [x] TLS, 
- [x] UDP
- [x] Quic

### DIFF
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -28,8 +28,9 @@
   ///
   /// For TCP and TLS links, it is possible to specify the TCP buffer sizes:
   /// E.g. tcp/192.168.0.1:7447#so_sndbuf=65000;so_rcvbuf=65000
-  /// For TCP, UDP, Quic and TLS links, it is possible to specify a bind address for the local socket:
+  /// For TCP, UDP, Quic and TLS links, it is possible to specify a `bind` address for the local socket:
   /// E.g. tcp/192.168.0.1:7447#bind=192.168.0.1:0
+  //  Note!: Currently it is unsupported to specify both `bind` and `iface`. 
   connect: {
     /// timeout waiting for all endpoints connected (0: no retry, -1: infinite timeout)
     /// Accepts a single value (e.g. timeout_ms: 0)
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -28,6 +28,8 @@
   ///
   /// For TCP and TLS links, it is possible to specify the TCP buffer sizes:
   /// E.g. tcp/192.168.0.1:7447#so_sndbuf=65000;so_rcvbuf=65000
+  /// For TCP, UDP, Quic and TLS links, it is possible to specify a bind address for the local socket:
+  /// E.g. tcp/192.168.0.1:7447#bind=192.168.0.1:0
   connect: {
     /// timeout waiting for all endpoints connected (0: no retry, -1: infinite timeout)
     /// Accepts a single value (e.g. timeout_ms: 0)
@@ -74,8 +76,6 @@
   ///
   /// For TCP and TLS links, it is possible to specify the TCP buffer sizes:
   /// E.g. tcp/192.168.0.1:7447#so_sndbuf=65000;so_rcvbuf=65000
-  /// For TCP, UDP, Quic and TLS it is possible to specify a bind endpoint both in IPV4 and IPV6:
-  /// E.g. tcp/192.168.0.1:7447#bind=192.168.0.1:0
   listen: {
     /// timeout waiting for all listen endpoints (0: no retry, -1: infinite timeout)
     /// Accepts a single value (e.g. timeout_ms: 0)
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -74,6 +74,8 @@
   ///
   /// For TCP and TLS links, it is possible to specify the TCP buffer sizes:
   /// E.g. tcp/192.168.0.1:7447#so_sndbuf=65000;so_rcvbuf=65000
+  /// For TCP, UDP, Quic and TLS it is possible to specify a bind endpoint both in IPV4 and IPV6:
+  /// E.g. tcp/192.168.0.1:7447#bind=192.168.0.1:0
   listen: {
     /// timeout waiting for all listen endpoints (0: no retry, -1: infinite timeout)
     /// Accepts a single value (e.g. timeout_ms: 0)
--- a/commons/zenoh-protocol/src/core/endpoint.rs
+++ b/commons/zenoh-protocol/src/core/endpoint.rs
@@ -880,4 +880,8 @@ fn endpoints() {
     assert_eq!(i.next(), Some("224.0.0.2"));
     assert_eq!(i.next(), Some("224.0.0.3"));
     assert_eq!(i.next(), None);
+
+    let endpoint = EndPoint::from_str("udp/127.0.0.1:7447#bind=224.0.0.1:8080").unwrap();
+    let c = endpoint.config();
+    assert_eq!(c.get("bind"), Some("224.0.0.1:8080"));
 }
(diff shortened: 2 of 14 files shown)


io/zenoh-link-commons/src/tcp.rs
```diff
@@ -60,6 +63,28 @@ impl<'a> TcpSocketConfig<'a> {
         dst_addr: &SocketAddr,
     ) -> ZResult<(TcpStream, SocketAddr, SocketAddr)> {
         let socket = self.socket_with_config(dst_addr)?;
```
OlivierHecart, line 65: It seems that the socket will be bound twice here. Once in `socket_with_config` and once below.

DEFAULT_CONFIG.json5
```diff
@@ -28,6 +28,8 @@
   ///
   /// For TCP and TLS links, it is possible to specify the TCP buffer sizes:
   /// E.g. tcp/192.168.0.1:7447#so_sndbuf=65000;so_rcvbuf=65000
+  /// For TCP, UDP, Quic and TLS links, it is possible to specify a bind address for the local socket:
```
OlivierHecart, line 31: Bind option is for connect endpoints, not listen endpoints.

OlivierHecart:
This is I believe where the bind address could be used for multicast case: [multicast.rs#L254](https://github.com/ZettaScaleLabs/zenoh/blob/ea33f35d7b2983a19630cefe5de0e9697769a3a8/io/zenoh-links/zenoh-link-udp/src/multicast.rs#L254).
Note: we already extract the ip address from the network interface. But, with the bind option, the user could specify the port number and select a particular address for multi address interfaces.
I propose the following: If the bind address is not specified, we do as today. If the bind address is `IpvXAddr::UNSPECIFIED`, we do as today for the address but use the provided port number. If the bind address is a specific address then we use it to bind.

commons/zenoh-protocol/src/core/endpoint.rs
```diff
@@ -880,4 +880,8 @@ fn endpoints() {
     assert_eq!(i.next(), Some("224.0.0.2"));
     assert_eq!(i.next(), Some("224.0.0.3"));
     assert_eq!(i.next(), None);
+
+    let endpoint = EndPoint::from_str("udp/127.0.0.1:7447#bind=224.0.0.1:8080").unwrap();
```
Mallets, line 884: Nice but not necessary since in `endpoint.rs` `key=value` is purely syntactic and does not carry any semantics.

io/zenoh-links/zenoh-link-quic/src/unicast.rs
```diff
@@ -264,17 +265,19 @@ impl LinkManagerUnicastTrait for LinkManagerUnicastQuic {
         client_crypto.client_config.alpn_protocols =
             ALPN_QUIC_HTTP.iter().map(|&x| x.into()).collect();
 
-        let ip_addr: IpAddr = if dst_addr.is_ipv4() {
-            Ipv4Addr::UNSPECIFIED.into()
+        let src_addr = if let Some(bind_socket_str) = epconf.get(BIND_SOCKET) {
+            SocketAddr::from_str(bind_socket_str)?
```
Mallets, line 269: Here we are limiting binding a socket to an IP only. Would it make sense to use `get_quic_addr` instead to retrive the IP address potentially from a local dns name?

io/zenoh-links/zenoh-link-tcp/src/utils.rs
```diff
@@ -1,3 +1,5 @@
+use std::{net::SocketAddr, str::FromStr};
```
Mallets, line 1: Use slipped before the header.

io/zenoh-links/zenoh-link-tcp/src/utils.rs
```diff
@@ -45,14 +48,21 @@ pub(crate) struct TcpLinkConfig<'a> {
...
     pub(crate) tx_buffer_size: Option<u32>,
     pub(crate) bind_iface: Option<&'a str>,
+    pub(crate) bind_socket: Option<SocketAddr>,
 }
 
 impl<'a> TcpLinkConfig<'a> {
     pub(crate) fn new(config: &'a Config) -> ZResult<Self> {
+        let mut bind_socket = None;
+        if let Some(bind_socket_str) = config.get(BIND_SOCKET) {
+            bind_socket = Some(SocketAddr::from_str(bind_socket_str)?)
```
Mallets, line 58: Same comment as in Quic. Use `get_tcp_addrs`?

io/zenoh-links/zenoh-link-tls/src/utils.rs
```diff
@@ -271,6 +271,10 @@ impl<'a> TlsServerConfig<'a> {
                     .map_err(|_| zerror!("Unknown TCP write buffer size argument: {}", size))?,
             );
         };
+        let mut bind_socket = None;
+        if let Some(bind_socket_str) = config.get(BIND_SOCKET) {
+            bind_socket = Some(SocketAddr::from_str(bind_socket_str)?);
```
Mallets, line 276: Same comment as in Quic. Use `get_tls_addr`?

Mallets:
Tests for the `bind` feature should be added as part of the transport tests.

Moreover, it is unclear how potential conflicts between `iface` and `bind` are handled. 
It should be at least documented how incompatible configurations are handled.
E.g., how to handle connecting to `tcp/foo.com:7447#iface=wifi0;bind=10.0.0.1` where `10.0.0.1` belongs to a different interface than `wifi0`?

io/zenoh-links/zenoh-link-quic/src/unicast.rs
Charles-Schleich: Ive the added calling of this function to retrieve the IP address from a dns name. 
Is this more or less what you meant @Mallets  ?

io/zenoh-links/zenoh-link-tcp/src/utils.rs
Charles-Schleich: added, Same as above.

io/zenoh-links/zenoh-link-tls/src/utils.rs
```diff
@@ -439,13 +444,20 @@ impl<'a> TlsClientConfig<'a> {
             );
         };
 
+        // TODO: Do we want this option for a TLS Client ?
+        let mut bind_socket = None;
+        if let Some(bind_socket_str) = config.get(BIND_SOCKET) {
+            bind_socket = Some(SocketAddr::from_str(bind_socket_str)?);
```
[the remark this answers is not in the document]
Charles-Schleich: added, Same as above.

io/zenoh-links/zenoh-link-quic/src/unicast.rs
Mallets: Yes. Although the CI is still failing.

Charles-Schleich:
@Mallets @OlivierHecart 
Would it be acceptable to not support both `iface` and `bind` at the same time, or is this something that we would want in the future ? 

The way I see it is that if `iface` is present, we will nee to search through all ip's that are associated with that `iface` and verify that one of them matches the bind address.

Mallets:
I believe `iface` and `bind` should be mutually exclusive.

commons/zenoh-protocol/src/core/endpoint.rs
```diff
@@ -120,7 +120,7 @@ impl fmt::Debug for ProtocolMut<'_> {
 // Address
 #[repr(transparent)]
 #[derive(Copy, Clone, PartialEq, Eq, Hash)]
-pub struct Address<'a>(pub(super) &'a str);
+pub struct Address<'a>(pub &'a str);
```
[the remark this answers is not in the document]
Charles-Schleich: I agree, updated `Address` to implement from `From<&'a str>`

OlivierHecart:
Ideally the tests should check that the bind address and port are indeed used by checking (on both sides) that the created link indeed uses the correct src/dst addresses. I don't believe it's the case yet.

OlivierHecart (changes requested):
Ideally the tests should check that the bind address and port are indeed used by checking (on both sides) that the created link indeed uses the correct src/dst addresses. I don't believe it's the case yet.

Issue #1823, opened 2025-03-12
Handle the addition or suppression of network interfaces/addresses


Hugal31:
### Describe the feature

# Problem

Right now, the node [caches the UP and RUNNING interfaces on startup](https://github.com/eclipse-zenoh/zenoh/blob/b661454b65ac5a5b1b06f619b3a1b1ab69fe2f34/commons/zenoh-util/src/net/mod.rs#L33), starts its listeners and [caches their locators](https://github.com/eclipse-zenoh/zenoh/blob/b661454b65ac5a5b1b06f619b3a1b1ab69fe2f34/zenoh/src/net/runtime/orchestrator.rs#L584) (e.g. `tcp/[::]:7447` becomes `tcp/192.168.0.1:7447`).

However, if the node is started before a network interface, or if its IP address changes, it is not taken in account.

My particular use case is a handled device connecting over Wi-Fi to one of our multiple robots.

# Solution

Automatically detect new/deleted IP addresses and update the locator list as well as the socket writing on the multicast address. On Linux, this could be done using Netlink, but it could work using a simple timer.

# Alternative solution

Maybe have the user detect the network change themselves and call a function to update the locator list/multicast socket.

# Additional context

I have no clue on how this would work on system other than Unix-like systems.

PR #1824, opened 2025-03-12
Detect change on network interface to update locators and scouting

Hugal31:
On Unix systems, use netlink to detect added or delete IPv4 addresses to:
  * Renew the interface list.
  * Update the node locators.
  * Restart the scouting to use the new addresses/discard the old addresses.

This is still quite hacky, with the following shortcomings:
* We do not handle IPv6 addresses, because the IPv6 addresses are "renewed" every few seconds, causing unecessary closing and re-opening of the sockets.
* We perform the locator update and the scouting reset even if the new/old addresses are not used as per the configuration.
* Only works on Linux because of the Netlink dependency.
* The code overall is not the best quality.

This would solve #1823 (with the mentioned caveats).

Feel free to close this, I opened it more for reference and I might not update it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2241,6 +2241,71 @@ dependencies = [
 ]
 
 [[package]]
+name = "netlink-packet-core"
+version = "0.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "72724faf704479d67b388da142b186f916188505e7e0b26719019c525882eda4"
+dependencies = [
+ "anyhow",
+ "byteorder",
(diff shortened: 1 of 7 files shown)


Hugal31:
A few changes:

* I did a bigger refactor to extract the scouting procedure.
* I completely ignore IPv6 changes.
* I now only open/close socket that are missing/no longer working because of the interface up/down, instead of closing and re-opening everything.

There's an additional caveat, the "reset" of the scouting routine *may* interrupt something important (e.g. the connection to a peer).

patrickelectric:
this is quite important!

PR #1825, opened 2025-03-12
add qos overwrite interceptor

DenisBiryukov91:
add qos overwrite interceptor

### DIFF
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -401,34 +401,33 @@
   //  /// Configure QoS overwrite rules.
   //  qos_overwrite: [
   //     {
-  //       /// Id has to be unique within the rule set.
-  //       id: "rule1",
+  //       /// Optional Id, has to be unique.
+  //       id: "lo0_en0_qos_overwrite",
   //       // Optional list of interfaces, if not specified, will be applied to all interfaces.
   //       interfaces: [
   //         "lo0",
   //         "en0",
   //       ],
-  //       messages: ["put", "delete"],
-  //       flows: ["egress"],
+  //       /// List of message types to apply to.
+  //       messages: [
+  //          "put", // put publicaitons
+  //          "delete" // delete publications
+  //          "query", // get queries
+  //          "reply", // replies to to queries
+  //       ],
+  //       /// Optional list of data flows messages will be processed on ("egress" and/or "ingress").
+  //       /// If absent, the rules will be applied to both flows.
+  //       flows: ["egress", "ingress"],
   //       key_exprs: ["test/demo"],
   //       overwrite: {
   //          /// Optional new priority value, if not specified priority of the messages will stay unchanged.
   //          priority: "real_time",
   //          /// Optional new congestion control value, if not specified congestion control of the messages will stay unchanged.
   //          congestion_control: "block",
-  //          /// Optional new priority value, if not specified priority of the messages will stay unchanged.
+  //          /// Optional new express value, if not specified express flag of the messages will stay unchanged.
   //          express: true
   //        },
-  //     },
-  //     {
-  //        id: "rule2",
-  //        messages: ["query", "reply"],
-  //        flows: ["ingress"],
-  //        key_exprs: ["**"],
-  //        overwrite: {
-  //          priority: "interactive_high",
-  //        },
-  //     },
+  //     }
   //   ],
 
   /// Configure internal transport parameters
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -398,6 +398,39 @@
   //   ]
   //},
 
+  //  /// Configure QoS overwrite rules.
+  //  qos_overwrite: [
+  //     {
+  //       /// Id has to be unique within the rule set.
+  //       id: "rule1",
+  //       // Optional list of interfaces, if not specified, will be applied to all interfaces.
+  //       interfaces: [
+  //         "lo0",
+  //         "en0",
+  //       ],
+  //       messages: ["put", "delete"],
+  //       flows: ["egress"],
+  //       key_exprs: ["test/demo"],
+  //       overwrite: {
+  //          /// Optional new priority value, if not specified priority of the messages will stay unchanged.
+  //          priority: "real_time",
+  //          /// Optional new congestion control value, if not specified congestion control of the messages will stay unchanged.
+  //          congestion_control: "block",
+  //          /// Optional new priority value, if not specified priority of the messages will stay unchanged.
+  //          express: true
+  //        },
+  //     },
+  //     {
+  //        id: "rule2",
+  //        messages: ["query", "reply"],
+  //        flows: ["ingress"],
+  //        key_exprs: ["**"],
+  //        overwrite: {
+  //          priority: "interactive_high",
+  //        },
+  //     },
+  //   ],
+
   /// Configure internal transport parameters
   transport: {
     unicast: {
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -249,6 +249,39 @@
   //        },
   //      },
   //    ],
+  //    /// Overwrite QoS options for messages sent and received from/to the network
+  //    /// This allows more fine grained rules (per network card, etc...) but is
+  //    /// less performant than the publication option above.
+  //    network: [
+  //      {
+  //        /// Optional Id, has to be unique.
+  //        id: "lo0_en0_qos_overwrite",
+  //        // Optional list of interfaces, if not specified, will be applied to all interfaces.
+  //        interfaces: [
+  //          "lo0",
+  //          "en0",
+  //        ],
+  //        /// List of message types to apply to.
+  //        messages: [
+  //          "put", // put publications
+  //          "delete" // delete publications
+  //          "query", // get queries
+  //          "reply", // replies to queries
+  //        ],
+  //        /// Optional list of data flows messages will be processed on ("egress" and/or "ingress").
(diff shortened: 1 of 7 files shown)


zenoh/src/net/routing/interceptor/qos_overwrite.rs
```diff
@@ -0,0 +1,288 @@
...
+                if self.filter.delete {
+                    self.overwrite_qos(ext_qos);
+                    tracing::trace!(
+                        "Overwriting QoS for {:?} to {:?}",
+                        QosOverwriteMessage::Delete,
+                        ext_qos
+                    );
+                }
+            }
+            _ => {}
```
oteffahi, line 284: Explicitly add all other variants of pattern match, just in case we ever change the patterns and need to update interceptors.

zenoh/tests/qos_overwrite.rs
```diff
@@ -0,0 +1,233 @@
...
+    assert_eq!(
+        reply.result().unwrap().priority(),
+        Priority::InteractiveHigh
+    );
+    assert_eq!(
+        reply.result().unwrap().congestion_control(),
+        CongestionControl::Drop
+    );
+    assert!(!reply.result().unwrap().express());
+}
```
oteffahi, line 233: Add tests for configuration format when opening session:
- empty flows list should error out
- flows set to None should not error out
- etc...

For reference: https://github.com/eclipse-zenoh/zenoh/blob/main/zenoh/tests/interceptors.rs#L287-L319
(doesn't necessarily need to be a `should_panic` test, it can just be an assertion on the returned error's content)

DEFAULT_CONFIG.json5
```diff
@@ -398,6 +398,39 @@
...
+  //     },
+  //     {
+  //        id: "rule2",
+  //        messages: ["query", "reply"],
+  //        flows: ["ingress"],
+  //        key_exprs: ["**"],
+  //        overwrite: {
+  //          priority: "interactive_high",
+  //        },
+  //     },
```
oteffahi, line 431: It's better to just have one config example with all the possible variants of its fields

DEFAULT_CONFIG.json5
```diff
@@ -398,6 +398,39 @@
   //   ]
   //},
 
+  //  /// Configure QoS overwrite rules.
+  //  qos_overwrite: [
+  //     {
+  //       /// Id has to be unique within the rule set.
```
oteffahi, line 404: Please mention that it's optional

oteffahi (changes requested):

zenoh/src/net/routing/interceptor/qos_overwrite.rs
DenisBiryukov91: I'm not sure this right approach for an interceptor that is specifically meant only to affect certain kinds of messages (i.e. put, delete, query, reply) - we will have to come back and fix it every time a new control message type is added.

DEFAULT_CONFIG.json5
DenisBiryukov91: done

DEFAULT_CONFIG.json5
DenisBiryukov91: done

zenoh/tests/qos_overwrite.rs
DenisBiryukov91: done

DEFAULT_CONFIG.json5
```diff
@@ -398,6 +398,39 @@
...
+  //       ],
+  //       messages: ["put", "delete"],
+  //       flows: ["egress"],
+  //       key_exprs: ["test/demo"],
+  //       overwrite: {
+  //          /// Optional new priority value, if not specified priority of the messages will stay unchanged.
+  //          priority: "real_time",
+  //          /// Optional new congestion control value, if not specified congestion control of the messages will stay unchanged.
+  //          congestion_control: "block",
+  //          /// Optional new priority value, if not specified priority of the messages will stay unchanged.
```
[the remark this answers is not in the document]
DenisBiryukov91: done

zenoh/src/net/routing/interceptor/qos_overwrite.rs
```diff
@@ -0,0 +1,288 @@
...
+    let mut id_set = HashSet::new();
+    for q in config {
+        // check unicity of rule id
+        if let Some(id) = &q.id {
+            if !id_set.insert(id.clone()) {
+                bail!("Invalid QosOverwrite config: id '{id}' is repeated");
+            }
+        }
+        let q = q.clone();
+        // check for undefined flows and initialize them
```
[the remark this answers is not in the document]
DenisBiryukov91: done

commons/zenoh-config/src/lib.rs
```diff
@@ -201,6 +201,32 @@ pub enum Permission {
     Deny,
 }
 
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, Eq, Hash, PartialEq)]
+#[serde(rename_all = "snake_case")]
+pub enum QosOverwriteMessage {
```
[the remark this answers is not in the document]
DenisBiryukov91: agree, @OlivierHecart, @Mallets, @milyin do you have any suggestions ?

zenoh/src/net/routing/interceptor/qos_overwrite.rs
Mallets: The addition of a new control message is a rare thing since it means a breaking change in the protocol.
By  explicitly enumerating all the variant, it allows to immediately see the impact of a new addition throughout the whole codebase and react accordingly to the new behaviour. Therefore, I think it's good to have all variants explicitly enumerated.

commons/zenoh-config/src/lib.rs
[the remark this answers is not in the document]
DenisBiryukov91: @Mallets  with this pr we now have 2 similarly named config options: 
1. qos_overwrite (for this interceptor),
2. qos (to overwrite default qos values for publisher).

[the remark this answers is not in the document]
Mallets: I see. What about having this config inside `qos`?

Something like:
```
qos: {
  /// Overwrite QoS options at API-level for PUT and DELETE messages
  //    publication: [ ... ],
  /// Overwrite QoS options at interceptor-level for PUT, DELETE, QUERY, REPLY messages
  interceptor: [ ... ],
}
```

zenoh/src/net/routing/interceptor/qos_overwrite.rs
DenisBiryukov91: fixed

commons/zenoh-config/src/lib.rs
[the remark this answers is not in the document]
oteffahi: My comment was more about the struct naming in the code than the config field naming 😅

DEFAULT_CONFIG.json5
```diff
@@ -398,6 +398,38 @@
...
+  //       /// Optional Id, has to be unique.
+  //       id: "lo0_en0_qos_overwrite",
+  //       // Optional list of interfaces, if not specified, will be applied to all interfaces.
+  //       interfaces: [
+  //         "lo0",
+  //         "en0",
+  //       ],
+  //       /// List of message types to apply to.
+  //       messages: [
+  //          "put", // put publicaitons
```
oteffahi, line 413:
```suggestion
  //          "put", // put publications
```

DEFAULT_CONFIG.json5
```diff
@@ -398,6 +398,38 @@
...
+  //          "delete" // delete publications
+  //          "query", // get queries
+  //          "reply", // replies to to queries
```
oteffahi, line 416:
```suggestion
  //          "reply", // replies to queries
```

zenoh/src/net/routing/interceptor/qos_overwrite.rs
```diff
@@ -0,0 +1,331 @@
...
+            .flows
+            .get_or_insert(vec![InterceptorFlow::Ingress, InterceptorFlow::Egress]);
+        if flows.is_empty() {
+            bail!("Invalid Qos Overwrite config: flows list must not be empty");
+        }
+        // check for empty messages list
+        if q.messages.is_empty() {
+            bail!("Invalid Qos Overwrite config: messages list must not be empty");
+        }
+        res.push(Box::new(QosOverwriteFactory::new(q)));
```
oteffahi, line 61: Add a check for non-empty list of interfaces if provided.

oteffahi (changes requested):

zenoh/src/net/routing/interceptor/qos_overwrite.rs
DenisBiryukov91: fixed

commons/zenoh-config/src/lib.rs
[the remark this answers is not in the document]
DenisBiryukov91: I do not think that internal struct naming is an issue, on the contrary it would be better if they could be reused between the two as much as possible. The config entry naming seems to be more important to me ...

oteffahi (approved):
Maybe the name of the config field needs to change or be moved.
Other than that, LGTM.

oteffahi:
This interceptor needs to be updated with the feature added in https://github.com/eclipse-zenoh/zenoh/pull/1850

commons/zenoh-config/src/lib.rs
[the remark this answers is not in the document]
Mallets: I'm fine with `network`.

PR #1836, opened 2025-03-17
fix: add rust 1.75 CI check

diogomatsubara:
With the toolchain bump proposed in https://github.com/eclipse-zenoh/zenoh/pull/1803, we need to guarantee zenoh can still be built on rust 1.75


OlivierHecart (approved):

fuzzypixelz:
@diogomatsubara @OlivierHecart Why check with the release profile? It needlessly increases the job execution time.

PR #1837, opened 2025-03-17
Block drop/undeclare for subscribers/queryables/listeners until callbacks return

### LINKED ISSUE #1818: Ensure that callbacks are never called when subscriber/queryable go out of scope.
### Describe the bug

Ensure that callbacks are never called when subscriber/queryable go out of scope. Also ensure that these callbacks will be dropped before corresponding undeclare call returns.
This is not relevant to rust, but in other binding like C and C++, calling callbacks after subscriber/queryable destruction can result in UB, since there is no borrow checker to ensure that data captured by callbacks is still alive at the moment of callback invocation/destruction.

### To reproduce

It is sufficient to undeclare subscriber/queryable roughly at the same moment when a new message/query arrives.

### System info

any

DenisBiryukov91:
Block drop/undeclare for subscribers/queryables/listeners until callbacks return.
Resolves #1818

### DIFF
--- a/zenoh/src/api/handlers/callback.rs
+++ b/zenoh/src/api/handlers/callback.rs
@@ -14,7 +14,9 @@
 
 //! Callback handler trait.
 
-use std::{sync::Arc, time::Duration};
+use std::sync::Arc;
+
+use flume::{Receiver, Sender};
 
 use crate::api::handlers::IntoHandler;
 
@@ -128,24 +130,17 @@ where
 }
 
 #[cfg(not(feature = "internal"))]
+#[derive(Clone)]
 pub(crate) struct WeakCallback<T> {
-    use_guard: Arc<()>,
+    guard: Option<Sender<()>>,
     cb: Callback<T>,
 }
 
 #[zenoh_macros::internal]
+#[derive(Clone)]
 pub struct WeakCallback<T> {
-    use_guard: Arc<()>,
+    guard: Option<Sender<()>>,
     cb: Callback<T>,
-}
-
-impl<T> Clone for WeakCallback<T> {
-    fn clone(&self) -> Self {
-        Self {
-            use_guard: self.use_guard.clone(),
-            cb: self.cb.clone(),
-        }
-    }
 }
 
 impl<T> WeakCallback<T> {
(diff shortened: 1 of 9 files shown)


zenoh/src/api/handlers/callback.rs
```diff
@@ -126,3 +126,66 @@ where
         (move |evt| (self.callback)(evt), ()).into_handler()
     }
 }
+
+#[doc(hidden)]
+pub struct WeakCallback<T> {
```
OlivierHecart, line 131: Shouldn't this be marked internal ?

zenoh/src/api/handlers/callback.rs
```diff
@@ -126,3 +126,66 @@ where
...
+impl<T> WeakCallback<T> {
+    #[inline]
+    pub fn call(&self, arg: T) {
+        self.cb.call(arg);
+    }
+}
+
+#[doc(hidden)]
+// A Callback that blocks on Drop operation until all of its weak clones are dropped
+pub struct StrongCallback<T> {
```
OlivierHecart, line 154: Shouldn't this be marked internal ?

zenoh/src/api/handlers/callback.rs
```diff
@@ -126,3 +126,77 @@ where
...
+                cb: value,
+            },
+        }
+    }
+}
+
+impl<T> Drop for StrongCallback<T> {
+    fn drop(&mut self) {
+        const SLEEP_PERIOD: Duration = Duration::from_millis(1);
+        while Arc::strong_count(&self.cb.use_guard) > 1 {
```
OlivierHecart, line 196: This is extremely dangerous. Called in an async context, it could block the executor.

DenisBiryukov91: Any non-async synchronization (i.e. mutex, etc) can block the executor. The way we use weak callbacks today (except probably advanced pub-sub) guarantees that they are destroyed in the same thread which acquired them without any async calls in between, so this is safe, but obviously not future-proof.

wyfo:
Please, no sleep loop.

DenisBiryukov91:
> Please, no sleep loop.

The only alternative I see is to use mutex/rwLock when cloning/destroying a callback, but this would badly affect performance. I'm wondering if there is a way to use atomic_wait here to get rid of sleep loop ...

wyfo:
I've an idea that I'm testing, give me a few minutes.

wyfo:
I've implemented an event based approach in #1871

PR #1841, opened 2025-03-18
Stop plugins on runtime close.

Jezza:
Plugins don't receive a kill signal so cleanup is difficult, if not straight-up impossible.

This just stops the plugins when the runtime is closed.

(Also fixed a small error message)


FedorSmirnov89:
Bump. Would also be interested in having this.

Jezza:
Any actionable advice for this PR?

Jezza:
This actually leads to bugs internally within zenoh.

PR #1842, opened 2025-03-18
feat: Add TTL connection param to multicast

jmonticelli:
Hi, I am using Zenoh for a project and I use a multicast listen endpoint in peer mode.

Using a multicast endpoint for multicast communication works fine on flat networks, but once a router is involved, the multicast traffic is lost, [due to the default TTL of 1 in multicast connections in the Tokio library](https://docs.rs/tokio/latest/tokio/net/struct.UdpSocket.html#method.set_multicast_ttl_v4).

See also: https://docs.rs/tokio/latest/tokio/net/struct.UdpSocket.html#method.set_multicast_ttl_v4

Note I have minimal rust programming experience, and used online resources and the code surrounding this addition to help me craft this PR, so feel free to make changes if I did something suboptimally or wrong.

I was able to build zenoh and run the zenoh pub client like so:

```
./target/release/examples/z_pub -m peer -l 'udp/239.1.3.37:9001#iface=myinterface;ttl=32' -k 'testkey' -p 'Hello from beyond the router' --no-multicast-scouting
```

I confirmed that the outgoing multicast TTL is set to what I specify it to, and that the traffic is no longer dropped on the other end of the network.

This is a critical feature when handling multicast traffic that is routed.

P.S., very nice library :)

### DIFF
--- a/io/zenoh-links/zenoh-link-udp/src/multicast.rs
+++ b/io/zenoh-links/zenoh-link-udp/src/multicast.rs
@@ -326,7 +326,7 @@ impl LinkManagerMulticastUdp {
             match &local_addr {
                 IpAddr::V4(_) => {
                     let ttl = match ttl_str.parse::<u32>() {
-                        Ok(ttl)  => ttl,
+                        Ok(ttl) => ttl,
                         Err(e) => bail!("Can not parse TTL '{}' to a u32: {}", ttl_str, e),
                     };
 
--- a/io/zenoh-links/zenoh-link-udp/src/multicast.rs
+++ b/io/zenoh-links/zenoh-link-udp/src/multicast.rs
@@ -323,29 +323,23 @@ impl LinkManagerMulticastUdp {
 
         // If TTL is specified, add set the socket's TTL
         if let Some(ttl_str) = config.get(UDP_MULTICAST_TTL) {
-            let ttl = match ttl_str.parse::<u32>() {
-                Ok(ttl) => ttl,
-                Err(e) => bail!("Can not parse TTL '{}' to a u32: {}", ttl_str, e),
-            };
-
             match &local_addr {
                 IpAddr::V4(_) => {
-                    ucast_sock
-                        .set_multicast_ttl_v4(ttl)
-                        .map_err(|e| zerror!("{}: {}", mcast_addr, e))?;
+                    let ttl = match ttl_str.parse::<u32>() {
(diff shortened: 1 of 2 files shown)


io/zenoh-links/zenoh-link-udp/src/multicast.rs
```diff
@@ -321,6 +321,34 @@ impl LinkManagerMulticastUdp {
         // https://docs.rs/tokio/latest/tokio/net/struct.UdpSocket.html#notes
         mcast_sock.set_nonblocking(true)?;
 
+        // If TTL is specified, add set the socket's TTL
+        if let Some(ttl_str) = config.get(UDP_MULTICAST_TTL) {
+            let ttl = match ttl_str.parse::<u32>() {
```
Mallets, line 326: IPv6 hop limit is not yet supported in the Rust library.
I've opened https://github.com/rust-lang/rust/pull/138744 to address the issue.

In the meanwhile, I suggest to simply log a warning in the case of an IPv6 socket.

```rust
        // If TTL is specified, add set the socket's TTL
        if let Some(ttl_str) = config.get(UDP_MULTICAST_TTL) {
            match &local_addr {
                IpAddr::V4(_) => {
                    let ttl = match ttl_str.parse::<u32>() {
                        Ok(ttl) => ttl,
                        Err(e) => bail!("Can not parse TTL '{}' to a u32: {}", ttl_str, e),
                    };

                    ucast_sock.set_multicast_ttl_v4(ttl).map_err(|e| {
                        zerror!("Can not set multicast TTL {} on {}: {}", ttl, mcast_addr, e)
                    })?;
                }
                IpAddr::V6(_) => {
                    tracing::warn!(
                            "UDP multicast hop limit not supported for v6 socket: {}. See https://github.com/rust-lang/rust/pull/138744.",
                            mcast_addr
                        );
                }
            }
        }
```

jmonticelli: Wasn't sure how tokio handled this under the hood, I just saw is said it "may not have any effect" and figured it was a shot in the dark to attempt to set the TTL.


Sounds great, I will make this change. Thanks!

jmonticelli: I have taken your changes, feel free to re-review that they were taken correctly.
It passed a local build for me.

Mallets (approved):

Issue #1835, opened 2025-03-17
Sending large messages with `CongestionControl::Drop`


evshary:
### Describe the bug

As we discussed in [rmw_zenoh](https://github.com/ros2/rmw_zenoh/issues/543#issuecomment-2723860300), if we send a large message (10 MB) with `CongestionControl::Drop`, the message might never be received.

I'm not sure if this is the limitation, we should always force users to use `CongestionControl::Block` when it comes to the large payload. If yes, perhaps it's worth mentioning this somewhere in the documentation.

### To reproduce

Modifying the congestion control to DROP in z_ping and z_pong. **Note that we only send 1 message here.**

Topology: (connected with WiFi)

```
     host 1                   host 2
[ping ----- zenohd] ----- [zenohd ----- pong]
```

* Host 1

```
./z_ping -c DEFAULT_RMW_ZENOH_SESSION_CONFIG.json5 -n 1 10000000
```

* Host 2

```
./z_pong -c DEFAULT_RMW_ZENOH_SESSION_CONFIG.json5
```

Here are the configurations we used for z_ping / z_pong / zenohd
https://github.com/ros2/rmw_zenoh/tree/rolling/rmw_zenoh_cpp/config

### System info

- Platform: Ubuntu 24.04
- Zenoh version: 1.2.3

evshary:
Should be solved by https://github.com/eclipse-zenoh/zenoh/pull/1861

PR #1843, opened 2025-03-19
Improve batching wait on multifragment messages

yellowhatter:
Refactor for batching wait:
- start exponential growth from the first advance (vs second previously)
- improve some edge cases behavior

### DIFF
--- a/io/zenoh-transport/src/common/pipeline.rs
+++ b/io/zenoh-transport/src/common/pipeline.rs
@@ -167,14 +167,26 @@ impl WaitTime {
     }
 
     fn advance(&mut self, instant: &mut Instant) {
+        // grow wait_time exponentially
+        self.wait_time = self.wait_time.saturating_mul(2);
+
+        // check for waiting limits
         match &mut self.max_wait_time {
-            Some(max_wait_time) => {
-                if let Some(new_max_wait_time) = max_wait_time.checked_sub(self.wait_time) {
-                    *instant += self.wait_time;
-                    *max_wait_time = new_max_wait_time;
-                    self.wait_time *= 2;
-                }
+            // if we have reached the waiting limit, we do not increase wait instant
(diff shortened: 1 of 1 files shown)


yellowhatter:
Before:
2025:03:20-15:08:17 throughput aggregated

After:
2025:03:20-14:43:51 throughput aggregated

Mallets:
What is the improvement w.r.t. https://github.com/eclipse-zenoh/zenoh/issues/1835?

yellowhatter:
> What is the improvement w.r.t. #1835?

This is more general improvement I made after reviewing batching code. It makes things better for #1835 : 
- exponential time increase starts from 1st fragment (originally it started from 2nd)
- time leftover for max_wait_before_drop is appreciated
...but doesn't solve the full problem.

For #1835 we with @YuanYuYuan  are preparing more comprehensive concept to be discussed.

Mallets (approved):

Jezza:
This PR contains uncontrolled exponential growth.

image

yellowhatter:
> Reverting this change, fixes the crash.
> 
> It was only doubling the wait_time when there was a max. Either we max out the wait_time or we drop the constant doubling.

What exactly is panicking?
The `wait_time` doubling doesn't make sense as it is not applied on `instant` when limit is reached

Jezza:
> What exactly is panicking?

`*instant += self.wait_time;` is the code that's overflowing.
Because it's non-conditionally doubling it, it shoots through the roof, and overflows.

I'm not saying that previous code was perfect, just that it only doubled it in some cases.

image

Jezza:
Throwing something like:
```
self.wait_time = self.wait_time.saturating_mul(2)
            .min(Duration::from_secs(60 * 60));
```
at the start, solves the issue.

Anything needing to wait longer than an hour is probably fine waking up a little bit.

PR #1844, opened 2025-03-19
Add log for better trace

evshary:
The PR will add more log for better analysis on the system like rmw_zenoh.

### DIFF
--- a/zenoh/src/net/routing/dispatcher/queries.rs
+++ b/zenoh/src/net/routing/dispatcher/queries.rs
@@ -591,6 +591,15 @@ pub(crate) fn route_send_response(
 
     match face.pending_queries.get(&qid) {
         Some((query, _)) => {
+            tracing::debug!(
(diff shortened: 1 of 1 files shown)


OlivierHecart:
It would be good if when the first message is dropped by the downsampling interceptor a message like this is traced at info level:
```
Some messages have been dropped by the downsampling interceptor. Enable trace level tracing for more details.
```
This can be achieved by a static (atomic)bool.

evshary:
> It would be good if when the first message is dropped by the downsampling interceptor a message like this is traced at info level:
> 
> ```
> Some messages have been dropped by the downsampling interceptor. Enable trace level tracing for more details.
> ```
> 
> This can be achieved by a static (atomic)bool.

Sounds good. I've added that. Should we add a trace log for every downsampled data? Not sure if it generates too many messages or not.

OlivierHecart (approved):

PR #1846, opened 2025-03-20
Pin zenoh internal dependencies

JEnoch:
In its workspace [`Cargo.toml`](https://github.com/eclipse-zenoh/zenoh/blob/16ba77c83d292e3108ad65bc2f39c5c74778d434/Cargo.toml#L197C1-L229C1), Zenoh depends on a non-pinned version of its "internal" crates.

Those "internal" crates have been made to accelerate the build time. They are released on crates.io but should not be used as direct dependency by any other project that zenoh, as we don't guarantee that their API will remain unchanged in any version, including patch updates.

A problem that could occur is if a project depends on a pinned version of `zenoh`, and we release a patch version of Zenoh with some changes in the APIs of those internal crates. Cargo will automatically bump the internal crates, leading to compile errors with the pinned `zenoh` crate.
Thus we have to pin the internal dependencies of Zenoh.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -99,7 +99,7 @@ derive-new = "0.7.0"
 tracing-subscriber = { version = "0.3", features = ["json", "env-filter"] }
 event-listener = "5.3.1"
 flume = "0.11"
-form_urlencoded = "1.2.1"
+form_urlencoded = "=1.2.1"
 futures = "0.3.30"
 futures-util = { version = "0.3.30", default-features = false } # Default features are disabled due to some crates' requirements
 git-version = "0.3.9"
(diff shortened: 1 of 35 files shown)


Cargo.toml
```diff
@@ -99,7 +99,7 @@ derive-new = "0.7.0"
 tracing-subscriber = { version = "0.3", features = ["json", "env-filter"] }
 event-listener = "5.3.1"
 flume = "0.11"
-form_urlencoded = "1.2.1"
+form_urlencoded = "=1.2.1"
```
Mallets, line 102: This is not a zenoh internal dependency.

JEnoch: Oups... blind search/replace 😅

PR #1849, opened 2025-03-21
Cache invalidation

DenisBiryukov91:
Support for interceptors cache update and invalidation.

### DIFF
--- a/commons/zenoh-sync/src/cache.rs
+++ b/commons/zenoh-sync/src/cache.rs
@@ -52,14 +52,21 @@ impl<T> Cache<T> {
     }
 
     /// Tries to retrive value for the specified version.
+    /// Returns a result either containing a cached value, or an f (which is guaranteed to be not invoked by function call in this case).
     /// If requested version corresponds to the value currently stored in cache - the value is returned.
     /// If requested version is older None will be returned.
-    /// If requested version is newer, the new value will be computed and stored by calling f, and then returned.
-    pub fn value(&self, version: usize, f: impl Fn() -> T) -> Option<CacheValueType<T>> {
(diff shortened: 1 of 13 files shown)


commons/zenoh-sync/src/cache.rs
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    /// Tries to retrieve value for the specified version.
+    /// Returns a result either containing a cached value, or an f (which is guaranteed to be not invoked by function call in this case).
+    /// If requested version corresponds to the value currently stored in cache - the value is returned.
+    /// If requested version is older None will be returned.
+    /// If requested version is newer, the new value will be computed and stored by calling f, and then returned,
+    /// unless the value is being currently updated - in this case None will be returned.
+    /// If None is returned it is guaranteed that f was not called.
+    pub fn value(
```
Mallets, line 61: I believe this implementation is not resistant to concurrent updates and the ABA problem.
Is it the desired behaviour?

DenisBiryukov91: Well, it is supposed to be resistant to concurrent updates: i.e. no concurrent updates are possible due to atomic is_updating.

DenisBiryukov91: The user always requests a value for **specific version** of data, and not for the most up-to-date one. So he can calculate the value for the requested version by calling `f` he provides in the `.value` call, since it is returned if no value corresponding to the requested version was found in cache and update is not currently possible (either the version is too old, or there is another update in progress).

Mallets: Thanks, the use case is more clear now.

PR #1850, opened 2025-03-21
Add link protocols subject to interceptors

oteffahi:
Link protocols can be used to identify transports that should have interceptor rules applied to them.
For example: 
- Downsampling can be configured to limit throughput to 10Hz on all transports that have a TCP connection.
- Access Control can be configured to deny confidential messages from passing on transports with clear-text communication (i.e not TLS and QUIC).

The configuration field is named `link_protocols`. Possible values are all of Zenoh's currently supported link protocols: `tcp`, `udp`, `tls`, `quic`, `ws`, `serial`, `unixsock-stream`, `unixpipe`, `vsock`.


oteffahi:
Tracked in Jira as [ZEN-530]

PR #1851, opened 2025-03-21
Maintain keyexpr type in interceptors

fuzzypixelz:
When profiling subscribers in high throughput scenarios with small messages, I noticed that a significant amount of time is spent checking that cached keyexprs are indeed valid keyexprs.

With this patch, the `keyexpr::new` call disappears from the flamegraph.

cache-typed-keyexpr

### DIFF
--- a/zenoh/src/net/routing/dispatcher/face.rs
+++ b/zenoh/src/net/routing/dispatcher/face.rs
@@ -22,6 +22,7 @@ use std::{
 
 use arc_swap::ArcSwap;
 use tokio_util::sync::CancellationToken;
+use zenoh_keyexpr::keyexpr;
 use zenoh_protocol::{
     core::{ExprId, Reliability, WhatAmI, WireExpr, ZenohIdProto},
     network::{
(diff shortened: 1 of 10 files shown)


DenisBiryukov91 (approved):

oteffahi (approved):

wyfo (approved):

fuzzypixelz:
Now that there is no more key-expression in the cache, this performance improvements of this pull request is null. 

Instead, I simply make sure to use `keyexpr` instead of `str` in order not to loose type information. I also removed all `str::is_empty` calls on key-expressions because they always return `false` by construction.

zenoh/src/net/routing/dispatcher/face.rs
```diff
@@ -157,8 +155,8 @@ impl FaceState {
             .map(|itor| itor.load())
             .and_then(|is| is.is_empty().not().then_some(is))
         {
-            if let Ok(expr) = KeyExpr::try_from(res.expr().to_string()) {
-                let cache = interceptor.compute_keyexpr_cache(&expr);
+            if let Ok(expr) = keyexpr::new(res.expr()) {
```
DenisBiryukov91, line 158: @OlivierHecart  btw, should not res.expr() always return a valid ke string ? I.e. could we use from_str_unchecked?

fuzzypixelz: I honestly can't say for sure if the keyexpr has been validated at this stage or not. Technically if `res` is the root then it will be empty and thus a non-valid keyexpr. 

@OlivierHecart?

OlivierHecart: Right all Resources in the tree should resolve to a vlid keyexpr except the root.

DenisBiryukov91 (approved):

PR #1858, opened 2025-03-26
Fixes #1820

### LINKED ISSUE #1820: Clarification on value delimiter for key expression parameters
### Describe the bug

There appears to be a discrepancy between the docs and the implementation regarding the separation char for param name/value pairs used in key expressions.

The docs specify `&`.

From the [source code docs](https://github.com/eclipse-zenoh/zenoh/blob/41ac4a7072036d786854f2be12cc3bb58098ec74/zenoh/src/api/selector.rs#L38):

> When in string form, selectors look a lot like a URI, with similar semantics:
> ...
> parameters are separated by `&`

This is consistent with the referenced description in the [roadmap repo](https://github.com/eclipse-zenoh/roadmap/tree/main/rfcs/ALL/Selectors#the-parameters).

However, the [implementation](https://github.com/eclipse-zenoh/zenoh/blob/41ac4a7072036d786854f2be12cc3bb58098ec74/commons/zenoh-protocol/src/core/parameters.rs#L32) uses `;`.

I can confirm that params specified as `foo?bar=123&baz=456` results in a single param `bar` of `123&baz=456`.

`foo?bar=123;baz=456` yields the two params `foo` and `bar`.

I didn't see any user facing examples that call attention to this issue one way or another.

### To reproduce

See description.

### System info

Applies to any system.

Mallets:
Fix selector doc.

### DIFF
--- a/zenoh/src/api/selector.rs
+++ b/zenoh/src/api/selector.rs
@@ -35,7 +35,7 @@ use crate::api::{key_expr::KeyExpr, queryable::Query};
 /// When in string form, selectors look a lot like a URI, with similar semantics:
 /// - the `key_expr` before the first `?` must be a valid key expression.
 /// - the `parameters` after the first `?` should be encoded like the query section of a URL:
-///     - parameters are separated by `&`,
+///     - parameters are separated by `;`,
 ///     - the parameter name and value are separated by the first `=`,
 ///     - in the absence of `=`, the parameter value is considered to be the empty string,
 ///     - both name and value should use percent-encoding to escape characters,


yellowhatter:
Probably we need to fix implementation instead of doc? Just because `&` as "and" selector allows also adding `|` for "or" selector in the future

yellowhatter:
There are also a lot of places like [rfc](https://github.com/eclipse-zenoh/roadmap/tree/main/rfcs/ALL/Selectors) where we use `&` symbol. And it is intended to be the same as URL encoding manner....

yellowhatter (changes requested):
I definitely vote for fixing impl rather than doc

Mallets:
The docs refer to the pre `1.0.0` implementation. It was then refactored and changed in `1.0.0`.

yellowhatter (approved):
Agreed, the rfc I related to is out-of-date

PR #1859, opened 2025-03-26
Make sure that UDP transports remain open when UDP listener is closed

OlivierHecart:


io/zenoh-transport/tests/unicast_transport.rs
```diff
@@ -559,10 +559,15 @@ async fn run_single(
     .await;
     println!("Transports kept working after closing the router listeners...");
 
-    // Open transport against closed endpoints -> This should fail
+    // Open transport against closed endpoints -> This should fail or timeout
     for e in client_endpoints.iter() {
-        let result = ztimeout!(client_manager.open_transport_unicast(e.clone()));
-        assert!(result.is_err());
+        // Short timeout for connectionless protocols like udp and quick
+        let x = tokio::time::timeout(
```
Mallets, line 565: If it should fail and we are expecting a different timeout, I rather change the timeout in the config.
The expected error should be clear and not ambiguous.

PR #1866, opened 2025-03-31
Make SHM segment cleanup routine more robust

yellowhatter:


yellowhatter:
Previously routine relied on `Segment::open` and that was not good as open might fail (if, for example, cleaning-up zero-sized segment) and cleanup will not happen

commons/zenoh-shm/src/shm/mod.rs
```diff
@@ -63,10 +63,9 @@ impl<ID: SegmentID> Segment<ID> {
         Ok(Self { inner })
     }
```
YuanYuYuan, line 65: Use `id` with `allow(unused)` to follow Rust's convention.

commons/zenoh-shm/src/shm/unix.rs
```diff
@@ -213,6 +258,17 @@ impl<ID: SegmentID> SegmentImpl<ID> {
...
     }
+
+    fn cleanup_segment(id: ID) {
+        let id = Self::id_str(id);
+        tracing::trace!("shm_unlink(name={})", id);
+        let _ = shm_unlink(id.as_str());
+        #[cfg(shm_external_lockfile)]
+        {
+            let lockpath = std::env::temp_dir().join(id);
+            let _ = std::fs::remove_file(lockpath);
```
YuanYuYuan, line 269: The same. Should we do the error handling?

yellowhatter:
> I've left some comments. And I have a question irrelevant to this PR. Why we use `fn is_dangling_segment(id: ID)` instead of `fn is_dangling_segment(&self, id: ID)`? (and the other similarly)

Because this code is sometimes run as part of cleanup routine, that operates not on Segment instances that can be created by `create()` or `open()` but just on ID's.

PR #1867, opened 2025-04-01
Fix memory leak in shm segment cleanup

yellowhatter:
Based on #1866

### DIFF
--- a/commons/zenoh-shm/src/cleanup.rs
+++ b/commons/zenoh-shm/src/cleanup.rs
@@ -45,15 +45,14 @@ impl Cleanup {
 
     pub(crate) fn unregister_cleanup<ID: crate::shm::SegmentID>(&self, id: ID) {
         let mut lock = zlock!(self.cleanups);
-        let _ = lock.remove(&id.into());
+        lock.remove(&id.into());
     }
 
     fn cleanup(&self) {
-        let mut ids: HashSet<u64> = Default::default();
-        {
+        let ids = {
             let mut lock = zlock!(self.cleanups);
-            std::mem::swap(lock.deref_mut(), &mut ids);
-        }
+            std::mem::take(lock.deref_mut())
+        };
 
         for id in ids {
             Segment::ensure_not_persistent(id);
--- a/commons/zenoh-shm/src/cleanup.rs
+++ b/commons/zenoh-shm/src/cleanup.rs
@@ -12,10 +12,12 @@
 //   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
 //
 
-use crossbeam_queue::SegQueue;
-use static_init::dynamic;
+use std::{collections::HashSet, ops::DerefMut, sync::Mutex};
 
-use crate::posix_shm::cleanup::cleanup_orphaned_segments;
+use static_init::dynamic;
+use zenoh_core::zlock;
+
+use crate::{posix_shm::cleanup::cleanup_orphaned_segments, shm::Segment};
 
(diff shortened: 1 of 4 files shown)


yellowhatter:
Fixes https://github.com/eclipse-zenoh/zenoh-c/issues/944

To be merged after https://github.com/eclipse-zenoh/zenoh/pull/1866

commons/zenoh-shm/src/cleanup.rs
```diff
@@ -36,15 +38,25 @@ impl Cleanup {
...
-    pub(crate) fn register_cleanup(&self, cleanup_fn: Box<dyn FnOnce() + Send>) {
-        self.cleanups.push(Some(cleanup_fn));
+    pub(crate) fn register_cleanup<ID: crate::shm::SegmentID>(&self, id: ID) {
+        let mut lock = zlock!(self.cleanups);
+        lock.insert(id.into());
+    }
+
+    pub(crate) fn unregister_cleanup<ID: crate::shm::SegmentID>(&self, id: ID) {
+        let mut lock = zlock!(self.cleanups);
+        let _ = lock.remove(&id.into());
```
YuanYuYuan, line 48: It's nice to have the error handling.

commons/zenoh-shm/src/cleanup.rs
```diff
@@ -36,15 +38,25 @@ impl Cleanup {
...
     }
 
     fn cleanup(&self) {
-        while let Some(cleanup) = self.cleanups.pop() {
-            if let Some(f) = cleanup {
-                f();
-            }
+        let mut ids: HashSet<u64> = Default::default();
```
YuanYuYuan, line 52: Rust is ergonomic :)

```rust
let ids = {
    let mut lock = zlock!(self.cleanups);
    std::mem::swap(lock.deref_mut())
};
```

commons/zenoh-shm/src/posix_shm/segment.rs
```diff
@@ -30,6 +30,17 @@ where
     shmem: shm::Segment<ID>,
 }
 
+impl<ID> Drop for Segment<ID>
+where
+    rand::distributions::Standard: rand::distributions::Distribution<ID>,
+    ID: shm::SegmentID,
+{
+    fn drop(&mut self) {
+        // Unregister unnecessary cleanup routine as Segment will be unlinked here
```
YuanYuYuan: will be unlinked or has been unlinked? IIRC, the member field has been dropped.

YuanYuYuan (changes requested):

commons/zenoh-shm/src/posix_shm/segment.rs
yellowhatter: the member fields are alive within `fn drop(...)` and will be dropped right after `drop(...)` finishes. So yes, "will be unlinked"

YuanYuYuan: Oh, yes. I think I misunderstood somewhere.

commons/zenoh-shm/src/cleanup.rs
yellowhatter: I can't do that as CLEANUP routines are executed in no-threadlocal environment  and logger panics because of this

commons/zenoh-shm/src/cleanup.rs
yellowhatter: Nice! This is just an evidence of my C++ past  :)

PR #1870, opened 2025-04-02
Pass messages by reference instead of by value

fuzzypixelz:
Supersedes #1828. Patch is by @wyfo; I rebased against main.


wyfo:
#1828 has been updated.

fuzzypixelz:
> #1828 has been updated

OK.

PR #1871, opened 2025-04-02
wait for callbacks before undeclaration return

wyfo:
supersede #1837

### DIFF
--- a/zenoh/src/api/handlers/callback.rs
+++ b/zenoh/src/api/handlers/callback.rs
@@ -14,7 +14,15 @@
 
 //! Callback handler trait.
 
-use std::sync::Arc;
+use std::{
+    ops::Deref,
+    sync::{
+        atomic::{AtomicUsize, Ordering},
+        Arc,
+    },
+};
+
+use tokio::sync::Notify;
 
 use crate::api::handlers::IntoHandler;
 
@@ -124,5 +132,62 @@ where
 
     fn into_handler(self) -> (Callback<Event>, Self::Handler) {
         (move |evt| (self.callback)(evt), ()).into_handler()
+    }
+}
+
+struct CallbackTracker {
+    counter: AtomicUsize,
+    notify: Notify,
+}
+
+pub(crate) struct TrackedCallback<T> {
+    callback: Callback<T>,
+    tracker: Arc<CallbackTracker>,
+}
+
+impl<T> TrackedCallback<T> {
+    pub(crate) fn new(callback: Callback<T>) -> Self {
+        let tracker = Arc::new(CallbackTracker {
+            counter: AtomicUsize::new(0),
+            notify: Notify::new(),
+        });
+        Self { callback, tracker }
+    }
+
+    pub(crate) fn get_callback(&self) -> CallbackGuard<T> {
+        self.tracker.counter.fetch_add(1, Ordering::Relaxed);
+        CallbackGuard {
+            callback: self.callback.clone(),
+            tracker: self.tracker.clone(),
+        }
+    }
+
+    pub(crate) async fn wait_callbacks(&self) {
+        let notified = self.tracker.notify.notified();
+        if self.tracker.counter.load(Ordering::Relaxed) != 0 {
+            notified.await;
+        }
+    }
+}
+
+#[derive(Clone)]
+pub(crate) struct CallbackGuard<T> {
+    callback: Callback<T>,
+    tracker: Arc<CallbackTracker>,
+}
+
+impl<T> Drop for CallbackGuard<T> {
+    fn drop(&mut self) {
+        if self.tracker.counter.fetch_sub(1, Ordering::Relaxed) == 1 {
+            self.tracker.notify.notify_waiters();
+        }
+    }
+}
+
+impl<T> Deref for CallbackGuard<T> {
+    type Target = Callback<T>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.callback
     }
 }
--- a/zenoh/src/api/handlers/callback.rs
+++ b/zenoh/src/api/handlers/callback.rs
@@ -14,16 +14,7 @@
 
 //! Callback handler trait.
 
-use std::{
-    ops::Deref,
-    sync::{
-        atomic::{AtomicUsize, Ordering},
-        Arc,
-    },
-};
-
-use futures::executor::block_on;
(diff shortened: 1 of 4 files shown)


zenoh/src/api/handlers/callback.rs
```diff
@@ -126,3 +134,60 @@ where
...
+
+    pub(crate) fn get_callback(&self) -> CallbackGuard<T> {
+        self.tracker.counter.fetch_add(1, Ordering::Relaxed);
+        CallbackGuard {
+            callback: self.callback.clone(),
+            tracker: self.tracker.clone(),
+        }
+    }
+
+    pub(crate) async fn wait_callbacks(&self) {
```
DenisBiryukov91, line 165: I think we need to put wait_callback inside TrackedCallback drop function. Otherwise it is very difficult to make it future proof. If I'm not wrong, even now if we declare background subscribers, the wait will not be called upon session destruction (which leads to the same problem -> usage of resources after destructor call) and putting it manually (and similar stuff for all listeners and queryables) everywhere is a pain.

zenoh/src/api/handlers/callback.rs
```diff
@@ -126,3 +134,60 @@ where
...
+        let notified = self.tracker.notify.notified();
+        if self.tracker.counter.load(Ordering::Relaxed) != 0 {
+            notified.await;
+        }
+    }
+}
+
+#[derive(Clone)]
```
DenisBiryukov91, line 173: is clone really needed here ? Besides its default implementation is not compatible with TrackedCallback.

zenoh/src/api/handlers/callback.rs
```diff
@@ -126,3 +134,60 @@ where
...
+        self.tracker.counter.fetch_add(1, Ordering::Relaxed);
+        CallbackGuard {
+            callback: self.callback.clone(),
+            tracker: self.tracker.clone(),
+        }
+    }
+
+    pub(crate) async fn wait_callbacks(&self) {
+        let notified = self.tracker.notify.notified();
+        if self.tracker.counter.load(Ordering::Relaxed) != 0 {
```
DenisBiryukov91, line 167: what if call get callback from another thread, right after load here ?

wyfo: If you call `get_callback` from another thread, you are doing something wrong, but it still work as expected. By definition, `wait_callbacks` can return several time, as it takes a `&self`.

zenoh/src/api/handlers/callback.rs
wyfo: If we do so, then we just admit once for all that the async API is a joke. I'm ok with that.

zenoh/src/api/handlers/callback.rs
wyfo: No, I forgot to remove it.

wyfo:
@DenisBiryukov91 Is it ok for you now? In this case, I will use `TrackedCallback` everywhere and add a bit of comments.
By the way, if you have a better naming proposal, I'm not really satisfied.

zenoh/src/api/handlers/callback.rs
DenisBiryukov91: Except it is not, it will return, and will assume that there are no more callbacks.
I agree that given session.state lock, this should not happen for the case of subscriber, but unfortunately there is nothing that can ensure this will always be the case and we did not forget to acquire a lock prior to calling get_callback (Please see AdvancedPubSub code for reference).

wyfo: No, you are assuming that there will be no more callback, but that's not the contract of this function. To do what you want, it needs an extra synchronization, which would indeed be the session lock.

wyfo: Anyway, we do not care about the asynchronous API, so it's in the destructor now.

DenisBiryukov91:
> @DenisBiryukov91 Is it ok for you now? In this case, I will use `TrackedCallback` everywhere and add a bit of comments. By the way, if you have a better naming proposal, I'm not really satisfied.

I'm not really see how it is better than what was done in #1837. I agree that it looks more neat but have exactly the same blocking behavior in drop. And as I pointed out it is not automatically concurrency-safe. 

If the idea is to only hide sleep, instead of tracking arcs we could probably use a a pair of flume senders/receivers which are cheap to clone, and we will simply wait on receiver in the drop until all senders die - thus hiding a sleep loop behind a more sophisticated flume implementation.

wyfo:
Be aware that putting the wait in `Drop` is not as obviously synchronized as you seem to say, because callbacks are often put into `Arc`, which drop timing is not always obvious.

DenisBiryukov91:
Also, regarding performance, does notify_waiters have any measurable overhead (that was the reason, why I did not go for mutex + condvar) ?

wyfo:
> I do not really see how it is better than what was done in https://github.com/eclipse-zenoh/zenoh/pull/1837.

Euh... no sleep loop?

And I don't think using flume stuff is as simple as a simple `Notify`, because it forces you to keep both a `Sender` and a `Receiver`, and you need to put the `Sender` into a `Mutex<Option<>>`, because you will need to drop the sender that you are cloning each time you instantiate a guard.

wyfo:
> Also, regarding performance, does notify_waiters have any measurable overhead (that was the reason, why I did not go for mutex + condvar) ?

1 seqcst load + 1 seqcst FAA (if there is no waiter)
I think it should be ok regarding zenoh performance.

DenisBiryukov91:
> And I don't think using flume stuff is as simple as a simple `Notify`, because it forces you to keep both a `Sender` and a `Receiver`, and you need to put the `Sender` into a `Mutex<Option<>>`, because you will need to drop the sender that you are cloning each time you instantiate a guard.

I'm not sure I understand why, I do not even need to ever use sender. I'll just wait till all sender clones die.
So the only time I would need to call drop on it, is inside TrackedCallback drop, prior to calling receiver.recv().

wyfo:
> I'm not sure I understand why, I do not even need to ever use sender. I'll just wait till all sender clones die. So the only time I would need to call drop on it, is inside TrackedCallback drop, prior to calling receiver.recv().

Indeed, I was confusing with the asynchronous stuff I did, sorry.

wyfo:
Is it ok with flume?

PR #1878, opened 2025-04-03
remove Display implementation for TransportMessage

wyfo:
Display implementation was used by dissector, but it has been broken with lazy frame deserialization.
The implementation should be written directly where it's used, e.g. in dissector, where deserialization can be done.


OlivierHecart:
The display is used in various logs. Ket's keep it.

PR #1879, opened 2025-04-03
Cache link selection in unicast transport

yellowhatter:
Improves TX perf  (~100K msgs\sec)

### DIFF
--- a/io/zenoh-transport/src/unicast/universal/transport.rs
+++ b/io/zenoh-transport/src/unicast/universal/transport.rs
@@ -86,8 +86,7 @@ impl TransportLinks {
         zwrite!(self.links)
     }
 
-    #[allow(private_interfaces)]
-    pub(crate) fn read(&self) -> RwLockReadGuard<'_, Box<[TransportLinkUnicastUniversal]>> {
+    pub(super) fn read(&self) -> RwLockReadGuard<'_, Box<[TransportLinkUnicastUniversal]>> {
         zread!(self.links)
     }
 }
--- a/io/zenoh-transport/src/unicast/universal/transport.rs
+++ b/io/zenoh-transport/src/unicast/universal/transport.rs
@@ -13,7 +13,10 @@
 //
 use std::{
     fmt::DebugStruct,
-    sync::{Arc, RwLock},
+    sync::{
+        atomic::{AtomicUsize, Ordering},
+        Arc, RwLock, RwLockReadGuard, RwLockWriteGuard,
+    },
     time::Duration,
 };
 
@@ -22,7 +25,7 @@ use tokio::sync::{Mutex as AsyncMutex, MutexGuard as AsyncMutexGuard};
 use zenoh_core::{zasynclock, zcondfeat, zread, zwrite};
 use zenoh_link::Link;
 use zenoh_protocol::{
-    core::{Priority, WhatAmI, ZenohIdProto},
+    core::{Priority, Reliability, WhatAmI, ZenohIdProto},
     network::NetworkMessage,
     transport::{close, Close, PrioritySn, TransportMessage, TransportSn},
 };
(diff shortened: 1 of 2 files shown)


io/zenoh-transport/src/unicast/universal/transport.rs
```diff
@@ -42,6 +45,51 @@ use crate::{
...
 
+/// Transport link storage with link selection caching functionality
+pub(crate) struct TransportLinks {
+    links: RwLock<Box<[TransportLinkUnicastUniversal]>>,
+    index_cache: [AtomicUsize; Priority::NUM * 2],
+}
+
+impl TransportLinks {
+    fn new(links: RwLock<Box<[TransportLinkUnicastUniversal]>>) -> Self {
+        let index_cache = [const { AtomicUsize::new(usize::MAX) }; 16];
```
wyfo, line 56: `const` expressions are 1.79, it makes the CI fails

wyfo, line 56:
```suggestion
        let index_cache = [const { AtomicUsize::new(usize::MAX) }; Priority::NUM * 2];
```

io/zenoh-transport/src/unicast/universal/transport.rs
```diff
@@ -42,6 +45,51 @@ use crate::{
...
+    fn write(&self) -> RwLockWriteGuard<'_, Box<[TransportLinkUnicastUniversal]>> {
+        // on write access to the links we clear link cache
+        for atomic in &self.index_cache {
+            atomic.store(usize::MAX, Ordering::Relaxed);
+        }
+        zwrite!(self.links)
+    }
+
+    #[allow(private_interfaces)]
+    pub(crate) fn read(&self) -> RwLockReadGuard<'_, Box<[TransportLinkUnicastUniversal]>> {
```
wyfo, line 88:
```suggestion
    pub(super) fn read(&self) -> RwLockReadGuard<'_, Box<[TransportLinkUnicastUniversal]>> {
```

yellowhatter: That was made because of:
```
warning: type `TransportLinkUnicastUniversal` is more private than the item `TransportLinks::read`
  --> io/zenoh-transport/src/unicast/universal/transport.rs:90:5
   |
90 |     pub(crate) fn read(&self) -> RwLockReadGuard<'_, Box<[TransportLinkUnicastUniversal]>> {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ method `TransportLinks::read` is reachable at visibility `pub(crate)`
   |
note: but type `TransportLinkUnicastUniversal` is only usable at visibility `pub(universal)`
  --> io/zenoh-transport/src/unicast/universal/link.rs:39:1
   |
39 | pub(super) struct TransportLinkUnicastUniversal {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   = note: `#[warn(private_interfaces)]` on by default

```

io/zenoh-transport/src/unicast/universal/transport.rs
yellowhatter: fixed

yellowhatter: changed

wyfo:
By caching only the indices, you sill need to acquire the lock. I would do it like this instead https://github.com/eclipse-zenoh/zenoh/commit/baced61f071bca76f6d2bd466e4c8a96031caaa4

yellowhatter:
> By caching only the indices, you sill need to acquire the lock. I would do it like this instead [baced61](https://github.com/eclipse-zenoh/zenoh/commit/baced61f071bca76f6d2bd466e4c8a96031caaa4)

Yes! Good idea! And this can also be joined with [this](https://github.com/eclipse-zenoh/zenoh/pull/1880) as it already returns shallow copy

wyfo:
Actually, #1880 wouldn't need to be reused here, as the cache I made already use an outer `Arc` to solve the same issue you solved in #1880.

OlivierHecart:
Superseded by https://github.com/eclipse-zenoh/zenoh/pull/1888

PR #1880, opened 2025-04-03
Optimize transmission pipeline operation

yellowhatter:
Improves TX perf (~300k msps)

### DIFF
--- a/io/zenoh-transport/src/common/pipeline.rs
+++ b/io/zenoh-transport/src/common/pipeline.rs
@@ -13,7 +13,7 @@
 //
 use std::{
     fmt,
-    ops::Add,
+    ops::{Add, Deref},
     sync::{
         atomic::{AtomicBool, AtomicU32, AtomicU8, Ordering},
         Arc, Mutex, MutexGuard,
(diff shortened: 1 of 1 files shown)


wyfo:
Numbers is not enough, you should able to explain why it does improve things. Looking quickly at the PR, I've no idea how adding an additional indirection could cause such an improvement.

yellowhatter:
> Numbers is not enough, you should able to explain why it does improve things. Looking quickly at the PR, I've no idea how adding an additional indirection could cause such an improvement.

Very simple: [at this line](https://github.com/eclipse-zenoh/zenoh/blob/76eadaf3f357a462393d8acc27d47fc27d17ce3e/io/zenoh-transport/src/unicast/universal/tx.rs#L104) we clone pipeline on every message. These shallow clones and then destructions are very visible on the flamegraph and take around 7% of time (don't have one - already removed, sorry). Adding indirection reduces the number of actual clones and destructions dramatically (as we just clone only one `Arc` instead of 2 `Arc` and 3 `Duration`).

wyfo:
> Very simple: [at this line](https://github.com/eclipse-zenoh/zenoh/blob/76eadaf3f357a462393d8acc27d47fc27d17ce3e/io/zenoh-transport/src/unicast/universal/tx.rs#L104) we clone pipeline on every message.

Sorry, it was late and I was coming back from an afterwork 😅

PR #1881, opened 2025-04-03
Bug/fix batching wait regression

### LINKED ISSUE #1889: Wait timeouts not fully respected by Congestion Control
### Describe the bug

[Here](https://github.com/eclipse-zenoh/zenoh/blob/94e917948914af093eae340ee52118b19fb1671f/io/zenoh-transport/src/common/pipeline.rs#L172) the maximum wait time is intended to limit actual waits. However, the `max_wait_time` leftover in this line is not considered if it is less then `wait_time` which can be quite big because it grows exponentially. Because of this algo, zenoh doesn't fully respect the `max_wait_before_drop` config parameter and can Drop data earlier than configured.

### To reproduce

-

### System info

any

yellowhatter:
Fixes regression #1843 
Adds tests for that regression :)

[closes #1889 ]


yellowhatter:
[closes #1889 ]

OlivierHecart (approved):

PR #1888, opened 2025-04-07
cache pipeline selection

wyfo:
supersede #1880 and #1879

### DIFF
--- a/io/zenoh-transport/src/unicast/universal/transport.rs
+++ b/io/zenoh-transport/src/unicast/universal/transport.rs
@@ -23,7 +23,7 @@ use tokio::sync::{Mutex as AsyncMutex, MutexGuard as AsyncMutexGuard};
 use zenoh_core::{zasynclock, zcondfeat, zread, zwrite};
 use zenoh_link::Link;
 use zenoh_protocol::{
-    core::{Locator, Priority, Reliability, WhatAmI, ZenohIdProto},
+    core::{Priority, Reliability, WhatAmI, ZenohIdProto},
     network::NetworkMessageMut,
     transport::{close, Close, PrioritySn, TransportMessage, TransportSn},
 };
@@ -54,7 +54,7 @@ use crate::{
 #[derive(Default)]
 pub(crate) struct TransportLinks {
     links: RwLock<Box<[TransportLinkUnicastUniversal]>>,
-    cache: ArcSwap<[[Option<Arc<ScheduledLink>>; Priority::NUM]; 2]>,
+    cache: ArcSwap<[[Option<Arc<TransmissionPipelineProducer>>; Priority::NUM]; 2]>,
 }
 
 impl TransportLinks {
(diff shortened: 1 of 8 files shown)


io/zenoh-transport/src/unicast/universal/transport.rs
```diff
@@ -41,10 +45,54 @@ use crate::{
...
+        select: impl FnOnce(&[TransportLinkUnicastUniversal]) -> Option<usize>,
+    ) -> Option<Arc<ScheduledLink>> {
+        if let Some(link) = self.cache.load()[reliability as usize][priority as usize].clone() {
+            return Some(link);
+        }
+        let guard = self.read();
+        let transport_link = &guard[select(&guard)?];
+        let link = Arc::new(ScheduledLink {
+            pipeline: transport_link.pipeline.clone(),
+            dst: transport_link.link.link.get_dst().clone(),
```
wyfo, line 82: @yellowhatter doesn't like this string copy which is only used for logging

yellowhatter: Yes, I don't like it because removing it we can:
- lower memory usage in large-scale scenario
- get rid of extra type `ScheduledLink` in the codebase

yellowhatter: Our 200-nodes scaling scenario will have 200*200*(19+8+8)=1.3Mb extra memory used. Not that much - but just for logging???

wyfo: I agree with all your points. However, I cannot take the decision to remove this log. @OlivierHecart What do you think?

wyfo: Another solution would be to acquire the `RWLock` just for the log, I would be fine with that.

yellowhatter (approved):
Very good!
Let's wait for @OlivierHecart word on logging and anyway it is OK to merge

fuzzypixelz (approved):

io/zenoh-transport/src/unicast/universal/transport.rs
OlivierHecart: I'd either:
- Simply remove this info in the trace.
- Add a link id in the trace that user could map to link details through a link creation trace.

wyfo: Solution 1 accepted.

yellowhatter (approved):
Great!

yellowhatter:
@fuzzypixelz

fuzzypixelz (approved):

fuzzypixelz:
@wyfo @yellowhatter Can you provide any performance data?

yellowhatter:
> @wyfo @yellowhatter Can you provide any performance data?

on my machine for 8-byte payload this improves perf 4.9 -> 5.3 M messages per second.

OlivierHecart (approved):

PR #1894, opened 2025-04-08
Lowlatency async finalize

yellowhatter:
Async finalize lowlatency transport when triggered from link to avoid deadlock (the same approach as it is done in unicast transport)


yellowhatter:
According to #1893 , this PR  fixes #1859

Issue #1886, opened 2025-04-04
Reconnection can fail with `Received router declaration with unknown routing context id`


Hugal31:
#### Description

When disconnecting and reconnecting routers, if one router is busy and the timing right, it can fail the reconnection and display
```
ERROR  rx-0 ThreadId(06) zenoh::net::routing::hat::router: Received router declaration with unknown routing context id 0
```

#### Expected

The reconnection goes well and messages are routed between the two hosts.

#### My setup

![Setup with A and B computer, A has one router, two z_pub and one z_sub, B has one router and one z_sub](https://private-user-images.githubusercontent.com/9070803/429981208-c6898403-d9a1-4c0e-b96f-3a78a113e8f5.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NDM3NTYxNzYsIm5iZiI6MTc0Mzc1NTg3NiwicGF0aCI6Ii85MDcwODAzLzQyOTk4MTIwOC1jNjg5ODQwMy1kOWExLTRjMGUtYjk2Zi0zYTc4YTExM2U4ZjUucG5nP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MDQwNCUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTA0MDRUMDgzNzU2WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9YmI0NzQxZWJiMWQ0OWIzN2ExMDMwZDc1MWMxODliMWMxNDNkZTNmZDI5ZDIwNWY5MDI1NGUxNTdlMTkzNjZhNCZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.CK2UEO8Dp1LLGY3EBiwqjc59Ttihuw8k7Q_364PL1n4)

* Two computers, A and B, connected through Wi-Fi.
* Each has a router configured to connect to each others in TCP.
[...]

PR #1896, opened 2025-04-09
Do not wait on rx_task and tx_task when closing the socket.

Hugal31:
When loosing the connection to a peer, if the send buffer happens to fill up before the keep alive timeout, the tx_task can be stuck in the `send_batch` operation. On my computer, it can be stuck for 15 minutes before returning with EHOSTUNREACH ("No route to host").

Now this is not that bat itself, it's just a dangling task, `TransportLinkUnicastUniversal` and `TransportUnicastUniversal`. But this make the race condition mentionned in #1886 very likely: if the peer reconnect within those 15 minutes, the reconnection is bogus.

This "fix" make `TransportLinkUnicastUniversal::close()` not wait for the rx and tx task before returning. It doesn't seem to cause any issue. I didn't do the same in `TransportUnicastLowlatency` because I am not sure it is necessary.

Now I am not really happy with this hack, but I would like to have a solution for #1886 and it's a bit too involved for me alone.

An alternative would be to watch for the cancellation token while calling `send_batch`.

### DIFF
--- a/io/zenoh-transport/src/unicast/universal/link.rs
+++ b/io/zenoh-transport/src/unicast/universal/link.rs
@@ -163,7 +163,6 @@ impl TransportLinkUnicastUniversal {
         self.tracker.close();
         self.token.cancel();
         self.pipeline.disable();
-        self.tracker.wait().await;
 
         self.link.close(None).await
     }


josephduchesne:
I'm running into this on systems using Wifi Halow I believe (out at the edge of connectivity ~1.1km in my case). Is there any way to detect it happening and force a reconnect? I'd rather not have to add watchdogs to all systems using zenoh if I can avoid it.

Hugal31:
> I'm running into this on systems using Wifi Halow I believe (out at the edge of connectivity ~1.1km in my case). Is there any way to detect it happening and force a reconnect? I'd rather not have to add watchdogs to all systems using zenoh if I can avoid it.

Not that I am aware of. And nothing would prevent the bug from happening again right after...

I'm glad (and sorry!) to hear we're not the only ones with this bug. Could you add some logs, details (or just a thumbsup) on #1886?

Hugal31:
I noticed this hack caused issues elsewhere (i.e. it made an unrelated keep_alive TX loop miss deadline for some reasons). So that's not an option.

PR #1901, opened 2025-04-10
use ahash in critical path

wyfo:


fuzzypixelz (commented):
You'll have to set `default-features = false` for `ahash` and set the std feature of `zenoh-collections` to `["ahash/std"]`.

fuzzypixelz (approved):

PR #1905, opened 2025-04-15
TDD: Track session state post-close access in tests

yellowhatter:
This PR is for discussion and adds panic in tests if Session internal state is accessed after Session is being closed. I believe that to have proper design we should make this work


wyfo:
Accessing `SessionState` after session close is perfectly valid. `Session::close` takes a shared reference, so when the session will be dropped after having been closed, it will access the state and raise an error with your PR — I guess that's not expected.

PR #1912, opened 2025-04-18
feat: add hash support for Encoding

alokpr:
Hash trait was missing for Encoding which prevented them from being used as keys of a HashMap.

### DIFF
--- a/commons/zenoh-buffers/src/zslice.rs
+++ b/commons/zenoh-buffers/src/zslice.rs
@@ -222,6 +222,12 @@ impl<Rhs: AsRef<[u8]> + ?Sized> PartialEq<Rhs> for ZSlice {
 
 impl Eq for ZSlice {}
 
+impl std::hash::Hash for ZSlice {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.as_slice().hash(state);
+    }
+}
+
 impl fmt::Display for ZSlice {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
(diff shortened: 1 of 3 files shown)


alokpr:
@YuanYuYuan I have signed Eclipse ECA

OlivierHecart:
Please fix CI failures.

alokpr:
> Please fix CI failures.

Looks like the CI is mainly failing due to formatting. I tried reformatting using "cargo fmt", but that did not change anything. Is the CI using different rules than those specified in rustfmt.toml?

alokpr:
I also do not seem to have permission to add labels to this PR

alokpr:
Could a Zenoh maintainer help merge this PR?

alokpr:
> Could you please try: `cargo fmt -- --config "unstable_features=true,imports_granularity=Crate,group_imports=StdExternalCrate"`

Done. I am curious - why not add these settings to `rustfmt.toml`?

JEnoch:
> Done. I am curious - why not add these settings to `rustfmt.toml`?

For the reason explained in https://github.com/rust-lang/rustfmt/issues/4353 wrt. unstable features.
However don't understand why those unstable features work with a stable `rustfmt` when passed in command line args... 🤔

JEnoch:
Although `no_std` is not supported by Zenoh, we still would like to eventually support it. So we're making sure that the base crates for the protocol stack support `no_std`.
I updated the PR to have `Hash` trait on `ZSlice` and `Encoding` only if `std` feature is set.

PR #1914, opened 2025-04-18
[ZEN-547] Add support for link weights

DenisBiryukov91:
* [x] Add support for link weights on routers net
* [x] Add weights to linkstate message
* [x] Add ability to specify link weights through config
* [x] Add internal api allowing to update link weights
* [x] Add support for link weights on linkstate_peers net
* [x] Expose link weights to adminspace


DEFAULT_CONFIG.json5
```diff
@@ -217,11 +217,27 @@
       /// The failover brokering only works if gossip discovery is enabled
       /// and peers are configured with gossip target "router".
       peers_failover_brokering: true,
+      /// Optional weights of the outgoing links on routers net.
+      /// For non-specified destination nodes the corresponding link weight will be set to 100,
+      /// unless the weight on the same link is provided in the config of destination node, in which case
```
OlivierHecart, line 222: Unclear. What is the case this is referring to ?

DEFAULT_CONFIG.json5
```diff
@@ -217,11 +217,27 @@
...
+//      destination: "2", weight: "200",
+//    ]
     },
     /// The routing strategy to use in peers and it's configuration.
     peer: {
       /// The routing strategy to use in peers. ("peer_to_peer" or "linkstate").
       mode: "peer_to_peer",
+      /// Optional weights of the outgoing links on peers net in linkstate mode.
+      /// For non-specified destination nodes the corresponding link weight will be set to 100,
+      /// unless the weight on the same link is provided in the config of destination node, in which case
```
OlivierHecart, line 235: Unclear. What is the case this is referring to ?

OlivierHecart:
To be coherent with our global naming. We should speak about transports and transport weights in all the visible areas.

DEFAULT_CONFIG.json5
```diff
@@ -217,11 +217,27 @@
       /// The failover brokering only works if gossip discovery is enabled
       /// and peers are configured with gossip target "router".
       peers_failover_brokering: true,
+      /// Optional weights of the outgoing links on routers net.
+      /// For non-specified destination nodes the corresponding link weight will be set to 100,
+      /// unless the weight on the same link is provided in the config of destination node, in which case
+      /// it will be used as a link weight. If both link endpoints set a link weight, the greater value will be used.
+//    link_weights: [
```
OlivierHecart, line 224: For backward compatibility, I believe we should have something like this:
```
transport_weights: {
  /// This needs to be set to the same value in all routers of the system.
  enabled: false,
  weights: [...],
}
```

DEFAULT_CONFIG.json5
```diff
@@ -217,11 +217,27 @@
...
     },
     /// The routing strategy to use in peers and it's configuration.
     peer: {
       /// The routing strategy to use in peers. ("peer_to_peer" or "linkstate").
       mode: "peer_to_peer",
+      /// Optional weights of the outgoing links on peers net in linkstate mode.
+      /// For non-specified destination nodes the corresponding link weight will be set to 100,
+      /// unless the weight on the same link is provided in the config of destination node, in which case
+      /// it will be used as a link weight. If both link endpoints set a link weight, the greater value will be used.
+//    link_weights: [
```
OlivierHecart, line 237: For backward compatibility, I believe we should have something like this:
```
transport_weights: {
  /// This needs to be set to the same value in all routers and peers of the sub-system.
  enabled: false,
  weights: [...],
}
```

DEFAULT_CONFIG.json5
DenisBiryukov91: The case when weight is not set on source node, but set on destination node.

DEFAULT_CONFIG.json5
DenisBiryukov91: fixed

PR #1927, opened 2025-04-30
expose routing trees in adminspace

wyfo:


zenoh/src/net/protocol/network.rs
```diff
@@ -1145,6 +1146,41 @@ impl Network {
...
+
+    pub(crate) fn compute_path(&self, src: ZenohIdProto, dst: ZenohIdProto) -> Option<NetworkPath> {
+        let (mut src_idx, mut dst_idx) = (None, None);
+        for idx in self.graph.node_indices() {
+            if self.graph[idx].zid == src {
+                src_idx = Some(idx);
+            }
+            if self.graph[idx].zid == dst {
+                dst_idx = Some(idx);
+            }
```
OlivierHecart, line 1158:
```suggestion
            if self.graph[idx].zid == src {
                src_idx = Some(idx);
                if dst_idx.is_some() {
                    break;
                }
            }
            if self.graph[idx].zid == dst {
                dst_idx = Some(idx);
                if src_idx.is_some() {
                    break;
                }
            }
```

zenoh/src/net/protocol/network.rs
```diff
@@ -1159,3 +1195,104 @@ pub(crate) fn shared_nodes(net1: &Network, net2: &Network) -> Vec<ZenohIdProto>
         })
         .collect()
 }
+
+#[derive(Debug, Clone, PartialEq, serde::Serialize)]
+pub(crate) struct NetworkPathNode {
+    pub(crate) zid: ZenohIdProto,
+    pub(crate) locators: Option<Vec<Locator>>,
```
OlivierHecart, line 1202: I believe this is not the place to put those informations. I'd rather have the paths be simple lists of zids and let an eventual tooling get the locators of each node if needed.

zenoh/src/net/runtime/orchestrator.rs
```diff
@@ -1253,4 +1260,18 @@ impl Runtime {
         let hat_code = tables.hat_code.clone();
         hat_code.links_info(&tables)
     }
+
+    pub(crate) fn get_router_path(&self, src: ZenohId, dst: ZenohId) -> Option<NetworkPath> {
```
OlivierHecart, line 1264: I don't see why this is in orchestrator.rs`. I'd rather have admin space directly call hat functions like it's done for `routers_linkstate_data` for coherency.

wyfo:
Superseded by #1932

PR #1937, opened 2025-05-12
support DSCP link config

### LINKED ISSUE #1924: Configuration of DiffServ DSCP class
### Describe the feature

[DiffServ](https://en.wikipedia.org/wiki/Differentiated_services) defines classes of traffic which can be configured for a socket via a DSCP value.
Zenoh could support the configuration of a DSCP class in the same way than other socket options can be configured on endpoints.
e.g.: `"tcp/192.168.0.1:7447#dscp=48"`.

Note that the way to set DSCP class differ between IPv4 and IPv6:
- **IPv4**: it can be set via [ToS](https://en.wikipedia.org/wiki/Type_of_service#DSCP_and_ECN), calling `setsockopt(sockfd, IPPROTO_IP, IP_TOS,...)`. 
  In Rust, call [`socket2::Socket::set_tos(&self, tos: u32)`](https://docs.rs/socket2/0.5.9/socket2/struct.Socket.html#method.set_tos)
- **IPv6**: it can be set via Traffic Class, calling `setsockopt(sockfd, IPPROTO_IPV6, IPV6_TCLASS,...)`.
  In Rust, it seems to be implemented only for UNIX in [`socket2::sys::unix::Socket::set_tclass_v6(&self, tclass: u32)`](https://github.com/rust-lang/socket2/blob/36dec54c30c5ca0c711709b9e1f03b8cb8d155fa/src/sys/unix.rs#L2399)

wyfo:
Resolves #1924

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5385,6 +5385,7 @@ dependencies = [
  "rustls",
  "rustls-webpki",
  "serde",
+ "socket2 0.5.7",
  "time 0.3.36",
  "tokio",
  "tokio-util",
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -30,7 +30,10 @@
   /// E.g. tcp/192.168.0.1:7447#so_sndbuf=65000;so_rcvbuf=65000
   /// For TCP, UDP, Quic and TLS links, it is possible to specify a `bind` address for the local socket:
   /// E.g. tcp/192.168.0.1:7447#bind=192.168.0.1:0
-  //  Note!: Currently it is unsupported to specify both `bind` and `iface`. 
+  ///  Note!: Currently it is unsupported to specify both `bind` and `iface`.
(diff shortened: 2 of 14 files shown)


wyfo:
@JEnoch Could you test that it gives the expected outcome? It seems to work on my laptop, but I would like a more expert review. Also, do you think I went too far in the parsing capabilities?

JEnoch:
> Where should this feature be documented? 

Currently, the only place where we document such endpoint config is the `DEFAULT_CONFIG.json5`, on `listen` and/or `connect` sections.
Please add a short description here.

JEnoch (changes requested):
Please add a minimal doc (endpoint examples) in DEFAULT_CONFIG.json5, in a similar way than `bind` or `iface` options.

gstangel:
❤️

JEnoch (approved):

PR #1944, opened 2025-05-15
fix: use 0.0.0.0 instead of [::] default address when ipv6 disabled

ninousf:


milyin:
@ninousf could you please fix formatting issue and sign eclipse agreement to finalize the PR ?

PR #1954, opened 2025-05-20
remove unstable connect/endpoints update

wyfo:


zenoh/src/api/session.rs
```diff
@@ -807,32 +807,6 @@ impl Session {
     }
 
     /// Get the current configuration of the zenoh [`Session`](Session).
-    ///
```
OlivierHecart, line 810: Shouldn't we keep at least an example of how to read the conf ?

OlivierHecart (approved):

PR #1955, opened 2025-05-20
prevent config update at runtime

wyfo:
They are not taken into account, except for plugins in adminspace.


OlivierHecart:
ZEN-582 was about making [Notifier::insert_json5](https://github.com/eclipse-zenoh/zenoh/blob/fc36c1a3c1c000f7443e4f011a67f571f735d7b0/zenoh/src/api/config.rs#L249-L250) return an error. Not only the admin space (Note: the admin space already propagate errors from `Notifier::insert_json5`. So the current proposed change should not be needed.)

PR #1959, opened 2025-05-20
Keyed handlers

fuzzypixelz:


evshary:
@fuzzypixelz I've already resolved the conflict in the branch from ZettaScaleLabs. Perhaps you can use that one directly
https://github.com/eclipse-zenoh/zenoh/compare/main...ZettaScaleLabs:zenoh:dev/keyed_handlers

evshary (approved):
I'm not sure if we should increase the test coverage, but it looks good to me anyway.
btw, the CI failure is due to out of space

PR #1963, opened 2025-05-23
chore: unpin home dependency

simpleton:
The rust toolchain version has be upgraded to 1.8.5 in (#1803).
we don't need to pin the home to 0.5.9 (#1673)


simpleton:
cc @diogomatsubara, do you have any concern on this unpin?

PR #1966, opened 2025-05-27
expose node current locators

wyfo:


zenoh/src/api/info.rs
```diff
@@ -82,4 +86,20 @@ impl SessionInfo {
...
+    /// ```
+    /// # #[tokio::main]
+    /// # async fn main() {
+    ///
+    /// let session = zenoh::open(zenoh::Config::default()).await.unwrap();
+    /// println!("{:?}", session.info().locators().await);
+    /// // print ["tcp/127.0.0.1:7447"]
+    /// # }
+    /// ```
+    pub fn locators(&self) -> impl Resolve<Vec<Locator>> + '_ {
```
OlivierHecart, line 105: Please mark as unstable

PR #1968, opened 2025-05-27
fix handling of port 0 on multicast locator

wyfo:
When binding on port 0, the OS assigns the actual port. However, the multicast locator was stored without taking this potential port change.


yellowhatter (approved):
Nice catch!

PR #1975, opened 2025-06-02
SHM API improvement: ProtocolID and others...

yellowhatter:
Draft because it requires API alignment for C and C++ (to be done after we agree on this)

- hide ProtocolID
- simplify ShmProvider builder APIs
- add minimal example for SHM
- simplify some SHM examples using simplified SHM API

### DIFF
--- a/commons/zenoh-shm/src/api/client_storage/mod.rs
+++ b/commons/zenoh-shm/src/api/client_storage/mod.rs
@@ -68,6 +68,7 @@ impl ShmClientSetBuilder {
     /// Include default clients
     #[zenoh_macros::unstable_doc]
     pub fn with_default_client_set(self) -> ShmClientStorageBuilder {
+        let client = PosixShmClient {};
         let clients = BTreeMap::from([(client.id(), Arc::new(client) as Arc<dyn ShmClient>)]);
         ShmClientStorageBuilder::new(clients)
     }
--- a/commons/zenoh-shm/src/api/provider/shm_provider.rs
+++ b/commons/zenoh-shm/src/api/provider/shm_provider.rs
@@ -22,7 +22,7 @@ use std::{
 };
 
 use async_trait::async_trait;
-use zenoh_core::{zlock, Resolvable, Wait};
+use zenoh_core::{zlock, zresult::ZResult, Resolvable, Wait};
 
 use super::{
     chunk::{AllocatedChunk, ChunkDescriptor},
--- a/commons/zenoh-shm/tests/shm_provider.rs
+++ b/commons/zenoh-shm/tests/shm_provider.rs
@@ -13,37 +13,16 @@
 //
 
 use zenoh_core::Wait;
-use zenoh_shm::api::{
-    protocol_implementations::posix::{
-        posix_shm_provider_backend::PosixShmProviderBackend, protocol_id::POSIX_PROTOCOL_ID,
-    },
-    provider::shm_provider::ShmProviderBuilder,
-};
+use zenoh_shm::api::provider::shm_provider::ShmProviderBuilder;
 
 #[test]
 fn shm_provider_create() {
-    let backend = PosixShmProviderBackend::builder()
-        .with_size(65536)
-        .unwrap()
-        .wait()
-        .unwrap();
-    let _provider = ShmProviderBuilder::builder()
-        .protocol_id::<POSIX_PROTOCOL_ID>()
-        .backend(backend)
-        .wait();
+    let _provider = ShmProviderBuilder::default_backend(65536).wait().unwrap();
 }
 
 #[test]
 fn shm_provider_alloc_and_drop_provider() {
-    let backend = PosixShmProviderBackend::builder()
-        .with_size(65536)
-        .unwrap()
-        .wait()
-        .unwrap();
-    let provider = ShmProviderBuilder::builder()
-        .protocol_id::<POSIX_PROTOCOL_ID>()
-        .backend(backend)
-        .wait();
+    let provider = ShmProviderBuilder::default_backend(65536).wait().unwrap();
 
     let mut shm_buffer = provider.alloc(1024).wait().unwrap();
 


examples/examples/z_queryable_shm.rs
```diff
@@ -45,10 +43,7 @@ async fn main() {
         .wait()
         .unwrap();
     // ...and an SHM provider
-    let provider = ShmProviderBuilder::builder()
-        .protocol_id::<POSIX_PROTOCOL_ID>()
-        .backend(backend)
-        .wait();
+    let provider = ShmProviderBuilder::backend(backend).wait();
```
evshary: Why can't we use `let provider = ShmProviderBuilder::default_backend().wait().unwrap();` in this example?

examples/examples/z_alloc_shm.rs
```diff
@@ -23,22 +23,44 @@ use zenoh::{
 async fn main() {
     // Initiate logging
     zenoh::init_log_from_env_or("error");
-    run().await.unwrap()
+    run_minimal().await.unwrap();
+    run().await.unwrap();
+}
+
+async fn run_minimal() -> zenoh::Result<()> {
```
evshary: In fact, I'm not pretty sure if we want this. The basic usage should be already in other examples.

examples/examples/z_alloc_shm.rs
evshary, line 1: It took me some time to understand what the example was trying to convey. I suggest adding the description in the example README.  (`z_alloc_shm.rs` is missing there)

evshary, line 1: I'm new to the Zenoh shared memory usage. While we provide two ways to allocate SHM (direct allocation and layout allocation), I would suggest using a block to isolate each way to avoid confusion.

```rust
// Option 1:
let _direct_allocation = {
  ...
};
// Option 2:
let _layout_allocation = {
  let buffer_layout = {...};
  let sbuf = {...};
  sbuf
};
```

I'm just taking an example. You could choose a better variable name.

commons/zenoh-shm/src/api/protocol_implementations/posix/posix_shm_client.rs
```diff
@@ -19,14 +19,24 @@ use zenoh_result::ZResult;
...
 };
 
 /// Client factory implementation for particular shared memory protocol
 #[zenoh_macros::unstable_doc]
 #[derive(Debug)]
 pub struct PosixShmClient;
 
+impl WithProtocolID for PosixShmClient {
+    fn id(&self) -> ProtocolID {
+        POSIX_PROTOCOL_ID
```
evshary, line 36: Just out of curiosity, do we plan to provide more ways to allocate SHM? I only see `POSIX_PROTOCOL_ID` now, so I'm not sure why we need to specify the ID.

evshary (commented):
I'm happy that the ID disappeared and made the usage simpler :+1: 
I think we can polish the examples for better understanding.

commons/zenoh-shm/src/api/protocol_implementations/posix/posix_shm_client.rs
yellowhatter: There are reasons to keep the ID:
1. Some day we can implement some specific shared-memory backends like ones backed by SystemV shared memory or some device vendor things like CUDA shared-memory.
2. `ShmProviderBackend` trait is made public so that any user can implement specific backend.
3. IDs + file access mode flags can be used to artificially split SHM domains.

SHM is architectured to be extremely flexible. You might find some things excessive, for example  all that stuff with Layout and Alignment - and it was made to smoothly support allocator API (like C++ STL-style allocators) and more specific stuff for GPU. Now I'm polishing it to make things easier for basic usage and hide too much specific things yet leaving them accessible for advanced use case.

evshary: I believe CUDA shared-memory is something useful. Make sense in this case.

examples/examples/z_alloc_shm.rs
```diff
@@ -23,22 +23,44 @@ use zenoh::{
...
+        // NOTE: For extended PosixShmProviderBackend API please check z_posix_shm_provider.rs
+        // create backend...
+        let comprehencive = PosixShmProviderBackend::builder()
+            .with_size(65536)
+            .unwrap()
+            .wait()
+            .unwrap();
+        // ...and an SHM provider
+        ShmProviderBuilder::backend(comprehencive).wait()
+    };
```
kydos, line 51: I think we should define the alignment for the provider and completely remove the layout type.

examples/examples/z_alloc_shm.rs
```diff
@@ -23,22 +23,44 @@ use zenoh::{
...
+    // there is a simple way to create default ShmProvider initialized with default-configured
+    // SHM backend (PosixShmProviderBackend) and more comprehensive way
+    let provider = {
+        // OPTION: simple ShmProvider creation
+        let _simple = ShmProviderBuilder::default_backend().wait().unwrap();
+
+        // OPTION: comprehensive ShmProvider creation
+        // NOTE: For extended PosixShmProviderBackend API please check z_posix_shm_provider.rs
+        // create backend...
+        let comprehencive = PosixShmProviderBackend::builder()
```
kydos: s/comprehencive/comprehensive

examples/examples/z_bytes_shm.rs
```diff
@@ -26,10 +26,7 @@ fn main() {
         .wait()
         .unwrap();
     // ...and an SHM provider
-    let provider = ShmProviderBuilder::builder()
-        .protocol_id::<POSIX_PROTOCOL_ID>()
-        .backend(backend)
-        .wait();
+    let provider = ShmProviderBuilder::backend(backend).wait();
```
kydos: For POSIX SHM it would be nice to be able to combine the two steps into one.

examples/examples/z_queryable_shm.rs
kydos: ...and provide the size.

zenoh/tests/bytes.rs
```diff
@@ -27,10 +27,7 @@ fn shm_bytes_single_buf() {
...
         .unwrap();
     // ...and an SHM provider
-    let provider = ShmProviderBuilder::builder()
-        .protocol_id::<POSIX_PROTOCOL_ID>()
-        .backend(backend)
-        .wait();
+    let provider = ShmProviderBuilder::backend(backend).wait();
 
     // Prepare a layout for allocations
     let layout = provider.alloc(1024).into_layout().unwrap();
```
kydos, line 33: The _layout_ has confused many users. I suggest we fix the layout at the creation of the provider and then only expose a simple alloc. This would make it more user friendly and will streamline the implementation of the SHM. I don't think we should allow to allocate with different alignment within the same provider.

examples/examples/z_bytes_shm.rs
yellowhatter: Yes, it is possible to do like this:
```
let provider = ShmProviderBuilder::default_backend(16 * 1024 * 1024).wait()?;
let mut shm_buffer = provider.alloc(512).wait()?;
```
No layouts by default, nothing, just pure sizes.

However, I want to keep the layout API both for provider and allocations - just hide it behind streamlined defaults. The rationale to do that is:
- allocator APIs
- the integration of GPU and FPGA devices into our shared-memory - there you might easily have 1024 byte alignment for certain cases and sometimes operate with different differently-aligned arrays of data within the same project - that's where the layout API will have it's use.
- custom provider backends for specific needs
 
So yes, I already take measures to hide layouts away from general use-case, and I vote to keep it for advanced scenario.

zenoh/tests/bytes.rs
yellowhatter: Pls check my answer above

examples/examples/z_queryable_shm.rs
yellowhatter: done

examples/examples/z_alloc_shm.rs
yellowhatter: removed minimal

examples/examples/z_alloc_shm.rs
yellowhatter: added readme

yellowhatter: Reorganized options in this example

evshary (approved):

PR #1990, opened 2025-06-11
Publish localhost locator if no other locators are avalable or when binding to the localhost interface.

### LINKED ISSUE #1989: Zenoh never advertise localhost locator even if it's the only one available
### Describe the bug

When Zenoh runs on an offline computer, no locator are advertised when gossiping with peers. This leads to peer not connecting with each others and the router doing failover brokering.

### To reproduce

1. Disconnect the computer from any network.
2. Run `zenohd --listen tcp/0.0.0.0:7447`
3. Run `z_put --listen tcp/0.0.0.0:0 --connect tcp/127.0.0.1:7447`
3. Run `z_sub --listen tcp/0.0.0.0:0 --connect tcp/127.0.0.1:7447`

If you look at the network or the trace logs, you can see the router is doing failover brokering.

Connect the computer to a network and locators (e.g. the private address 192.168.X.X) are advertised, and the peers can connect to each others.

### System info

- Platform: Ubuntu 20.04 64-bit
- Zenoh 1.4.0.

Hugal31:
This PR does two things:

When listening on an unspecified IP address (i.e. 0.0.0.0 and [::]):

1. If a binding interface is specified, do not filter out localhost addresses. Rational: we'll only get localhost addresses if the bound interface is the loopback interface (at least on Unix systems, someone correct me on Windows?)
2. If the list of IP locator is empty and we didn't specify any interface, manually add a localhost locator.

Note that is has the side effect of adding the localhost locator to the list even if there is already another non-ip locator (e.g. Unix socket, ...). I think that's all-right.

Closes #1989

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/commons/zenoh-util/src/net/mod.rs
+++ b/commons/zenoh-util/src/net/mod.rs
@@ -11,7 +11,7 @@
 // Contributors:
 //   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
 //
-use std::net::{IpAddr, Ipv6Addr};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 
 #[cfg(unix)]
 use lazy_static::lazy_static;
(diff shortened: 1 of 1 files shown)


Hugal31:
I am not sure how to add unit tests for that, since it requires knowing the available network interfaces.

Hugal31:
PR #2671 broke this one. It allows for loopback addresses to be listed in the locator for multicast **scouting**, but it doesn't allow **gossip** to pass localhost locators...

One option would be to publish the localhost locator only to loopback peers when gossiping, but I'm not sure how to do that. Any other idea?

Issue #1706, opened 2025-01-13
Scanning network interfaces should be made configurable


Mallets:
Shouldn't this be made configurable?
For example, in the case of VPN or something similar, interfaces can appear and disappear.
Some interfaces may not be there at the start, but will appear later, if the user wants to take this case into account, then such optimization can harm him.

_Originally posted by @sashacmc in https://github.com/eclipse-zenoh/zenoh/issues/1704#issuecomment-2583206778_

patrickelectric:
We are having this problem in https://github.com/bluerobotics/BlueOS where network configuration changes when the system is starting. If the network configuration changes by the user a restart of zenohd is also necessary.

PR #2004, opened 2025-06-19
Update network interfaces periodically

patrickelectric:
Helps #1706 

I did try to make it configurable, but it failed with a recursive dependency issue between the modules. 

This is not the proper solution but I believe that is a better approach than the original concept of never updating.


patrickelectric:
ping @OlivierHecart

patrickelectric:
ping @OlivierHecart

patrickelectric:
It would be nice if someone could take a look in this PR, it's quite important!

PR #2013, opened 2025-06-24
use dynamic runtime in plugins

### LINKED ISSUE #1745: plugin incompatibility is not always detected
### Describe the bug

There are 2 problems in plugin compatibility detection:
- information compared  between plugin and zenohd (in `get_vtable` function) at this moment is not enough. It's not easy to implement full solution, but at least it makes sense to compare zenoh release version
- list of features for compare listed in `FEATURES` global variable may not fully cover all features existing. There was a test to check this, but it seems to be removed. Need to recover the test or find other way to check validity of this list

The problems were found when fixing plugin crash in ts : https://github.com/eclipse-zenoh/zenoh-ts/pull/109 

### To reproduce

-

### System info

-

### LINKED ISSUE #1957: Plugin compatibility checking uses non-ABI-stable types
### Describe the bug

When loading a plugin the loader and plugin assume that the `zenoh_plugin_trait::Compatibility` type is ABI-stable enough to pass between arbitrary versions to then verify if they are compatible.

This type is itself `repr(C)`, but it contains `repr(Rust)` types:

https://github.com/eclipse-zenoh/zenoh/blob/2f955cc8749a8451d193eb4b4060e46e30165804/plugins/zenoh-plugin-trait/src/compatibility.rs#L57-L66

There's a few potentially unstable type layouts here and in sub-types:

 * `&'static str`: still undefined, it may be e.g. `(len, ptr)` or `(ptr, len)`
 * `Option<RustcVersion>`/`Option<PluginStructVersion>`: both of these sub-structs, while being `repr(C)`, contain multiple fields with niches, the `Option` discriminant may be packed in different niches, or kept as a separate tag field. If stored as a tag field the byte value used may be different in different compilations.

### To reproduce

Probably requires compiling with different compiler versions with different niche heuristics and/or `-Zbuild-std`/`-Zrandomize-layout`. I'm not certain there are existing compilers that have different enough behavior to trigger these problems, but they're at least allowed by the current specifications.

### System info

irrelevant

DenisBiryukov91:
- Pass dynamic runtime to plugins instead of statically typed Runtime to reduce the probability of conflicts due to unstable rust ABI.
- Make Compatibility struct ABI-stable.

Resolves #1957 
Resolves #1745 

corresponding plugin api update PRs:
- https://github.com/eclipse-zenoh/zenoh-ts/pull/316
- https://github.com/eclipse-zenoh/zenoh-plugin-dds/pull/573
- https://github.com/eclipse-zenoh/zenoh-plugin-ros2dds/pull/543
- https://github.com/eclipse-zenoh/zenoh-plugin-webserver/pull/401

corresponding backend api update PRs:
- https://github.com/eclipse-zenoh/zenoh-backend-filesystem/pull/429
- https://github.com/eclipse-zenoh/zenoh-backend-s3/pull/413
- https://github.com/eclipse-zenoh/zenoh-backend-rocksdb/pull/442
- https://github.com/eclipse-zenoh/zenoh-backend-influxdb/pull/473


plugins/zenoh-plugin-trait/src/manager/dynamic_plugin.rs
```diff
@@ -69,19 +69,15 @@ impl<StartArgs: PluginStartArgs, Instance: PluginInstance>
...
             &plugin_compatibility_record
         );
-        if !plugin_compatibility_record.compare(&mut host_compatibility_record) {
-            bail!(
-                "Plugin compatibility mismatch:\nHost:\n{}Plugin:\n{}",
-                host_compatibility_record,
-                plugin_compatibility_record
-            );
+        if let Err(e) = host_compatibility_record.check(&plugin_compatibility_record) {
+            bail!("Plugin compatibility mismatch:\n {}", e);
```
milyin, line 80: Do it always display full difference information or `e` is just an error message?
I think the panic message should contain enough information to resolve the issue (e.g. which features are different. etc)

plugins/zenoh-plugin-trait/src/vtable.rs
```diff
@@ -13,7 +13,7 @@
 //
 use zenoh_result::ZResult;
 
-use crate::{Plugin, StructVersion};
+use crate::Plugin;
 
 pub type PluginLoaderVersion = u64;
 pub const PLUGIN_LOADER_VERSION: PluginLoaderVersion = 1;
```
milyin: It's necessary to increase this version - it's checked before loading to make sure that the loading mechanism is the same

plugins/zenoh-plugin-trait/src/manager/dynamic_plugin.rs
DenisBiryukov91: it will show the values that do not match: i.e. either both rust c versions, both zenoh versions or both sets of zenoh features

plugins/zenoh-plugin-trait/src/vtable.rs
DenisBiryukov91: done

PR #2014, opened 2025-06-24
add CongestionControl::BlockFirst

wyfo:
This congestion controlled is implemented with a special "slot" inside `TransportUnicastUniversal`. When a `BlockFirst` message is sent, the slot is acquired and a task is spawned to send the message, as if it was blocking. If a second `BlockFirst` message arrives, but the slot is not free, then it is dropped (after waiting for the configured time).
The slot is materialized by a event pair (notifier, waiter). Waiting for the event means acquiring the slot, and notifying after having sent the message is means releasing the slot.

### DIFF
--- a/commons/zenoh-config/src/qos.rs
+++ b/commons/zenoh-config/src/qos.rs
@@ -53,6 +53,7 @@ pub struct PublisherQoSConfig {
 pub enum CongestionControlConf {
     Drop,
     Block,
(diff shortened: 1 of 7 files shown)


OlivierHecart:
LGTM
Setting this as Draft until we decide if we integrate that or not.
Need also to discuss tasks.

PR #2021, opened 2025-06-26
Add `application/flatbuffers` encoding

fuzzypixelz:
See https://github.com/eclipse-zenoh/roadmap/discussions/193.


DenisBiryukov91:
We need to add the same predefined encoding support to all other bindings + zenoh-pico

DenisBiryukov91:
Also this is a breaking change for those who already pass application/flatbuffers encoding as a string

PR #2024, opened 2025-06-30
Polish shm (for zshm examle)

yellowhatter:


yellowhatter:
Lints failure should be ignored (upstream issue)

PR #2030, opened 2025-07-03
Add more options to QoSOverwrite interceptor

OlivierHecart:


commons/zenoh-config/src/lib.rs
```diff
@@ -176,6 +183,82 @@ pub struct AclConfigSubjects {
     pub link_protocols: Option<NEVec<InterceptorLink>>,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
```
wyfo, line 186: use `#[serde(deserialize_with = "deserialize_range", serialize_with = "serialize_range")]` instead

wyfo (changes requested):

PR #2034, opened 2025-07-07
SHM typed API

yellowhatter:
This PR:
- makes shm buffer traits have generic parameter of underlying type 
- adds `Typed` generic that is a higher-level wrapper around all shm buffer types. The `Typed` generic implements some of shm buffer trait interfaces and provides safe typed access to underlying data


commons/zenoh-shm/tests/shm_buffer.rs
```diff
@@ -0,0 +1,50 @@
...
+struct SharedByteData {
+    data: [u8; 64],
+}
+
+fn make_shm_buffer() -> ZShmMut {
+    let provider = ShmProviderBuilder::default_backend(65536).wait().unwrap();
+    provider.alloc(64).wait().unwrap()
+}
+
+fn validate_raw_buffer_consistency(buffer: &mut ZShmMut) {
```
kydos, line 32: Is this really necessary? Isn't memory on modern OS always set to 0 for security reasons?

commons/zenoh-shm/tests/shm_buffer.rs
```diff
@@ -0,0 +1,50 @@
...
+    buffer.as_mut().data.fill(0);
+}
+
+#[test]
+fn shm_buffer_morph() {
+    let mut buffer = make_shm_buffer();
+    validate_raw_buffer_consistency(&mut buffer);
+
+    let mut buffer: Typed<SharedByteData, _> = buffer.try_into().unwrap();
+    validate_typed_buffer_consistency(&mut buffer);
```
kydos, line 99: The validate functions fill the buffer with 0s using either the typed or untyped API. It would be interesting to write something using one API and read it using the other API and show consistency in values. Otherwise we are just showing that morth the SHMBuf into the SharedData type. What do you think?

yellowhatter: Yes, fully agree with you! The tests for typed API to be improved.

Cargo.toml
```diff
@@ -164,7 +164,7 @@ serde_json = "1.0.128"
 serde_with = "3.12.0"
 serde_yaml = "0.9.34"
 static_init = "1.0.3"
-stabby = "36.1.1"
+stabby = "72.1.1"
```
YuanYuYuan: I understand the stabby version is old, but it's better to put it in a separate PR if it's unnecessary.

commons/zenoh-shm/src/api/buffer/typed.rs
```diff
@@ -0,0 +1,164 @@
+//
+// Copyright (c) 2023 ZettaScale Technology
```
YuanYuYuan: 2023 -> 2025

commons/zenoh-shm/src/api/buffer/zshmmut.rs
```diff
@@ -2,11 +2,11 @@
...
 // This program and the accompanying materials are made available under the
-// terms of the Eclipse Public License 2.0 which is available at
-// http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
-// which is available at https://www.apache.org/licenses/LICENSE-2.0.
+// terms of the Eclipse Public License 2.inner which is available at
+// http://www.eclipse.org/legal/epl-2.inner, or the Apache License, Version 2.inner
+// which is available at https://www.apache.org/licenses/LICENSE-2.inner.
 //
-// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+// SPDX-License-Identifier: EPL-2.inner OR Apache-2.inner
```
YuanYuYuan: These might be messed up by auto-naming.

commons/zenoh-shm/src/api/provider/types.rs
```diff
@@ -296,4 +297,5 @@ impl From<ZLayoutAllocError> for zenoh_result::Error {
 
 /// SHM buffer layouting and allocation result
 #[zenoh_macros::unstable_doc]
+#[allow(type_alias_bounds)]
```
YuanYuYuan: No type alias bounds being used here?

commons/zenoh-shm/src/api/provider/types.rs
```diff
@@ -273,6 +273,7 @@ pub type ChunkAllocResult = Result<AllocatedChunk, ZAllocError>;
 
 /// SHM buffer allocation result
 #[zenoh_macros::unstable_doc]
+#[allow(type_alias_bounds)]
```
YuanYuYuan: No type alias bounds being used here?

commons/zenoh-shm/src/api/buffer/traits.rs
```diff
@@ -11,14 +11,14 @@
...
-
-use std::{
-    num::NonZeroUsize,
-    ops::{Deref, DerefMut},
-};
+use std::{num::NonZeroUsize, ptr};
 
 use crate::api::provider::types::MemoryLayout;
 
+pub trait ResideInShm: zerocopy::KnownLayout {}
```
YuanYuYuan: Why don't we use the zerocopy::KnowLayout instead?

commons/zenoh-shm/src/api/buffer/traits.rs
```diff
@@ -28,10 +28,60 @@ pub enum BufferRelayoutError {
 }
 
 #[zenoh_macros::unstable_doc]
-pub trait ShmBuf: Deref<Target = [u8]> + AsRef<[u8]> {
+pub trait ShmBuf<T: ?Sized>: Sized + AsRef<T> {
```
YuanYuYuan, line 37: 1. The `Sized` is required by `ShmConvert` in the returning result.
2. `ShmBuf` is a trait. Since we already have lots of struct names containing this word, it's better to make some distinction. If the purpose here is only for validation, naming it as `ValidShmBuf` will be better.
3. It seems that we usually set T to a `[u8]`. Is there any other case?

commons/zenoh-shm/src/api/buffer/traits.rs
```diff
@@ -28,10 +28,60 @@ pub enum BufferRelayoutError {
...
     #[zenoh_macros::unstable_doc]
     fn is_valid(&self) -> bool;
+}
 
+pub trait ShmConvert<T: ?Sized, Tother: ?Sized> {
```
YuanYuYuan: 1. `ShmConvert<T: ?Sized, Tother: ?Sized>: Sized`
2. This trait is neither tested nor used publicly. Should we include it in this PR?

commons/zenoh-shm/src/api/buffer/traits.rs
```diff
@@ -28,10 +28,60 @@ pub enum BufferRelayoutError {
...
+    /// Performs the conversion.
+    fn shm_convert<Tdst: ShmBuf<Tother>>(self) -> Result<Tdst, Self>
```
YuanYuYuan: If we want to follow the camel case naming convention, we should change Tdst to TDst, and so on and so forth.

commons/zenoh-shm/src/api/buffer/typed.rs
```diff
@@ -0,0 +1,164 @@
...
+        zshmmut::{zshmmut, ZShmMut},
+    },
+    ShmBufInner,
+};
+
+fn can_transmute<T: zerocopy::KnownLayout + zerocopy::FromBytes>(
+    value: &ShmBufInner,
+) -> ZResult<()> {
+    let slice = value.as_ref();
+    let _ = T::read_from_bytes(slice).map_err(|e| format!("Error transmutting: {e}"))?;
```
YuanYuYuan: We waste a copy here.

commons/zenoh-shm/src/api/buffer/typed.rs
```diff
@@ -0,0 +1,164 @@
...
+use crate::{
+    api::buffer::{
+        traits::{ResideInShm, ShmBuf, ShmBufMut, ShmBufUnsafeMut},
+        zshm::{zshm, ZShm},
+        zshmmut::{zshmmut, ZShmMut},
+    },
+    ShmBufInner,
+};
+
+fn can_transmute<T: zerocopy::KnownLayout + zerocopy::FromBytes>(
```
YuanYuYuan: I believe this is equivalent to `T::read_from_bytes` error branch. You can handle the error if you want to make it verbose. For instance, `SizeMismatch(usize, usize)`.

```rust
fn can_transmute<T>(value: &ShmBufInner) -> bool {
    value.len().get() == size_of::<T>()
}
```

commons/zenoh-shm/src/api/buffer/typed.rs
```diff
@@ -0,0 +1,164 @@
...
+    Ok(())
+}
+
+pub struct Typed<T: ?Sized, Tbuf: ShmBuf<[u8]>> {
+    buf: Tbuf,
+    _phantom: PhantomData<T>,
+}
+
+impl<T: ?Sized, Tbuf: ShmBuf<[u8]>> Typed<T, Tbuf> {
+    pub fn unwrap(self) -> Tbuf {
```
YuanYuYuan: 1. `unwrap` is commonly used for unwrapping an Option. Suggest a renaming: unwrap => get or inner.
2. Regarding the trait bounds, we can write like this.
```rust
impl<T: ?Sized, B> Typed<T, B> {
    pub fn unwrap(self) -> B {
        self.buf
    }
}
```

commons/zenoh-shm/src/api/buffer/typed.rs
```diff
@@ -0,0 +1,164 @@
...
+    fn try_from(value: &'a zshm) -> Result<Self, Self::Error> {
+        can_transmute::<T>(&value.inner).map(|_| Self {
+            buf: value,
+            _phantom: PhantomData,
+        })
+    }
+}
+
+impl<'a, T: ResideInShm + zerocopy::FromBytes> TryFrom<&'a mut zshm> for Typed<T, &'a mut zshm> {
+    type Error = Box<dyn IError + Send + Sync + 'static>;
```
YuanYuYuan, line 184: Why do we include zshm in the returning error?

commons/zenoh-shm/src/api/buffer/typed.rs
```diff
@@ -0,0 +1,164 @@
...
+
+fn can_transmute<T: zerocopy::KnownLayout + zerocopy::FromBytes>(
+    value: &ShmBufInner,
+) -> ZResult<()> {
+    let slice = value.as_ref();
+    let _ = T::read_from_bytes(slice).map_err(|e| format!("Error transmutting: {e}"))?;
+    Ok(())
+}
+
+pub struct Typed<T: ?Sized, Tbuf: ShmBuf<[u8]>> {
```
YuanYuYuan: Typed it too generic and hard to debug. I recommend using `TypedShmBuf` with the following definition

```rust
pub struct TypedShmBuf<T: ?Sized, B> {
    buf: B,
    _phantom: PhantomData<T>,
}
```

We don't need so many bounds in the definition. Instead, we can limit it while implementing.

commons/zenoh-shm/src/api/buffer/typed.rs
```diff
@@ -0,0 +1,164 @@
...
+    _phantom: PhantomData<T>,
+}
+
+impl<T: ?Sized, Tbuf: ShmBuf<[u8]>> Typed<T, Tbuf> {
+    pub fn unwrap(self) -> Tbuf {
+        self.buf
+    }
+}
+
+impl<T: ResideInShm + zerocopy::FromBytes> TryFrom<ZShm> for Typed<T, ZShm> {
```
YuanYuYuan: If we use the `can_transmute` proposed above, the trait bounds `ResideInShm` and `FromBytes` are unnecessary.

commons/zenoh-shm/src/api/buffer/zshm.rs
```diff
@@ -94,64 +113,59 @@ impl AsRef<[u8]> for ZShm {
...
 }
 
 impl TryFrom<ZShm> for ZShmMut {
     type Error = ZShm;
 
     fn try_from(value: ZShm) -> Result<Self, Self::Error> {
-        match value.0.is_unique() && value.0.is_valid() {
-            true => Ok(Self(value.0)),
+        match value.inner.is_unique() && value.inner.is_valid() {
+            true => Ok(unsafe { std::mem::transmute::<ZShm, ZShmMut>(value) }),
```
YuanYuYuan, line 133: The unsafe claim is missing.

commons/zenoh-shm/src/api/buffer/typed.rs
yellowhatter: For owned `ZShm` and `ZShmMut` types we need to do this in order not to drop the instance of the buffer in case of error. This is an optimization for `ZShm` - no need to make mandatory copy before making `try_*` and absolutely needed for `ZShmMut` because it even doesn't have copy

commons/zenoh-shm/src/api/buffer/typed.rs
yellowhatter: I agree with the idea of removing trait bounds.
However, with naming I want to discuss.
The idea is that Typed API is planned to become zenoh-wide, so that non-shm zenoh buffers will also be "typable". In this case, the naming for `Typed`  should be neutral...

Cargo.toml
yellowhatter: rolled back!

commons/zenoh-shm/src/api/buffer/typed.rs
yellowhatter: fixed

commons/zenoh-shm/src/api/buffer/zshmmut.rs
yellowhatter: fixed

commons/zenoh-shm/src/api/provider/types.rs
yellowhatter: removed

commons/zenoh-shm/src/api/buffer/typed.rs
yellowhatter: ResideInShm is a trait that collects bounds that make type SHM-safe.
For `FromBytes` - yes, removed

commons/zenoh-shm/src/api/buffer/zshm.rs
yellowhatter: fixed

commons/zenoh-shm/src/api/buffer/typed.rs
YuanYuYuan: I'm a bit confused. Isn't the input a borrowed `&'a mut zshm`? I believe it won't be dropped.

commons/zenoh-shm/src/api/provider/types.rs
yellowhatter: fixed

commons/zenoh-shm/src/api/buffer/typed.rs
YuanYuYuan: I see. It makes more sense if it aims to work for the whole zenoh.

commons/zenoh-shm/src/api/buffer/typed.rs
yellowhatter: For `&` versions Error does not contain the value

commons/zenoh-shm/src/api/buffer/typed.rs
yellowhatter: renamed to `into_inner`

commons/zenoh-shm/tests/shm_buffer.rs
yellowhatter: fixed!

commons/zenoh-shm/src/api/protocol_implementations/posix/posix_shm_provider_backend.rs
```diff
@@ -66,52 +66,44 @@ impl PartialEq for Chunk {
 
 /// Builder to create posix SHM provider
 #[zenoh_macros::unstable_doc]
-pub struct PosixShmProviderBackendBuilder;
+pub struct PosixShmProviderBackendBuilder<What> {
+    what: What,
```
YuanYuYuan: I believe it deserves a meaningful name.

examples/examples/z_alloc_shm.rs
```diff
@@ -77,8 +72,7 @@ async fn run() -> zenoh::Result<()> {
 
         // Option 2: Comprehensive configuration:
         let comprehensive_layout = provider
-            .alloc(512)
-            .with_alignment(AllocAlignment::new(2)?)
+            .alloc((512, AllocAlignment::ALIGN_2_BYTES))
```
YuanYuYuan, line 78: Why is this a tuple? What do the 1st and the 2nd fields represent?

commons/zenoh-shm/src/api/provider/types.rs
```diff
@@ -275,6 +198,11 @@ pub type ChunkAllocResult = Result<AllocatedChunk, ZAllocError>;
 #[zenoh_macros::unstable_doc]
 pub type BufAllocResult = Result<ZShmMut, ZAllocError>;
 
+/// SHM buffer allocation result
+#[zenoh_macros::unstable_doc]
+#[allow(type_alias_bounds)]
```
YuanYuYuan: Do we need `type_alias_bounds`?

commons/zenoh-shm/src/api/provider/types.rs
```diff
@@ -297,3 +237,8 @@ impl From<ZLayoutAllocError> for zenoh_result::Error {
 /// SHM buffer layouting and allocation result
 #[zenoh_macros::unstable_doc]
 pub type BufLayoutAllocResult = Result<ZShmMut, ZLayoutAllocError>;
+
+/// Typed SHM buffer layouting and allocation result
+#[zenoh_macros::unstable_doc]
+#[allow(type_alias_bounds)]
```
YuanYuYuan: Do we need `type_alias_bounds`?

commons/zenoh-shm/src/api/provider/memory_layout.rs
```diff
@@ -0,0 +1,203 @@
...
+    }
+}
+
+impl<T> From<LayoutForType<T>> for MemoryLayout {
+    fn from(value: LayoutForType<T>) -> Self {
+        value.layout().into()
+    }
+}
+
+pub struct StaticLayout<T> {
```
YuanYuYuan, line 208: What's this used for?

examples/examples/z_alloc_shm.rs
yellowhatter: `alloc` accepts arguments convertible into MemoryLayout and also LayoutForType<T> for typed allocations

There are various ways to construct MemoryLayout:
`(usize, AllocAlignment)`
`(NonZeroUsize, AllocAlignment)`
`usize`
`NonZeroUsize`
`&StaticLayout<T>`
So what you see is a tuple for easy MemoryLayout description

commons/zenoh-shm/src/api/provider/memory_layout.rs
yellowhatter: This type is inner for LayoutForType and represents a statically-known layout WITHOUT type information

The idea is simple:

```
// typed layout
let typed_layout = BuildLayout::for_type::<SharedData>();

 // allocate **typed** SHM buffer
 let typed_buf = shm_provider.alloc(typed_layout).wait().unwrap();

 // allocate **untyped** SHM buffer
 let untyped_buf = shm_provider.alloc(typed_layout.layout()).wait().unwrap();
```

yellowhatter: So, `LayoutForType` is used as statically-known layout in context of typed operations, while `StaticLayout` is the same but explicitly untyped

yellowhatter: Statically-known layouts are always correct, zero-sized & zero-cost while `MemoryLayout` is dynamic one, thus it has size and fallible ctor

commons/zenoh-shm/src/api/protocol_implementations/posix/posix_shm_provider_backend.rs
yellowhatter: renamed to `Layout`

commons/zenoh-shm/src/api/provider/types.rs
yellowhatter: removed

commons/zenoh-shm/src/api/provider/types.rs
yellowhatter: removed

commons/zenoh-shm/src/api/provider/memory_layout.rs
YuanYuYuan: I see. I think we should document this in the comments.

YuanYuYuan (changes requested):

examples/examples/z_alloc_shm.rs
YuanYuYuan: Okay. It's better to have documentation for this, as it serves as an example for API usage.

YuanYuYuan:
Hi @yellowhatter, please address the two documentation issues above. You can merge this once you finish it.

PR #2035, opened 2025-07-09
Avoid key expression checks in local query handling

OlivierHecart:


zenoh/src/api/session.rs
```diff
@@ -1725,7 +1724,7 @@ impl SessionInner {
                     body: DeclareBody::UndeclareQueryable(UndeclareQueryable {
                         id: qable_state.id,
                         ext_wire_expr: WireExprType {
-                            wire_expr: qable_state.key_expr.clone(),
+                            wire_expr: WireExpr::empty(),
```
evshary, line 1727: Just out of curiosity, why do we change `wire_expr` here?

evshary (approved):
LGTM overall, but just left one question

zenoh/src/api/session.rs
```diff
@@ -1933,17 +1932,14 @@ impl SessionInner {
...
-            MatchingStatusType::Queryables(true) => state.queryables.values().any(|q| {
-                q.complete
-                    && state
-                        .local_wireexpr_to_expr(&q.key_expr)
-                        .is_ok_and(|ke| ke.includes(key_expr))
-            }),
+            MatchingStatusType::Queryables(false) => state
+                .queryables
+                .values()
+                .any(|q| q.key_expr.includes(key_expr)),
```
fuzzypixelz:
```suggestion
                .any(|q| q.key_expr.intersects(key_expr)),
```

fuzzypixelz (changes requested):
Good for me, aside from what I think is a typo.

On an unrelated note, I couldn't find any documentation of how complete query/reply uses _include_ and not _intersect_, is this documented somewhere?

zenoh/src/api/session.rs
OlivierHecart: We don't need the wire expr here as we have the queryable id. Let's not spend time computing it.

PR #2045, opened 2025-07-18
Support QUIC unreliable datagrams

fuzzypixelz:
This commit makes it possible to use unreliable datagrams in QUIC by setting `rel=0` in QUIC endpoints. The initial MTU value and the MTU discovery interval are both configurable via endpoint configuration.

This is the first link to reuse an exiting protocol name, so the transport manager has to be updated to support it. Now links are identified by both their protocol name and reliability.

The transport_quic and transport_quic_datagram can be independently enabled/disabled.

I took the opportunity refactor TLS configuration options into zenoh-link-commons; QUIC and TLS use the exact same values except for the handshake timeout duration (not used by QUIC links).


oteffahi (approved):
Looks good for the release, however I still think we should refactor this into a single Quic link.

PR #2046, opened 2025-07-18
optimize hashmap for integer keys

wyfo:
In the case of a small number of integer keys starting from zero, hashmap can be replaced by a vector to eliminate hash cost. It must of course fallback to a regular hashmap if keys overflow a given small constant.


wyfo:
~~It would also be possible to use a dynamic variable instead of a generic constant for the small limit~~

PR #2047, opened 2025-07-18
use vector instead of hashmap for routes

wyfo:
Routes were built with hashmap to avoid duplicate, but then stored as it, even if the map aspect was not used after built. The map was iterated, which not very performant.
By storing routes in a vector instead, and using a hashset for deduplication, performance are improved significantly.


zenoh/src/net/routing/dispatcher/queries.rs
```diff
@@ -244,11 +243,11 @@ fn compute_final_route(
             if let Some(qabl) = qabls.iter().find(|qabl| {
                 qabl.direction.0.id != src_face.id && qabl.info.is_some_and(|info| info.complete)
             }) {
-                let mut route = HashMap::new();
+                let mut route = QueryRouteBuilder::new();
 
                 let mut direction = qabl.direction.clone();
                 let qid = insert_pending_query(&mut direction.0, query);
-                route.insert(direction.0.id, (direction, qid));
+                route.insert(direction.0.id, || (direction, qid));
```
OlivierHecart: Please reformat as the above cases. It will make the code more homogeneous and readable.

PR #2053, opened 2025-07-21
Improve interceptors performances

OlivierHecart:
Improve interceptors performances for interceptors that don't need to access cache nor key_expr.

On a host where 
```
z_pub_thr 8
z_sub_thr

=> 7.2 Mmsgs/s
```

Using the following config:
```
{
   qos: {
     network: [
       {
         messages: ["put"],
         overwrite: {
           priority: "data",
         },
       },
     ],
   },
}
```

### Before the changes,
```
z_pub_thr 8 -c configs/qosoverwrite.js
z_sub_thr

=> 5.3 Mmsgs/s
```

```
z_pub_thr 8
z_sub_thr -c configs/qosoverwrite.js

=> 5.5 Mmsgs/s
```

### After the changes,
```
z_pub_thr 8 -c configs/qosoverwrite.js
z_sub_thr

=> 7.1 Mmsgs/s
```

```
z_pub_thr 8
z_sub_thr -c configs/qosoverwrite.js

=> 6.8 Mmsgs/s
```

### DIFF
--- a/zenoh/src/net/primitives/demux.rs
+++ b/zenoh/src/net/primitives/demux.rs
@@ -11,13 +11,13 @@
 // Contributors:
 //   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
 //
-use std::{any::Any, sync::Arc};
+use std::{any::Any, cell::OnceCell, sync::Arc};
 
 use arc_swap::ArcSwap;
 use zenoh_link::Link;
 use zenoh_protocol::network::{
-    ext, response, Declare, DeclareBody, DeclareFinal, NetworkBodyMut, NetworkMessageMut,
-    ResponseFinal,
+    ext, response, Declare, DeclareBody, DeclareFinal, NetworkBodyMut, NetworkMessageExt as _,
+    NetworkMessageMut, ResponseFinal,
 };
 use zenoh_result::ZResult;
 use zenoh_transport::{unicast::TransportUnicast, TransportPeerEventHandler};
@@ -25,7 +25,8 @@ use zenoh_transport::{unicast::TransportUnicast, TransportPeerEventHandler};
 use super::Primitives;
 use crate::net::routing::{
     dispatcher::face::Face,
-    interceptor::{InterceptorTrait, InterceptorsChain},
+    interceptor::{InterceptorContext, InterceptorTrait, InterceptorsChain},
+    router::{InterceptorCacheValueType, Resource},
     RoutingContext,
 };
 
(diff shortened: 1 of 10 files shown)


zenoh/src/net/primitives/demux.rs
```diff
@@ -49,26 +50,73 @@ impl DeMux {
     }
 }
 
+struct DeMuxContext<'a> {
+    demux: &'a DeMux,
+    cache: OnceCell<InterceptorCacheValueType>,
+    expr: OnceCell<String>,
```
wyfo, line 56: You have the same logic duplicated 3 times. Could you abstract it?

zenoh/src/net/routing/interceptor/mod.rs
```diff
@@ -162,56 +173,72 @@ impl InterceptorTrait for InterceptorsChain {
...
-impl<T: InterceptorTrait> InterceptorTrait for ComputeOnMiss<T> {
-    #[inline]
-    fn compute_keyexpr_cache(&self, key_expr: &keyexpr) -> Option<Box<dyn Any + Send + Sync>> {
-        self.interceptor.compute_keyexpr_cache(key_expr)
+pub(crate) enum ComputedOnMiss<'a> {
+    Hit(Option<&'a Box<dyn Any + Send + Sync>>),
+    Computed(Option<Box<dyn Any + Send + Sync>>),
+}
+
+impl ComputedOnMiss<'_> {
```
wyfo, line 219: What's the point of computing a cache to not store it anywhere?

OlivierHecart: That's to abstract the process of computing the needed value if it was not cached from the interceptors. This abstraction was intended to be used in multiple interceptors but is currently only used in downsampling interceptor.

zenoh/src/net/primitives/mux.rs
```diff
@@ -46,33 +51,81 @@ impl Mux {
...
+        };
         let mut ctx = RoutingContext {
-            msg: NetworkMessageMut {
-                body: NetworkBodyMut::Interest(ctx.msg),
-                reliability: Reliability::Reliable,
-            },
-            inface: ctx.inface,
-            outface: ctx.outface,
-            prefix: ctx.prefix,
+            msg: (),
```
wyfo, line 119: Could you just pass the RoutingContext as an additional parameter to `send_interest` instead of having to do that after?

zenoh/src/net/routing/interceptor/mod.rs
wyfo: I see that, but it doesn't make any sense to have to allocate a `Box<Any>` for something that is not even stored. In downsampling interceptor, why do we allocate this cache instead of just compute the id if it's not cached?

zenoh/src/net/primitives/demux.rs
OlivierHecart: It could be further abstracted. But it's not trivial and we have some time constraints.

zenoh/src/net/primitives/mux.rs
OlivierHecart: We could but that's for another PR.

zenoh/src/net/routing/interceptor/mod.rs
OlivierHecart: Maybe I'm missing something but: if downsampling interceptor has to do the job it's not abstracted. As the `compute_keyexpr_cache` fn from the `InterceptorTrait` returns an `Option<Box<dyn Any + Send + Sync>>` that's what we store.
It is indeed not optimal, but simplicity has been chosen here.

PR #2057, opened 2025-07-22
Format TOML files using Taplo

fuzzypixelz:
Taplo is a TOML toolkit. It is the default LSP implementation used in Zed and the popular "Even Better TOML" VS Code extension.

This commit leverages Taplo to ensure uniform TOML formatting in CI and locally (through precommit hooks), thus removing patch pollution due to TOML formatting changes.

This is a temporary solution until the question of TOML formatting is resolved upstream: https://github.com/rust-lang/rustfmt/issues/4091.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -21,5 +21,5 @@ test(=header_check_memory_concurrent) |
 test(=header_link_concurrent) |
 test(=header_link_failure_concurrent)
 """
-threads-required = 'num-cpus'
 slow-timeout = { period = "60s", terminate-after = 6 }
+threads-required = 'num-cpus'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,6 +106,9 @@ jobs:
 
       - name: Install latest cargo-semver-checks
         uses: taiki-e/install-action@cargo-semver-checks
+        
+      - name: Install latest taplo
+        uses: taiki-e/install-action@taplo
 
       - name: Code format check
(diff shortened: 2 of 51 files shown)


wyfo:
Could you add alphabetical sorting? I remember having tried taplo for zenoh a while ago, and having trouble with sorting, but I'm don't remember exactly what the problem was (maybe dependency feature sorting).
Also, we should do the same in all other zenoh repository.

fuzzypixelz:
> Could you add alphabetical sorting? I remember having tried taplo for zenoh a while ago, and having trouble with sorting, but I'm don't remember exactly what the problem was (maybe dependency feature sorting).

Done. I chose not to set `reorder_inline_tables` (makes "version" go after "features").

> Also, we should do the same in all other zenoh repository.

Eventually! But let's not move the goalpost and instead see how this goes before we generalize to all repos.

I'm also making this into a draft so it wouldn't interfere with the coming release (cc @diogomatsubara).

diogomatsubara:
> > Could you add alphabetical sorting? I remember having tried taplo for zenoh a while ago, and having trouble with sorting, but I'm don't remember exactly what the problem was (maybe dependency feature sorting).
> 
> Done. I chose not to set `reorder_inline_tables` (makes "version" go after "features").
> 
> > Also, we should do the same in all other zenoh repository.
> 
> Eventually! But let's not move the goalpost and instead see how this goes before we generalize to all repos.
> 
> I'm also making this into a draft so it wouldn't interfere with the coming release (cc @diogomatsubara).

@fuzzypixelz release is done, so feel free to merge whenever it's ready.

wyfo:
Use a pre-commit hook like https://github.com/ComPWA/taplo-pre-commit to have automatic installation.

PR #2060, opened 2025-07-24
feat: make CongestionControl and Priority hashable

alokpr:
We would like to use these enums as hash keys and discovered that they were not marked as Hashable


alokpr:
CC @JEnoch

PR #2062, opened 2025-07-24
avoid copying shm buffer when sending it on network

wyfo:
The buffer copy was useless, as the content of the buffer was written anyway.

The increment of the shm refcount has also been moved to the serialization step, as it avoids incrementing uselessly in case of message being dropped (even if the watchdog catch it eventually).

Last but not least, the code has been refactored to remove macros.

### DIFF
--- a/commons/zenoh-codec/src/core/zbuf.rs
+++ b/commons/zenoh-codec/src/core/zbuf.rs
@@ -141,7 +141,9 @@ mod shm {
                                 ZSliceKind::ShmPtr => {
                                     self.codec.write(&mut *writer, SHM_PTR)?;
                                     let shmb = zs.downcast_ref::<ShmBufInner>().unwrap();
-                                    Zenoh080::new().write(&mut *writer, &shmb.info)?;
+                                    let mut info = vec![];
+                                    Zenoh080::new().write(&mut &mut info, &shmb.info)?;
+                                    self.codec.write(&mut *writer, &*info)?;
                                     // Increase the reference count so to keep the ShmBufInner
                                     // valid until it is received.
                                     unsafe { shmb.inc_ref_count() };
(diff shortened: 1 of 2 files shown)


io/zenoh-transport/src/shm.rs
```diff
@@ -128,33 +128,40 @@ trait MapShm {
...
+    zbuf: &mut ZBuf,
+    ext_shm: &mut Option<ShmType<ID>>,
+    partner_shm_cfg: &Option<ShmCfg>,
+) -> ZResult<()> {
+    if let Some(shm_cfg) = partner_shm_cfg {
+        for zs in zbuf.zslices_mut() {
+            if let Some(shmb) = zs.downcast_ref::<ShmBufInner>() {
+                if shm_cfg.supports_protocol(shmb.protocol()) {
+                    // Increase the reference count so to keep the ShmBufInner valid
+                    unsafe { shmb.inc_ref_count() };
```
yellowhatter: there already is an increment at serialization stage! no need it here

wyfo: Oups, forgot to remove this one after moving it...

commons/zenoh-codec/src/core/zbuf.rs
```diff
@@ -164,7 +176,7 @@ mod shm {
                                     reader.read_zslices(len, |s| zbuf.push_zslice(s))?;
                                 }
                                 SHM_PTR => {
-                                    let mut zslice: ZSlice = self.codec.read(&mut *reader)?;
+                                    let mut zslice: ZSlice = dbg!(self.codec.read(&mut *reader))?;
```
YuanYuYuan, line 179: Shouldn't we remove this dbg macro?

YuanYuYuan: Addressed in #2066

PR #2063, opened 2025-07-24
Remove ComputedOnMiss

OlivierHecart:


zenoh/src/net/routing/interceptor/downsampling.rs
```diff
@@ -213,63 +213,65 @@ impl DownsamplingInterceptor {
...
         }
     }
+    fn compute_id(&self, key_expr: &keyexpr) -> Option<usize> {
+        let ke_id = zlock!(self.ke_id);
+        if let Some(node) = ke_id.intersecting_keys(key_expr).next() {
+            if let Some(id) = ke_id.weight_at(&node) {
+                return Some(*id);
+            }
+        }
+        None
```
wyfo:
```suggestion
        let node = ke_id.intersecting_keys(key_expr).next()?;
        ke_id.weight_at(&node).copied()
```

zenoh/src/net/routing/interceptor/downsampling.rs
```diff
@@ -213,63 +213,65 @@ impl DownsamplingInterceptor {
...
+        Some(Box::new(self.compute_id(key_expr)))
     }
 
     fn intercept(&self, msg: &mut NetworkMessageMut, ctx: &mut dyn InterceptorContext) -> bool {
-        let cache = ComputedOnMiss::new(self, msg, ctx);
+        let id = ctx
+            .get_cache(msg)
+            .and_then(|c| c.downcast_ref::<Option<usize>>())
+            .cloned()
+            .unwrap_or_else(|| ctx.full_keyexpr(msg).and_then(|k| self.compute_id(&k)));
```
wyfo:
```suggestion
            .unwrap_or_else(|| self.compute_id(&ctx.full_keyexpr(msg)?));
```

zenoh/src/net/routing/interceptor/downsampling.rs
```diff
@@ -213,63 +213,65 @@ impl DownsamplingInterceptor {
...
+
         if self.is_msg_filtered(msg) {
-            if let Some(cache) = cache.as_ref() {
-                if let Some(id) = cache.downcast_ref::<Option<usize>>() {
-                    if let Some(id) = id {
-                        let mut ke_state = zlock!(self.ke_state);
-                        if let Some(state) = ke_state.get_mut(id) {
-                            let timestamp = tokio::time::Instant::now();
+            if let Some(id) = &id {
```
wyfo: we can see here that there is no point to retrieve or compute the id if the message is not filtered, so the check should be moved at the beginning of the function, like:
```rust
if !self.is_msg_filtered(msg) {
    return true;
}
```
Also, to avoid one additional indentation level, we should use `let Some(id) = ... else { return true; };`

wyfo (changes requested):

PR #2068, opened 2025-07-25
Box WBatch

wyfo:
WBatch is a 56B struct (64B with stats) that is moved and reused all across the pipeline code.
Boxing it makes moves less costly.

### DIFF
--- a/io/zenoh-transport/src/common/batch.rs
+++ b/io/zenoh-transport/src/common/batch.rs
@@ -171,22 +171,6 @@ pub enum Finalize {
     Buffer,
 }
 
-#[derive(Debug, Clone)]
-struct WBatchInner {
-    // The buffer to perform the batching on
-    buffer: BBuf,
-    // The batch codec
-    codec: Zenoh080Batch,
-    // It contains 1 byte as additional header, e.g. to signal the batch is compressed
-    config: BatchConfig,
-    // Statistics related to this batch
-    #[cfg(feature = "stats")]
-    stats: WBatchStats,
-    // an ephemeral batch will not be recycled in the pipeline
-    // it can be used to push a stop fragment when no batch are available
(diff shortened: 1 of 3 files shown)


yellowhatter:
@wyfo and why not just boxing WBatch internals?
Instead of having Box<>.. everywhere in the code, we can just make smth like this:
``` Rust
struct WBacth {
    inner: Box<WBacthInner>
}
```

wyfo:
> why not just boxing WBatch internals?

Because `WBatch` has a few methods, and if I boxed a inner struct in it, I would have to use `self.0` everywhere. So between `self.0` and `Box<WBatch>`, I chose the later — implementing `Deref` is not an option for me here. And instead of wrapping an inner box, I would rather use a type alias, like `BoxedWBtach`, but honestly, I'm don't have issue to use `Box<WBtach>` directly.

yellowhatter:
> > why not just boxing WBatch internals?
> 
> Because `WBatch` has a few methods, and if I boxed a inner struct in it, I would have to use `self.0` everywhere. So between `self.0` and `Box<WBatch>`, I chose the later — implementing `Deref` is not an option for me here. And instead of wrapping an inner box, I would rather use a type alias, like `BoxedWBtach`, but honestly, I'm don't have issue to use `Box<WBtach>` directly.

@wyfo 
`Box<WBtach>` everywhere is less readable and makes a wish to refactor this removing `Box` for those devs who don't know the rationale. As opposed, `Box` in struct :
- provides a way to document the `Box` rationale in one place
- gives flexibility to WBatch layout control (we might want to put some small fields to root struct or do other stuff)
- makes code more readable
- eliminates `Vec<Box<>>` clippy error
- opens an easier way to colocalize WBatch internals in the future (I want to colocalize WBatch fields and other objects in one array struct in memory)

PR #2070, opened 2025-07-28
Optimize KeyExpr::try_from

wyfo:
Previous code was doing a first iteration for chunk split, and a second one for chunk validation. Raw chunk comparison are also integrated into the main iteration. Last but not least, ASCII ranges are used to use only one branch for alphabetical characters.


commons/zenoh-keyexpr/src/key_expr/borrowed.rs
```diff
@@ -726,96 +727,88 @@ impl fmt::Display for keyexpr {
...
-                        "Invalid Key Expr `{}`: `**/*` must be replaced by `*/**` to reach canon-form",
-                        value
-                    ),
-                    _ => {}
+        if value.is_empty() {
+            return Err(KeyExprError::EmptyChunk.err(value));
+        }
+        let bytes = value.as_bytes();
+        let mut i = 0;
+        let mut chunk_start = 0usize.wrapping_sub(1);
```
YuanYuYuan: Why not `usize::MAX`?

commons/zenoh-keyexpr/src/key_expr/borrowed.rs
```diff
@@ -726,96 +727,88 @@ impl fmt::Display for keyexpr {
...
+        let mut chunk_stars = 0;
+        let mut prev_chunk_double_star = false;
+        while i < bytes.len() {
```
YuanYuYuan, line 773: It's better to document the rules.

YuanYuYuan (changes requested):

commons/zenoh-keyexpr/src/key_expr/borrowed.rs
wyfo: Because it's strictly the same, but I find it a lot more explicit written this way: the chunk start is one character behind the first character.

commons/zenoh-keyexpr/src/key_expr/borrowed.rs
wyfo: Indeed, it's a first POC but I need to add some comments.

wyfo: Comments have been added.

commons/zenoh-keyexpr/src/key_expr/borrowed.rs
wyfo: I've changed to have `chunk_start` starting at zero (it's the character after the `/` now)

wyfo:
Here are some quick benchmarks using 2 keyexprs from a typical RMW use case:
```
ke1 = "0/pub_21/describe_parameters/rcl_interfaces::srv::dds_::DescribeParameters_/RIHS01_845b484d71eb0673dae682f2e3ba3c4851a65a3dcfb97bddd82c5b57e91e4cff"
ke2 = "@ros2_lv/0/3563123a6382a559d344b0c864c92223/0/6/SS/%/%/pub_21/%pub_21%describe_parameters/rcl_interfaces::srv::dds_::DescribeParameters_/RIHS01_845b484d71eb0673dae682f2e3ba3c4851a65a3dcfb97bddd82c5b57e91e4cff/:2:,1000:,:,:,,"
```
|     | before | after |
|-----|--------|-------|
| ke1 | 118ns  | 60ns  |
| ke2 | 210ns  | 100ns |

commons/zenoh-keyexpr/src/key_expr/borrowed.rs
```diff
@@ -726,96 +727,129 @@ impl fmt::Display for keyexpr {
...
-    ContainsSharpOrQMark = -7,
-    ContainsUnboundDollar = -8,
+    StarInChunk = -5,
+    DollarAfterDollar = -6,
+    SharpOrQMark = -7,
+    UnboundDollar = -8,
+}
+
+impl KeyExprError {
+    #[cold]
```
evshary, line 742: Good to know the hint :+1:

commons/zenoh-keyexpr/src/key_expr/borrowed.rs
```diff
@@ -1081,6 +1119,7 @@ mod tests {
     #[test_case("demo/example#/test", false; "Contain sharp")]
     #[test_case("demo/example?/test", false; "Contain mark")]
     #[test_case("demo/$/test", false; "Contain unbounded dollar")]
+    #[test_case("demo/$*?/test", false; "Contain mark after dollar start")]
```
evshary: I think `?` is forbidden, so we don't need to test `$*?`.
Instead, testing `$*$` might be a good idea.

commons/zenoh-keyexpr/src/key_expr/borrowed.rs
```diff
@@ -726,96 +727,129 @@ impl fmt::Display for keyexpr {
...
+        let mut chunk_start = 0;
+        // Use a while loop to scan the string because it requires to advance the iteration
+        // manually for some characters, e.g. '$'.
+        let mut i = 0;
+        while i < bytes.len() {
+            match bytes[i] {
+                // In UTF-8, all keyexpr special characters are lesser or equal to '/', except '?'
+                // This shortcut greatly reduce the number of operations for alphanumeric
+                // characters, which are the most common in keyexprs.
+                c if c > b'/' && c != b'?' => i += 1,
```
evshary, line 778: It's a good idea to have this shortcut. I thought we didn't allow some special characters like `%` after adding this check, but it only avoids the unnecessary checks for alphanumeric characters.

commons/zenoh-keyexpr/src/key_expr/borrowed.rs
```diff
@@ -726,96 +727,129 @@ impl fmt::Display for keyexpr {
...
-                            value
-                        );
+                // The first encountered '*' must be at the beginning of a chunk
+                b'*' if i != chunk_start => return Err(StarInChunk.into_err(value)),
+                // When a '*' is match, it means this is a wildcard chunk, possibly with two "**",
+                // which must be followed by a '/' or the end of the string.
+                // So the next character is checked, and the cursor
+                b'*' => match bytes.get(i + 1) {
+                    // Break if end of string is reached.
+                    None => break,
```
evshary, line 793: Perhaps we can add this to the test case to increase the test coverage.

commons/zenoh-keyexpr/src/key_expr/borrowed.rs
```diff
@@ -726,96 +727,129 @@ impl fmt::Display for keyexpr {
...
-                        if chunk_bytes.get(i + 1) == Some(&b'*') {
-                            if chunk_bytes.get(i + 2) == Some(&b'$') {
-                                bail!((KeyExprConstructionError::DollarAfterDollarOrStar)
-                                    "Invalid Key Expr `{}`: `$` is not allowed after `$*`",
-                                    value
-                                )
+                    // If a second '*' is found, the next character must be a slash.
+                    Some(&b'*') => match bytes.get(i + 2) {
+                        // Break if end of string is reached.
+                        None => break,
```
evshary, line 803: Same here

commons/zenoh-keyexpr/src/key_expr/borrowed.rs
```diff
@@ -726,96 +727,129 @@ impl fmt::Display for keyexpr {
...
+                    _ => return Err(StarInChunk.into_err(value)),
+                },
+                // A '$' must be followed by '*'.
+                b'$' if bytes.get(i + 1) != Some(&b'*') => {
+                    return Err(UnboundDollar.into_err(value))
                 }
+                // "$*" has some additional rules to check.
+                b'$' => match bytes.get(i + 2) {
+                    // Break if end of string is reached.
+                    None => break,
```
evshary: Same here

commons/zenoh-keyexpr/src/key_expr/borrowed.rs
evshary: I'm thinking if we should add some special character tests in our test cases to avoid unintentional changes in the future. (Although we are using a blacklist for special characters...)

wyfo: Actually most of the performance improvement comes from this shortcut, the whole PR is based on this.
I don't see why we should prevent unintentional changes in the future, as the keyexpr specification has been fixed since 1.0. You can't for example start forbidding `:` or you will destroy RMW keyexpr model.

commons/zenoh-keyexpr/src/key_expr/borrowed.rs
wyfo: Actually there was a bug here, which I've fixed thanks to your comment.

evshary (approved):
LGTM. I guess the macos failure is a flaky one.

PR #2080, opened 2025-08-04
Apply clippy lints from `rustc 1.91.0-nightly`

fuzzypixelz:
This is useful for two reasons:
1. The code coverage job step uses nightly Rust and is hard to read because of the many warnings (e.g. https://github.com/eclipse-zenoh/zenoh/actions/runs/16727083609/job/47345480281).
2. Using nightly Rust locally (e.g. for `-Zthreads=0`) is cumbersome.

### DIFF
--- a/commons/zenoh-keyexpr/src/key_expr/intersect/mod.rs
+++ b/commons/zenoh-keyexpr/src/key_expr/intersect/mod.rs
@@ -38,22 +38,22 @@ pub trait Intersector<Left, Right> {
 }
 
 pub(crate) mod restriction {
-    use core::ops::Deref;
+    // use core::ops::Deref;
 
-    #[repr(transparent)]
-    #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-    pub struct NoBigWilds<T>(pub T);
+    // #[repr(transparent)]
+    // #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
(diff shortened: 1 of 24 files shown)


commons/zenoh-keyexpr/src/key_expr/intersect/mod.rs
```diff
@@ -38,22 +38,22 @@ pub trait Intersector<Left, Right> {
 }
 
 pub(crate) mod restriction {
-    use core::ops::Deref;
+    // use core::ops::Deref;
```
evshary: Why not just remove them?

fuzzypixelz: There is an entire commented out module that relates to this code and I don't have context for we chose to comment it out instead of leaving it to git history.

fuzzypixelz: https://github.com/eclipse-zenoh/zenoh/blob/23c9b929e1101032a9d8745ab860b12e09211359/commons/zenoh-keyexpr/src/key_expr/intersect/mod.rs#L20-L28

It seems that these modules were deprecated before. I think you're right, I will delete it and add an issue so we can revisit it later.

fuzzypixelz: https://github.com/eclipse-zenoh/zenoh/issues/2083

evshary (approved):

PR #2082, opened 2025-08-05
explicitly specify versions on some dependent crates to make project buildable without Cargo.lock

### LINKED ISSUE #2081: Cargo.toml is incorrect for rust 1.75
### Describe the bug

without Cargo.lock build with rust 1.75 fails. 
This is the cause of issue https://github.com/eclipse-zenoh/zenoh-c/issues/1076

### To reproduce

```sh
rm Cargo.lock
cargo +1.75.0 check
```
result is
```sh
milyin@Mac zenoh % cargo +1.75 check
error: failed to download `base64ct v1.8.0`

Caused by:
  unable to get packages from source

Caused by:
  failed to download replaced source registry `crates-io`

Caused by:
  failed to parse manifest at `/Users/milyin/.cargo/registry/src/index.crates.io-6f17d22bba15001f/base64ct-1.8.0/Cargo.toml`

Caused by:
  feature `edition2024` is required

  The package requires the Cargo feature called `edition2024`, but that feature is not stabilized in this version of Cargo (1.75.0 (1d8b05cdd 2023-11-20)).
  Consider trying a newer version of Cargo (this may require the nightly release).
  See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#edition-2024 for more information about the status of this feature.
```

### System info

-

milyin:
Fixes #2081 

This update adds crate `zenoh-pinned-deps-1-75` to workspace. This crate pins exact versions of dependencies necessary to build zenoh with rust 1.75.

Current build process is unaffected: these crates are already pinned by Cargo.lock which we are keeping in the repository for the same purpose

`zenoh` distributed by crates.io is unaffected: the dependencies in cargo.toml of zenoh itself are not pinned and workspace doesn't exists on crates.io level. No crates depends on this `zenoh-pinned-deps-1-75`, it affects build inside the workspace just by the fact that it's included into it.

The puprose of this `zenoh-pinned-deps-1-75` is:
- allow to safely remove and rebuild Cargo.lock if it's convenient
- allow end user to apply restrictions necessary to build zenoh with rust 1.75 by importing this crate. Now crate is marked as internal, but this can be removed if we decide to make it part of official API

CI test controlling that we can remove Cargo.lock without breaking the build is added


milyin:
Ok, agree. I tried the appproach to put all pinned dependencies to separate crate and conditionallly include it, but this doesn't work too: when crate with pinned dependencies is visible to cargo, the dependencies are processed and pin is applied, no matter is feature enabled or not.
So seems that we have to stay with Cargo.lock for now :-(

milyin:
I think I found not only the solution to the problem, but also discovered that the problem is a bit more serious than just small inconvenience of keeping generated Cargo.lock.

Yes, that's true that the zenoh itself must not lock it's dependencies to avoid conflicts to dependent crates. But as we declare compatibility with Rust 1.75, we should provide users a way to build zenoh published on crates.io with Rust 1.75.
This doesn't work now: I create empty project, use zenoh in it, trying to compile it with rust 1.75 and fail:

Cargo.toml:
```toml
[package]
name = "zenoh_rust175_test"
version = "0.1.0"
edition = "2021"
[dependencies]
zenoh = "1.5.0"
```

```sh
milyin@Mac zenoh_rust175_test % cargo +1.75.0 run
error: failed to download `base64ct v1.8.0`

Caused by:
  unable to get packages from source

Caused by:
  failed to download replaced source registry `crates-io`

Caused by:
  failed to parse manifest at `/Users/milyin/.cargo/registry/src/index.crates.io-6f17d22bba15001f/base64ct-1.8.0/Cargo.toml`

Caused by:
  feature `edition2024` is required
```

The problem is that base64ct v1.8.0 doesn't support Rust 1.75

User have to manually tweak dependencies to make it compile. But we can provide the crate with already tweaked dependencies:
```toml
version = "0.1.0"
edition = "2021"
[dependencies]
zenoh = "1.5.0"
zenoh-pinned-deps = { path="../../zenoh/commons/zenoh-pinned-deps" } # will be just "1.x.0" after publishing of course
```

```sh
milyin@Mac zenoh_rust175_test % cargo +1.75.0 run
    Updating crates.io index
    Finished dev [unoptimized + debuginfo] target(s) in 2.33s
     Running `target/debug/zenoh_rust175_test`
```

```sh
milyin@Mac zenoh_rust175_test % cargo tree -i base64ct
base64ct v1.6.0
├── pem-rfc7468 v0.7.0
│   └── der v0.7.10
│       ├── pkcs1 v0.7.5
│       │   └── rsa v0.9.8
│       │       └── zenoh-transport v1.5.0
│       │           └── zenoh v1.5.0
│       │               └── zenoh_rust175_test v0.1.0 (/Users/milyin/ZS/tmp/zenoh_rust175_test)
│       ├── pkcs8 v0.10.2
│       │   ├── pkcs1 v0.7.5 (*)
│       │   └── rsa v0.9.8 (*)
│       └── spki v0.7.3
│           ├── pkcs1 v0.7.5 (*)
│           ├── pkcs8 v0.10.2 (*)
│           └── rsa v0.9.8 (*)
└── zenoh-pinned-deps v1.5.0 (/Users/milyin/ZS/zenoh/commons/zenoh-pinned-deps)
    └── zenoh_rust175_test v0.1.0 (/Users/milyin/ZS/tmp/zenoh_rust175_test)
```

The correct version of base64ct v1.6.0 is selected.

This also solves our problem with Cargo.lock as a source - now important dependency information is moved out from Cargo.lock to `zenoh-pinned-deps`, so Cargo.lock can be safely regenerated from scratch when necessary. The `zenoh-pinned-deps` is not included into other crates dependencies, so crates published on crates.io are not affected by version locks, they are applied inside workspace only

fuzzypixelz (commented):
I think this works nicely and is quite an clever hack. I would say that if we need to use `zenoh-pinned-deps` to build with 1.75 then the MSRV is simply not 1.75. We might want to mention that `zenoh-pinned-deps` exists and how to use it in the README.

To overcome the need to support 1.75 we should revisit https://github.com/ros/rosdistro/pull/44640 and build RMW Zenoh with `rustup` instead of the system toolchain.

And in order to better manage the MSRV, we could revisit #1026. It has stalled because it's quite complex:
- MSRV depends on the build target
- MSRV depends on whether you delete the lockfile or not. This is what you encountered here
- Each crate in the workspace has its own distinct MSRV

PR #2097, opened 2025-08-22
refactor and optimize downsampling interceptor

wyfo:
The mutex was used for read only, so it has been removed.

---
## 🏷️ Label-Based Checklist

**Labels:** `enhancement`

## ✨ Enhancement Requirements

**Instructions:**

### DIFF
--- a/zenoh/src/net/routing/interceptor/downsampling.rs
+++ b/zenoh/src/net/routing/interceptor/downsampling.rs
@@ -195,7 +195,10 @@ impl DownsamplingFilters {
             },
             NetworkBodyMut::Request(_) => self.query,
             NetworkBodyMut::Response(_) => self.reply,
-            _ => false,
+            NetworkBodyMut::ResponseFinal(_) => false,
+            NetworkBodyMut::Interest(_) => false,
+            NetworkBodyMut::Declare(_) => false,
+            NetworkBodyMut::OAM(_) => false,
         }
     }
 }
--- a/zenoh/src/net/routing/interceptor/downsampling.rs
+++ b/zenoh/src/net/routing/interceptor/downsampling.rs
@@ -19,10 +19,10 @@
 //! [Click here for Zenoh's documentation](https://docs.rs/zenoh/latest/zenoh)
 
 use std::{
-    collections::{HashMap, HashSet},
+    collections::HashSet,
     sync::{
         atomic::{AtomicBool, Ordering},
-        Arc, Mutex,
+        Mutex,
     },
 };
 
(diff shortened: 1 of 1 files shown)


zenoh/src/net/routing/interceptor/downsampling.rs
```diff
@@ -296,41 +273,35 @@ const NANOS_PER_SEC: f64 = 1_000_000_000.0;
...
 impl DownsamplingInterceptor {
     pub fn new(
-        messages: Arc<DownsamplingFilters>,
+        filters: DownsamplingFilters,
         rules: &NEVec<DownsamplingRuleConf>,
-        #[cfg(feature = "stats")] flow: InterceptorFlow,
+        flow: InterceptorFlow,
         #[cfg(feature = "stats")] stats: Arc<TransportStats>,
     ) -> Self {
         let mut ke_id = KeBoxTree::default();
```
oteffahi, line 286: For future reference: we should move the computation of the key-expr tree outside of the interceptor, and wrap it in an `Arc` in order to reuse the same tree for both Ingree/Egress interceptors constructed from the same Interceptor Factory.

zenoh/src/net/routing/interceptor/downsampling.rs
```diff
@@ -168,72 +156,64 @@ impl InterceptorFactoryTrait for DownsamplingInterceptorFactory {
...
+    fn is_msg_filtered(&self, msg: &mut NetworkMessageMut) -> bool {
+        match msg.body {
+            NetworkBodyMut::Push(Push { payload, .. }) => match payload {
+                PushBody::Put(_) => self.put,
+                PushBody::Del(_) => self.delete,
+            },
+            NetworkBodyMut::Request(_) => self.query,
+            NetworkBodyMut::Response(_) => self.reply,
+            _ => false,
+        }
```
oteffahi, line 202: The match with all enum variants was intentional to cause a compile-time error if the enum is ever changed. IMO we should keep that.

oteffahi (approved):
LGTM.
Left a few nitpick comments.

zenoh/src/net/routing/interceptor/downsampling.rs
wyfo: Done

zenoh/src/net/routing/interceptor/downsampling.rs
wyfo: > For future reference

Yes, because I don't want to do it now

oteffahi (approved):

PR #2107, opened 2025-08-27
additional pinned 1.75 dependencies

milyin:
Several crates were recently got new versions incompatible with rust 1.75. Locking them to last compatible versions
```
cargo_metadata = "=0.19.0"
cargo-platform = "=0.1.8"
safe_arch = "=0.7.4"
wide = "=0.7.33"
```


milyin:
not needed, since https://github.com/eclipse-zenoh/zenoh/pull/2114 this crate should contain only depencencies needed by zenoh itself, it's not it's purpose to serve as helper for other zenoh projects

PR #2110, opened 2025-08-29
Do not allow links between instances with the same zid

### LINKED ISSUE #1508: Peers and routers can connect to themselves
### Describe the bug

It's possible to configure a router or a peer to listen to and to connect to the same endpoint. This breaks pub/sub: all publications are received twice (except the first one for some reason?).

### To reproduce

1. Configure an peer or a router with `connect/endpoints:["tcp/localhost:10000"]` and `listen/endpoints:["tcp/localhost:10000"]`
2. Start a subscriber on key-expression `k`
3. Start a publisher on key-expression `k` with payloads `0, 1, 2, 3, ...`

The subscriber will receive `0, 1, 1, 2, 2, 3, 3, ...`

### System info

- ref: 82869fb87

DenisBiryukov91:
do not allow links between instances with the same zid
resolves #1508

### DIFF
--- a/commons/zenoh-protocol/src/transport/close.rs
+++ b/commons/zenoh-protocol/src/transport/close.rs
@@ -27,6 +27,7 @@ pub mod reason {
     pub const MAX_LINKS: u8 = 0x04;
     pub const EXPIRED: u8 = 0x05;
     pub const UNRESPONSIVE: u8 = 0x06;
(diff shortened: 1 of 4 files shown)


fuzzypixelz:
`zenoh_session_multicast` fails quite consistently on macOS; merging as is.

PR #2112, opened 2025-08-29
use IntHashSet for RouteBuilder

wyfo:


fuzzypixelz:
`zenoh_session_multicast` fails quite consistently on macOS; merging as is.

PR #2113, opened 2025-08-29
Implicit SHM optimization

yellowhatter:


yellowhatter:
The CI issues are upstream and not related to this PR

DEFAULT_CONFIG.json5
```diff
@@ -730,6 +730,15 @@
       /// - "init": SHM subsystem internals will be initialized upon Session opening. This setting sacrifices
       /// startup time, but guarantees no latency impact when first SHM buffer is processed.
       mode: "lazy",
+      transport_optimization: {
+          /// Enables transport optimization for large messages (default `true`).
+          /// Implicitly puts large messages into shared memory for transports with SHM-compatible connection.
+          enabled: true,
+          /// SHM cache size in bytes used for transport optimization (default `16 * 1024 * 1024`).
+          cache_size: 16777216,
```
fuzzypixelz: I find it confusing to call this a "cache". There is no key-value store here, nor any notion of a cache hit/miss. I would simply call it `buffer_size`.

io/zenoh-transport/src/common/pipeline.rs
```diff
@@ -835,13 +843,17 @@ impl TransmissionPipelineStatus {
     }
 }
 
+#[derive(Clone)]
+struct Waits {
```
fuzzypixelz, line 846: I propose splitting `wait_before_drop` into `wait_before_drop` and `max_wait_before_drop_fragments` and adding a small comment saying that `u32` represents microseconds.

DEFAULT_CONFIG.json5
wyfo: I think that the word "pool" is more suited than "buffer" in the context of a memory allocator, so I would advise to use `pool_size` instead.

io/zenoh-transport/src/common/pipeline.rs
wyfo: Why not just store duration instead? To avoid always converting them back to duration in the code?

io/zenoh-transport/src/multicast/establishment.rs
```diff
@@ -77,10 +77,21 @@ pub(crate) async fn open_link(
...
         sn_resolution,
         initial_sns,
-        #[cfg(feature = "shared-memory")]
-        is_shm: manager.config.multicast.is_shm,
     };
-    let ti = Arc::new(TransportMulticastInner::make(manager.clone(), config)?);
+
+    #[cfg(feature = "shared-memory")]
+    let shm_context = manager.state.shm_context.as_ref().map(|context| {
+        use crate::shm_context::MulticastTransportShmContext;
```
wyfo: Why importing the type here instead using the fully qualified path directly?

DEFAULT_CONFIG.json5
yellowhatter: But this is really a memory used for caching purpose (in future that will be different caching purposes) on-demand. I think calling it `buffer` will be misleading

fuzzypixelz: What could the caching purpose be if there is no store?

yellowhatter: Okay, let's call it  `arena_size` because it is technically an arena for allocations
I don't like the 'buffer" term here because the purpose and data stored here is heterogeneous 

UPD: Joseph suggested `pool_size`

io/zenoh-transport/src/unicast/lowlatency/tx.rs
```diff
@@ -29,9 +27,12 @@ impl TransportUnicastLowlatency {
     #[inline(always)]
     pub(crate) fn internal_schedule(&self, mut msg: NetworkMessageMut) -> ZResult<()> {
         #[cfg(feature = "shared-memory")]
-        {
-            if let Err(e) = map_zmsg_to_partner(&mut msg, &self.config.shm) {
-                bail!("Failed SHM conversion: {}", e);
```
wyfo, line 34: Why removing the error?

io/zenoh-transport/src/common/pipeline.rs
yellowhatter: okay, I agree, will go back to Duration!

io/zenoh-transport/src/unicast/lowlatency/tx.rs
yellowhatter: UPD: I made the whole `map_zmsg_to_partner` infallible

io/zenoh-transport/src/multicast/establishment.rs
yellowhatter: Fixed!

fuzzypixelz (approved):

PR #2115, opened 2025-09-01
Add put and delete downsampling message types

### LINKED ISSUE #2111: Config: downsampling/messages inconsistency
### Describe the feature

The `downsampling/messages` config accepts `"push"` which corresponds to `"put"` and `"delete"` messages.
While in other configs  (`access_control` and `low_pass_filter`), `"put"` and `"delete"` are accepted, but not `"push".`

For downsampling, I think it's important to be able to downsample the `put` messages, but NOT the `delete` messages. For publications to a storage, a user probably would probably like to downsample the publications, but never lose the `delete` messages.

My suggestion for consistency and keep backward compatibility:
- add `Put` and `Delete` to `DownsamplingMessage`
- add `Push` to `AclMessage` and `LowPassFilterMessage`

Bonus: add an `All` case for all messages configs.

DenisBiryukov91:
add put and delete downsampling message types;
undocument push downsampling message type;

Resolves #2111

### DIFF
--- a/commons/zenoh-config/src/lib.rs
+++ b/commons/zenoh-config/src/lib.rs
@@ -19,6 +19,8 @@
 //! [Click here for Zenoh's documentation](https://docs.rs/zenoh/latest/zenoh)
 //!
 //! Configuration to pass to `zenoh::open()` and `zenoh::scout()` functions and associated constants.
+#![allow(deprecated)]
+
 pub mod defaults;
 mod include;
 pub mod qos;
@@ -109,6 +111,7 @@ pub enum InterceptorFlow {
 #[serde(rename_all = "snake_case")]
 pub enum DownsamplingMessage {
     Delete,
+    #[deprecated = "Use `Put` or `Delete` instead."]
     Push,
     Put,
     Query,
--- a/commons/zenoh-config/src/lib.rs
+++ b/commons/zenoh-config/src/lib.rs
@@ -108,7 +108,9 @@ pub enum InterceptorFlow {
 #[derive(Clone, Copy, Debug, Serialize, Deserialize, Eq, PartialEq)]
 #[serde(rename_all = "snake_case")]
 pub enum DownsamplingMessage {
+    Delete,
     Push,
+    Put,
     Query,
     Reply,
 }
--- a/commons/zenoh-config/src/lib.rs
+++ b/commons/zenoh-config/src/lib.rs
@@ -147,6 +147,19 @@ pub struct DownsamplingItemConf {
     pub flows: Option<NEVec<InterceptorFlow>>,
 }
 
+fn downsampling_validator(d: &Vec<DownsamplingItemConf>) -> bool {
+    for item in d {
+        if item
+            .messages
+            .iter()
+            .any(|m| *m == DownsamplingMessage::Push)
+        {
+            tracing::warn!("In 'downsampling/messages' configuration: 'push' is deprecated and may not be supported in future versions, use 'put' and/or 'delete' instead");
+        }
+    }
+    true
+}
+
 #[derive(Serialize, Debug, Deserialize, Clone)]
 #[serde(deny_unknown_fields)]
 pub struct LowPassFilterConf {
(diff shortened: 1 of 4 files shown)


JEnoch:
Does this PR imply that it was decided to deprecate the `"push"` message for `downsampling` ?
If yes, I think there should be a warning log when used in config, and a clear description in next CHANGELOG.

commons/zenoh-config/src/lib.rs
```diff
@@ -108,7 +108,9 @@ pub enum InterceptorFlow {
 #[derive(Clone, Copy, Debug, Serialize, Deserialize, Eq, PartialEq)]
 #[serde(rename_all = "snake_case")]
 pub enum DownsamplingMessage {
+    Delete,
     Push,
```
JEnoch, line 115: Can `Push` be tagged as deprecated here ?

zenoh/src/net/routing/interceptor/downsampling.rs
```diff
@@ -177,7 +181,12 @@ impl From<&NEVec<DownsamplingMessage>> for DownsamplingFilters {
         let mut res = Self::default();
         for v in value {
             match v {
-                DownsamplingMessage::Push => res.push = true,
+                DownsamplingMessage::Delete => res.put = true,
+                DownsamplingMessage::Push => {
```
JEnoch, line 186: Maybe add a deprecation warning log here.

commons/zenoh-config/src/lib.rs
DenisBiryukov91: It can surely be, but it will produce warnings that will be treated as errors in most builds - we better just remove `Push` entirely in this case.

JEnoch: The [PublicationCache](https://github.com/eclipse-zenoh/zenoh/blob/8761235f78b04117c2213c89ba2a32d44cd930da/zenoh-ext/src/publication_cache.rs#L34) and [QueryingSubscriber](https://github.com/eclipse-zenoh/zenoh/blob/8761235f78b04117c2213c89ba2a32d44cd930da/zenoh-ext/src/querying_subscriber.rs#L36) are already tagged as deprecated, but still supported in [zenoh-c](https://github.com/eclipse-zenoh/zenoh-c/blob/b45367e0dce6687eff915d3571d536a3b78682c1/src/publication_cache.rs#L14). No warning or error building it. 
The benefit is that it's marked as deprecated in [Rust doc](https://docs.rs/zenoh-ext/latest/zenoh_ext/struct.PublicationCache.html).

zenoh/src/net/routing/interceptor/downsampling.rs
DenisBiryukov91: fixed

commons/zenoh-config/src/lib.rs
DenisBiryukov91: fixed

zenoh/src/net/routing/interceptor/downsampling.rs
```diff
@@ -177,7 +181,14 @@ impl From<&NEVec<DownsamplingMessage>> for DownsamplingFilters {
         let mut res = Self::default();
         for v in value {
             match v {
-                DownsamplingMessage::Push => res.push = true,
+                DownsamplingMessage::Delete => res.put = true,
+                #[allow(deprecated)]
+                DownsamplingMessage::Push => {
+                    tracing::warn!("DownsamplingMessage::Push is deprecated and may not be supported in future versions, use DownsamplingMessage::Put and/or DownsamplingMessage::Delete instead");
```
JEnoch: Such a message would be more clear for most of users dealing only with the config file:
`"In 'downsampling/messages' configuration: 'push' is deprecated and may not be supported in future versions, use 'put' and/or 'delete' instead"`

DenisBiryukov91: fixed

JEnoch (approved):

Issue #2122, opened 2025-09-04
Build failure with token-cell 1.6.0 - zenoh-keyexpr has breaking borrowing errors


nisseknudsen:
### Describe the bug

zenoh 1.5.0 fails to compile when using token-cell 1.6.0 (released September 2, 2025). The issue occurs in `zenoh-keyexpr` with numerous borrowing/mutability errors related to `TokenCell::get()`.

### To reproduce

1. Create a new project with `zenoh = "1.5.0"`
2. Delete `Cargo.lock` if present
3. Run `cargo build`

*Expected behavior*: Build succeeds
*Actual behavior*: Build fails with multiple errors in `zenoh-keyexpr-1.5.0/src/keyexpr_tree/arc_tree.rs` and related files:

- `error[E0596]: cannot borrow as mutable`
- `error[E0596]: cannot borrow data in an Arc as mutable`
- Issues with `TokenCell::get()` and `Arc<TokenCell<...>>`

*Workaround*: Pin token-cell to 1.5.0 in Cargo.toml:

```toml
[dependencies]
zenoh = "1.5.0"
token-cell = "=1.5.0"
```

### System info

zenoh version: 1.5.0
rustc version: 1.88

Can repro on both MacOS (aarch64) and Ubuntu 24.04 (x86_64)

evshary:
Hi @nisseknudsen 

Thank you for reporting the issue. This is because of the breaking change in `token-cell` 1.6.0.
If you don't have `Cargo.lock`, it will use the latest version of `token-cell` below the `2.0.0`
Now `token-cell` bumps to `2.0.0`, and `1.6.0` is yanked, so it will not be an issue anymore.
Please let me know if you still face the issue.

PR #2123, opened 2025-09-04
Fix the token-cell version.

evshary:
~~As https://github.com/eclipse-zenoh/zenoh/issues/2122 reported, bumping `token-cell` to 1.6.0 will break the build process.~~
~~I submitted a fix to upstream (https://github.com/p-avital/token-cell-rs/pull/4), but I think we should fix the version for the time being.~~

I've bumped the version to 2.0.0 now, ~~but it might still cause issues if users build the old version without `Cargo.lock`.~~

`token-cell` is yanked now, and it should not be an issue anymore.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -187,14 +187,8 @@ jobs:
       - name: Check unused dependencies
         run: cargo machete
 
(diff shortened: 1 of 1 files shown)


evshary:
Will be addressed there
https://github.com/eclipse-zenoh/zenoh/pull/2126

PR #2133, opened 2025-09-08
add missing impl for PosixShmProviderBackendTalcBuilder

wyfo:


wyfo:
After discussion with @yellowhatter, this is intended to not have the builder consuming the `MemoryLayout`, which doesn't implement `Copy`.

PR #2135, opened 2025-09-08
Fix treatment of complete field for same ke queryable on the same face

### LINKED ISSUE #1616: QueryableInfoType.complete is not properly used for queryables with identical key expressions on the same face.
### Describe the bug

QueryableInfoType.complete (in SessionContext and other structs) is not properly updated when declaring/undeclaring queryables if more than one queryable with the same key expression is declared for the same face.  I.e on declaration of new queryable the .complete field is just overwritten to correspond to that of newly declared queryable, and when undeclaring queryable, this field is ignored if there are still remaining queryable with the same key expression.

### To reproduce

The bug is easy to reproduce when using querier matching status/listener api introduced by #1591. 
1. declare incomplete queryable on "some/keyexpr"
2. declare complete queryable on same keyexpr
3. declare complete querier on same keyexpr
4. undeclare complete queryable
5. matching status for querier will be true (despite the fact that the only remaining queryable is incomplete)

### System info

any

DenisBiryukov91:
Fixes #1616

### DIFF
--- a/zenoh/src/net/routing/dispatcher/resource.rs
+++ b/zenoh/src/net/routing/dispatcher/resource.rs
@@ -163,6 +163,25 @@ impl SessionContext {
             e_interceptor_cache: InterceptorCache::empty(),
         }
     }
+
+    pub(crate) fn remove_queryable(&mut self, complete_only: bool) {
+        if let Some(q) = &mut self.qabl {
+            if complete_only {
+                q.complete = false;
+            } else {
+                self.qabl = None;
+            }
+        }
+    }
+
+    pub(crate) fn add_queryable(&mut self, q_info: &QueryableInfoType) {
+        if let Some(q) = &mut self.qabl {
+            q.complete = q.complete | q_info.complete;
+            q.distance = q.distance.min(q_info.distance);
+        } else {
+            self.qabl = Some(*q_info);
+        }
+    }
 }
 
 /// Global version number for route computation.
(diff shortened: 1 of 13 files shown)


zenoh/src/net/routing/dispatcher/resource.rs
```diff
@@ -178,6 +178,15 @@ impl SessionContext {
     }
 }
 
+pub(crate) fn merge_qabl_infos(
+    mut this: QueryableInfoType,
+    info: &QueryableInfoType,
+) -> QueryableInfoType {
+    this.complete = this.complete || info.complete;
+    this.distance = std::cmp::min(this.distance, info.distance);
```
OlivierHecart: As you suggested in another discussion it would be better to keep the smallest distance of the completes Queryables if there are some.

```suggestion
    this.distance = match (this.complete, info.complete) {
        (true, true) | (false, false) => std::cmp::min(this.distance, info.distance),
        (true, false) => this.distance,
        (false, true) => info.distance,
    };
```

zenoh/src/net/routing/dispatcher/resource.rs
```diff
@@ -178,6 +178,15 @@ impl SessionContext {
     }
 }
 
+pub(crate) fn merge_qabl_infos(
```
OlivierHecart: I would move both those changes (`merge_qabl_infos` and `update_queryable_info`) in dispatcher/queries.rs.

zenoh/src/net/routing/dispatcher/resource.rs
DenisBiryukov91: done

zenoh/src/net/routing/dispatcher/resource.rs
DenisBiryukov91: done

PR #2136, opened 2025-09-09
rework shm API and implementation

wyfo:


commons/zenoh-shm/src/api/provider/memory_layout.rs
```diff
@@ -149,87 +158,42 @@ impl TryFrom<(usize, AllocAlignment)> for MemoryLayout {
...
+    pub fn for_value(_: &T) -> Self {
+        Self::new()
     }
 }
 
-/// A statically-known layout WITHOUT type information.
-///
-/// Statically-known layouts are always correct, zero-sized & zero-cost.
-#[zenoh_macros::unstable_doc]
-pub struct StaticLayout<T> {
```
yellowhatter, line 208: `StaticLayout` was needed to express "allocate me untyped buffer with size and alignment similar to T"
AFAIU there is no more way to do that

wyfo: Yes there is: you can allocated a typed buffer and then retrieve the inner untyped buffer.

yellowhatter: Okay, I see, nice way to simplify

wyfo: Simplification was indeed the goal.

yellowhatter:
One more thing, there is no way to make things like this:
```
let comprehensive =
                PosixShmProviderBackend::builder(TypedLayout::<SharedByteData>::new()).wait()?;
```
In other words, I want to create a provider to fit one exact big structure to use it then. This is the exact case of "SHM channel" usecase and already used in in @kydos 's zshm example

wyfo:
As discussed `PosixShmProviderBackend::builder(MemoryLayout::for_type::<SharedByteData>())` does the job.

PR #2138, opened 2025-09-10
fix(links): correct IPv6 address parsing for TLS and QUIC locators

steils:
TLS and QUIC protocols sent malformed IPv6 locators like "tls/[:7447" instead of "tls/[::1]:7447" in HELLO messages due to incorrect string formatting in `get_*_host()` functions.
It's working correctly for TCP and UDP though.

Fix it by replacing manual formatting with `local_addr.to_string()` as in TCP and UDP.


wyfo:
Just use `rsplit_once`  instead of `split` in host computation

wyfo:
In fact, I'm stupid, `rsplit` alone was better than `rsplit_once`.

PR #2146, opened 2025-09-15
add stats per key interceptor

wyfo:
I chose in the end to use an interceptor instead of reusing the existing stat incrementing code because it was a lot easier to reuse the interceptor caching system than build a new one (that must support runtime invalidation and regeneration).


wyfo:
Superseded by #2284

PR #2147, opened 2025-09-15
MatchingListener deadlock regression test

OlivierHecart:
A test that would hang without this merged PR: https://github.com/eclipse-zenoh/zenoh/pull/2124

### DIFF
--- a/zenoh/tests/matching.rs
+++ b/zenoh/tests/matching.rs
@@ -240,3 +240,28 @@ async fn zenoh_publisher_matching_status() -> ZResult<()> {
     zenoh_publisher_matching_status_inner(Locality::SessionLocal, false).await;
     Ok(())
 }
+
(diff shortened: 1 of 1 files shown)


zenoh/tests/matching.rs
```diff
@@ -240,3 +243,32 @@ async fn zenoh_publisher_matching_status() -> ZResult<()> {
...
+            .declare_querier("zenoh_matching_listener_drop_deadlock")
+            .await
+            .unwrap(),
+    );
+    let matching_listener = querier
+        .matching_listener()
+        .callback({
+            let querier = querier.clone();
+            move |_| {
+                querier.key_expr();
```
DenisBiryukov91: maybe `println!(querier.key_expr())` to ensure that it is not being optimized out by compiler ?

DenisBiryukov91 (approved):

PR #2149, opened 2025-09-16
Fix `get_interface_names_by_addr` implementation for IPv4-mapped IPv6 addresses

### LINKED ISSUE #2143: Interfaces in `access_control` doesn't work
### Describe the bug

`access_control` works well when the subject doesn't include `interfaces`.
However, when we add `interfaces`, then the `access_contol` will become invalid.

### To reproduce

1. Run the router with the following config (You can list all the interfaces you have).

```json5
access_control: {
  "enabled": true,
  "default_permission": "allow",
  "rules":
  [
    {
      "id": "rule1",
      "messages": [
        "put", "delete", "declare_subscriber",
        "query", "reply", "declare_queryable",
        "liveliness_token", "liveliness_query", "declare_liveliness_subscriber",
      ],
      "flows":["egress","ingress"],
      "permission": "deny",
      "key_exprs": [
        "camera/points"
      ],
    },
  ],
  "subjects":
  [
    {
      "id": "subject1",
      "interfaces": [
        "lo0",
        "en0",
      ],
    },
  ],
  "policies":
  [
    {
      "id": "policy1",
      "rules": ["rule1"],
      "subjects": ["subject1"],
    },
  ]
},
```

2. Run `z_pub -k "camera/points"`

3. Run `z_sub -m client -k "camera/points"`

4. You can see that z_sub can receive the data. If we remove `interfaces`, then z_sub will not receive anything.

### System info

Ubuntu: 22.04
Zenoh commit: 5ea9d618035f302603e582e26616a1e1811207c4

oteffahi:
This PR fixes an issue with interface name lookup when the address is an IPv4-mapped IPv6 address.
This notably occurs when connecting to an IPv6 listener (ex: `tcp/[::]:7447`) via an IPv4 endpoint.

Closes #2143

### DIFF
--- a/zenoh/tests/acl.rs
+++ b/zenoh/tests/acl.rs
@@ -77,6 +77,13 @@ async fn test_acl_liveliness() {
     test_liveliness_deny_allow_query(27450).await;
 }
 
+#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+async fn test_acl_interface_names() {
+    zenoh::init_log_from_env_or("error");
+
+    test_pub_sub_network_interface(27451).await;
+}
+
 async fn get_basic_router_config(port: u16) -> Config {
     let mut config = Config::default();
     config.set_mode(Some(WhatAmI::Router)).unwrap();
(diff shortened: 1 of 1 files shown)


evshary (commented):
I'm just wondering how the test can help us catch the issue.
Perhaps, we should have two cases: One is using `lo`, and we can receive the data; another one is `ethx` (or any other name), and we can't receive the data.

oteffahi:
> I'm just wondering how the test can help us catch the issue. Perhaps, we should have two cases: One is using `lo`, and we can receive the data; another one is `ethx` (or any other name), and we can't receive the data.

The issue in question was that no interface was returned for the transport in question, so it wouldn't even match the loopback interface if the issue is present. I think the test added covers this case. If I'm missing something please elaborate.

PR #2152, opened 2025-09-19
Fix bug leading to reliable messages drop when sent along best effort messages

wyfo:
zenoh-buffers traits were implemented on mutable reference instead of the type itself, which is an anti-pattern and prevented to use `FrameReader<ZSlice>` (and `FrameReader<&mut ZSlice>` doesn't work for lifetime reasons). That's why I've rewritten them.

### DIFF
--- a/commons/zenoh-buffers/tests/readwrite.rs
+++ b/commons/zenoh-buffers/tests/readwrite.rs
@@ -30,6 +30,7 @@ const WBSN: [u8; 4] = [u8::MAX, u8::MAX, u8::MAX, u8::MAX];
 macro_rules! run_write {
     ($buffer:expr) => {
         println!(">>> Write");
+        #[allow(unused_mut)]
         let mut writer = $buffer.writer();
 
         writer.write_u8(WBS0).unwrap();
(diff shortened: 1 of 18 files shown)


commons/zenoh-buffers/tests/readwrite.rs
```diff
@@ -30,6 +30,7 @@ const WBSN: [u8; 4] = [u8::MAX, u8::MAX, u8::MAX, u8::MAX];
 macro_rules! run_write {
     ($buffer:expr) => {
         println!(">>> Write");
+        #[allow(unused_mut)]
```
OlivierHecart, line 33: Why not remove the mut ?
(Same below)

examples/examples/z_put.rs
```diff
@@ -27,7 +31,18 @@ async fn main() {
 
     println!("Putting Data ('{key_expr}': '{payload}')...");
     // Refer to z_bytes.rs to see how to serialize different types of message
-    session.put(&key_expr, payload).await.unwrap();
+    for i in 0..usize::MAX {
```
OlivierHecart: The `z_put` example should remain "1 shot".
(Btw, the `if i % 2` thing is an interesting test but has nothing to do in an example imho)

commons/zenoh-buffers/tests/readwrite.rs
wyfo: Because it won't compile. The macro is used with different types, some do require `mut`, other don't.

examples/examples/z_put.rs
wyfo: It's an unwanted committed file... my bad.

PR #2153, opened 2025-09-19
pin toml_parser dependency for 1.75 compatibility

wyfo:


commons/zenoh-pinned-deps-1-75/Cargo.toml
```diff
@@ -40,6 +40,7 @@ potential_utf = "=0.1.0"
 static_init = "=1.0.3"
 time = "=0.3.41"
 tinystr = "=0.8.0"
+toml_parser = "1.0.2"
```
OlivierHecart:
```suggestion
toml_parser = "=1.0.2"
```

PR #2158, opened 2025-09-23
fix client mode with multiple connect endpoints

### LINKED ISSUE #1602: Problem in client mode with multiple connect endpoints and non zero connect timeout_ms
### Describe the bug

Client fail at startup if configured with multiple connect endpoints and non zero connect timeout_ms when the first endpoint is unreachable.

### To reproduce

1. `zenohd`
2. `z_sub -m client --cfg=connect/timeout_ms:1000 --cfg=connect/endpoints:[\"tcp/127.0.0.1:7448\",\"tcp/127.0.0.1:7447\"]`

Fails

### System info

Zenoh 1.0.2

DenisBiryukov91:
ffix client mode with multiple connect endpoints in case of finite connection timeout
Resolves #1602

### DIFF
--- a/zenoh/src/net/runtime/orchestrator.rs
+++ b/zenoh/src/net/runtime/orchestrator.rs
@@ -865,41 +865,7 @@ impl Runtime {
     }
 
     async fn peer_connector_retry(&self, peer: EndPoint) -> ZResult<ZenohIdProto> {
-        let retry_config = self.get_connect_retry_config(&peer);
-        let mut period = retry_config.period();
-        let cancellation_token = self.get_cancellation_token();
-        loop {
-            tracing::trace!("Trying to connect to configured peer {}", peer);
-            let endpoint = peer.clone();
-            tokio::select! {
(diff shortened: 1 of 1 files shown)


zenoh/src/net/runtime/orchestrator.rs
```diff
@@ -783,42 +791,81 @@ impl Runtime {
...
     }
 
-    async fn peer_connector_retry(&self, peer: EndPoint) -> ZResult<ZenohIdProto> {
-        let retry_config = self.get_connect_retry_config(&peer);
-        let mut period = retry_config.period();
-        let cancellation_token = self.get_cancellation_token();
-        loop {
-            tracing::trace!("Trying to connect to configured peer {}", peer);
-            let endpoint = peer.clone();
+    async fn peers_exclusive_connector_retry(&self, peers: Vec<EndPoint>) -> ZResult<ZenohIdProto> {
```
milyin, line 794: Do we really need `Vec` here? Maybe slice is enough? This will let avoid constructing a vector in `peer_connector_retry`

PR #2160, opened 2025-09-23
Respect liveliness subscriber history when processing token declarations

### LINKED ISSUE #2071: When declaring a liveliness subscriber with history=true, the callback for known token is called on all subscribers independently of their history option
### Describe the bug

When declaring a liveliness subscriber with history=true, the callback for known token is called on all subscribers independently of their history option

### To reproduce

declare liveliness token on "token_ke"
create liveliness subscriber on "token_ke", with history = false,
create another liveliness subscriber on "token_ke" with history = true;

both subscriber callbacks will be triggered (although according to docs on the second one should)

### System info

Any

oteffahi:
Closes #2071

### DIFF
--- a/zenoh/src/api/session.rs
+++ b/zenoh/src/api/session.rs
@@ -338,6 +338,7 @@ impl SessionState {
             key_expr: key_expr.clone().into_owned(),
             origin,
             callback,
+            history: None,
         };
 
         let declared_sub = origin != Locality::SessionLocal;
@@ -1709,6 +1710,7 @@ impl SessionInner {
             key_expr: key_expr.clone().into_owned(),
             origin,
             callback: callback.clone(),
+            history: Some(history),
         };
 
         let sub_state = Arc::new(sub_state);
@@ -1975,6 +1977,7 @@ impl SessionInner {
         wire_expr: &WireExpr,
         qos: push::ext::QoSType,
         msg: impl FnOnce() -> &'a mut PushBody,
+        liveliness_interest_current: Option<bool>, // Some(_) for liveliness sub, None otherwise
         #[cfg(feature = "unstable")] reliability: Reliability,
     ) {
         let mut callbacks = SingleOrVec::default();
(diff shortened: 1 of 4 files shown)


zenoh/src/api/session.rs
```diff
@@ -1975,17 +1977,25 @@ impl SessionInner {
...
         #[cfg(feature = "unstable")] reliability: Reliability,
     ) {
         let mut callbacks = SingleOrVec::default();
         let state = zread!(self.state);
         if state.primitives.is_none() {
             return; // Session closing or closed
         }
+        // For LivelinessSubscriber, Interest::Current Tokens should only run liveliness subs callbacks with history=true
+        let filter_by_interest =
+            if matches!(kind, SubscriberKind::LivelinessSubscriber) && msg_interest_current {
```
OlivierHecart: If history flags are properly set, we should not need to match the subscriber kind

zenoh/src/api/session.rs
```diff
@@ -1975,17 +1977,25 @@ impl SessionInner {
         wire_expr: &WireExpr,
         qos: push::ext::QoSType,
         msg: impl FnOnce() -> &'a mut PushBody,
+        msg_interest_current: bool, // Currently only relevant for LivelinessSub due to history config
```
OlivierHecart: I would use a more generic name like "historical".

zenoh/src/api/session.rs
```diff
@@ -1975,17 +1977,25 @@ impl SessionInner {
...
+            if matches!(kind, SubscriberKind::LivelinessSubscriber) && msg_interest_current {
+                |sub: &&Arc<SubscriberState>| sub.history
+            } else {
+                |_: &&Arc<SubscriberState>| true
+            };
         if wire_expr.suffix.is_empty() {
             match state.get_res(&wire_expr.scope, wire_expr.mapping, local) {
                 Some(Resource::Node(res)) => {
-                    for sub in res.subscribers(kind) {
+                    for sub in res.subscribers(kind).iter().filter(filter_by_interest) {
```
OlivierHecart: It is strange to use a filter on the iterator and then an if in the for block. Please check both conditions in the filter or in the if.

OlivierHecart (changes requested):

zenoh/src/api/session.rs
oteffahi: I wanted to make it clear for future that this is only relevant to Liveliness Subscribers.

I'll apply the changes you requested but will add a comment to clarify this.

zenoh/src/api/session.rs
```diff
@@ -2550,6 +2556,9 @@ impl Primitives for WeakSession {
                                 &m.wire_expr,
                                 Default::default(),
                                 || body.insert(Put::default().into()),
+                                // interest_id is set if the Token is an Interest::Current.
+                                // This is used to decide if subs with history=false should be called or not
+                                msg.interest_id.is_some(),
```
wyfo, line 2561: If I've declared a subscriber 10min ago with history enabled (still active), and then a declare a new one with the same keyexpr, then the first subscriber will receive the message intended for the second one, won't it? Don't we need to also match the interest id with the subscriber one to ensure only the concerned subscriber will receive its history?

OlivierHecart: This code is only run if the `remote_tokens` entry is vacant and so will only run once for the first subscriber.
Note: the second subscriber history is built through the same local `remote_tokens` map here: https://github.com/ZettaScaleLabs/zenoh/blob/a6aa9aba59aff5bb6fc6444b6feff989c6679e32/zenoh/src/api/session.rs#L1744 .

PR #2165, opened 2025-09-23
make FrameReader consume messages on drop

wyfo:
In case of multicast, it happens that frame's messages are not read, but they still need to be deserialized to advance the reader cursor.


oteffahi:
For clarity: the issue was introduced following changes in https://github.com/eclipse-zenoh/zenoh/pull/2152

PR #2174, opened 2025-09-29
Improve scalability

OlivierHecart:
This PR brings optimizations by changing conditions order:
- Avoid matching interest when possible
- Avoid computing local queryable info when possible

And internal behavior changes:
- Ignore aggregate flag in token interests (that are anyway not allowed in the protocol)
- Push all tokens to local faces (local sessions) in peers (avoids matching interests)

### DIFF
--- a/zenoh/src/net/routing/hat/client/queries.rs
+++ b/zenoh/src/net/routing/hat/client/queries.rs
@@ -122,10 +122,10 @@ fn propagate_simple_queryable(
             })
             .unwrap_or(true)
         {
-            if let Some(current) = face_hat!(dst_face).local_qabls.get(res) {
+            if let Some(&(current_id, current_info)) = face_hat!(dst_face).local_qabls.get(res) {
                 let info = local_qabl_info(tables, res, &dst_face);
-                if current.1 != info {
-                    let id = current.0;
+                if current_info != info {
+                    let id = current_id;
                     send_declare_queryable(&mut dst_face, res, id, info, send_declare);
                 }
             } else {
--- a/zenoh/src/net/routing/hat/client/queries.rs
+++ b/zenoh/src/net/routing/hat/client/queries.rs
@@ -50,6 +50,7 @@ fn merge_qabl_infos(mut this: QueryableInfoType, info: &QueryableInfoType) -> Qu
     this
 }
 
+#[inline]
 fn local_qabl_info(
     _tables: &Tables,
     res: &Arc<Resource>,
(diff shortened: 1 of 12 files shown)


zenoh/src/net/routing/hat/client/queries.rs
```diff
@@ -82,44 +114,25 @@ fn propagate_simple_queryable(
...
-                        body: DeclareBody::DeclareQueryable(DeclareQueryable {
-                            id,
-                            wire_expr: key_expr,
-                            ext_info: info,
-                        }),
-                    },
-                    res.expr().to_string(),
-                ),
-            );
+            if let Some(current) = face_hat!(dst_face).local_qabls.get(res) {
```
DenisBiryukov91: I would suggest to deconstruct current directly as `if let Some((current_id, current_info)) = ...` to improve readability

zenoh/src/net/routing/hat/linkstate_peer/queries.rs
```diff
@@ -174,44 +207,32 @@ fn propagate_simple_queryable(
...
-        let current = face_hat!(dst_face).local_qabls.get(res);
         if src_face
             .as_ref()
             .map(|src_face| dst_face.id != src_face.id)
             .unwrap_or(true)
-            && (current.is_none() || current.unwrap().1 != info)
             && dst_face.whatami == WhatAmI::Client
-            && face_hat!(dst_face)
+        {
+            if let Some(current) = face_hat!(dst_face).local_qabls.get(res) {
```
DenisBiryukov91: I would suggest to deconstruct current directly as `if let Some((current_id, current_info)) = ...` to improve readability

zenoh/src/net/routing/hat/p2p_peer/queries.rs
```diff
@@ -87,49 +119,36 @@ fn propagate_simple_queryable_to(
...
-        .map(|src_face| dst_face.id != src_face.id)
+        .map(|src_face| {
+            dst_face.id != src_face.id
+                && (src_face.whatami == WhatAmI::Client || dst_face.whatami == WhatAmI::Client)
+        })
         .unwrap_or(true)
-        && (current.is_none() || current.unwrap().1 != info)
-        && (dst_face.whatami != WhatAmI::Client
+    {
+        if let Some(current) = face_hat!(dst_face).local_qabls.get(res) {
```
DenisBiryukov91: I would suggest to deconstruct current directly as `if let Some((current_id, current_info)) = ...` to improve readability

zenoh/src/net/routing/hat/router/queries.rs
```diff
@@ -268,31 +297,26 @@ fn propagate_simple_queryable(
...
-                        body: DeclareBody::DeclareQueryable(DeclareQueryable {
-                            id,
-                            wire_expr: key_expr,
-                            ext_info: info,
-                        }),
-                    },
-                    res.expr().to_string(),
-                ),
-            );
+            if let Some(current) = face_hat!(dst_face).local_qabls.get(res) {
```
DenisBiryukov91: I would suggest to deconstruct current directly as if let `Some((current_id, current_info)) = ...` to improve readability

PR #2175, opened 2025-09-30
Precommit SHM pages

yellowhatter:
Make sure all SHM pages are committed and locked in RAM.

### DIFF
--- a/commons/zenoh-shm/src/shm/unix.rs
+++ b/commons/zenoh-shm/src/shm/unix.rs
@@ -27,7 +27,7 @@ use nix::fcntl::open;
 use nix::{
     fcntl::OFlag,
     sys::{
-        mman::{mmap, munmap, shm_open, shm_unlink, MapFlags, ProtFlags},
+        mman::{mlock, mmap, munmap, shm_open, shm_unlink, MapFlags, ProtFlags},
         stat::{fstat, Mode},
     },
     unistd::ftruncate,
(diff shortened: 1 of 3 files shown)


commons/zenoh-shm/src/shm/unix.rs
```diff
@@ -269,8 +269,12 @@ impl<ID: SegmentID> SegmentImpl<ID> {
             flags,
             fd.as_raw_fd()
         );
+        let ptr = unsafe { mmap(None, len, prot, flags, fd, 0) }?;
 
-        unsafe { mmap(None, len, prot, flags, fd, 0) }
+        tracing::trace!("mlock(addr={:?}, length={})", ptr, len,);
+        unsafe { mlock(ptr, len.get()) }?;
+
+        Ok(ptr)
```
YuanYuYuan, line 279: It's worth documenting the SAFETY.

YuanYuYuan:
@yellowhatter I noticed an inconsistency between segment creation of windows and unix. Should we lock the memory on Windows? 


| Concept      |  Windows      |  Unix/Linux                          |  Description                               |
| -------------|---------------|--------------------------------------|-------------------------------------------|
| Reserve      |  SEC_RESERVE  |  mmap                                |  Reserve virtual address space only       |
| Commit       |  SEC_COMMIT   |  mmap+MAP_POPULATEor touching pages  |  Allocate physical memory backing (commit)|
| Lock in RAM  |  VirtualLock  |  mlock                               |  Prevent pages from being swapped out     |

YuanYuYuan (changes requested):

commons/zenoh-shm/src/shm/unix.rs
yellowhatter: Added safety comments

yellowhatter:
> @yellowhatter I noticed an inconsistency between segment creation of windows and unix. Should we lock the memory on Windows?
> Concept 	Windows 	Unix/Linux 	Description
> Reserve 	SEC_RESERVE 	mmap 	Reserve virtual address space only
> Commit 	SEC_COMMIT 	mmap+MAP_POPULATEor touching pages 	Allocate physical memory backing (commit)
> Lock in RAM 	VirtualLock 	mlock 	Prevent pages from being swapped out

Agreed, we need VirtualLock on Windows. Added

PR #2189, opened 2025-10-09
support quic streams per priority

wyfo:
Links' internal design has been modified to take priority in arguments. For link supporting priorities, like Quic, read/write have been split by priority.
Quic implement stream per priority using a single bidirectional stream for control messages (and opening), and unidirectional streaming for the other priorities.

---
## 🏷️ Label-Based Checklist

**Labels:** `new feature`

## 🆕 New Feature Requirements

**Instructions:**

### DIFF
--- a/io/zenoh-links/zenoh-link-quic/src/unicast.rs
+++ b/io/zenoh-links/zenoh-link-quic/src/unicast.rs
@@ -51,9 +51,9 @@ use super::{QUIC_ACCEPT_THROTTLE_TIME, QUIC_DEFAULT_MTU, QUIC_LOCATOR_PREFIX};
 
 /// Quic endpoint `multistream` config
 enum MultiStreamConfig {
-    /// `multistream=false`
+    /// `multistream=0`
     Disabled,
-    /// `multistream=true`
+    /// `multistream=1`
     Enabled,
     /// default, or `multistream=auto`
     Auto,
--- a/io/zenoh-transport/tests/unicast_transport.rs
+++ b/io/zenoh-transport/tests/unicast_transport.rs
@@ -1964,13 +1964,10 @@ async fn transport_unicast_multistream_quic_default() {
 async fn transport_unicast_multistream_quic_enabled() {
     zenoh_util::init_log_from_env_or("error");
 
-    let endpoint_quic = quic_endpoint("quic/localhost:10469?multistream=true");
+    let endpoint_quic = quic_endpoint("quic/localhost:10469?multistream=1");
     let endpoint = std::slice::from_ref(&endpoint_quic);
     let is_multistream = run_multistream_test(endpoint, endpoint, false).await;
-    assert!(
-        is_multistream,
-        "'?multistream=true' should enable multistream"
-    );
+    assert!(is_multistream, "'?multistream=1' should enable multistream");
 }
 
 #[cfg(feature = "transport_quic")]
@@ -1978,12 +1975,12 @@ async fn transport_unicast_multistream_quic_enabled() {
 async fn transport_unicast_multistream_quic_disabled() {
     zenoh_util::init_log_from_env_or("error");
 
-    let endpoint_quic = quic_endpoint("quic/localhost:10470?multistream=false");
+    let endpoint_quic = quic_endpoint("quic/localhost:10470?multistream=0");
     let endpoint = std::slice::from_ref(&endpoint_quic);
     let is_mutlistream = run_multistream_test(endpoint, endpoint, false).await;
     assert!(
         !is_mutlistream,
-        "'?multistream=false' should disable multistream"
+        "'?multistream=0' should disable multistream"
     );
 }
 
@@ -1995,7 +1992,7 @@ async fn transport_unicast_multistream_quic_auto_explicit() {
     let port = 10471;
     let is_mutlistream = run_multistream_test(
         &[quic_endpoint(&format!(
-            "quic/localhost:{port}?multistream=true"
+            "quic/localhost:{port}?multistream=1"
         ))],
         &[quic_endpoint(&format!("quic/localhost:{port}"))],
         false,
@@ -2003,13 +2000,13 @@ async fn transport_unicast_multistream_quic_auto_explicit() {
     .await;
     assert!(
         is_mutlistream,
-        "'?multistream=true' with auto listener should enable multistream"
+        "'?multistream=1' with auto listener should enable multistream"
     );
 
     let port = 10472;
     let is_mutlistream = run_multistream_test(
         &[quic_endpoint(&format!(
-            "quic/localhost:{port}?multistream=false"
+            "quic/localhost:{port}?multistream=0"
         ))],
         &[quic_endpoint(&format!("quic/localhost:{port}"))],
         false,
@@ -2017,35 +2014,35 @@ async fn transport_unicast_multistream_quic_auto_explicit() {
     .await;
     assert!(
         !is_mutlistream,
-        "'?multistream=false' with auto listener should disable multistream"
+        "'?multistream=0' with auto listener should disable multistream"
     );
 
     let port = 10473;
     let is_mutlistream = run_multistream_test(
         &[quic_endpoint(&format!("quic/localhost:{port}"))],
         &[quic_endpoint(&format!(
-            "quic/localhost:{port}?multistream=true"
+            "quic/localhost:{port}?multistream=1"
         ))],
         false,
     )
     .await;
     assert!(
         is_mutlistream,
-        "'?multistream=true' with auto connect should enable multistream"
+        "'?multistream=1' with auto connect should enable multistream"
     );
 
     let port = 10474;
     let is_mutlistream = run_multistream_test(
         &[quic_endpoint(&format!("quic/localhost:{port}"))],
         &[quic_endpoint(&format!(
-            "quic/localhost:{port}?multistream=false"
+            "quic/localhost:{port}?multistream=0"
         ))],
         false,
     )
     .await;
     assert!(
         !is_mutlistream,
-        "'?multistream=false' with auto connect should disable multistream"
+        "'?multistream=0' with auto connect should disable multistream"
     );
 }
 
@@ -2074,10 +2071,10 @@ fn transport_unicast_multistream_quic_incompatible() {
             .unwrap()
             .block_on(run_multistream_test(
                 &[quic_endpoint(&format!(
-                    "quic/localhost:{port}?multistream=true"
+                    "quic/localhost:{port}?multistream=1"
                 ))],
                 &[quic_endpoint(&format!(
-                    "quic/localhost:{port}?multistream=false"
+                    "quic/localhost:{port}?multistream=0"
                 ))],
                 false,
             ))
@@ -2093,10 +2090,10 @@ fn transport_unicast_multistream_quic_incompatible() {
             .unwrap()
             .block_on(run_multistream_test(
                 &[quic_endpoint(&format!(
-                    "quic/localhost:{port}?multistream=false"
+                    "quic/localhost:{port}?multistream=0"
                 ))],
                 &[quic_endpoint(&format!(
-                    "quic/localhost:{port}?multistream=true"
+                    "quic/localhost:{port}?multistream=1"
                 ))],
                 false,
             ))
@@ -2112,7 +2109,7 @@ fn transport_unicast_multistream_quic_incompatible() {
 async fn transport_unicast_multistream_quic_lowlatency() {
     zenoh_util::init_log_from_env_or("error");
 
-    let endpoint_quic = quic_endpoint("quic/localhost:10478?multistream=true");
+    let endpoint_quic = quic_endpoint("quic/localhost:10478?multistream=1");
     let endpoint = std::slice::from_ref(&endpoint_quic);
     let is_multistream = run_multistream_test(endpoint, endpoint, true).await;
     assert!(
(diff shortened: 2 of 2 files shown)


io/zenoh-links/zenoh-link-quic/src/unicast.rs
```diff
@@ -13,68 +13,159 @@
...
+                PROTOCOL_SINGLE_STREAM.into(),
+                PROTOCOL_LEGACY.into(),
+            ],
+        }
+    }
+
+    fn max_concurrent_uni_streams(&self) -> quinn::VarInt {
+        match self {
+            Self::Disabled => 0u8.into(),
+            _ => (Priority::NUM as u8 - 1).into(),
```
oteffahi, line 96: Please comment to explain that a bidirectional stream exists for priority 0

io/zenoh-links/zenoh-link-quic/src/unicast.rs
```diff
@@ -13,68 +13,224 @@
...
+    /// `multistream=true`
+    Enabled,
+    /// default, or `multistream=auto`
+    Auto,
+}
+
+impl MultiStreamConfig {
+    /// Parse multistream configuration.
+    fn new(config: Config) -> ZResult<Self> {
+        Ok(match config.get("multistream") {
```
oteffahi: Maybe this key should be exposed in `zenoh_link_commons` so other links that will support multistream will use the same parameter.

io/zenoh-links/zenoh-link-quic/src/unicast.rs
```diff
@@ -13,68 +13,224 @@
...
+        mut priority_txs: HashMap<usize, oneshot::Sender<quinn::RecvStream>>,
+    ) -> ZResult<()> {
+        while !priority_txs.is_empty() {
+            let recv = connection.accept_uni().await?;
+            // Uni streams' indexes starts from zero, while priorities above Control starts from 1,
+            // hence the `+ 1`
+            let prio = recv.id().index() as usize + 1;
+            if let Some(tx) = priority_txs.remove(&prio) {
+                tx.send(recv).ok();
+            }
```
oteffahi, line 173: I think there's a potential silent failure here. Maybe we should handle the `else` and log an error.

io/zenoh-links/zenoh-link-quic/src/unicast.rs
```diff
@@ -83,13 +239,56 @@ impl LinkUnicastQuic {
...
+    /// Retrieved the write-stream mapped to the priority
+    ///
+    /// # Safety
+    ///
+    /// There should be only one caller per priority.
+    #[allow(clippy::mut_from_ref)]
+    unsafe fn write_stream(&self, priority: Priority) -> &mut quinn::SendStream {
+        unsafe { &mut *self.send[priority as usize].get() }
+            .as_mut()
+            .expect("multistream should have been started")
```
oteffahi, line 262: This error message would be confusing if it occurs in single-stream mode.

io/zenoh-links/zenoh-link-quic/src/unicast.rs
```diff
@@ -83,13 +239,56 @@ impl LinkUnicastQuic {
...
+    /// The stream may be pending, in which case we wait until it is accepted.
+    ///
+    /// # Safety
+    ///
+    /// There should be only one caller per priority.
+    #[allow(clippy::mut_from_ref)]
+    async unsafe fn read_stream(&self, priority: Priority) -> ZResult<&mut quinn::RecvStream> {
+        match unsafe { &mut *self.recv[priority as usize].get() }
+            .as_mut()
+            .expect("multistream should have been started")
```
oteffahi, line 279: This error message would be confusing if it occurs in single-stream mode.

io/zenoh-transport/src/multicast/link.rs
```diff
@@ -37,14 +37,13 @@ use crate::{
...
-            TransmissionPipeline, TransmissionPipelineConf, TransmissionPipelineConsumer,
-            TransmissionPipelineProducer,
+            PipelineConsumer, TransmissionPipeline, TransmissionPipelineConf,
+            TransmissionPipelineConsumer, TransmissionPipelineProducer,
         },
         priority::TransportPriorityTx,
     },
     multicast::transport::TransportMulticastInner,
 };
-
```
oteffahi, line 45: Please add this line-break back so I can sleep at night :')

io/zenoh-transport/src/unicast/universal/link.rs
```diff
@@ -184,59 +197,108 @@ impl TransportLinkUnicastUniversal {
...
-                            stats.inc_tx_bytes(n);
-                        }
-                    }
+            pull = pipeline.pull() => {
+                let Some((mut batch, priority)) = pull else {
+                    // The queue has been disabled: break the tx loop, drain the queue, and exit
+                    break
+                };
+                debug_assert!(write_priority == Priority::Control || priority == write_priority);
+                link.send_batch(&mut batch, write_priority).await?;
```
oteffahi: I think it's better here to `link.send_batch(&mut batch, priority)` and let the underlying link decide the scheduling based on if it is set to multistream or not. Writing everything under `Priority::Control` in single-stream links is a bit confusing, and could be cause for false assumptions later if links use the message's priority for features other than stream multiplexing.

Will the additional check on every write have a non-negligible impact on performance? Or I guess we can have some additional abstraction to avoid this check.

io/zenoh-transport/src/unicast/universal/link.rs
```diff
@@ -256,50 +318,167 @@ async fn rx_task(
...
+    if link.link.supports_priorities() {
+        let (res, _, _) = select_all((Priority::MAX as u8..Priority::MIN as u8).map(|prio| {
+            let mut link = link.clone();
+            let transport = transport.clone();
+            let token = token.clone();
+            let lease_tracker = lease_tracker.clone();
+            #[cfg(feature = "stats")]
+            let stats = stats.clone();
+            let pool = pool.clone();
+            zenoh_runtime::ZRuntime::RX.spawn(async move {
```
oteffahi, line 377: I think we can use `tokio::runtime::Handle::current()` to match the behavior of the `else` which schedules the future on whatever runtime called the `rx_task`.
I believe the same remark applies for `tx_task` and its `write_loop` futures.

io/zenoh-transport/src/unicast/universal/link.rs
```diff
@@ -256,50 +318,167 @@ async fn rx_task(
...
+            }
+            Poll::Pending
+        })
+        .await
+    }
+}
+
+impl Drop for TimeoutTracker {
+    fn drop(&mut self) {
+        if Arc::strong_count(&self.0) == 2 {
```
oteffahi: Couldn't this be avoided by passing a `Weak` to the task? Or can the upgrade call take too long and impact the lease?

io/zenoh-transport/src/unicast/link.rs
```diff
@@ -92,7 +92,7 @@ impl TransportLinkUnicast {
 
     pub(crate) async fn send(&self, msg: &TransportMessage) -> ZResult<usize> {
         let mut link = self.tx();
-        link.send(msg).await
+        link.send(msg, Priority::Control).await
```
oteffahi: May be worth to comment that this is only used by the Zenoh handshake or to close the link, and therefore can be hardcoded to `Priority::Control`.

io/zenoh-transport/src/unicast/universal/link.rs
wyfo: I made the priority optional to avoid confusion

io/zenoh-transport/src/unicast/universal/link.rs
wyfo: Everywhere else, RX runtime is used

io/zenoh-transport/src/unicast/universal/link.rs
wyfo: done

io/zenoh-transport/src/unicast/link.rs
wyfo: Priority is optional now

io/zenoh-transport/src/multicast/link.rs
wyfo: done

io/zenoh-links/zenoh-link-quic/src/unicast.rs
wyfo: It should not occurs in single-stream mode

io/zenoh-links/zenoh-link-quic/src/unicast.rs
wyfo: It should not occurs in single-stream mode

io/zenoh-links/zenoh-link-quic/src/unicast.rs
wyfo: I added a comment

io/zenoh-links/zenoh-link-quic/src/unicast.rs
wyfo: Done

io/zenoh-links/zenoh-link-quic/src/unicast.rs
oteffahi: My problem isn't about the channel, it's about the `if let Some(tx) = priority_txs.remove(&prio)`. If the `remove` is expected to always return `Some`, then it should be clear in the code and panic/error-out if this guarantee is lost in the future. An `if let Some` without an `else` here feels to me like a ticking bomb.

io/zenoh-links/zenoh-link-quic/src/unicast.rs
```diff
@@ -13,68 +13,226 @@
...
+impl MultiStreamConfig {
+    /// Parse multistream configuration.
+    fn new(config: Config) -> ZResult<Self> {
+        Ok(match config.get(MULTISTREAM_CONFIG_KEY) {
+            Some("false") => Self::Disabled,
+            Some("true") => Self::Enabled,
+            None | Some("auto") => Self::Auto,
+            Some(s) => bail!("Invalid multistream config: {s}"),
+        })
+    }
```
oteffahi: This should be an endpoint parameter instead of an endpoint config. (`quic/127.0.0.1:7447?multistream=0` instead of `quic/127.0.0.1:7447#multistream=0`). It would enable the listener's multistream mode to be propagated in scouting.

oteffahi:
@wyfo I have pushed tests for this PR. Please review them.

wyfo:
LGTM

oteffahi (approved):

io/zenoh-links/zenoh-link-quic/src/unicast.rs
```diff
@@ -83,13 +245,59 @@ impl LinkUnicastQuic {
     async fn close(&self) -> ZResult<()> {
         tracing::trace!("Closing QUIC link: {}", self);
         // Flush the QUIC stream
-        let mut guard = zasynclock!(self.send);
-        if let Err(e) = guard.finish() {
-            tracing::trace!("Error closing QUIC stream {}: {}", self, e);
+        for (i, send) in self.send.iter().enumerate() {
+            if let Some(send) = unsafe { &mut (*send.get()) }.as_mut() {
```
oteffahi, line 243: Correct me if I'm wrong, but since `LinkUnicastTrait::close` can be called concurrently with `LinkUnicastTrait::write` and `LinkUnicastTrait::write_all`, the `UnsafeCell` usage here is a UB.

Maybe we can skip closing the streams and just close the connection?

io/zenoh-links/zenoh-link-quic/src/unicast.rs
```diff
@@ -325,6 +544,8 @@ impl LinkManagerUnicastTrait for LinkManagerUnicastQuic {
             })?
             .await
             .map_err(|e| zerror!("Can not create a new QUIC link bound to {}: {}", host, e))?;
+        quic_conn.set_max_concurrent_bi_streams(1u8.into());
+        quic_conn.set_max_concurrent_uni_streams(multistream.max_concurrent_uni_streams());
```
oteffahi: This changes the config after the QUIC handshake, by sending dedicated messages. These limits should be set on the `quinn::TransportConfig` passed to `quic_endpoint.set_default_client_config`.
Line 529 can be changed like so:
```rust
        quic_endpoint.set_default_client_config({
            let mut client_config = quinn::ClientConfig::new(Arc::new(quic_config));
            let mut transport_config = quinn::TransportConfig::default();
            
            transport_config.max_concurrent_bidi_streams(1u8.into());
            transport_config.max_concurrent_uni_streams(multistream.max_concurrent_uni_streams());
            
            client_config.transport_config(transport_config.into());
            client_config
        });
```

oteffahi (changes requested):

io/zenoh-link-commons/src/quic.rs
```diff
@@ -42,8 +42,13 @@ use crate::{
     ConfigurationInspector, LinkAuthId, BIND_INTERFACE,
 };
 
-// Default ALPN protocol
-pub const ALPN_QUIC_HTTP: &[&[u8]] = &[b"hq-29"];
+// ALPN protocols
+/// Protocol used by zenoh <= 1.6.2
```
oteffahi:
```suggestion
/// Protocol used by zenoh <= 1.8.0
```

io/zenoh-links/zenoh-link-quic/src/unicast.rs
```diff
@@ -13,68 +13,230 @@
...
+impl MultiStreamConfig {
+    /// Parse multistream configuration.
+    fn new(metadata: Metadata) -> ZResult<Self> {
+        let multistream = metadata.get(Metadata::MULTISTREAM).unwrap_or("auto");
+        if multistream == "auto" {
+            return Ok(Self::Auto);
+        }
+        if multistream
+            .parse()
+            .map_err(|_| zerror!("Invalid multistream config:  {multistream}"))?
```
oteffahi: My bad, I though boolean parsing would work for 0/1 values, but it turns out it doesn't 😅. Please change this back to checking for "0" and "1".

oteffahi (changes requested):

PR #2193, opened 2025-10-10
Add shm infos in adminspace and stats

OlivierHecart:


io/zenoh-transport/src/common/stats.rs
```diff
@@ -249,11 +254,21 @@ use serde::{Deserialize, Serialize};
 stats_struct! {
     #[derive(Clone, Debug, Deserialize, Serialize)]
     pub struct DiscriminatedStats {
```
wyfo: You should rename this one if there are multiple discriminations

commons/zenoh-protocol/src/network/mod.rs
```diff
@@ -31,6 +31,8 @@ pub use request::{AtomicRequestId, Request, RequestId};
 pub use response::{Response, ResponseFinal};
 
 use crate::core::{CongestionControl, Priority, Reliability, WireExpr};
+#[cfg(feature = "shared-memory")]
+use crate::zenoh::PushBody;
```
wyfo: Not consistent with other imports done in the function

PR #2225, opened 2025-10-24
Bypass commit check if it is unspecified for plugin or host

DenisBiryukov91:
Bypass commit check if it is unspecified for plugin or host

### DIFF
--- a/plugins/zenoh-plugin-trait/src/compatibility.rs
+++ b/plugins/zenoh-plugin-trait/src/compatibility.rs
@@ -53,12 +53,18 @@ impl Compatibility {
         }
 
         fn version_equals(left: &str, right: &str) -> bool {
-            const RELEASE_COMMIT: &str = "release";
+            const RELEASE_COMMIT: &str = "release"; // fallback from `zenoh::GIT_COMMIT`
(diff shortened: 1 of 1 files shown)


plugins/zenoh-plugin-trait/src/compatibility.rs
```diff
@@ -44,17 +44,28 @@ impl Compatibility {
...
+        }
+
+        fn version_equals(left: &str, right: &str) -> bool {
+            const RELEASE_COMMIT: &str = "release";
+            let (left_version, left_commit) = get_version_and_commit(left);
+            let (right_version, right_commit) = get_version_and_commit(right);
+            if left_version != right_version {
+                return false;
+            }
+            if left_commit != right_commit
```
milyin, line 68: I think here it's necessary to explain the logic behind and reasons for this decision.
Also it's slightly dangerous that the comparison logic is implicitly tied to our release process. Where this "release" string comes from and what if at some moment it changes or some other postfix variant appear?

DenisBiryukov91: fixed

milyin (approved):

PR #2246, opened 2025-11-03
ci: add workflow to check Eclipse IPDDP compliance

diogomatsubara:


diogomatsubara:
> You might want to define a reusable workflow in eclipse-zenoh/ci to share the handling of Cargo dependencies.

Added here: https://github.com/eclipse-zenoh/ci/pull/405 and updated this one to use the re-usable workflow.

PR #2255, opened 2025-11-10
Add tests and improve coverage

OlivierHecart:


zenoh/src/api/queryable.rs
```diff
@@ -109,54 +109,172 @@ pub struct Query {
...
     /// The full [`Selector`] of this Query.
+    ///
+    /// # Examples
+    /// ```
+    /// # #[tokio::main]
+    /// # async fn main() {
+    /// let session = zenoh::open(zenoh::Config::default()).await.unwrap();
+    /// let queryable = session
+    ///     .declare_queryable("key/expression")
+    ///     .callback(move |query| { query.selector(); })
```
DenisBiryukov91: maybe it is better to make this example and the ones below meaningful and at least print corresponding output ?
```suggestion
    ///     .callback(move |query| { println!("{}", query.selector()); })
```

PR #2256, opened 2025-11-10
Handle ignored cases in `Interest::options`

fuzzypixelz:
This function is meant to rectify the values of the `R`, `N` and `M` interest options during encoding.

Unforunately, it only sets an option if it should be set but isn't: i.e. it doesn't unset an option if it's not supposed to be set.

Consider the following scenario: node X receives an interest with `N` set and a non-empty suffix (as expected), then it forwards it to node Y with an empty suffix; the `Interest::options` function will not unset the original `N` flag during encoding on X, thus node Y will receive an invalid message.

Thus these options should be set _if and only if_ their respective conditions are met.


fuzzypixelz:
Perhaps these fields should not even be part of the `Interest` message definition.

PR #2260, opened 2025-11-13
Rework stats/metrics

wyfo:


zenoh/src/net/runtime/adminspace.rs
```diff
@@ -721,30 +699,48 @@ fn metrics(context: &AdminContext, query: Query) {
...
-    let mut metrics = format!(
-        r#"# HELP zenoh_build Information about zenoh.
-# TYPE zenoh_build gauge
-zenoh_build{{version="{}"}} 1
-"#,
-        context.version
-    );
-
+    #[cfg(not(feature = "stats"))]
+    let (metrics, encoding) = ("# EOF\n", Encoding::TEXT_PLAIN);
```
OlivierHecart: It was handy to have the `zenoh_build` metric even when the stats feature is disabled. It allowed to monitor the zenoh daemons statuses through prometheus without paying the full stats performance cost. 
Please reintroduce that.

zenoh/src/net/routing/interceptor/access_control.rs
```diff
@@ -620,44 +364,84 @@ impl InterceptorTrait for IngressAclEnforcer {
...
+            None => {
+                // Undeclarations in ingress are only filtered if the ext_wire_expr is set.
+                // If it's not set, we let the undeclaration pass, it will be rejected by the routing logic
+                // if its associated declaration was denied.
+                match key_expr {
+                    Some(ke) => self.action(action, log_msg, &ke),
+                    None => Permission::Allow,
+                }
+            }
+        }
```
oteffahi, line 441: Undeclarations without a key-expr should always be allowed through: no need to check the cache, since it's probably `None` due to the lack of a key-expr.

oteffahi: I realize now that this isn't part of the changes made by this PR, so it's not mandatory to fix this here.

zenoh/src/net/routing/interceptor/access_control.rs
```diff
@@ -439,9 +166,9 @@ impl InterceptorTrait for IngressAclEnforcer {
                     return false;
                 };
                 if self.cached_result_or_action(
-                    cache.map(|c| c.delete),
+                    cache.map(|c| c.put),
```
oteffahi, line 169: Looks like this mistake was already in the code, it wasn't introduced by this PR...
```suggestion
                    cache.map(|c| c.delete),
```

oteffahi: I think it's better to make another PR to fix this after this one is merged.
I opened an issue: https://github.com/eclipse-zenoh/zenoh/issues/2293

zenoh/src/net/routing/interceptor/access_control.rs
```diff
@@ -927,6 +694,281 @@ impl InterceptorTrait for EgressAclEnforcer {
...
+                "{zid} did not match any configured ACL subject. Default permission `{:?}` will be applied on all messages",
+                self.enforcer.default_permission
+            );
+        }
+        #[cfg(feature = "stats")]
+        let Ok(stats) = transport
+            .get_stats()
+            .map(|stats| stats.drop_stats(zenoh_stats::ReasonLabel::AccessControl))
+        else {
+            return (None, None);
```
oteffahi, line 821: I don't think it's a good idea to disable ACL because stats feature failed to initialize. ACL being a security feature aside, the same can be said about other interceptors, they should still work even without stats.

I think it would be better to just log the failure, rather than disable all interceptors silently and have zenoh start without them.

wyfo: The only reason why `get_stats` could fail is the transport being closed, as in other interceptors.

OlivierHecart: That's the case today. What if someone changes `get_stats` to fail for other reasons ?

oteffahi: Let's add this to the doc of `get_stats`, so future changes will mind it.

wyfo: I can make the `DropStats` optional in all interceptors if you want.

OlivierHecart: Well I'd prefer an error on `get_stats` to indeed not disable ACL if possible ...

wyfo: As you wish.

OlivierHecart: It's coherent with other Transport methods... Nevermind.

PR #2267, opened 2025-11-18
Support one BlockFirst message per link per priority

wyfo:


commons/zenoh-sync/src/event.rs
```diff
@@ -104,6 +104,7 @@ impl fmt::Debug for NotifyError {
 impl std::error::Error for NotifyError {}
 
 // Inner
+#[derive(Debug)]
```
OlivierHecart: Do we need those derives ?

wyfo: No, we don't

PR #2272, opened 2025-11-20
chore: undo pins for ntex crates

diogomatsubara:
ntex crates are a dependency of zenoh-plugin-mqtt and due to a misunderstanding on how to use the zenoh-pinned-deps-1-75 I ended up adding them here. The correct process is to pin them directly in zenoh-plugin-mqtt Cargo.toml, which will be done on https://github.com/eclipse-zenoh/zenoh-plugin-mqtt/pull/535

To summarize I quote @milyin:
 - zenoh itself doesn't have locks, it can be used freely, it doesn't break user's projects by dependency locks
 - zenoh-pinned-deps-1-75 guarantees that zenoh is buildable with 1.75
 - plugins are not supposed to be used as dependencies, so they can contain needed locks right it their cargo.toml's

### DIFF
--- a/commons/zenoh-pinned-deps-1-75/Cargo.toml
+++ b/commons/zenoh-pinned-deps-1-75/Cargo.toml
@@ -37,12 +37,6 @@ librocksdb-sys = "=0.17.1"
 litemap = "=0.7.4"
 lz4_flex = "=0.11.3"
 nonempty-collections = "=0.3.1"
-ntex = { version = "=2.7.0", features = ["rustls", "tokio"] }
-ntex-bytes = "=0.1.28"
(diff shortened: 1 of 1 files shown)


diogomatsubara:
Lint error is related to the just released tls-listener crate `0.11.1`. Looks like this new version doesn't compile on a windows runner but it does compile on mac and linux.

diogomatsubara:
I submitted an issue: https://github.com/tmccombs/tls-listener/issues/64 for tls-listener upstream.

diogomatsubara:
@milyin A new version of `tls-listener` has been released and CI now passes

PR #2275, opened 2025-11-21
Remove rocksdb dependencies

evshary:
Now we always compile RocksDB to make sure the compatibility of Cargo 1.75.
https://github.com/eclipse-zenoh/zenoh/pull/2157
However, the issue comes from [zenoh-backend-filesystem](https://github.com/eclipse-zenoh/zenoh-backend-filesystem), and we can pin the version there.
While Zenoh itself doesn't use RocksDB, it slows down the compile process significantly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -559,44 +559,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "bindgen"
-version = "0.69.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


evshary:
I think the Windows lint issue will be fixed by https://github.com/eclipse-zenoh/zenoh/pull/2276

diogomatsubara (approved):

OlivierHecart (approved):

PR #2276, opened 2025-11-21
chore: pin tls-listener to 0.11.0

diogomatsubara:
tls-listener 0.11.1 doesn't compile on windows and newer versions `>0.11.1` will require rust edition 2024

~~Fix https://github.com/eclipse-zenoh/zenoh/actions/runs/19555241578/job/55996089049~~


diogomatsubara:
Actually it doesn't fix CI lint, for that we'll have to wait a new release or pin on zenoh's Cargo.toml

milyin (approved):

Issue #1760, opened 2025-02-06
Expose liveliness token in adminspace


gabrik:
### Describe the feature

It would be nice to have access to the declared liveliness tokens via the adminspace, similarly to what we already have for publisher/subscribers/queriers and queryables.

PR #2294, opened 2025-12-03
liveliness token support in adminspace

milyin:
fixes https://github.com/eclipse-zenoh/zenoh/issues/1760 : adds `@/{zid}/{mode}/token/**` to keyexpr space

Adminspace queryable refactored to avoid code and keyexpr duplication


zenoh/src/net/runtime/adminspace.rs
```diff
@@ -63,14 +69,17 @@ pub struct AdminContext {
...
-type Handler = Arc<dyn Fn(&AdminContext, Query) + Send + Sync>;
+type Handler = Arc<dyn Fn(&keyexpr, &AdminContext, Query) + Send + Sync>;
 
 pub struct AdminSpace {
     zid: ZenohId,
     queryable_id: QueryableId,
     primitives: Mutex<Option<Arc<Face>>>,
     mappings: Mutex<HashMap<ExprId, String>>,
-    handlers: HashMap<OwnedKeyExpr, Handler>,
+    // key is full key expression (root_key/key[/glob]), value is (handler, glob_len)
```
fuzzypixelz: I suggest you make a proper Rustdoc comment with `///`.

zenoh/src/net/runtime/adminspace.rs
```diff
@@ -502,9 +479,12 @@ impl Primitives for AdminSpace {
                     attachment: query.ext_attachment.take().map(Into::into),
                 };
 
-                for (key, handler) in &self.handlers {
-                    if key_expr.intersects(key) {
-                        handler(&self.context, query.clone());
+                for (full_key, (handler, glob_len)) in &self.handlers {
+                    if key_expr.intersects(full_key) {
+                        // Extract prefix by removing the glob suffix
+                        let prefix_str = &full_key.as_str()[..full_key.as_str().len() - glob_len];
```
fuzzypixelz: Having to coerce into `str`, perform slicing and then coerce back into `keyexpr` makes the code more error prone. The reader is left to wonder if the slice indexing can fail, if the subtraction can overflow, if `keyexpr::new` can fail; all are extra headaches for little benefit in return.

I suggest you store the glob separately; this can be a `&'static keyexpr` since it can only ever be `*` or `**`.

zenoh/src/net/runtime/adminspace.rs
milyin: It's internal code, why make doc comments here? It's for code reading only

fuzzypixelz: In an LSP-powered editor, hovering over a reference of the variable would show you the Rustdoc.

zenoh/src/net/runtime/adminspace.rs
milyin: I'd prefer to avoid building new key expression for each test in the loop. It's more effective to build full key expression with glob once. But I've simplified the code by storing prefix length instead of glob length. This allows to avoid substraction.
It's possible to store prefix separately as another owned key expression, but this is less clear: it will make the one who read the code wonder, is it the same prefix as in key value or not

zenoh/src/net/runtime/adminspace.rs
milyin: Fixed

zenoh/src/net/runtime/adminspace.rs
fuzzypixelz: I'm starting to wonder if changing the data structure at all is a good idea. Today a GET on `@/*/*/linkstate` returns nothing; what if we want to make it queryable in the future, for whatever reason?

Do we not want any two handlers to share the same non-wild prefix?

milyin: I don't get it. The refactoring doesn't change the original logic: as in the original implementation the handlers are stored in the map: `@/{zid}/{whatami}/{key}[/**]` -> `Fn(context,query)`
The only change is that now it's `Fn(prefix,context,query)` where prefix is `@/{zid}/{whatami}/{key}`. Originally each handler function recreated this prefix internally. Result was exactly the same, but with risk of inconsistent changes in `AdminSpace::start` and handler function.

fuzzypixelz: Okay, so even though two handlers might share the same prefix, their map keys are not the same.

PR #2310, opened 2025-12-09
Count payloads stats at routing level

wyfo:
Payload stats histograms were previously updated with a mix of link-level and router-level informations. It requires a complex system using thread-local variables to pass informations between both levels, but this system was bugged, as it didn't support adminspace routing for example.
The decision was taken to remove payload histograms from link stats, counting them only for transport, so they can be updated in routing alone.
However, for TX direction, it still requires to pass back the information of the message being actually sent or not, to avoid updating stats otherwise.

### DIFF
--- a/commons/zenoh-stats/src/link.rs
+++ b/commons/zenoh-stats/src/link.rs
@@ -38,8 +38,12 @@ impl LinkStats {
                 .transport_message(StatsDirection::from_index(dir))
                 .get_or_create_owned(transport, Some(&link), &labels)
         });
-        let tx_congestion =
-            DropStats::new(registry.clone(), transport.clone(), ReasonLabel::Congestion);
+        let tx_congestion = DropStats::new(
+            registry.clone(),
+            transport.clone(),
+            ReasonLabel::Congestion,
+            Some(protocol.clone()),
+        );
         Self(Arc::new(LinkStatsInner {
             transport_stats,
             link,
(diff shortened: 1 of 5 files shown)


zenoh/src/api/session.rs
```diff
@@ -3025,33 +3025,39 @@ impl Primitives for WeakSession {
 
 impl crate::net::primitives::EPrimitives for WeakSession {
     #[inline]
-    fn send_interest(&self, ctx: crate::net::routing::RoutingContext<&mut Interest>) {
-        (self as &dyn Primitives).send_interest(ctx.msg)
+    fn send_interest(&self, ctx: crate::net::routing::RoutingContext<&mut Interest>) -> bool {
+        (self as &dyn Primitives).send_interest(ctx.msg);
+        false
```
fuzzypixelz: Why is this `false`?

io/zenoh-transport/src/unicast/lowlatency/transport.rs
```diff
@@ -237,8 +237,9 @@ impl TransportUnicastTrait for TransportUnicastLowlatency {
     /*************************************/
     /*                TX                 */
     /*************************************/
-    fn schedule(&self, msg: NetworkMessageMut) -> ZResult<()> {
-        self.internal_schedule(msg)
+    fn schedule(&self, msg: NetworkMessageMut) -> ZResult<bool> {
+        self.internal_schedule(msg)?;
```
DenisBiryukov91, line 241: should not congestion inside internal_schedule be taken into account in return value ?

OlivierHecart (approved):

io/zenoh-transport/src/unicast/lowlatency/transport.rs
wyfo: `internal_schedule` returns `ZResult<()>` for low latency, there is no drop in case of congestion.

PR #2317, opened 2025-12-11
Close the advanced subscriber with the session 

wyfo:
It also fixes the bug of the session not being closed when an advanced
subscriber is alive.

WeakSession in the code becomes a simple alias to Arc<SessionInner>
thanks to #1641. A new WeakSessionInternal, exported as WeakSession, can
be used in internal crates, such as zenoh-ext, to keep a weak session
reference. So the advanced subscriber only captures a weak session now.

Moreover, in order to properly close the advanced subscriber, its
callback must be dropped. It is done in the drop callback of the
underlying subscriber, so the advanced subscriber is now really bound
to the real registered subscriber.

Based on #1641

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5867,7 +5867,6 @@ dependencies = [
  "petgraph 0.8.3",
  "phf",
  "rand 0.8.5",
- "ref-cast",
  "rustc_version 0.4.1",
  "serde",
  "serde_json",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -142,7 +142,6 @@ quote = "1.0.40"
 rand = { version = "0.8.5", default-features = false } # Default features are disabled due to usage in no_std crates
 rand_chacha = "0.3.1"
 rcgen = "0.14.4"
-ref-cast = "1.0.24"
 regex = "1.11.2"
 ringbuffer-spsc = "0.1.15"
 ron = "0.12.0"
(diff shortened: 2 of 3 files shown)


zenoh/src/api/session.rs
```diff
@@ -620,61 +626,36 @@ impl Drop for Session {
...
-        Self(self.0.clone())
-    }
-}
+/// A weak reference to the session.
+// `WeakSession` cannot be exposed as it gives access to the session internal,
+// but we need  a type that deref to a regular session (in order to call session method).
+// It is reexported and exposed as `WeakSession` in the API.
+#[zenoh_macros::internal]
+#[derive(Clone)]
+pub struct WeakSessionInternal(WeakSession);
```
DenisBiryukov91: I think it is better to get rid of WeakSession type alias, rename WeakSessionInternal -> WeakSession and just access Arc<SessionInner> via .0 or .inner() method of WeakSession to avoid confusion

zenoh-ext/src/advanced_subscriber.rs
```diff
@@ -822,10 +827,22 @@ impl<Handler> AdvancedSubscriber<Handler> {
             }
         };
 
+        // When the underlying subscriber is undeclared (for example when the session is closed)
+        // the advanced subscriber callback must be dropped to "close" the receiver.
+        let drop_callback = {
+            let statesref = statesref.clone();
+            move || {
+                statesref.lock().unwrap().callback.take();
```
DenisBiryukov91: We might probably also want to clean miss_handlers callbacks

DenisBiryukov91 (approved):

OlivierHecart (approved):

Issue #2142, opened 2025-09-12
Multilink establishment in client mode


OlivierHecart:
### Describe the bug

Defining multiple connect endpoints with different QoS classes to establish multiple links does not work in client mode.

### To reproduce

```
RUST_LOG=zenoh_transport::unicast::establishment=debug zenohd --cfg=transport/unicast/max_links:2
RUST_LOG=zenoh_transport::unicast::establishment=debug z_sub --cfg=transport/unicast/max_links:2 -m peer -e tcp/127.0.0.1:7447?rel=1 -e tcp/127.0.0.1:7447?r
el=0
```
OK: 2 links established

```
RUST_LOG=zenoh_transport::unicast::establishment=debug zenohd --cfg=transport/unicast/max_links:2
RUST_LOG=zenoh_transport::unicast::establishment=debug z_sub --cfg=transport/unicast/max_links:2 -m client -e tcp/127.0.0.1:7447?rel=1 -e tcp/127.0.0.1:7447?r
el=0
```
KO: only one link established

### System info

Zenoh 1.5.1

PR #2176, opened 2025-10-02
replace tide with axum in rest plugin and remove async_std dependencies

wyfo:
---
## 🏷️ Label-Based Checklist

Based on the labels applied to this PR, please complete these additional requirements:

**Labels:** `enhancement`, `dependencies`

## ✨ Enhancement Requirements

Since this PR enhances existing functionality:

- [ ] **Enhancement scope documented** - Clear description of what is being improved
- [ ] **Minimum necessary code** - Implementation is as simple as possible, doesn't overcomplicate the system
- [ ] **Backwards compatible** - Existing code/APIs still work unchanged
- [ ] **No new APIs added** - Only improving existing functionality
- [ ] **Tests updated** - Existing tests pass, new test cases added if needed
- [ ] **Performance improvement measured** - If applicable, before/after metrics provided
- [ ] **Documentation updated** - Existing docs updated to reflect improvements
- [ ] **User impact documented** - How users benefit from this enhancement

**Remember:** Enhancements should not introduce new APIs or breaking changes.

## 📦 Dependency Updates

Since this PR updates dependencies:

- [ ] **Changelog reviewed** - Breaking changes in dependency identified and addressed
- [ ] **Security advisories checked** - No known vulnerabilities in new version
- [ ] **License compatible** - New version license is compatible with project
- [ ] **Tests pass** - All tests pass with new dependency versions
- [ ] **Transitive deps reviewed** - New transitive dependencies checked
- [ ] **Version pinned appropriately** - Semver range is intentional and correct
- [ ] **Update scope limited** - Updating one major dependency at a time when possible

**Best practice:** Keep dependency updates focused and test thoroughly.

**Instructions:**
1. Check off items as you complete them (change `- [ ]` to `- [x]`)
2. The PR checklist CI will verify these are completed

*This checklist updates automatically when labels change, but preserves your checked boxes.*

PR #2320, opened 2025-12-17
Support multiple connections in the client mode.

evshary:
Solve https://github.com/eclipse-zenoh/zenoh/issues/2142

We added the `EndPoints` to accept a single `EndPoint` or `EndPoint` vector.

To test the multiple connections:

```bash
RUST_LOG=zenoh_transport::unicast::establishment=debug cargo run --bin zenohd -- --cfg=transport/unicast/max_links:2
RUST_LOG=zenoh_transport::unicast::establishment=debug cargo run --example z_sub -- --cfg=transport/unicast/max_links:2 --cfg="connect/endpoints:[{strategy:'allOf',locators:['tcp/127.0.0.1:7447?rel=0','tcp/127.0.0.1:7447?rel=1']}]" -m client
```

Note that only 'allOf' is implemented in the PR. We leave 'oneOf' for the future.

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**


examples/src/lib.rs
```diff
@@ -55,8 +55,14 @@ impl From<&CommonArgs> for Config {
         }
 
         if !args.connect.is_empty() {
+            // Able to parse multiple endpoints, e.g. "tcp/127.0.0.1:7447?rel=0,tcp/127.0.0.1:7447?rel=1"
```
OlivierHecart: This is a change in behavior. 
Coma separated values were already accepted here and were the equivalent of passing multiple `-e` arguments. 
Let's decide if:
- we assume this behavior change and add this in a minor release
- find another separator for this case

zenoh/src/net/runtime/orchestrator.rs
```diff
@@ -338,72 +341,90 @@ impl Runtime {
...
-                self.get_global_connect_timeout(),
-                retry_config
-            );
-            if retry_config.timeout().is_zero() || self.get_global_connect_timeout().is_zero() {
-                // try to connect and exit immediately without retry
-                if let Err(e) = self.peer_connector(endpoint).await {
-                    if retry_config.exit_on_failure {
+    async fn connect_peers_multiply_links(&self, peers: &[EndPoints]) -> ZResult<()> {
+        for peer_group in peers {
+            for peer in peer_group.as_vec() {
```
OlivierHecart: That is not the behavior we want. We would like Zenoh to connect to one and only one peer of each group. In case of background connection we want Zenoh to spawn a single task that tries to connect to the group.

OlivierHecart:
I also realize that we had plans to create locator groups for peers, where peer connect to a single endpoint of each group. We need to decide if this is still something we consider and how to articulate that with the multilink thing here.

examples/src/lib.rs
evshary: I personally prefer the first one, while the second one will make usage complicated.
However, it's still fine with me if we decide on the second one.

OlivierHecart:
Another point, I don't see any changes in `DEFAULT_CONFIG.json5`. It would be good to have this roughly documented there.

evshary:
> Another point, I don't see any changes in `DEFAULT_CONFIG.json5`. It would be good to have this roughly documented there.

I thought it would make the description too complicated, so I just skipped that.
I've added it to the config now.

evshary:
@OlivierHecart I also added some tests for the feature.
I think the CI failure is related to the outdated tide in the rest plugin, which should be fixed here.
https://github.com/eclipse-zenoh/zenoh/pull/2176

OlivierHecart (approved):

PR #2322, opened 2025-12-18
zenoh: api: encoding: Add MAVLink

patrickelectric:
MAVLink is a messaging protocol for communicating with drones (and between onboard drone components).

It would be great for the drone industry (and zenoh adoption) to have the MAVLink protocol as a native encoding option in Zenoh.
I'm working with MAVLink/ArduPilot/BlueOS/PX4 and zenoh integration in the ecosystem, this patch would move things forward exceptionally.

For more information:
  - https://mavlink.io/en/
  - https://github.com/mavlink/mavlink
  - https://ardupilot.org/dev/docs/mavlink-commands.html
  - https://docs.px4.io/main/en/mavlink/#mavlink-overview
    - That already has a initial zenoh integration (https://github.com/PX4/PX4-Autopilot/tree/3e425210e0a625539268709c3ca8b670014785f2/src/modules/zenoh)

### DIFF
--- a/zenoh/src/api/encoding.rs
+++ b/zenoh/src/api/encoding.rs
@@ -471,6 +471,13 @@ impl Encoding {
         id: 52,
         schema: None,
     });
+    /// A Micro Air Vehicle Link (MAVLink)-encoded data.
+    ///
+    /// Constant alias for string: `"application/mavlink"`.
+    pub const APPLICATION_MAVLINK: Encoding = Self(zenoh_protocol::core::Encoding {
+        id: 53,
+        schema: None,
+    });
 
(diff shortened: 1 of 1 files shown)


patrickelectric:
Ping

DenisBiryukov91:
@patrickelectric Unfortunately with current implementation, adding a new predefined encoding would break wire compatibility with previous versions (i.e. you would set encoding to `application/mavlink` on updated zenoh instance and transmit only numerical id 53 on the wire, but the older zenoh instance would not be able to decode, since it does not have `application/mavlink` on its predefined encoding list).

patrickelectric:
Thanks for the info @DenisBiryukov91. If that's the case I'll close the PR.

PR #2323, opened 2025-12-18
Introduce label-specific checklists instead of check labels

sashacmc:
Introduce label-specific checklists instead of check labels to improve the quality of PR and facilitate PR review
Adds PR description template


.github/pull_request_template.md
```diff
@@ -0,0 +1,12 @@
+## Description
+<!-- TODO: Add a clear description of what this PR does and why -->
+**⚠️ Please replace this section with your PR description**
```
fuzzypixelz, line 3:
```suggestion
```

fuzzypixelz: If comments are sufficient for the other sections, then they should be so for this one as well.

sashacmc:
> I don't think it's a good idea for automated workflows to read/write a pull request body. This is inherently error-prone. 

My original idea (and implementation) was to have a separate PR's comment and everything in it, but unfortunately, during testing, we turned out that only committers could edit (check checkboxes) bot comments, meaning half of our developers and all external developers wouldn't be able to do this. So, I had to move the list to the PR description. This is the only thing guaranteed to be editable by the PR author (unless we consider using third-party services, which I think would be even more difficult).

> * The author risks loosing edits that happen in the critical section where the `generate` workflow reads/writes the body.
> * I haven't audited the workflow code—something quite daunting, but I suspect that even an honest text editing mistake can cause parsing errors. For example, what if a user deletes the checklist marker?

generate workflow should not corrupt anything except the section marked with markers, if it does not find it, it will add it.

> * Is the checklist marker `` `---\n## ${marker}` `` or `` `<!-- ${marker} START -->` ``? Both appear in the `generate` workflow and I'm not sure I understand the logic.

One is the actual hidden markers, the second is the section title that the user sees. In theory, you could make two separate titles, two constants, but I think it's better this way.

> * The `verify` workflow "Counts all checkboxes in the entire PR description", what if a user adds their own checklists? This is common practice.
 
Yes, absolutely, this is the advantage (if the user has created another list themselves, or we add a list to the templates, it also needs to be checked), that verify workflow will check them too, this was not in the original version (we had already been testing this on zenoh-pico for 2 weeks), but after feedback I specifically added it.

> I suggest a list of checklists (i.e. per label) within the pull request template—this way we avoid any unpleasant surprises whilst not giving up on anything: we still have to trust contributors to take the checklists seriously. We don't stand to gain anything form dynamically generating checklists.

How are you going to do this? A PR template must be selected BEFORE creating a PR, and by default there's only one; if you need multiple, you'll have to select them manually. Additionally, only committers can add labels.


> If there is a concern about maintaining a single source of truth—something that the `pr-checklists.json` file under eclipse-zenoh/ci undoubtedly provides—then that can be solved with a simple sync workflow à la `sync-lockfiles.yml`.

I don't quite understand why we need to add synchronization and complicate the logic if the workflow that uses it is already in the same repository. 

> Ultimately, reviewers are the only effective guarantors of contribution quality. A [`CODEOWNERS`](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#example-of-a-codeowners-file) is something we're missing in that regard.

And they, too, should check this list. The main purpose of the list is to clearly reiterate the key points, since, unfortunately, reviewers sometimes forget to check certain things.

.github/pull_request_template.md
sashacmc: The comment is hidden, this text will remain and will be an eyesore, so make sure it is filled in.

fuzzypixelz:
> This is the only thing guaranteed to be editable by the PR author

> How are you going to do this? A PR template must be selected BEFORE creating a PR, and by default there's only one; if you need multiple, you'll have to select them manually. Additionally, only committers can add labels.

I agree. That's why I'm proposing to have the pull request template contain _all_ the checklists. This way the pull request body is read-only for the workflows. The extra checklists for which the required label is not present would be ignored by the user and the workflows.

> I don't quite understand why we need to add synchronization and complicate the logic if the workflow that uses it is already in the same repository.

I think synchronizing pull request tempplates is less complicated than writing into a contributor's pull request body.

sashacmc:
Ok, @fuzzypixelz, I got your idea, but in this case checklist will be too big, I don't want to have non relevant items, long checklist will be ignored. I prefer to have it compact and label related, to have only relevant items.

fuzzypixelz:
> Ok, @fuzzypixelz, I got your idea, but in this case checklist will be too big, I don't want to have non relevant items, long checklist will be ignored. I prefer to have it compact and label related, to have only relevant items.

So it comes down to whether the complexity of mutable pull request descriptions outweighs the benefit of "focused checklists".

I tried to wrap checklists in `<details>` elements, but quickly realized that the GitHub Markdown renderer doesn't process checklists inside `<details>` and sanitizes `<input>` elements (for security reasons). [In fact, the GFM spec doesn't support `<input>` elements at all](https://github.github.com/gfm/#html-blocks):

<details><summary>🐛 Bug Fix Requirements (using HTML)</summary>

<div><input type='checkbox'><label><b>Root cause documented</b> - Explain what caused the bug in the PR description</label></div>
<div><input type='checkbox'><label><b>Reproduction test added</b> - Test that fails on main branch without the fix</label></div>
<div><input type='checkbox'><label><b>Test passes with fix</b> - The reproduction test passes with your changes</label></div>
<div><input type='checkbox'><label><b>Regression prevention</b> - Test will catch if this bug reoccurs in the future</label></div>
<div><input type='checkbox'><label><b>Fix is minimal</b> - Changes are focused only on fixing the bug</label></div>
<div><input type='checkbox'><label><b>Related bugs checked</b> - Verified no similar bugs exist in related code</label></div>

</details>

<details><summary>✨ Enhancement Requirements (using Markdown)</summary>

[ ] - **Enhancement scope documented** - Clear description of what is being improved
[ ] - **Minimum necessary code** - Implementation is as simple as possible, doesn't overcomplicate the system
[ ] - **Backwards compatible** - Existing code/APIs still work unchanged
[ ] - **No new APIs added** - Only improving existing functionality
[ ] - **Tests updated** - Existing tests pass, new test cases added if needed
[ ] - **Performance improvement measured** - If applicable, before/after metrics provided
[ ] - **Documentation updated** - Existing docs updated to reflect improvements
[ ] - **User impact documented** - How users benefit from this enhancement

</details>

I don't think the complexity is justified for the reasons I provided above and that's the extent of what I can add to this discussion. Since I don't have any solution to address the would be dozens of lines of obnoxious checklists (even if one could argue that some of them are superfluous), you may merge this as is—as an interesting experiment where many things might go wrong.

PR #2329, opened 2025-12-22
Add retention period to RecoveryConfig

wyfo:
Each source state get an additional latest access timestamp which is incremented at each access. A garbage collection task is spawned to periodically check the least recently used source states, to decide if they should be garbage collected.

Alive publishers, as per spawned liveliness subscriber, are never garbage collected; there latest access is updated instead.

`lru` library is used to avoid making a full scan of the source states' lists at each garbage collection.

Removing a source state must also clean up the periodic query associated to it. `zenoh_util::Timer` is not built for this use case (and it is badly designed anyway), so it has been replaced by raw tokio tasks. Using `tokio_util::task::AbortOnDropHandle` makes tasks cleanup very easy.

---
## 🏷️ Label-Based Checklist

**Labels:** `enhancement`

## ✨ Enhancement Requirements

**Instructions:**


zenoh-ext/src/advanced_subscriber.rs
```diff
@@ -732,32 +808,32 @@ impl<Handler> AdvancedSubscriber<Handler> {
...
+                    .and_then(|h| h.sample_depth)
+                    .unwrap_or_default(),
+                query_target: conf.query_target,
+                query_timeout: conf.query_timeout,
+                callback: callback.clone(),
+                miss_handlers: HashMap::new(),
+                token: None,
+                _gc_task: AbortOnDropHandle::new(ZRuntime::Application.spawn(gc_task(
+                    weak.clone(),
+                    retransmission.as_ref().and_then(|r| r.retention_period),
```
DenisBiryukov91: Why not unwrap_or retention period immediately instead of passing an option which will be unwrapped anyways inside gc_task ?

DenisBiryukov91 (approved):

OlivierHecart (approved):

PR #2333, opened 2025-12-23
Raise an error if max_samples is zero in advanced pub/sub

wyfo:
A history depth of zero makes no sense, both on publisher cache side, and on subscriber query side.

There was a bug in the subscriber, as not having `sample_depth` set was preventing proper buffering. It has been fixed by considering that no `sample_depth` set means an arbitrary number of responses to the query, so responses must be buffered till the end.

---
## 🏷️ Label-Based Checklist

**Labels:** `enhancement`

## ✨ Enhancement Requirements

**Instructions:**


zenoh-ext/src/advanced_subscriber.rs
```diff
@@ -75,9 +75,16 @@ impl HistoryConfig {
     }
 
     /// Specify how many samples to query for each resource.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `depth` is zero.
     #[zenoh_macros::unstable]
     pub fn max_samples(mut self, depth: usize) -> Self {
```
DenisBiryukov91, line 81: why not accept depth as NonZeroUsize and avoid panicking inside the setter ?

zenoh-ext/src/advanced_subscriber.rs
```diff
@@ -404,7 +411,7 @@ struct State {
     key_expr: KeyExpr<'static>,
     retransmission: bool,
     period: Option<Period>,
-    history_depth: usize,
+    history_depth: Option<NonZeroUsize>,
```
DenisBiryukov91: why not keep unwrap_or'ed value of history depth here, instead unwrapping it every time in handle_sample ?

zenoh-ext/src/advanced_subscriber.rs
wyfo: Because it would make the API not so nice: I don't think that people like writing `NonZero::new(x).unwrap()`/`x.try_into().unwrap()` everywhere. Panicking on zero would give exactly the same result as users' `unwrap`.

zenoh-ext/src/advanced_subscriber.rs
wyfo: Because I don't think it matters a lot, but I will do the change.

zenoh-ext/src/advanced_subscriber.rs
wyfo: Fixed by removing the panic

DenisBiryukov91 (approved):

OlivierHecart (approved):

PR #2341, opened 2025-12-30
[ZEN-641] Undeclare key expression on drop

### LINKED ISSUE #1518: z_declare_publisher and then z_undeclare_publisher results in memory leak (zenoh-c and Rust)
### Describe the bug

The memory allocated by calling z_declare_publisher is not released when z_undeclare_publisher or z_publisher_drop is called.
I have an application which has high numbers of objects, each with a high number of updates. Publishers being created and destroyed over time results in the application memory growing and never being returned.

### To reproduce

To reproduce:
Loop calling z_declare_publisher and z_undeclare_publisher with a changing keyexpr every iteration. Monitor memory used which will increase by more than 2KB per iteration.

Example main from z_pub.c:

```
int main(int argc, char** argv) {
    zc_init_log_from_env_or("error");

    z_owned_config_t config;
    struct args_t args = parse_args(argc, argv, &config);

    printf("Opening session...\n");
    z_owned_session_t s;
    if (z_open(&s, z_move(config), NULL) < 0) {
        printf("Unable to open session!\n");
        exit(-1);
    }

    for (int idx = 0; 1; ++idx)
    {
        z_sleep_ms(10);

        z_owned_publisher_t pub;
        z_view_keyexpr_t ke;

        // Add idx to key expr
        char theKeyExp[100] = {0};
        sprintf(theKeyExp, "%s%d", args.keyexpr, idx);
        z_view_keyexpr_from_str(&ke, theKeyExp);

        if (idx%1000 == 0)
            printf("Declaring Publisher on '%s'...\n", theKeyExp);

        if (z_declare_publisher(&pub, z_loan(s), z_loan(ke), NULL) < 0) {
            printf("Unable to declare Publisher for key expression!\n");
            exit(-1);
        }
        z_sleep_ms(10);
        z_undeclare_publisher(z_move(pub));
    }
    z_close(z_move(s), NULL);
    return 0;
}

```

### System info

Platform: Intel i7 Windows 10 VS2019
Zenoh-c tag: 1.0.0.10
Reproduced in Zenoh Rust 1.0.0-beta4 on Ubuntu 22

DenisBiryukov91:
- undeclare key expression on drop to prevent temporary memory leaks when undeclaring zenoh entities
- do not undeclare key expression on wire if there are other key expressions still referencing the same declaration
- remove BorrowedWire and OwnedWire key expression variants (Borrowed and Owned now contain optional wire declaration data) to simplify the code
- Fix bug allowing to undeclare key expression using session different from the one it was declared on

Resolves #1518

### DIFF
--- a/zenoh/src/api/key_expr.rs
+++ b/zenoh/src/api/key_expr.rs
@@ -449,7 +449,7 @@ impl std::ops::Div<&keyexpr> for KeyExpr<'_> {
     fn div(self, rhs: &keyexpr) -> Self::Output {
         KeyExpr(KeyExprInner::Owned {
             key_expr: self.key_expr() / rhs,
-            declaration: self.extract_declaration(),
+            declaration: self.into_declaration(),
         })
     }
 }
@@ -504,7 +504,7 @@ impl<'a> KeyExpr<'a> {
         Ok(self)
     }
 
-    fn extract_declaration(self) -> Option<Arc<KeyExprWireDeclaration>> {
+    fn into_declaration(self) -> Option<Arc<KeyExprWireDeclaration>> {
         match self.0 {
             KeyExprInner::Borrowed { declaration, .. } => declaration,
             KeyExprInner::Owned { declaration, .. } => declaration,


zenoh/src/api/key_expr.rs
```diff
@@ -469,132 +447,132 @@ impl std::ops::Div<&keyexpr> for KeyExpr<'_> {
...
+                Some(prefix) => {
+                    KeyExprWireDeclaration::new(prefix.as_str(), session, force)?.map(Arc::new)
+                }
+                None => None,
+            }
+        }
+        Ok(self)
+    }
+
+    fn extract_declaration(self) -> Option<Arc<KeyExprWireDeclaration>> {
```
OlivierHecart: I'd call that `into_declaration`.

OlivierHecart:
I wonder if we couldn't integrate things even more by setting a `Arc<KeyExprWireDeclaration>` in the `local_resource` and reusing the `Arc` counter. But this might go in another PR ...

PR #2301, opened 2025-12-05
Connectivity API: get information about links and transports through Session::info()

milyin:
implementation for https://github.com/eclipse-zenoh/zenoh/issues/692

- implemented transports() and links() methods
- implemented transport_events_listener() and link_events_listener() methods
- test added
- z_info example demonstrates new functionality
- adminspace uses the connectivity api, no separate implementation
- the adminspace keys `@/<local_zid>/session/transport/(multicast|unicast)/<remote_zid>` and `@/<local_zid>/session/transport/(multicast|unicast)/<remote_zid>/link/<link_id>` works as before, but uses the new API. The keeping the same adminspace behavior is ensured by test https://github.com/eclipse-zenoh/zenoh/pull/2334
- the priority and reliability for `Link` are none if transport doesn't support qos
- the reliability is derived from the link protocol if not configured explicitly

---
## 🏷️ Label-Based Checklist

Based on the labels applied to this PR, please complete these additional requirements:

**Labels:** `new feature`

## 🆕 New Feature Requirements

Since this PR adds a new feature:

- [x] **Feature scope documented** - Clear description of what the feature does and why it's needed
- [x] **Minimum necessary code** - Implementation is as simple as possible, doesn't overcomplicate the system
- [x] **New APIs well-designed** - Public APIs are intuitive, consistent with existing APIs
- [x] **Comprehensive tests** - All functionality is tested (happy path + edge cases + error cases)
- [x] **Examples provided** - Usage examples in code comments or separate example files
- [x] **Documentation added** - New docs explaining the feature, its use cases, and API
- [x] **Feature flag considered** - Can the feature be enabled/disabled for gradual rollout?
- [x] **Performance impact assessed** - Memory, CPU, storage implications measured
- [x] **Integration tested** - Feature works with existing features

**Consider:** Can this feature be split into smaller, incremental PRs?

**Instructions:**
1. Check off items as you complete them (change `- [ ]` to `- [x]`)
2. The PR checklist CI will verify these are completed

*This checklist updates automatically when labels change, but preserves your checked boxes.*

PR #2367, opened 2026-01-15
fix: Delete link events were duplicated when closing the transport

milyin:
### What does this PR do?

It was found that when the `Transport` is disconnected the `LinkEvent` message may be duplicated. The PR adds test for this behavior and fixes it

Event duplication is incorrect

### Related Issues
The connectivity api was added in https://github.com/eclipse-zenoh/zenoh/pull/2301
The incorrect behavior was found in `z_api_info` test in https://github.com/eclipse-zenoh/zenoh-c/pull/1171

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -222,7 +222,7 @@ dependencies = [
  "nom 7.1.3",
  "num-traits",
  "rusticata-macros",
- "thiserror 2.0.17",
+ "thiserror 2.0.18",
  "time 0.3.41",
 ]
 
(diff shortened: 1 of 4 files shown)


io/zenoh-transport/src/unicast/universal/transport.rs
```diff
@@ -177,16 +172,17 @@ impl TransportUnicastUniversal {
             }) {
                 let is_last = guard.len() == 1;
                 if is_last {
+                    // even if closing the whole transport, still need to remove the link from the list
+                    // because multiple concurrent del_link calls could be staying on this guard
+                    *guard = vec![].into_boxed_slice();
```
OlivierHecart, line 177: Shouldn't we still drop the guard before invoking the callback ?

OlivierHecart: Guard is indeed dropped before invoking the callback. Forget that.

PR #2380, opened 2026-01-20
Make primitives consume the pushed message

wyfo:
---
## 🏷️ Label-Based Checklist

**Labels:** `internal`

## 🏠 Internal Change

**Instructions:**

### DIFF
--- a/zenoh/src/api/session.rs
+++ b/zenoh/src/api/session.rs
@@ -2460,17 +2460,13 @@ impl Session {
                 reliability,
             );
         }
+        // ext_unknown is not touched by routing/callbacks, so it must be empty
+        // we let the compiler knows it so it can optimize its drop out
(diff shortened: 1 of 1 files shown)


zenoh/src/api/session.rs
```diff
@@ -2351,66 +2388,90 @@ impl Session {
...
+                );
+            }
+            call_local(
+                callbacks,
+                &mut push,
                 #[cfg(feature = "unstable")]
                 reliability,
             );
         }
+        match push.payload {
```
OlivierHecart, line 2466: As far as I understand this piece of code is purely optimisation. If this is indeed the case could you please add a comment that informs the reader that this is purely optimisation as well as a quick explanation of why/how it optimizes.
For example I'm not sure to understand what are the benefits of the 2 match cases with identical code.

PR #2382, opened 2026-01-20
Force the response QoS to the original query QoS

wyfo:
## Description
See title

### What does this PR do?
Force the response QoS to the original query QoS. Setting the QoS (except `express`) of the response is deprecated and disabled.

Having multiple QoS for responses, especially for `ResponseFinal`, could cause priority inversion in response handling, causing responses to be lost.

### Related Issues
https://zettascale.atlassian.net/browse/ZEN-860

---
## 🏷️ Label-Based Checklist

**Labels:** `enhancement`

## ✨ Enhancement Requirements

**Instructions:**

### DIFF
--- a/zenoh/src/api/builders/querier.rs
+++ b/zenoh/src/api/builders/querier.rs
@@ -27,11 +27,7 @@ use super::sample::QoSBuilderTrait;
 #[cfg(feature = "unstable")]
 use crate::api::cancellation::{CancellationTokenBuilderTrait, SyncGroup};
 #[cfg(feature = "unstable")]
-use crate::api::query::ReplyKeyExpr;
-#[cfg(feature = "unstable")]
 use crate::api::sample::SourceInfo;
-#[cfg(feature = "unstable")]
-use crate::query::ZenohParameters;
 use crate::{
     api::{
         builders::sample::{EncodingBuilderTrait, SampleBuilderTrait},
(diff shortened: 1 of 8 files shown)


wyfo:
CI is failing because I'm adding a `#[deprecated]` attribute, and we have `CARGO_SEMVER_CHECKS_RELEASE_TYPE: patch` in the CI. 
@diogomatsubara It's wrong to block the CI like this, preventing to merge necessary minor changes. Could we refine the semver check? For example, adding a "minor" label, and change the semver check according to its presence? Of course, the CI for `main` branch would always have it set to `minor`.

zenoh/src/api/builders/reply.rs
```diff
@@ -140,14 +140,16 @@ impl<T> SampleBuilderTrait for ReplyBuilder<'_, '_, T> {
 
 #[zenoh_macros::internal_trait]
 impl<T> QoSBuilderTrait for ReplyBuilder<'_, '_, T> {
-    fn congestion_control(self, congestion_control: CongestionControl) -> Self {
-        let qos = self.qos.congestion_control(congestion_control);
-        Self { qos, ..self }
+    #[doc(hidden)]
```
OlivierHecart: Shouldn't those be still documented but marked as deprecated in doc ?
And indicated it has no effect.

zenoh/src/net/routing/dispatcher/queries.rs
OlivierHecart: The response should be sent with the qos of the query. 
This has no impact if the query has been propagated to all transports with the same qos but is important if the query qos has been changed by some interceptors or any other reasons and propagated to different transports with different qos.

zenoh/src/api/builders/reply.rs
wyfo: Why keeping documenting them if they have no effect?

OlivierHecart: So that users (in particular historical users) know that they are deprecated and that they have no effect.

PR #2387, opened 2026-01-22
Account for declared ke in Query::reply

DenisBiryukov91:
## Description
Account for declared ke in Query::reply

### What does this PR do?
If declared ke is passed to Query::reply, the optimization will be taken into account, instead of completely unwrapping ke string.

Removes unnecessary copy of ke string in Query::reply if ke is fully optimized (this is in particular always the case for advanced publisher cache).

### Related Issues
None

---
## 🏷️ Label-Based Checklist

**Labels:** `internal`

## 🏠 Internal Change

**Instructions:**


OlivierHecart:
Failing tests are independent from this PR.

Issue #2400, opened 2026-02-02
zenoh-link-tls panics when Certificate Common Name is missing


Dryvnt:
### Describe the bug

Zenoh uses the Certificate Common Name as means of identification under authentication. But not all valid TLS certificates have a Common Name property. See e.g. [Let's Encrypt's tlsserver profile](https://letsencrypt.org/docs/profiles/#tlsserver) not supporting the property.

While such certificates are not suited for TLS *zenoh authentication*, they are however still valid for TLS *zenoh transport*.

But zenoh clients cannot connect to a router that uses such a certificate, due to an unwrap in zenoh-link-tls.

### To reproduce

1. Attempt to connect to a tls endpoint that uses a certificate without a Common Name.
    - I have prepared `cert-sample.gravgaard.dev` for this purpose. There is no zenoh router on that server, but that is not required to trigger the panic.
    - i.e. `./zenohd --connect tls/cert-sample.gravgaard.dev:443`
2. See error message about thread panic, e.g.:
```
thread 'net-0' panicked at io\zenoh-links\zenoh-link-tls\src\unicast.rs:599:14:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### System info

- Zenoh version 1.7.2

oteffahi:
Fixed by https://github.com/eclipse-zenoh/zenoh/pull/2401

PR #2401, opened 2026-02-02
Don't panic on missing certificate CN

Dryvnt:
## Description
Stop zenoh-link-tls from panicking when connecting to a server that uses a certificate without a Certificate Common Name property.

### What does this PR do?
Rather than assume that every certificate has a CN, and unwrapping to generate a `TlsAuthId` value, instead return empty auth value if the certificate has no CN.

Zenoh cannot connect to routers via TLS if the certificate does not have a CN. See related issue.

### Related Issues
https://github.com/eclipse-zenoh/zenoh/issues/2400

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/io/zenoh-transport/tests/unicast_openclose.rs
+++ b/io/zenoh-transport/tests/unicast_openclose.rs
@@ -965,13 +965,13 @@ const fn get_tls_certs_without_common_name() -> (&'static str, &'static str, &'s
     //           openssl req -new -x509 -noenc -days 365000 -batch -newkey ec -pkeyopt ec_paramgen_curve:prime256v1 -sha384 -keyout ca-key.pem -out ca.pem
     //           openssl req -new -x509 -noenc -days 365000 -batch -newkey ec -pkeyopt ec_paramgen_curve:prime256v1 -sha384 -keyout server-key.pem -out server-cert.pem -CA ca.pem -CAkey ca-key.pem -addext "authorityKeyIdentifier=keyid,issuer" -addext "basicConstraints=CA:FALSE" -addext "subjectAltName=DNS:localhost" -addext "extendedKeyUsage=serverAuth,clientAuth"
 
-    const SERVER_KEY_NO_CN: &'static str = "-----BEGIN PRIVATE KEY-----
+    const SERVER_KEY_NO_CN: &str = "-----BEGIN PRIVATE KEY-----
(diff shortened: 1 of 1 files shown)


Dryvnt:
Workflow checklist issues: I can't figure out how to get the ECA check to work despite signing up, linking my GitHub account, and signing the ECA. And I can't seem to label the PR.

oteffahi:
> Workflow checklist issues: I can't figure out how to get the ECA check to work despite signing up, linking my GitHub account, and signing the ECA. And I can't seem to label the PR.

It does take a bit of time for the API to correctly expose new ECA signatures. Looks like it's fixed now.
Regarding labels, I believe only members of the project can do that. I have added the appropriate label to your PR, please review the checklist and complete your PR to the best of your abilities.
Thanks again 😄

Dryvnt:
@oteffahi I have fixed the issue in zenoh-link-quic as well, and added integration tests for TLS and QUIC 👍

io/zenoh-transport/tests/unicast_transport.rs
```diff
@@ -1803,3 +1955,40 @@ async fn transport_unicast_quic_datagram_only_server() {
...
+            reliability: Reliability::BestEffort,
+        },
+        Channel {
+            priority: Priority::RealTime,
+            reliability: Reliability::BestEffort,
+        },
+    ];
+    // Run
+    let endpoints = vec![endpoint];
+    run_with_universal_transport(&endpoints, &endpoints, &channel, &MSG_SIZE_NOFRAG).await;
```
oteffahi: Running the entire transport test-suite for this check is a bit overkill. It's enough to just open a listener transport and a client transport, and have the client not panic on connection.
Please make this change for all similar tests added in this PR.

Dryvnt: I'm wary of creating a new suite of tests, so I extended what seemed to make sense at the time. The fact that I'm not entirely sure what the tests suites do does not help my case :sweat_smile:
Is there an existing suite that would be more appropriate to use instead of unicast_transport? Perhaps unicase_openclose or unicast_bind?

Dryvnt:
Note: Rewriting the tests, I've found that the issue also applies to client certificates used during mTLS.

PR #2423, opened 2026-02-19
always close link removed from transport

milyin:
## Description

Followup to the https://github.com/eclipse-zenoh/zenoh/pull/2367/ : that fix removed problem of double event senting but caused not calling `close()` on the link. 

### What does this PR do?
Always call the `close()` method on the deleted link

This returns the correct behavior which existed before mentioned fix: the link deletion was skipped and it's `close()` was called on transport deletion. The fix #2367 added removing link without calling `close()` on it

### Related Issues
https://github.com/eclipse-zenoh/zenoh/pull/2367/

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

- **Reproduction test added** - Test that fails on main branch without the fix
- **Test passes with fix** - The reproduction test passes with your changes
- **Regression prevention** - Test will catch if this bug reoccurs in the future

**Instructions:**

The goal of the update is to restore the original state, there is no test for it, checkboxes removed

### DIFF
--- a/io/zenoh-transport/src/unicast/universal/transport.rs
+++ b/io/zenoh-transport/src/unicast/universal/transport.rs
@@ -189,11 +189,13 @@ impl TransportUnicastUniversal {
         if let Some(callback) = cb {
             callback.del_link(link);
         }
-        stl.close().await?;
         if is_last {
+            let r = stl.close().await; // do not return early to ensure that the transport is deleted even if the link close fails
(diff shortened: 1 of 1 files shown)


milyin:
Discussed with @DenisBiryukov91 . Copiloit comments are valid, but this requires deeper rework, this is out of scope of this specific fix

PR #2424, opened 2026-02-19
Use TaskController in TransportLinkUnicastUniversal

DenisBiryukov91:
## Description

Use TaskController in TransportLinkUnicastUniversal instead of TaskTracker + CancellationToken

### What does this PR do?
Replaces TaskTracker + CancellationToken in TransportLinkUnicastUniversal with TaskTracker. In addition tx_task
is made properly abortable.

Without this change cancelling tx_task could take an arbitrarily long time making the closure of a transport arbitrarily long.

### Related Issues

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/io/zenoh-transport/src/unicast/universal/link.rs
+++ b/io/zenoh-transport/src/unicast/universal/link.rs
@@ -252,9 +252,11 @@ async fn tx_task(
                 }
             }
         }
-        Ok(())
+        ZResult::Ok(())
     };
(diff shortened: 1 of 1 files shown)


OlivierHecart:
This problem is relatively hard to reproduce in a simple environment.
The fix has been manually tested.
A non regression test will hopefully come up in a set of connectivity loss tests but will not happen in this PR.

Issue #2425, opened 2026-02-19
QUIC and TLS listeners error-out on certain connection-specific errors


oteffahi:
### Describe the bug

QUIC and TLS links' `accept_task` should not be stopped when encountering errors related only to a new connection, as stopping the acceptor means no longer being able to accept new connections.

Currently this can be caused by errors getting the peer certificate's CN or expiration info. 
https://github.com/eclipse-zenoh/zenoh/blob/ccbf047725f144373398735897830f015147bfac/io/zenoh-links/zenoh-link-tls/src/unicast.rs#L508
https://github.com/eclipse-zenoh/zenoh/blob/ccbf047725f144373398735897830f015147bfac/io/zenoh-links/zenoh-link-tls/src/unicast.rs#L513

Fix: should not return, but instead log when encountering errors related to a newly accepted connection.

### To reproduce

Can happen according to the code's semantics. Didn't bother reproducing.

### System info

Zenoh >= 1.0.0

PR #2429, opened 2026-02-19
Refactor QUIC links, fix known issues

### LINKED ISSUE #2427: QUIC listener socket is missing its DSCP config
### Describe the feature

DSCP was added to all links using TCP or UDP sockets. QUIC listener does not apply the DSCP config to its UDP socket, because we forgot to add it.
https://github.com/eclipse-zenoh/zenoh/blob/ccbf047725f144373398735897830f015147bfac/io/zenoh-links/zenoh-link-quic/src/unicast.rs#L408-L428

### LINKED ISSUE #2428: QUIC client does not use MTU and PMTUD configs
### Describe the feature

QUIC datagram link added the ability to configure QUIC's MTU and PMTUD config. The config was added to the QUIC sever via `quinn::ServerConfig`, but not to the client.
https://github.com/eclipse-zenoh/zenoh/blob/c438e86b2f570db616c81070b58f6a96db7b090d/io/zenoh-links/zenoh-link-quic_datagram/src/lib.rs#L87-L90

oteffahi:
## Description
This PR refactors the QUIC codebase to eliminate the current copy/pasted chunks of code between `zenoh-link-quic` and `zenoh-link-quic_datagram` crates, and facilitates upcoming changes that rely on the `quinn` crate.
Also fixes certain issues that were discovered during this rewrite.

Depends on #2189.
Partially supersedes #2079 (does not completely remove the `zenoh-link-quic_datagram` crate).

Improve maintainability of the code.

### Related Issues
Closes #2428 (also extends MTU/PMTUD config support to streamed QUIC link)
Closes #2427
Partial fix for #2425

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`, `internal`

## 🐛 Bug Fix Requirements

## 🏠 Internal Change

**Instructions:**


io/zenoh-link-commons/src/quic/unicast.rs
```diff
@@ -0,0 +1,659 @@
...
+    pub make_link: F,
+}
+
+pub struct QuicAcceptor<F: AcceptorCallback> {
+    quic_endpoint: quinn::Endpoint,
+    tls_close_link_on_expiration: bool,
+    inner: QuicAcceptorParams<F>,
+}
+
+impl<F: AcceptorCallback> IntoFuture for QuicAcceptor<F> {
```
wyfo: This implementation is useless. Just call `quic_acceptor.accept_task().await` instead of `quic_acceptor.await`.
By the way, `async move { quic_acceptor.await }` is the same as `quic_acceptor`. So just use `quic_acceptor.accept_task()`

io/zenoh-links/zenoh-link-quic/src/unicast.rs
```diff
@@ -51,42 +38,34 @@ pub struct LinkUnicastQuic {
...
-            recv: AsyncMutex::new(recv),
+            streams,
             auth_identifier,
             expiration_manager,
         }
     }
 
     async fn close(&self) -> ZResult<()> {
         tracing::trace!("Closing QUIC link: {}", self);
-        // Flush the QUIC stream
```
wyfo, line 241: On TCP, we don't care about data race and access the `UnsafeCell`. 
If you look at the code, before `close` is called, the TX and RX tasks has been cancelled, so there should be no concurrent access to the stream.
So we should finish the streams properly.

oteffahi: This is not valid on QUIC, because the `ExpirationManager` can currently close the link on certificate expiration without the TX/RX tasks being stopped beforehand.

In hindsight maybe the `ExpirationManager` should be moved to the transport layer to avoid this issue, but for now this is an issue that makes this an undefined behavior.

PR #2432, opened 2026-02-19
Fix potential TLS acceptor crash due to malformed peer certificates

oteffahi:
## Description
Fixes potential TLS acceptor crash due to malformed peer identity when parsing its certificate.

### Related Issues
Partial fix for #2425

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/io/zenoh-links/zenoh-link-tls/src/unicast.rs
+++ b/io/zenoh-links/zenoh-link-tls/src/unicast.rs
@@ -516,7 +516,14 @@ async fn accept_task(
                         // Get certificate chain expiration
                         let mut maybe_expiration_time = None;
                         if tls_close_link_on_expiration {
-                            match get_cert_chain_expiration(&tls_conn.peer_certificates())? {
+                            let exp = match get_cert_chain_expiration(&tls_conn.peer_certificates()) {
+                                Ok(exp) => exp,
+                                Err(e) => {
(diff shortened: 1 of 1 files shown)


io/zenoh-links/zenoh-link-tls/src/unicast.rs
```diff
@@ -567,8 +580,8 @@ async fn accept_task(
 }
 
 fn get_client_cert_common_name(tls_conn: &rustls::CommonState) -> ZResult<TlsAuthId> {
-    if let Some(serv_certs) = tls_conn.peer_certificates() {
-        let (_, cert) = X509Certificate::from_der(serv_certs[0].as_ref())?;
+    if let Some(client_certs) = tls_conn.peer_certificates() {
+        let (_, cert) = X509Certificate::from_der(client_certs[0].as_ref())?;
```
fuzzypixelz, line 584: I'm still uneasy about this index access. The documentation does not clearly state that the result of `peer_certificates` is either `None` or `Some` with at least one entry. But the implementation appears to be correct:
https://github.com/rustls/rustls/blob/66c931a87dd5c32fb69cd62b69f6ca79fc9b4703/rustls/src/server/tls13.rs#L1040-L1053.

PR #2438, opened 2026-02-23
Ensure that no new transports to the same zid can be established while the current one is not fully closed

DenisBiryukov91:
## Description

Prevent opening new transports to the same zid while the current one is not fully closed.

### What does this PR do?

Prevent opening new transports to the same zid while the current one is not fully closed.

To prevent having 2 or more faces for the same zid.

### Related Issues

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/io/zenoh-transport/src/unicast/lowlatency/rx.rs
+++ b/io/zenoh-transport/src/unicast/lowlatency/rx.rs
@@ -82,7 +82,12 @@ impl TransportUnicastLowlatency {
 
             match msg.body {
                 zenoh_protocol::transport::TransportBodyLowLatency::Close(_) => {
-                    let _ = self.delete().await;
+                    // Spawn a task to avoid a deadlock waiting for this same task
+                    // to finish in the link close() joining the rx handle
+                    let c_transport = self.clone();
+                    zenoh_runtime::ZRuntime::Net.spawn(async move {
+                        let _ = c_transport.delete().await;
+                    });
                 }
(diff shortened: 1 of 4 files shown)


OlivierHecart:
This problem is relatively hard to reproduce in a simple environment.
The fix has been manually tested.
A non regression test will hopefully come up in a set of connectivity loss tests but will not happen in this PR.

Micbetter:
Thanks for the fix. 
Does this mean that transports with the same zid are not allowed, even if they connect to different routers? Such as this topology:
image

DenisBiryukov91:
> Thanks for the fix. Does this mean that transports with the same zid are not allowed, even if they connect to different routers? Such as this topology: image

This should be allowed, since the check is done only within single zenoh session/runtime, but I believe this kind of configuration might eventually  lead to undefined behavior and difficult to debug situations (if for example there is an eventual fallback from Router(A) to Router(B) if Router(A) fails)

Micbetter:
I agree with you. This situation should never happen, but for now we can only prevent it from happening via a high-level implementation, rather than a Zenoh implementation. Am I correct?

DenisBiryukov91:
> I agree with you. This situation should never happen, but for now we can only prevent it from happening via a high-level implementation, rather than a Zenoh implementation. Am I correct?

Yes you are correct, as of today router acts as a gateway for the clients/peers and abstracts the information about that subnet, so it's internal topology is not exposed to other parts of the network.

PR #2463, opened 2026-03-09
feat(cuda): add CUDA IPC transport via ZSliceKind::CudaPtr and zenoh-cuda crate

YuanYuYuan:
## Summary

- Adds `ZSliceKind::CudaPtr` (0x2) and `ZSliceKind::CudaTensor` (0x3) to `zenoh-buffers`,
  enabling GPU device memory to travel as a Zenoh payload via 64-byte CUDA IPC handles
  (zero CPU copies of tensor data)
- Introduces `zenoh-cuda` crate: `CudaBufInner` for owning/borrowed device allocations,
  IPC handle serialization, `copy_to_host` (DeviceToHost via `cudaMemcpy`), and DLPack
  `TensorMeta` for typed N-D tensors with PyTorch pool sub-allocation offset support
- Introduces `zenoh-mem-transport` crate: pluggable `ZeroMemTransport` backend trait,
  `MemRegistry` for session-scoped backend registration, `CudaIpcBackend` and `ShmBackend`,
  `NegotiatedMemCaps` for per-connection capability tracking, and `ext_mem` wire format
- Adds `ci_cuda.yml`: Docker-based CUDA SDK CI job using
  `ghcr.io/rust-gpu/rust-cuda-ubuntu24-cuda12` on `ubuntu-latest` (no physical GPU needed)

## Key Changes

- `commons/zenoh-buffers`: `ZSliceKind` variants, feature-gated under `cuda`
- `commons/zenoh-cuda`: new crate with CUDA Runtime FFI, `CudaBufInner`, `TensorMeta`
- `commons/zenoh-codec`: `CudaPtr`/`CudaTensor` encode/decode arms in `zbuf.rs`
- `commons/zenoh-mem-transport`: `ZeroMemTransport` trait, `MemRegistry`, backends, `ext_mem`
  TLV wire format, `NegotiatedMemCaps`
- `io/zenoh-transport`: `map_zmsg_to_partner_with_caps` — capability-aware TX entry point
- `.github/workflows/ci_cuda.yml`: new CI workflow for CUDA compilation and unit tests
- 16 unit tests + `registry_negotiate` example

## Architecture: why CUDA IPC is not a `ShmProviderBackend`

`ShmProvider` is a CPU heap in shared OS memory. Its contract assumes:
- `alloc()` returns a `ChunkDescriptor` with a **CPU-side pointer offset**
- `ShmBufInfo` wire format carries `(SegmentID: u32, offset, data_len, generation)` — no
  room for a 64-byte IPC handle
- `ShmClient::attach(SegmentID)` opens a POSIX `shm_open`/`mmap` segment
- The watchdog heartbeat is a CPU-mapped atomic — cannot be hosted in device memory
- `ZSliceBuffer::as_slice()` returns CPU-readable bytes

CUDA device memory has no CPU address (`as_slice()` returns `&[]`), the IPC handle is
64 opaque bytes + `device_id`, and `cudaIpcOpenMemHandle` has completely different
lifetime semantics. More fundamentally, Zenoh never owns the GPU allocation — the tensor
already exists in the framework allocator (PyTorch caching allocator, RAPIDS, etc.).
There is no allocation lifecycle for `ShmProvider` to manage.

`ShmProvider` appears exactly once in the CUDA flow: as a staging area for the
`DowngradePath::ToShm` fallback (`cudaMemcpy DeviceToHost` → raw bytes → wrap into an
SHM region for a receiver that supports POSIX SHM but not CUDA IPC).

## Architecture: why `ext_mem` (0x8) supersedes the planned `ext_cuda`

The original design had `ext_shm (0x5)` for POSIX SHM and a planned `ext_cuda` for CUDA
device IDs. `ext_cuda` was replaced by `ext_mem (0x8)` because both share the same
structural need: versioned, TLV-encoded, per-backend capability advertisement in
`InitSyn`/`InitAck`.

The critical difference from `ext_shm`: `ext_shm` has **boolean semantics** (present or
absent, with a side-channel auth segment ID). `ext_mem` has **typed payload per backend**,
which is what CUDA needs to advertise device IDs and what future RDMA backends will need
for queue pair info. `ext_mem` is a strict superset: it can carry the SHM segment ID
alongside CUDA device IDs and RDMA parameters in a single versioned TLV frame.

```
[version=1] [n_backends]
  [id=0, len=8,       segment_id: u64]              ← SHM (replaces ext_shm payload)
  [id=1, len=1+n*4,   n_devs, dev_ids: i32...]      ← CUDA IPC
  [id=2, len=0]                                      ← RDMA (reserved)
```

## Breaking Changes

None — new feature-gated crates and CI workflow, no changes to existing public APIs.
`ZSliceKind` variants are additive and guarded by `#[cfg(feature = "cuda")]`.

## Phased implementation

This PR covers Phases 1–2 (wire protocol, codec, negotiation framework). Remaining work:
- Phase 3: migrate `write_handle`/`read_handle` from the codec into the backend traits
- Phase 4: `DowngradePath::ToShm` staging (requires `ShmProvider` reference in backend)
- Phase 5: wire `map_zmsg_to_partner_with_caps` into the transport TX path

---
## 🏷️ Label-Based Checklist

**Labels:** `new feature`

## 🆕 New Feature Requirements

- [ ] **Feature scope documented** - Clear description of what the feature does and why it's needed
- [ ] **New APIs well-designed** - Public APIs are intuitive, consistent with existing APIs
- [ ] **Comprehensive tests** - All functionality is tested (happy path + edge cases + error cases)
- [ ] **Examples provided** - Usage examples in code comments or separate example files
- [ ] **Documentation added** - New docs explaining the feature, its use cases, and API
- [ ] **Feature flag considered** - Can the feature be enabled/disabled for gradual rollout?
- [ ] **Performance impact assessed** - Memory, CPU, storage implications measured
- [ ] **Integration tested** - Feature works with existing features

**Instructions:**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -434,7 +434,7 @@ jobs:
           RUST_LOG: None
         run: |
           sudo prlimit --memlock=unlimited --pid=$$
-          cargo llvm-cov test --features unstable --features test --features shared-memory \
+          cargo llvm-cov test --workspace --features unstable --features test --features shared-memory \
           --exclude zenoh-cuda \
           ${{ matrix.rust == 'nightly' && '--doctests' || '' }} --lcov --output-path lcov.info \
           --no-cfg-coverage --no-cfg-coverage-nightly --ignore-run-fail -- \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,14 +141,14 @@ jobs:
         run: cargo +stable clippy -p zenoh --all-targets --features unstable,shared-memory -- --deny warnings
 
       - name: Clippy workspace
-        run: cargo +stable clippy --all-targets --features test -- --deny warnings
+        run: cargo +stable clippy --all-targets --exclude zenoh-cuda --features test -- --deny warnings
 
       - name: Clippy workspace unstable
-        run: cargo +stable clippy --all-targets --features unstable,test -- --deny warnings
+        run: cargo +stable clippy --all-targets --exclude zenoh-cuda --features unstable,test -- --deny warnings
 
       - name: Clippy all features
         if: ${{ matrix.os == 'ubuntu-latest' || matrix.os == 'macos-latest' }}
-        run: cargo +stable clippy --all-targets --all-features -- --deny warnings
+        run: cargo +stable clippy --all-targets --exclude zenoh-cuda --all-features -- --deny warnings
 
       - name: Install generic no_std target
         # Generic no_std target architecture is x86_64-unknown-none
@@ -260,13 +260,13 @@ jobs:
         if: ${{ matrix.os == 'macos-latest' }}
         run: |
           sudo cargo build -p zenoh-plugin-rest --lib
-          sudo cargo nextest run -F test -F plugins,unstable,internal --exclude zenoh-examples --exclude zenoh-plugin-example --workspace
+          sudo cargo nextest run -F test -F plugins,unstable,internal --exclude zenoh-examples --exclude zenoh-plugin-example --exclude zenoh-cuda --workspace
 
       - name: Run tests with unstable
         if: ${{ matrix.os != 'macos-latest' }}
         run: |
           cargo build -p zenoh-plugin-rest --lib
-          cargo nextest run -F test -F plugins,unstable,internal --exclude zenoh-examples --exclude zenoh-plugin-example --workspace
+          cargo nextest run -F test -F plugins,unstable,internal --exclude zenoh-examples --exclude zenoh-plugin-example --exclude zenoh-cuda --workspace
 
       - name: Upload test results to Codecov
         if: ${{ !cancelled() }}
@@ -306,19 +306,19 @@ jobs:
 
       - name: Run tests with SHM (sudo macos)
         if: ${{ matrix.os == 'macos-latest' }}
-        run: sudo cargo nextest run -F test -F shared-memory -F unstable -E 'not (test(test_default_features) or test(test_adminspace_read))' --exclude zenoh-examples --exclude zenoh-plugin-example --workspace
+        run: sudo cargo nextest run -F test -F shared-memory -F unstable -E 'not (test(test_default_features) or test(test_adminspace_read))' --exclude zenoh-examples --exclude zenoh-plugin-example --exclude zenoh-cuda --workspace
         env:
           RUST_BACKTRACE: 1
 
       - name: Run tests with SHM
         if: ${{ matrix.os == 'windows-latest' }}
-        run: cargo nextest run -F test -F shared-memory -F unstable -E 'not (test(test_default_features) or test(test_adminspace_read))' --exclude zenoh-examples --exclude zenoh-plugin-example --workspace
+        run: cargo nextest run -F test -F shared-memory -F unstable -E 'not (test(test_default_features) or test(test_adminspace_read))' --exclude zenoh-examples --exclude zenoh-plugin-example --exclude zenoh-cuda --workspace
 
       - name: Run tests with SHM + unixpipe
         if: ${{ matrix.os == 'ubuntu-latest' }}
         run: |
           sudo prlimit --memlock=unlimited --pid=$$
-          cargo nextest run -F test -F shared-memory -F unstable -F transport_unixpipe -E 'not (test(test_default_features) or test(test_adminspace_read))' --exclude zenoh-examples --exclude zenoh-plugin-example --workspace
+          cargo nextest run -F test -F shared-memory -F unstable -F transport_unixpipe -E 'not (test(test_default_features) or test(test_adminspace_read))' --exclude zenoh-examples --exclude zenoh-plugin-example --exclude zenoh-cuda --workspace
 
       - name: Upload test results to Codecov
         if: ${{ !cancelled() }}
@@ -435,6 +435,7 @@ jobs:
         run: |
           sudo prlimit --memlock=unlimited --pid=$$
           cargo llvm-cov test --features unstable --features test --features shared-memory \
+          --exclude zenoh-cuda \
           ${{ matrix.rust == 'nightly' && '--doctests' || '' }} --lcov --output-path lcov.info \
           --no-cfg-coverage --no-cfg-coverage-nightly --ignore-run-fail -- \
           --skip test_default_features \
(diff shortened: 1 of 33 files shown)


yellowhatter:
> ## Architecture: why CUDA IPC is not a `ShmProviderBackend`
> 
> `ShmProvider` is a CPU heap in shared OS memory. Its contract assumes:
> 
>     * `alloc()` returns a `ChunkDescriptor` with a **CPU-side pointer offset**

```
pub struct ChunkDescriptor {
    pub segment: SegmentID,
    pub chunk: ChunkID,
    pub len: NonZeroUsize,
}
```
`ChunkID` is not necessarily an offset, it's interpretation is backend-defined. For CUDA backend, `ChunkDescriptor` can point either to null-state or to corresponding `PosixShmProviderBackend`-compatible segment, providing lazy init semantics.
In other words, by default CUDA buffer can have null-state `ChunkDescriptor`, on any access to CPU memory it will implicitly allocate POSIX SHM buffer and plug it in, filling itself with `PosixShmProviderBackend`-compatible `ChunkDescriptor` data (while still staying CUDA-SHM-buffer-typed). Such dualism for CUDA buffers give out-of-the-box compatibility with any other SHM mechanics we have: implicit CUDA->RAW conversion when crossing SHM domain boundary, CUDA->SHM conversion when crossing CUDA domain boundary, refcount and watchdog-based lifetime tracking for lazily-initialized CPU memory, safe write access, one-to-many buffer sharing, etc
 
>     * `ShmBufInfo` wire format carries `(SegmentID: u32, offset, data_len, generation)` — no
>       room for a 64-byte IPC handle

Any handles can be placed into SHM Metadata segment that already exists and designed to have backward-compatible extensibility.

>     * `ShmClient::attach(SegmentID)` opens a POSIX `shm_open`/`mmap` segment

This is backend-defined behavior.
 
>     * The watchdog heartbeat is a CPU-mapped atomic — cannot be hosted in device memory

Watchdog itself is useful for CUDA-to-host mappings, see my answer above
 
>     * `ZSliceBuffer::as_slice()` returns CPU-readable bytes

Perfect point for CUDA-to-CPU conversion, , see my answer above

> 
> CUDA device memory has no CPU address (`as_slice()` returns `&[]`), the IPC handle is 64 opaque bytes + `device_id`, and `cudaIpcOpenMemHandle` has completely different lifetime semantics. More fundamentally, Zenoh never owns the GPU allocation — the tensor already exists in the framework allocator (PyTorch caching allocator, RAPIDS, etc.). There is no allocation lifecycle for `ShmProvider` to manage.

For external mappings, `ShmProvider::map` method exist 

> `ShmProvider` appears exactly once in the CUDA flow: as a staging area for the `DowngradePath::ToShm` fallback (`cudaMemcpy DeviceToHost` → raw bytes → wrap into an SHM region for a receiver that supports POSIX SHM but not CUDA IPC).
> ## Architecture: why `ext_mem` (0x8) supersedes the planned `ext_cuda`

Covered with my answer above
 
> The original design had `ext_shm (0x5)` for POSIX SHM and a planned `ext_cuda` for CUDA device IDs. `ext_cuda` was replaced by `ext_mem (0x8)` because both share the same structural need: versioned, TLV-encoded, per-backend capability advertisement in `InitSyn`/`InitAck`.
> 
> The critical difference from `ext_shm`: `ext_shm` has **boolean semantics** (present or absent, with a side-channel auth segment ID). `ext_mem` has **typed payload per backend**, which is what CUDA needs to advertise device IDs and what future RDMA backends will need for queue pair info. `ext_mem` is a strict superset: it can carry the SHM segment ID alongside CUDA device IDs and RDMA parameters in a single versioned TLV frame.

All CUDA per-device accesses can be negotiated with auth segment negotiation routine, as it is also designed to have the same backward-compatibility as SHM Metadata segment.

YuanYuYuan:
Hi @yellowhatter. Thank you for the clarification. You're right. Let me update my understanding.

- ChunkID is backend-defined (u32) — a CUDA backend could use it as a handle-table index rather than a CPU offset.
- `ShmClient::attach` is backend-defined — a CUDA client could call `cudaIpcOpenMemHandle` instead of shm_open/mmap.
- The SHM metadata segment can hold a 64-byte IPC handle — the format already has extensibility headroom.
- `ShmProvider::map` exists for externally-allocated chunks — this is the right hook for PyTorch-owned tensors and we didn't account for it.
- The watchdog heartbeat is useful for the lazy CPU-mapped copy path, not just for pure POSIX SHM.

Therefore it's actually can be implemented via ShmProvider.


The approach you're describing yields a "CUDA buffer that lazily provides CPU access via SHM" — `as_slice()` triggers a `cudaMemcpy` `DeviceToHost` into a freshly allocated SHM buffer. This is the right design when the receiver might be either GPU-capable or CPU-only and SHM compatibility is the priority.

However, the current target use case is different: **GPU-to-GPU zero-copy** IPC where the receiver processes the tensor on GPU (ros-z/torch: PyTorch publisher → PyTorch subscriber, both on the same or peer CUDA device). On that path, `as_slice()` should never be called; if it is, it means the capability negotiation failed silently. I prefer the lazy-copy trigger to be explicit — `DowngradePath::ToRaw` or `ToShm` selected at handshake time — rather than implicit inside `as_slice()`.

Three concrete costs with the lazy-`as_slice()` approach in our context:

1. **Interior mutability on every CPU read.** `Mutex<Option<ShmBufInner>>` is held across every `as_slice()` call to guard the one-time copy. For the happy path (GPU-only receiver), this lock is taken and released on every message receive with no benefit.
2. **`PtrInSegment` assumes a CPU address.** `AllocatedChunk.data` is a `PtrInSegment { ptr: *mut u8, … }`. For pure device memory there is no CPU address; `ptr` would be null, and all existing SHM code that calls `.ptr()` or `.ptr_mut()` would require null-checks throughout.
3. **CUDA-specific fields in general SHM metadata.** `byte_offset` (PyTorch pool sub-allocation offset within the IPC pool block) and `TensorMeta` (DLPack shape/dtype/strides) are CUDA-specific. Putting them in the general SHM metadata segment means every SHM parser needs to understand or skip them.

On capability negotiation: whether device IDs live in an extended auth segment or in `ext_mem`, the wire-level change is structurally the same — a versioned typed payload per backend in InitSyn/InitAck. `ext_mem` provides one slot for all backends (SHM, CUDA, future RDMA) rather than one extension per backend.

godardt:
Getting support for this would be fantastic!

PR #2469, opened 2026-03-10
Remove callback generic parameter

wyfo:
This was an unwanted and useless breaking change

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`, `internal`, `api fix`

## 🐛 Bug Fix Requirements

## 🏠 Internal Change

**Instructions:**

### DIFF
--- a/zenoh/src/api/handlers/mod.rs
+++ b/zenoh/src/api/handlers/mod.rs
@@ -29,7 +29,7 @@ use crate::api::session::API_DATA_RECEPTION_CHANNEL_SIZE;
 /// while granting you access to the handler through the returned value via [`std::ops::Deref`] and [`std::ops::DerefMut`].
 ///
 /// Any closure that accepts `T` can be converted into a pair of itself and `()`.
-pub trait IntoHandler<T: CallbackParameter> {
+pub trait IntoHandler<T> {
     type Handler;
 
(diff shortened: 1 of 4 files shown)


zenoh/src/api/handlers/mod.rs
```diff
@@ -29,7 +29,7 @@ use crate::api::session::API_DATA_RECEPTION_CHANNEL_SIZE;
 /// while granting you access to the handler through the returned value via [`std::ops::Deref`] and [`std::ops::DerefMut`].
 ///
 /// Any closure that accepts `T` can be converted into a pair of itself and `()`.
-pub trait IntoHandler<T: CallbackParameter> {
```
fuzzypixelz, line 32: Could we add a regression test with a simple `IntoHandler` impl?

wyfo: What do you mean?

fuzzypixelz: Add a test that implements `IntoHandler`, like a user would. If there is such a breaking change in the future, the test would fail to compile.

fuzzypixelz: You have ~30mins before the auto-merge kicks in ;)

PR #2471, opened 2026-03-10
Fix race condition that could lead to having more links per transport than allowed by config

DenisBiryukov91:
## Description

Fix race condition that could lead to having more links per transport than allowed by config

### What does this PR do?
Fix race condition that could lead to having more links per transport than allowed by config

It fixes a race condition introduced by #2438 

### Related Issues

---
## 🏷️ Label-Based Checklist

**Labels:** `internal`

## 🏠 Internal Change

**Instructions:**

### DIFF
--- a/io/zenoh-transport/src/unicast/universal/transport.rs
+++ b/io/zenoh-transport/src/unicast/universal/transport.rs
@@ -210,7 +210,6 @@ impl TransportUnicastUniversal {
 
         match *status_guard {
             TransportStatus::Uninitialized => {
-                *status_guard = TransportStatus::Alive;
                 let csn = PrioritySn {
                     reliable: initial_sn_rx,
                     best_effort: initial_sn_rx,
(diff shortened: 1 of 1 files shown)


io/zenoh-transport/src/unicast/universal/transport.rs
```diff
@@ -210,14 +210,14 @@ impl TransportUnicastUniversal {
...
             TransportStatus::Uninitialized => {
-                *status_guard = TransportStatus::Alive;
                 let csn = PrioritySn {
                     reliable: initial_sn_rx,
                     best_effort: initial_sn_rx,
                 };
                 for c in self.priority_rx.iter() {
                     c.sync(csn)?;
                 }
+                *status_guard = TransportStatus::Alive;
```
oteffahi, line 220: Maybe this needs to be kept where it was?

DenisBiryukov91: I believe it makes more sense to set Alive status only in case of success. Otherwise, this value is nor being read anywhere except this function (and in case of failure the transport will be closed anyways), should this should be ok.

PR #2519, opened 2026-03-24
Fix RuntimeState leak

yellowhatter:
## Description

TDD to identify and fix RuntimeState leak problem

Root Cause:

The Network struct (in zenoh/src/net/protocol/network.rs) held a strong reference to Runtime via its runtime: Runtime field. This created a circular reference:

    Session holds Arc<Runtime> (strong reference) → Runtime.state: Arc<RuntimeState>
    Runtime wraps the router and protocol components that include Network
    Network.runtime: Runtime (strong reference back to Runtime)

This circular dependency prevented RuntimeState from being properly deallocated:

    When Session is dropped, its strong reference to Runtime is released
    However, Network still holds a strong reference to Runtime
    Since Network is part of the runtime's routing internals, RuntimeState cannot be freed
    Result: RuntimeState remains alive indefinitely (memory leak)

Solution:

Changed the Network.runtime field from a strong reference (Runtime) to a weak reference (WeakRuntime) defined in zenoh/src/net/runtime/mod.rs. This breaks the circular dependency:

    Network.runtime: WeakRuntime (weak reference)
    WeakRuntime wraps Weak<RuntimeState> instead of Arc<RuntimeState>
    When Session is dropped, Runtime can now be fully dropped
    Network can safely access Runtime via .upgrade().unwrap() when needed
    When the upgrade fails, it indicates the runtime has been cleaned up

The fix maintains all necessary functionality while eliminating the cycle that prevented proper cleanup of RuntimeState.

### What does this PR do?
Tests RuntimeState leakage and aims to fix this issue

RuntimeState leaks under some circumstances

### Related Issues
None

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`, `release`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2295,9 +2295,9 @@ checksum = "b6d2cec3eae94f9f509c767b45932f1ada8350c4bdb85af2fcab4a3c14807981"
 
 [[package]]
 name = "libredox"
-version = "0.1.14"
+version = "0.1.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1744e39d1d6a9948f4f388969627434e31128196de472883b39f148769bfe30a"
+checksum = "7ddbf48fd451246b1f8c2610bd3b4ac0cc6e149d89832867093ab69a17194f08"
 dependencies = [
  "libc",
 ]
@@ -3854,9 +3854,9 @@ dependencies = [
 
 [[package]]
 name = "serde_spanned"
-version = "1.0.4"
+version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f8bbf91e5a4d6315eee45e704372590b30e260ee83af6639d64557f51b067776"
+checksum = "876ac351060d4f882bb1032b6369eb0aef79ad9df1ea8bc404874d8cc3d0cd98"
 dependencies = [
  "serde_core",
 ]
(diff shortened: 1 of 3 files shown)


zenoh/src/net/protocol/network.rs
```diff
@@ -337,7 +337,7 @@ impl Network {
             whatami: self.graph[idx].whatami,
             locators: if details.locators {
                 if idx == self.idx {
-                    Some(self.runtime.get_locators())
+                    Some(self.runtime.upgrade().unwrap().get_locators())
```
OlivierHecart, line 340: I'd prefer avoid the unwrap. (unwrap_or ?)
With warning ?

zenoh/src/net/protocol/network.rs
```diff
@@ -617,8 +617,8 @@ impl Network {
     }
 
     fn connect_discovered_peer(&self, zid: ZenohIdProto, locators: Vec<Locator>) {
-        let runtime = self.runtime.clone();
-        self.runtime.spawn(async move {
+        let runtime = self.runtime.upgrade().unwrap();
```
OlivierHecart, line 620: I'd prefer avoid the unwraps. (if let ?)
With warning ?

zenoh/src/net/protocol/network.rs
yellowhatter: It is the same as `GossipNet` implementation:
https://github.com/eclipse-zenoh/zenoh/blob/2f403d4688497e3b9d5a193731e71696b5b6e137/zenoh/src/net/protocol/gossip.rs#L276

Network implementation is owned by the `Runtime`, so it is safe and this unwrap should never fail. I'd rather vote for spending time to switch from RuntimeWeak to &'a Runtime for Networks rather than adding warning handling and returning `ZResult` that never happen

zenoh/src/net/protocol/network.rs
yellowhatter: Answered above.

yellowhatter:
> Why changing the runtime in network.rs and n to in gossip.rs ? I think we should change both at least for coherency.

Gossip already has this `RuntimeWeak`:
https://github.com/eclipse-zenoh/zenoh/blob/2f403d4688497e3b9d5a193731e71696b5b6e137/zenoh/src/net/protocol/gossip.rs#L186

PR #2524, opened 2026-03-27
Fix unknown scope late push

### LINKED ISSUE #980: Sporadic "Route data with unknown scope"
### Describe the bug

If I start my example it sometimes prints an error:

zenoh-issue

On the subscriber side I then see a missing payload.

### To reproduce

1. I start my subscriber app
2. I run my publish example many times

Zenoh default config on both sides.

My app is fairly complex (and closed source) but I could try to make a minimal reproduction if I can get some context from maintainers what the error message means and what I should look for etc.

### System info

System is rolling endeavour OS.

Using Zenoh 0.11.0-rc.

### LINKED ISSUE #1050: Sending fails to stop properly after the session is closed. ERROR: Route data with unknown scope
### Describe the bug

```console
2024-05-24T09:16:37.076191Z  INFO main ThreadId(01) zenoh::net::runtime: Using ZID: c8fb696718026344604f9b074a3ff1e7
[ping] Started measuring for 5(s)
[ping] RTT(us) p05: 93393, p50: 769794, p95: 1409960, avg: 763211.63
[ping] Recv rate 520142/705828 = 73.69% messages
[ping] Done
2024-05-24T09:16:42.795357Z ERROR rx-0 ThreadId(07) zenoh::handlers: sending on a closed channel
2024-05-24T09:16:42.797147Z ERROR rx-0 ThreadId(07) zenoh::net::routing::dispatcher::pubsub: Route data with unknown scope 2!
2024-05-24T09:16:42.797181Z ERROR rx-0 ThreadId(07) zenoh::net::routing::dispatcher::pubsub: Route data with unknown scope 2!
2024-05-24T09:16:42.797188Z ERROR rx-0 ThreadId(07) zenoh::net::routing::dispatcher::pubsub: Route data with unknown scope 2!
...
2024-05-24T09:16:42.798251Z ERROR rx-0 ThreadId(07) zenoh::net::routing::dispatcher::pubsub: Route data with unknown scope 2!
2024-05-24T09:16:42.798258Z ERROR rx-0 ThreadId(07) zenoh::net::routing::dispatcher::pubsub: Route data with unknown scope 2!
2024-05-24T09:16:42.798264Z ERROR rx-0 ThreadId(07) zenoh::net::routing::dispatcher::pubsub: Route data with unknown scope 2!
2024-05-24T09:16:42.799943Z  WARN net-0 ThreadId(05) zenoh::net::runtime::orchestrator: Unable to connect to tcp/127.0.0.1:7447! Received a close message (reason MAX_LINKS) in response to an OpenSyn on: TransportLinkUnicast { link: Link { src: tcp/127.0.0.1:44314, dst: tcp/127.0.0.1:7447, mtu: 65535, is_reliable: true, is_streamed: true }, config: TransportLinkUnicastConfig { direction: Outbound, batch: BatchConfig { mtu: 65535, is_streamed: true, is_compression: false } } } at io/zenoh-transport/src/unicast/establishment/open.rs:444.
2024-05-24T09:16:42.800186Z ERROR net-0 ThreadId(05) zenoh::net::runtime::orchestrator: [] Unable to connect to any of [tcp/127.0.0.1:7447]!  at zenoh/src/net/runtime/orchestrator.rs:278.

```

### To reproduce

1. Add `chrono`
```console
cargo add chrono
```
2. Put the following file in _examples_ and name it as _z_ping_parallel_.
```rust
use clap::Parser;
use std::sync::{
    atomic::{AtomicBool, AtomicUsize, Ordering::SeqCst},
    Arc,
};
use std::time::Duration;
use zenoh::config::Config;
use zenoh::publication::CongestionControl;
use zenoh_examples::CommonArgs;


#[tokio::main]
async fn main() {
    // initiate logging
    zenoh_util::try_init_log_from_env();

    let (config, timeout, size) = parse_args();
    let session = zenoh::open(config).res().await.unwrap().into_arc();

    // The key expression to publish data on
    let key_expr_ping = keyexpr::new("test/ping").unwrap();

    // The key expression to wait the response back
    let key_expr_pong = keyexpr::new("test/pong").unwrap();


    let is_running = Arc::new(AtomicBool::new(true));
    let counter = Arc::new(AtomicUsize::new(0));
    let publisher = session
        .declare_publisher(key_expr_ping)
        .congestion_control(CongestionControl::Block)
        .res()
        .await
        .unwrap();
    tokio::task::spawn({
        let is_running = is_running.clone();
        let counter = counter.clone();
        async move {
            while is_running.load(SeqCst) {
                let start = chrono::offset::Utc::now().timestamp_nanos_opt().unwrap();
                let mut payload = vec![1u8; size];
                payload[..8].clone_from_slice(&start.to_le_bytes());
                publisher.put(payload.to_vec()).res().await.unwrap();
                counter.fetch_add(1, SeqCst);
                // std::thread::sleep(std::time::Duration::from_millis(50));
            }
        }
    });

    println!("[ping] Started measuring for {}(s)", timeout.as_secs());
    let sub = session.declare_subscriber(key_expr_pong).res().await.unwrap();
    let timer = std::time::Instant::now();
    let mut samples = vec![];
    while let Ok(sample) = sub.recv_async().await {
        let payload = sample.value.payload.contiguous();
        let start = i64::from_le_bytes(payload[..8].try_into().unwrap());
        let end = chrono::offset::Utc::now().timestamp_nanos_opt().unwrap();
        let elpased = (end - start) / 1000;
        samples.push(elpased);
        if timer.elapsed() > timeout {
            break;
        }
    }

    is_running.swap(false, SeqCst);
    samples.sort();
    let sum: i64 = samples.iter().sum();
    println!(
        "[ping] RTT(us) p05: {}, p50: {}, p95: {}, avg: {:.02}",
        samples[(samples.len() as f64 * 0.05) as usize],
        samples[(samples.len() as f64 * 0.50) as usize],
        samples[(samples.len() as f64 * 0.95) as usize],
        sum as f64 / samples.len() as f64,
    );

    let sent = counter.load(SeqCst);
    let recv = samples.len();
    let rate = recv as f64 / sent as f64 * 100.0;
    println!("[ping] Recv rate {recv}/{sent} = {rate:.02}% messages");

    println!("[ping] Done");

}

#[derive(Parser)]
struct Args {
    #[arg(short, long, default_value = "10")]
    /// The number of seconds to warm up (u64)
    timeout: u64,
    #[arg(short, long, default_value = "32")]
    /// Sets the size of the payload to publish
    payload_size: usize,
    #[command(flatten)]
    common: CommonArgs,
}

fn parse_args() -> (Config, Duration, usize) {
    let args = Args::parse();
    (
        args.common.into(),
        Duration::from_secs(args.timeout),
        args.payload_size,
    )
}
```
3. Execute the test
```console
cargo run --example z_pong -- --listen tcp/127.0.0.1:7447
cargo run --example z_ping_parallel -- --mode client --connect tcp/127.0.0.1:7447 --timeout 5
```

### System info

- zenoh: 25f06bd

eclipse0922:
## Description

  This PR fixes silent pub/sub data loss when a late in-flight `Push` arrives after
  its `ExprId` mapping has already been removed by `unregister_expr()` or
  `close_face()`.

  ### What does this PR do?

  - Adds a bounded per-face late-mapping cache so `route_data()` can recover the
    full key expression when normal `remote_mappings` lookup fails.
  - Clears stale cache entries on `ExprId` reuse to avoid misrouting.
  - Drains `remote_mappings` into the late cache during forced face teardown in
    all hat paths (`client`, `linkstate_peer`, `p2p_peer`, `router`).
  - Adds a per-face byte budget and eviction policy to keep the cache bounded.
  - Adds regression coverage for:
    - late push after undeclare
    - forced face close with queued data
    - late-cache byte-budget eviction
    - end-to-end repeated declare/publish/drop cycles

  Today, if control-path cleanup removes an `ExprId` mapping before all queued
  data-path messages for that `ExprId` have been drained, `route_data()` cannot
  resolve the scope and drops the sample with an `unknown scope` error. This
  creates silent subscriber-side data loss during disconnects, undeclares, and
  unstable network conditions.

  The fix keeps recovery local to the routing dispatcher and face teardown path,
  without changing the transport or protocol layers.

  ### Related Issues

  Fixes #980
  Fixes #1050
  Related to #2417

---
## 🏷️ Label-Based Checklist

Current labels: _No labels_

### DIFF
--- a/zenoh/src/net/primitives/demux.rs
+++ b/zenoh/src/net/primitives/demux.rs
@@ -69,6 +69,29 @@ impl DeMuxContext<'_> {
         }
         None
     }
+
+    fn full_expr_value(&self, msg: &NetworkMessageMut) -> Option<String> {
+        let wire_expr = msg.wire_expr()?.to_owned();
+        let tables = zread!(self.demux.face.tables.tables);
+
+        if let Some(prefix) = tables
+            .get_mapping(&self.demux.face.state, &wire_expr.scope, wire_expr.mapping)
+            .cloned()
+        {
+            return Some(prefix.expr().to_string() + wire_expr.suffix.as_ref());
+        }
+
+        if matches!(msg.body, NetworkBodyMut::Push(_)) {
+            return self
+                .demux
+                .face
+                .state
+                .get_late_mapping(&wire_expr.scope, wire_expr.mapping)
+                .map(|prefix| prefix.to_string() + wire_expr.suffix.as_ref());
+        }
+
+        None
+    }
 }
 
 impl InterceptorContext for DeMuxContext<'_> {
@@ -78,23 +101,8 @@ impl InterceptorContext for DeMuxContext<'_> {
 
     fn full_expr(&self, msg: &NetworkMessageMut) -> Option<&str> {
         if self.expr.get().is_none() {
-            if let Some(wire_expr) = msg.wire_expr() {
-                if let Some(prefix) = self.prefix(msg) {
-                    self.expr
-                        .set(prefix.expr().to_string() + wire_expr.suffix.as_ref())
-                        .ok();
-                } else if matches!(msg.body, NetworkBodyMut::Push(_)) {
-                    if let Some(prefix) = self
-                        .demux
-                        .face
-                        .state
-                        .get_late_mapping(&wire_expr.scope, wire_expr.mapping)
-                    {
-                        self.expr
-                            .set(prefix.to_string() + wire_expr.suffix.as_ref())
-                            .ok();
-                    }
-                }
+            if let Some(expr) = self.full_expr_value(msg) {
+                self.expr.set(expr).ok();
             }
         }
         self.expr.get().map(|x| x.as_str())
--- a/zenoh/src/net/routing/dispatcher/face.rs
+++ b/zenoh/src/net/routing/dispatcher/face.rs
@@ -204,22 +204,19 @@ impl FaceState {
         id
     }
 
-    pub(crate) fn insert_late_remote_mapping(
-        &mut self,
-        expr_id: ExprId,
-        expr: String,
-    ) -> Option<String> {
+    pub(crate) fn insert_late_remote_mapping(&mut self, expr_id: ExprId, expr: String) {
         let new_bytes = expr.len();
 
         // Skip entries that are unreasonably large (> half the byte budget)
         if new_bytes > MAX_LATE_REMOTE_MAPPINGS_BYTES / 2 {
             tracing::warn!(
-                "Late remote mapping for ExprId {} skipped: \
+                "{} Skip late remote mapping for ExprId {}: \
                  expression length {} exceeds per-entry limit",
+                self,
                 expr_id,
                 new_bytes
             );
-            return None;
+            return;
         }
 
         // Remove existing entry for this expr_id if present
@@ -234,7 +231,15 @@ impl FaceState {
         {
             if let Some(old_id) = self.late_remote_mappings_order.pop_front() {
                 if let Some(evicted) = self.late_remote_mappings.remove(&old_id) {
+                    let evicted_len = evicted.len();
                     self.late_remote_mappings_bytes -= evicted.len();
+                    tracing::warn!(
+                        "{} Evict late remote mapping ExprId {} ({} bytes) while caching ExprId {}",
+                        self,
+                        old_id,
+                        evicted_len,
+                        expr_id
+                    );
                 }
             } else {
                 break;
@@ -244,7 +249,6 @@ impl FaceState {
         self.late_remote_mappings_bytes += new_bytes;
         self.late_remote_mappings.insert(expr_id, expr);
         self.late_remote_mappings_order.push_back(expr_id);
-        None
     }
 
     pub(crate) fn remove_late_remote_mapping(&mut self, expr_id: &ExprId) -> Option<String> {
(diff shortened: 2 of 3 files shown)


zenoh/tests/unknown_scope_repro.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    received
+}
+
+async fn open(config: Config) -> Session {
+    ztimeout!(zenoh::open(config)).unwrap()
+}
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
```
fuzzypixelz, line 91: Have you actually observed a failure of this test?

eclipse0922:
@fuzzypixelz 

Thank you for the review. You're right — this patches a symptom at the wrong layer instead of enforcing the invariant where it belongs. I should have opened a design discussion before jumping into implementation, and I apologize for not doing so.

For context on how we encountered this: we run a fleet of 30+ robots connected through a Zenoh router. At some point robot data stopped flowing entirely. The router log showed:

```
Face{79, ...} Route data with unknown scope 481!
```

The robots were still publishing, but the router could no longer route their data — effectively silent data loss in production. This led us to #980 and #1050.

The reproduction test was synthetically constructed to force the race window; it doesn't represent a naturally valid call sequence under the current API contract. That said, the underlying problem is real and affects production deployments at scale.

Closing this PR. If it would be helpful, I'm happy to open a discussion or contribute to a proper fix at the transport/session layer.

fuzzypixelz:
@eclipse0922

I was unable to reproduce the issue with `publisher_undeclare_does_not_drop_inflight_data`.  On the other hand, the `tables`-level tests assume the issue lies in the routing layer, which I don't believe is true.

We would need a minimal reproducible example of the issue your observe in your deployments or at least more details of the failure scenario (i.e. version, topology, etc). That is necessary in order to precisely identify the root cause of the problem.

Please feel free to open an issue and we may continue discussion there.

PR #2534, opened 2026-03-30
Fix the rustc-hash version 2.1.1 to make it compatible with 1.75

evshary:
## Description

While rustc-hash bumps to 2.1.2, it requires building with 1.77.
Fix it to 2.1.1 to make it compatible with 1.75.

### What does this PR do?

In the rustc-hash commit, they include the `rust-version = "1.77"`
https://github.com/rust-lang/rustc-hash/commit/211455cb398aa0a6d7b12e81941d4a062af54d8e
The PR is to deal with this issue.

The CI fails due to the compatibility issue


evshary:
Superceded by https://github.com/eclipse-zenoh/zenoh/pull/2535

Issue #2506, opened 2026-03-20
Wrong supported link with quic when "transport_quic" and not "transport_quic_datagram"


Jonate:
### Describe the bug

https://github.com/eclipse-zenoh/zenoh/blob/main/io/zenoh-link/src/lib.rs#L111C1-L113C85

```rust
#[cfg(all(feature = "transport_quic", not(feature = "transport_quic_datagram")))]
QUIC_LOCATOR_PREFIX => supported_links.push(LinkKind::QuicDatagram),
instead of 
QUIC_LOCATOR_PREFIX => supported_links.push(LinkKind::Quic),
```

So when 
```json
{
  "transport": {
    "link": {
      "protocols": ["quic"]
    }
  }
}
```
And only "transport_quic" feature and not  "transport_quic_datagram" is enabled.

The only supported link becomes "QuicDatagram" instead of "Quic"

### To reproduce

use config
```json
{
  "transport": {
    "link": {
      "protocols": ["quic"]
    }
  }
}
```
on a build with transport_quic and not transport_quic_datagram

### System info

Tested with zenoh-c 1.8.0
osx64

PR #2536, opened 2026-03-30
Fix wrong link kind whitelisting when `QUIC_DATAGRAM` is compiled-out

oteffahi:
## Description
Fix bug in supported links computation for QUIC when QUIC_DATAGRAM transport is not compiled.

### Related Issues
Closes https://github.com/eclipse-zenoh/zenoh/issues/2506


fuzzypixelz (approved):
Good catch!

PR #2556, opened 2026-04-08
implement transport from fields constructor

milyin:
### What does this PR do?

Create internal (under internal feature) constructor for Transport object which constructs it from the fields values

For some language bindings (zenoh-go) it's more effective to make the "Transport" type language-native. We need a way to build rust Transport object back from fields for applying it to links filtering.

### Related Issues
https://github.com/eclipse-zenoh/zenoh/issues/2554

### DIFF
--- a/zenoh/src/api/info.rs
+++ b/zenoh/src/api/info.rs
@@ -560,26 +560,6 @@ mod tests {
     use zenoh_protocol::core::WhatAmI;
 
     #[test]
-    fn test_new_from_fields_stores_fields() {
-        let zid = ZenohId::default();
-        let whatami = WhatAmI::Peer;
-        let t = Transport::new_from_fields(
-            zid.clone(),
-            whatami,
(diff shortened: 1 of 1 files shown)


zenoh/src/api/info.rs
```diff
@@ -534,3 +553,54 @@ impl CallbackParameter for LinkEvent {
...
+            /*is_shm=*/ true,
+        );
+        assert_eq!(t.zid, zid);
+        assert_eq!(t.whatami, whatami);
+        assert!(t.is_qos);
+        assert!(!t.is_multicast);
+        #[cfg(feature = "shared-memory")]
+        assert!(t.is_shm);
+    }
+
```
milyin:
```suggestion
```

milyin: remove trivial test, next one is enough

DenisBiryukov91 (approved):

PR #2558, opened 2026-04-08
Disable `full_linkstate` in `peer::Hat::Network`

fuzzypixelz:
## Description

This was a typo introduced in #2096 leading to a regression where gateways would use the "full" linkstate protocol (i.e not the gossip protocol) with south-bound peers, incurring high overhead. This was functionally sound and thus uncovered by any existing e2e/intergration test. This patch adds a new `oam` module dedicated to testing linkstate behavior of which the first test is regression test `test_peer_singlehop_gossip_linkstate_propagation`.

### DIFF
--- a/zenoh/src/net/protocol/network.rs
+++ b/zenoh/src/net/protocol/network.rs
@@ -563,7 +563,7 @@ impl Network {
             .collect::<Vec<_>>()
     }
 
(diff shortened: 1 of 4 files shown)


OlivierHecart (commented):
LGTM

PR #2566, opened 2026-04-13
[Reconnect] Trigger reconnect on all error-paths

ciprianf:
## Description
Trigger a reconnect on all error-paths in the runtime orchestrator. The existing implementation is not handling the case when `if let Ok(Some(orch_transport)) = transport.get_callback()` or `if let Some(orch_transport) = orch_transport` return false, which leads to rare events when nodes do not reconnect. Even if the previous check was successful, by the time these checks are performed the transport could have already been closed.

### What does this PR do?
Fixes an existing bug, see https://github.com/ciprianf/zenoh/pull/1 on how to reproduce the issue on main. Root cause identified by Claude.

We have a deployment setup where two routers are bridging a WiFi connection. Router A has a connect endpoint to router B (which has no specific connect). The WiFi connection is flaky and sometimes drops & reconnects. We noticed that in some cases after the WiFi reconnects the subscriber nodes do not get data. 

https://github.com/ciprianf/zenoh/pull/2 tests that the fix solves indeed the issue. By re-running the repro steps, the issue no longer surfaces.

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/zenoh/src/net/runtime/orchestrator.rs
+++ b/zenoh/src/net/runtime/orchestrator.rs
@@ -795,20 +795,27 @@ impl Runtime {
                     self.get_global_connect_timeout(),
                     self.get_connect_retry_config(&peer)
                 );
-                match self.manager().open_transport_unicast(peer.clone()).await {
-                    Ok(transport) => {
+                let result = self
+                    .manager()
+                    .open_transport_unicast(peer.clone())
+                    .await
+                    .and_then(|transport| -> ZResult<_> {
+                        let zid = transport.get_zid()?;
+                        let cb = transport
+                            .get_callback()?
+                            .ok_or_else(|| zerror!("Transport closed immediately"))?;
+                        let session = cb
+                            .as_any()
+                            .downcast_ref::<super::RuntimeSession>()
+                            .ok_or_else(|| zerror!("Unexpected callback type"))?;
+                        zwrite!(session.endpoints).insert(peer.clone());
+                        Ok(zid)
+                    });
+
+                match result {
+                    Ok(zid) => {
                         tracing::debug!("Successfully connected to configured peer {}", peer);
-                        if let Ok(Some(orch_transport)) = transport.get_callback() {
-                            if let Some(orch_transport) = orch_transport
-                                .as_any()
-                                .downcast_ref::<super::RuntimeSession>()
-                            {
-                                zwrite!(orch_transport.endpoints).insert(peer);
-                            }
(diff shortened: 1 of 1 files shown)


ciprianf:
Reviewers please add the appropriate label to the PR, I lack permissions.

ciprianf:
I marked the checklist items related to tests. I provided some links with repro steps and confirmation that the fix works, however given this is a race-condition it is not easily testable.

oteffahi (changes requested):
Thank you for the contribution.

If I'm not mistaken, the same pattern can be found here in `peer_connector` method:

https://github.com/eclipse-zenoh/zenoh/blob/7b54cabaed2eff06d2e8e4bce92ddf20d84cebb0/zenoh/src/net/runtime/orchestrator.rs#L425-L441

I think it should also be updated with the same changes you proposed.

Let me know if I misunderstood something, I rarely work in this part of the codebase 😅

ciprianf:
@oteffahi thank you for the review. 

Good catch. I applied the same fix to `peer_connector`.

Btw, I also filled https://github.com/eclipse-zenoh/zenoh/issues/2569 as a potential separate issue. Maybe somebody familiar with this part of the codebase can take a look?

oteffahi (approved):

ciprianf:
@oteffahi Thanks for approving. I though cannot merge the PR "No applicable reviews submitted by reviewers with write access".
  * "oteffahi Approved these changes with read-only permissions". Could you approve with write permissions too?

oteffahi:
> @oteffahi Thanks for approving. I though cannot merge the PR "No applicable reviews submitted by reviewers with write access".
> 
>  * "oteffahi Approved these changes with read-only permissions". Could you approve with write permissions too?

I do not have write-access since I don't have the "committer" role on Eclipse-Zenoh.
@OlivierHecart please approve and merge this.

oteffahi (approved):

fuzzypixelz (approved):

ciprianf:
Thank you for merging the PR. Is there a release schedule I can follow?

fuzzypixelz:
> Thank you for merging the PR. Is there a release schedule I can follow?

I'm afraid the next release date has not been decided yet. @sashacmc may be able to provide more information.

sashacmc:
Considering that the release has just come out, the next one is not planned until the end of May.

PR #2087, opened 2025-08-10
Add FreeBSD support

mdyring:
This PR makes a few minor changes to add support for FreeBSD:

1. Remove dependency on filepath crate as it does not support FreeBSD. Filepath is only used to resolve a path from a file descriptor in one place, so it seems reasonable to just store original path instead.
2. Make bind_device() for sockets a no-op on FreeBSD as this is not supported.
3. Customize UDP receiver to handle IP_RECVDSTADDR/IPV6_PKTINFO on FreeBSD.
4. Only use MAP_NORESERVE on Linux (unsupported on FreeBSD and MacOS)

Let me know if any other changes are required for this to be merged. It would be great to run Zenoh on FreeBSD. 👍

---
## 🏷️ Label-Based Checklist

Based on the labels applied to this PR, please complete these additional requirements:

**Labels:** `enhancement`

## ✨ Enhancement Requirements

Since this PR enhances existing functionality:

- [x] **Enhancement scope documented** - Clear description of what is being improved
- [x] **Minimum necessary code** - Implementation is as simple as possible, doesn't overcomplicate the system
- [x] **Backwards compatible** - Existing code/APIs still work unchanged
- [x] **No new APIs added** - Only improving existing functionality
- [x] **Tests updated** - Existing tests pass, new test cases added if needed
- [ ] **Performance improvement measured** - If applicable, before/after metrics provided
- [ ] **Documentation updated** - Existing docs updated to reflect improvements
- [x] **User impact documented** - How users benefit from this enhancement

**Remember:** Enhancements should not introduce new APIs or breaking changes.

**Instructions:**
1. Check off items as you complete them (change `- [ ]` to `- [x]`)
2. The PR checklist CI will verify these are completed

*This checklist updates automatically when labels change, but preserves your checked boxes.*

PR #2570, opened 2026-04-15
feat(enhancement): resolve the problem of udp asymmetric routing

Micbetter:
## Description

 This PR adds the sender-side support for `IP_PKTINFO` (and `IPV6_PKTINFO`) in the `zenoh-link-udp` module on Unix platforms. 

### What does this PR do?

 1. Adds `send_with_src_inner` to `pktinfo_unix.rs`, utilizing `libc::sendmsg` to inject `IP_PKTINFO` (or `IPV6_PKTINFO`) into the ancillary data (`cmsg`) of outgoing UDP packets.
 2. Ensures the `cmsg` buffer is efficiently stack-allocated using a compile-time computed `CMSG_BUF_SIZE` to avoid heap allocations in the hot path.
 3. Exposes a new `send_to(buffer, dst_addr, src_addr)` API in `PktInfoUdpSocket` and gracefully falls back to the standard `send_to` on non-Unix platforms.
 4. Updates `LinkUnicastUdpUnconnected` in `unicast.rs` to leverage this new API, ensuring that unconnected UDP sockets (e.g., those listening on `0.0.0.0` or `[::]`) reply using the exact local IP address on which the incoming packet was received.

Prior to this PR, Zenoh's UDP link successfully utilized `IP_PKTINFO` via `recvmsg` to determine the specific local IP address a packet arrived on. However, when replying, Zenoh fell back to the standard `UdpSocket::send_to()`, ignoring the captured local IP. 

This created a severe issue known as **udp asymmetric routing** in multi-homed environments (e.g., a node with both `eth0` and `wlan0` listening on `0.0.0.0`). 
 When a client connected to the server's `wlan0` IP, the server would often reply via `eth0` due to OS routing table defaults, causing the reply packet to bear the `eth0` source IP. Because the client's socket was `connect()`ed to the `wlan0` IP, the client's OS kernel would strictly drop the mismatching reply packet.

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/io/zenoh-links/zenoh-link-udp/src/pktinfo/mod.rs
+++ b/io/zenoh-links/zenoh-link-udp/src/pktinfo/mod.rs
@@ -62,4 +62,73 @@ impl PktInfoUdpSocket {
         }
         Ok((res.0, res.1, src_addr))
     }
+
+    pub(crate) async fn send_to(
+        &self,
+        buffer: &[u8],
+        dst_addr: &SocketAddr,
+        src_addr: &SocketAddr,
+    ) -> io::Result<usize> {
+        #[cfg(target_family = "unix")]
+        {
+            send_with_src(&self.socket, buffer, dst_addr, src_addr).await
+        }
+        #[cfg(not(target_family = "unix"))]
+        {
+            // Fallback for non-unix platforms
+            let _ = src_addr;
+            self.socket.send_to(buffer, dst_addr).await
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::{
+        net::{IpAddr, Ipv4Addr, SocketAddr},
+        sync::Arc,
+    };
+
+    use tokio::net::UdpSocket;
+
+    use super::*;
+
+    #[tokio::test]
+    async fn test_pktinfo_source_ip_consistency() {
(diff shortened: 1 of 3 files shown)


Micbetter:
Hi @oteffahi, Adding tests to reproduce this bug is not easily in a standard CI environment, because we require a multi-IP host. Do you have any suggestions on how to approach the reproduction test for this?

oteffahi:
> Hi @oteffahi, Adding tests to reproduce this bug is not easily in a standard CI environment, because we require a multi-IP host. Do you have any suggestions on how to approach the reproduction test for this?

Correct me if I'm wrong, I believe the requirements are the same as for binding tests, which use `zenoh_util::net::get_ipv4_ipaddrs(None)` to get IP addresses of non-loopback interfaces of the runner.

mdyring:
Just dropping a note there that IP_PKTINFO will need to be handled differently for Linux and FreeBSD similar to PR #2087.

Specifically this commit: https://github.com/eclipse-zenoh/zenoh/pull/2087/changes/678c8649b50307681b843aead570ead60c38fe68

PR #2579, opened 2026-04-20
Check for expected ZID when opening scouted links

### LINKED ISSUE #2151: Locators in HELLO message are assumed to belong to the same ZID
### Describe the bug

The orchestrator assumes that the locators contained in a HELLO message are endpoints that lead to the same Zenoh instance (i.e same ZID). This assumption is invalid, and a manually crafted HELLO message containing locators for different Zenoh instances causes the following: 
- *For peers and routers:* the decision to attempt to connect to a scouted locator depends on the configuration of `scouting/multicast/autoconnect` and `scouting/multicast/autoconnect_strategy` (and respectively their `gossip` config counterparts). These configurations are applied depending on the HELLO message's mode, which does not necessarily match the mode of the instances behind each locator. Crafted HELLO messages can therefore lead to opening transports and links that do not respect the scouting configuration.
- *For clients:* if configured with multilink (`transport/unicast/max_links` >= 2), trusting the contents of HELLO messages and assuming that all the locators within it belong to the same Zenoh instance (i.e same ZID) causes the client to connect to different peers, and therefore have more than one transport open at a time, which invalidates many assumptions made in its routing layer and contradicts the design of Zenoh clients.

### To reproduce

Tested by changing the orchestrator's code, hardcoding two different endpoints in the HELLO message sent in response to a Scout message.

- Cause a client to open multiple transports:
Note that for this to work on a client with multilink, both locators in the HELLO message must have a parameter (ex: `tcp/127.0.0.1:7447?rel=1` and `tcp/127.0.0.1:7448?rel=0`).
1- Start the router with the custom code and with multicast scouting enabled, listening on one of the two endpoints contained in the crafted HELLO message. 
2- Start a non-modified router, with scouting disabled, and listening on the other locator.
3- Start a non-modified client with multicast scouting enabled and `transport/unicast/max_links` set to 2.
The client will establish a transport with both routers.
- Cause a router to connect to a peer despite its `scouting/multicast/autoconnect` config:
1- Start the router with the custom code and with multicast scouting enabled, listening on one of the two endpoints contained in the crafted HELLO message. 
2- Start a non-modified peer, with scouting disabled, and listening on the other locator.
3- Start a non-modified router with multicast scouting enabled, no listener endpoint, and `scouting/multicast/autoconnect: ["router"]`.
The router will connect to the peer despite its autoconnect config stating otherwise.

### Proposed solution

The Zenoh handshake state-machine should take as input the expected ZID when the locator we are connecting to comes from scouting. This way, the handshake can be aborted after reception of INIT_ACK containing a ZID that is different from expected ZID.

### System info

- MacOS 15.5
- Apple Silicon M2
- Zenoh release 1.5.1

oteffahi:
## Description
This PR makes scouted connections validate that the peer ZID matches what was advertised in scouting.
It can also be leveraged to improve multilink for `client` mode by making sure all endpoints belong to the same ZID (to be implemented as part of #2320) 

It fixes possible attacks described in #2151 

### Related Issues
Fixes #2151

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/io/zenoh-transport/tests/unicast_transport.rs
+++ b/io/zenoh-transport/tests/unicast_transport.rs
@@ -2836,3 +2836,171 @@ async fn test_multilink_max_links(
         .await;
     (router_manager, client_manager, client_transport)
 }
+
+#[cfg(feature = "transport_tcp")]
+#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+async fn transport_unicast_with_zid_matching() {
+    zenoh_util::init_log_from_env_or("error");
+
+    let endpoint: EndPoint = format!("tcp/127.0.0.1:{}", 16200).parse().unwrap();
+
+    let client_id = ZenohIdProto::try_from([1]).unwrap();
+    let router_id = ZenohIdProto::try_from([2]).unwrap();
+
+    let router_handler = Arc::new(SHRouter::default());
+    let unicast = make_transport_manager_builder(
+        #[cfg(feature = "transport_multilink")]
+        1,
+        false,
+    );
+    let router_manager = TransportManager::builder()
+        .zid(router_id)
+        .whatami(WhatAmI::Router)
+        .unicast(unicast)
+        .build_test(router_handler.clone())
+        .unwrap();
+
+    let _ = ztimeout!(router_manager.add_listener(endpoint.clone())).unwrap();
+
+    let unicast = make_transport_manager_builder(
+        #[cfg(feature = "transport_multilink")]
+        1,
+        false,
+    );
+    let client_manager = TransportManager::builder()
+        .whatami(WhatAmI::Client)
+        .zid(client_id)
+        .unicast(unicast)
+        .build_test(Arc::new(SHClient))
+        .unwrap();
+
+    // Open transport with expected ZID - should succeed
+    let transport =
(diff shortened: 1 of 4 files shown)


evshary (approved):
LGTM, but maybe you want to retrigger the CI. Something might go wrong in the CI server.

io/zenoh-transport/tests/unicast_transport.rs
```diff
@@ -2836,3 +2836,171 @@ async fn test_multilink_max_links(
         .await;
     (router_manager, client_manager, client_transport)
 }
+
+#[cfg(feature = "transport_tcp")]
+#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+async fn transport_unicast_with_zid_matching() {
+    zenoh_util::init_log_from_env_or("error");
+
+    let endpoint: EndPoint = format!("tcp/127.0.0.1:{}", 16200).parse().unwrap();
```
evshary: I think now we can use `get_free_tcp_port` to avoid using a fixed port
https://github.com/eclipse-zenoh/zenoh/blob/main/commons/zenoh-test/README.md#5-not-recommended-pre-allocating-free-ports

evshary (approved):

fuzzypixelz (approved):

PR #2590, opened 2026-05-05
Implement known_keys_file support for PubKey authentication

### LINKED ISSUE #1339: `transport/auth/pubkey/known_keys_file` is ignored
### Describe the bug

https://github.com/eclipse-zenoh/zenoh/blob/3579f12e8d8d12b30305a9801c8dfda6b4d8ecc2/io/zenoh-transport/src/unicast/establishment/ext/auth/pubkey.rs#L123

### To reproduce

N/A

### System info

- ref: 3579f12e8d8d12b30305a9801c8dfda6b4d8ecc2

milyin:
## Description

Improve `AuthPubKey::from_config`: add `known_keys_file` parameter support previously marked as TBD

### What does this PR do?

- Adds support of `known_keys_file` config parameter
- Adds test for `AuthPubKey::from_config`
- Allows to define one key locally and other from file. Not absolutely necessary but also no reason to forbid it
- Explicitly documented config elements and behavior if both inline and file keys defined (inline key wins, this behavior existed before)

Declared functionality was missing with TBD comment

### Related Issues
Fixes https://github.com/eclipse-zenoh/zenoh/issues/1339,  based on https://github.com/eclipse-zenoh/zenoh/pull/2211

---
## 🏷️ Label-Based Checklist

**Labels:** `enhancement`

## ✨ Enhancement Requirements

**Instructions:**


io/zenoh-transport/tests/unicast_authenticator_config.rs
```diff
@@ -0,0 +1,294 @@
...
+        let mut config = PubKeyConf::default();
+        config
+            .set_public_key_file(Some(pub_path.to_str().unwrap().to_owned()))
+            .unwrap();
+        assert!(AuthPubKey::from_config(&config).is_err());
+
+        // only private file → Err
+        let mut config = PubKeyConf::default();
+        config
+            .set_private_key_file(Some(pri_path.to_str().unwrap().to_owned()))
```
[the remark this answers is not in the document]
milyin: don't care. this is a test

io/zenoh-transport/tests/unicast_authenticator_config.rs
```diff
@@ -0,0 +1,294 @@
...
+        // single key in file → Some, lookup contains that key
+        let single_path = write_tmp_file("zenoh-test-cfg-known-single.pem", &pub_pem);
+        let mut config = PubKeyConf::default();
+        config.set_public_key_pem(Some(pub_pem.clone())).unwrap();
+        config.set_private_key_pem(Some(pri_pem.clone())).unwrap();
+        config
+            .set_known_keys_file(Some(single_path.to_str().unwrap().to_owned()))
+            .unwrap();
+        let auth = AuthPubKey::from_config(&config).unwrap().unwrap();
+        assert!(auth.contains_pubkey(&key1_pub));
```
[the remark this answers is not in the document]
milyin: The actual scope of the issue: unicast_authenticator.rs has the identical problem — it uses test_helpers::make_basic_transport_manager_builder which is gated #[cfg(feature = "test")] at the module level (unicast/mod.rs:19). That's pre-existing and accepted.                                                                                                
                                                                                                                                                                                                                                                                                                                                                                   Why it's not really a problem in practice: test is in default = ["test", "transport_multilink"]. cargo test always enables default features unless you explicitly pass --no-default-features. No one runs integration tests for this crate with --no-default-features --features auth_pubkey.

io/zenoh-transport/src/unicast/establishment/ext/auth/pubkey.rs
```diff
@@ -79,50 +79,119 @@ impl AuthPubKey {
...
+                    .into(),
+            ),
+            (None, None) => None,
+        };
+
+        let mut lookup: Option<HashSet<ZPublicKey>> = None;
+        if let Some(keys_file) = config.known_keys_file() {
+            let content = std::fs::read_to_string(keys_file)
+                .map_err(|e| zerror!("{S} Invalid known keys file: {}.", e))?;
+            let mut known_keys: HashSet<ZPublicKey> = HashSet::new();
```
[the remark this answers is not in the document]
milyin: Unimportant

io/zenoh-transport/src/unicast/establishment/ext/auth/pubkey.rs
```diff
@@ -79,50 +79,119 @@ impl AuthPubKey {
...
+                    let public_key = RsaPublicKey::from_pkcs1_pem(&current_key).map_err(|e| {
+                        zerror!("{S} Invalid RSA public key in known keys file: {}.", e)
+                    })?;
+                    known_keys.insert(public_key.into());
+                    current_key.clear();
+                }
+            }
+            if current_key.chars().any(|c| !c.is_whitespace()) {
+                bail!("{S} Truncated RSA public key in known keys file.");
             }
```
oteffahi, line 165: I don't agree with this file format where PKCS#1 PEMs are chained together. I'd rather have the base64 encoding of each public key in PKCS#1 PEM, so each line is a single encoded public key, which simplifies the parsing.
Furthermore, this logic should not be inlined here. Consider extracting this in a helper function or wrapping the public keys file logic in its own type.

io/zenoh-transport/src/unicast/establishment/ext/auth/pubkey.rs
```diff
@@ -79,50 +79,119 @@ impl AuthPubKey {
...
             (Some(_), None) => {
-                bail!("{S} Missing Rsa Private Key: file.")
+                bail!("{S} Missing Rsa Private Key: set private_key_pem or private_key_file.")
             }
             (None, Some(_)) => {
-                bail!("{S} Missing Rsa Public Key: file.")
+                bail!("{S} Missing Rsa Public Key: set public_key_pem or public_key_file.")
+            }
+            (None, None) if lookup.is_some() => {
+                bail!("{S} known_keys_file requires a public/private key pair to be configured.")
```
oteffahi: Clarify why this is necessary by mentioning that the extension performs mutual authentication.
```suggestion
                bail!("{S} known_keys_file requires a public/private key pair to be configured for mutual authentication.")
```

commons/zenoh-config/src/lib.rs
```diff
@@ -869,11 +869,21 @@ validated_struct::validator! {
...
+                    /// Path to a file containing the RSA public key in PKCS#1 PEM format.
+                    /// Ignored when `public_key_pem` is also set.
                     public_key_file: Option<String>,
+                    /// Path to a file containing the RSA private key in PKCS#1 PEM format.
+                    /// Ignored when `private_key_pem` is also set.
                     private_key_file: Option<String>,
                     key_size: Option<usize>,
+                    /// Path to a file containing one or more trusted peer RSA public keys in PKCS#1 PEM format.
+                    /// Requires private and public keys to also be set.
                     known_keys_file: Option<String>,
```
oteffahi, line 887: Please add this documentation to DEFAULT_CONFIG.json5

oteffahi (changes requested):
IMHO pubkey auth extension is a naive and very verbose attempt at what mTLS already performs in a much safer and scalable way (which is also already integrated in ACL). This extension is internally used for mutlilink, which does not require the known_keys file. The best course of action to me is to remove `auth/pub_key` from the config file.

If we still want to go along with exposing this, I've left a few comments.

commons/zenoh-config/src/lib.rs
```diff
@@ -869,11 +869,21 @@ validated_struct::validator! {
...
+                    /// RSA private key in PKCS#1 PEM format (inline string).
+                    /// If both `private_key_pem` and `private_key_file` are set, `private_key_pem` takes priority.
                     private_key_pem: Option<String>,
+                    /// Path to a file containing the RSA public key in PKCS#1 PEM format.
+                    /// Ignored when `public_key_pem` is also set.
                     public_key_file: Option<String>,
+                    /// Path to a file containing the RSA private key in PKCS#1 PEM format.
+                    /// Ignored when `private_key_pem` is also set.
                     private_key_file: Option<String>,
                     key_size: Option<usize>,
```
oteffahi, line 884: This does not seem to be used anywhere in the code. I believe it was added in the event that we make the size of the generated RSA key for multilink configurable. IMO It does not belong here and therefore should be removed.

oteffahi:
From testing it seems like the PR allows providing a pub/priv key-pair without providing a trusted_keys file. This should be considered an invalid config.

Since the extension requires mutual authentication, this config results in an instance that cannot connect or accept any connections, since it requires any peer to support auth_pubkey, and hold a trusted keypair (which is impossible since the trusted_keys file was not provided). This can be considered a breaking change, since we previously allowed this configuration (which doesn't make sense).

Additionally, this configuration is valid when used by the multilink extension which does not use a trusted_keys file (its requirement is that the same keypair is used for two links on the same transport, it doesn't require that said keypair is known ahead of time). Please take this into consideration when making this change.

jpace121:
Hi! I'm interested in this PR as it would allow for having a bunch of different processes accepting connections know which connecting processes they can talk to without having one password used by everyone or having the accepting processes be able to impersonate each of the connecting processes.

As @oteffahi mentioned, this can also be done using mTLS, BUT because Zenoh loads the default WebPKI trust store, you need to make sure each process is assigned a unique and controlled URL for the CN AND make sure that each CN is handled in the ACL file.

Issue #2592, opened 2026-05-07
Umbrella issue for fuzz tests in Zenoh


evshary:
This is an umbrella issue for fuzz tests in Zenoh
We decided to use cargo-fuzz to test the Zenoh protocol, and we will list all the related PRs and the found vulnerabilities here.

# To add the fuzz framework

* zenoh
  * parse / round-trip
    * [x] transport_message: https://github.com/eclipse-zenoh/zenoh/pull/2591
    * [x] network_message: https://github.com/eclipse-zenoh/zenoh/pull/2591
    * [x] scouting_message: https://github.com/eclipse-zenoh/zenoh/pull/2591
    * [x] endpoint_from_str: https://github.com/eclipse-zenoh/zenoh/pull/2591
  * [x] arbitrary: structured messages: https://github.com/eclipse-zenoh/zenoh/pull/2627
  * (optional) proptest: check the protocol state machine (We need to find out some oracles or invariants to do this)
* zenoh-pico
  * [x] run fuzz-test in zenoh-pico (Run with libFuzzer)
    * [x] scouting_message: https://github.com/eclipse-zenoh/zenoh-pico/pull/1235
    * [x] endpoint_str: https://github.com/eclipse-zenoh/zenoh-pico/pull/1249
    * [x] transport_message: https://github.com/eclipse-zenoh/zenoh-pico/pull/1251
  * [x] increase the test coverage: https://github.com/eclipse-zenoh/zenoh-pico/pull/1259
  * (optional) Test with qemu (Run with AFL++): It isn't worth doing so, while zenoh-pico can run on the host directly.

# CI 

* [x] (zenoh) local CI: https://github.com/eclipse-zenoh/zenoh/pull/2616
* [x] Add the project to OSS-Fuzz: https://github.com/google/oss-fuzz/pull/15544
* [x] (zenoh) Enable the coverage report upload: https://github.com/eclipse-zenoh/zenoh/pull/2665
* [ ] (zenoh-pico) Enable CI and add scheduled tests after all the bugfix are merged. https://github.com/eclipse-zenoh/zenoh-pico/pull/1263

# Found issues

## Zenoh

* The mismatch of encode and decode:
  * https://github.com/eclipse-zenoh/zenoh/pull/2593: Crash when encoding packet with cookie=0 in the Transport Message
    * We can decode the raw packet into the Transport Message with VLE value = 0
    * But we can’t encode the Transport Message with VLE value = 0 back to a raw packet
[...]

PR #2601, opened 2026-05-14
fix(codec): reject truncated bounded strings

evshary:
## Description

This PR hardens the bounded byte/string decoder in zenoh-codec so truncated length-prefixed inputs are rejected before any buffer allocation is attempted.

### What does this PR do?

The shared bounded `Vec<u8>` decode path now checks whether the reader actually has len bytes remaining before allocating, and a regression test covers the malformed string case directly.

Fuzzing exposed that the bounded `Vec<u8>` / String decoder could allocate using an attacker-controlled declared length before proving that enough input bytes were available. Rejecting truncated inputs up front avoids capacity overflows and makes malformed length-prefixed data fail as a normal decode error.

### Related Issues

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**


YuanYuYuan (approved):
LGTM

PR #2603, opened 2026-05-14
fix routers unregistering liveness tokens and queryables

### LINKED ISSUE #2602: Router liveliness tokens aren't properly cleaned up on link disconnects
### Describe the bug

two zenoh routers do not unregister each others liveliness tokens. 

i traced this to `zenoh::net::routing::hat::router::token::Hat::unregister_node_tokens` - it successfully receives `unregister_node_tokens(<expired ZID>)`, but `self.net()` has already had the graph pruned, and so `remove_link` doesn't find any zid's to unpropagate tokens.

i'm going to look into a patch now, i'll update this issue if i find the route cause.

### To reproduce

running the following script on two devices connected by a thunderbolt 5 cable (one macos, one linux if that's relevant)

```rust
use zenoh::{Result, Wait};

#[tokio::main]
async fn main() -> Result<()> {
    let mut cfg = zenoh::Config::default();
    cfg.insert_json5("mode", "\"router\"")?;
    cfg.insert_json5("listen/endpoints", &format!("[\"tcp/[::]:44444\"]"))?;
    cfg.insert_json5(
        "connect/endpoints",
        "[
            \"tcp/fe80::<device A LLA>%iface:44444\",
            \"tcp/fe80::<device B LLA>%iface:44444\"
        ]",
    )?;
    let session = zenoh::open(cfg).wait()?;
    let _tok = session
        .liveliness()
        .declare_token(session.zid().to_string())
        .wait()?;
    session
        .liveliness()
        .declare_subscriber("*")
        .history(true)
        .callback(|sample| println!("{}: {}", sample.kind(), sample.key_expr()))
        .background()
        .wait()?;

    // wait until ctrl c
    tokio::signal::ctrl_c().await?;
    Ok(())
}
```

produces on A
```
PUT: <zid A>
PUT: <zid B>
...ctrl c on B...
DELETE: <zid B>
```
as expected, however when pulling the cable, 
```
PUT: <zid A>
PUT: <zid B>
...cable pulled...
```
no such DELETE is ever propagated.

### System info

- Platform: NixOS 25.5, macOS 26.3
- CPU: AMD AI 9 HX 370, Apple M4
- zenoh commit 77507c12b883e80d1b5d08604935502725a8f00d, latest main at time of writing
- also seen on release v1.9.0

Evanev7:
## Description
Zenoh routers did not unregister liveness tokens or queryables on cold disconnects due to early graph mutation logic

### What does this PR do?
this pr removes the graph mutating logic and defers it until after all unregistering operations have completed

after cold disconnecting two routers, they would not register that the remote router had disconnected, and would have to re-build their queries to be informed of the disconnection. after this change, liveliness tokens are more likely to be correctly updated after a network interface goes down

### Related Issues
fixes #2602

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**


Evanev7:
happy to iterate on this more to bring it inline with the rest of the codebase

fuzzypixelz (commented):
Thanks for your contribution @Evanev7.

I think this is a regression introduced in #2096. I've confirmed that #2602 doesn't occur in Zenoh 1.7.1. I'm yet to check 1.8 but I'm positive that the issue is specific to 1.9.

`remove_link` is an old API: the root cause of the issue here is that the `remove_node_*` methods misuse its (undocumented) contract. So the proper fix would be to call it exactly once in a new `remove_node_entities` method such that:

```rust
fn remove_node_entities(&mut self, zid: &ZenohIdProto) -> UnregisterFaceEntitiesResult;
```

and:

```rust
struct UnregisterFaceEntitiesResult {
  subscribers: HashSet<Arc<Resource>>,
  queryables: HashSet<Arc<Resource>>,
  tokens: HashSet<Arc<Resource>>,
}
```

This also implies replacing `unregister_face_subscribers` with a new `unregister_face_entities` method.

The name `UnregisterFaceEntitiesResult` is a bit of a mouthful but it is consistent with other names such as `UnregisterEntityResult`.

fuzzypixelz (changes requested):

fuzzypixelz:
[And please don't forget to sign the Eclipse ECA.](https://api.eclipse.org/git/eca/status/gh/eclipse-zenoh/zenoh/2603)

fuzzypixelz (commented):
We would also need a reproduction test under `zenoh/src/net/tests/regions/declare.rs`. But the framework has no support for closing connections yet. This can be the subject a separate pull request.

Evanev7:
ECA signed, ill get to this when I have a minute. thanks for the review!

Evanev7:
alright - this was a larger change than i'd hoped, but the unregister_face_entities is now called once. i wasn't sure whether to split out remove_face_entities as a separate method as all the other hat (broker, client, peer) all remove the entities at the same time as they unregister them, so i only added the one method to the hat trait.

zenoh/src/net/protocol/network.rs
```diff
@@ -987,12 +998,19 @@ impl Network {
...
+        &self,
+        dropped_zid: Option<&ZenohIdProto>,
+    ) -> Vec<(NodeIndex, ZenohIdProto)> {
         let mut dfs_stack = vec![self.idx];
         let mut visit_map = self.graph.visit_map();
         while let Some(node) = dfs_stack.pop() {
             if visit_map.visit(node) {
                 for succzid in self.graph[node].links.keys() {
+                    if node == self.idx && dropped_zid.is_some_and(|it| it == succzid) {
+                        // skip the edge from us to dropped_zid.
```
fuzzypixelz:
```suggestion
                        // NOTE: Don't branch out to `dropped_zid` from self. 
                        // Either (1) there is no path between self and `dropped_zid` other than this one,
                        // in which case `dropped_zid` will never visited and thus considered a detached node. 
                        // Or, (2) there is a path between self and `dropped_zid` that goes through another node, 
                        // in which case self (even though present in `dropped_zid`'s links) will not be revisited thanks to `visit_map`.
```

zenoh/src/net/protocol/network.rs
```diff
@@ -949,6 +949,17 @@ impl Network {
         free_index
     }
 
+    pub(crate) fn find_disconnected_nodes_after_removing_link(
```
fuzzypixelz, line 952:
```suggestion
    pub(crate) fn find_disconnected_nodes_when_removing_link(
```

I find that saying `after` can be easily read as "node has already been removed from the graph."

zenoh/src/net/protocol/network.rs
```diff
@@ -987,12 +998,19 @@ impl Network {
         }
     }
 
-    fn remove_detached_nodes(&mut self) -> Vec<(NodeIndex, ZenohIdProto)> {
+    fn disconnected_nodes_after_removing(
```
fuzzypixelz:
```suggestion
    fn disconnected_nodes_when_removing(
```

zenoh/src/net/routing/hat/router/mod.rs
```diff
@@ -583,4 +571,59 @@ impl HatFace {
...
+                .router_tokens
+                .retain(|router| !removed_routers.contains(router));
+
+            if self.res_hat(&res).router_tokens.is_empty() {
+                self.router_tokens.retain(|r| !Arc::ptr_eq(r, &res));
+                removed_tokens.insert(res);
+            }
+        }
+
+        self.net_mut().remove_link(zid);
```
fuzzypixelz: Perhaps this is better moved to `close_face`, that would align us with how peer hats remove links as well.

zenoh/src/net/routing/hat/router/mod.rs
```diff
@@ -583,4 +571,59 @@ impl HatFace {
     }
 }
 
-impl HatTrait for Hat {}
+impl HatTrait for Hat {
+    fn unregister_face_entities(&mut self, ctx: DispatcherContext) -> UnregisterFaceEntitiesResult {
+        let zid = &ctx.src_face.zid;
+        let removed_routers = self
+            // doesnt need to be mut anymore
```
fuzzypixelz:
```suggestion
            // NOTE: a &mut is not taken here and the link graph is not yet modified.
```

fuzzypixelz (commented):
Looks good to me overall, thank you for the refactoring effort.

fuzzypixelz:
@Evanev7 I have written a regression test in 9023d80eb52a. You should be able to cherry-pick it without conflicts.

zenoh/src/net/routing/hat/router/mod.rs
fuzzypixelz: Unfortunately "anymore" won't make sense outside the context of this pull request.

zenoh/src/net/routing/hat/router/mod.rs
Evanev7: makes sense! felt a little out of place there.

Evanev7:
so quick point, my github has three attached emails, one of which is my work email which i have used to sign the ECA. I'm going to reauthor all my commits with my work email in a sec, hopefully that addresses things. tests are running locally now - thanks for writing them, it took a while to get into this codebase.

Evanev7:
cargo test isnt picking up the new tests on my local machine..

fuzzypixelz:
> cargo test isnt picking up the new tests on my local machine..

I suspect that you're missing the `test` feature flag, or maybe the `--lib` cargo-test flag:

```
$ cargo test -p zenoh --lib -F test [test-name] -- --nocapture
```

fuzzypixelz:
@Evanev7 If you like, I can take it from here.

Evanev7:
thanks for bumping this. i'll take a crack at it this weekend, i appreciate the patience massively.

Evanev7:
alright, fixed an issue introduced by my refactor, the three tests added in https://github.com/eclipse-zenoh/zenoh/commit/9023d80eb52ac57b1ad1c8e353772d16c11dfd0c are now green

Evanev7:
and rebased onto latest main

Evanev7:
and hopefully that should fix the eca issues

godardt:
Hopefully we can get a release 1.9.1 soon, this issue is really annoying. Thanks for the fix!

fuzzypixelz:
> Hopefully we can get a release 1.9.1 soon, this issue is really annoying. Thanks for the fix!

1.10 has been released on the 14th of this month!

PR #2605, opened 2026-05-15
fix(codec): avoid extension count truncation on re-encode

evshary:
## Description

This PR fixes re-encoding of Zenoh messages that carry large numbers of unknown extensions.

### What does this PR do?

Several Zenoh message writers tracked the number of remaining extensions in a u8, which truncated ext_unknown.len() for messages with more than 255 unknown extensions. This PR switches those counters to usize so the “more extensions follow” bookkeeping remains correct during re-encode.

```raw
# Encode to result will crash because it's using u8 to count extensions internally
raw packet --decode--> Messages with more than 255 extension --encode--> result
```

Fuzzing found inputs that decode into messages with very large ext_unknown vectors. Those messages were accepted by the decoder but could panic during re-encode because the writer truncated the extension count to u8, then underflowed the decrementing counter while iterating over the full extension list. Using usize keeps the counter aligned with the actual number of extensions.

### Related Issues

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/commons/zenoh-codec/src/zenoh/del.rs
+++ b/commons/zenoh-codec/src/zenoh/del.rs
@@ -46,9 +46,9 @@ where
         if timestamp.is_some() {
             header |= flag::T;
         }
-        let mut n_exts = (ext_sinfo.is_some()) as u8
-            + (ext_attachment.is_some()) as u8
-            + (ext_unknown.len() as u8);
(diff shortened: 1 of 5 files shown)


YuanYuYuan (approved):
LGTM

PR #2606, opened 2026-05-18
fix(codec): reject impossible scouting locator counts

evshary:
## Description

This PR hardens scouting HELLO decoding against malformed locator counts that could otherwise trigger a large allocation during decode.

### What does this PR do?

The locator-list decoder now avoids preallocating from the declared locator count and instead grows the vector only as locators are actually decoded, while preserving clean decode failure on truncated input.

HELLO messages are decoded from untrusted input, so the declared locator count must not directly determine memory allocation size. The decoder should allocate only in proportion to the locators that are actually present and successfully parsed, and it should fail cleanly when the input is truncated or malformed. This keeps the decode path robust against malformed locator lists.

### Related Issues

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/commons/zenoh-codec/src/core/locator.rs
+++ b/commons/zenoh-codec/src/core/locator.rs
@@ -69,11 +69,8 @@ where
     type Error = DidntRead;
 
     fn read(self, reader: &mut R) -> Result<Vec<Locator>, Self::Error> {
-        let len = self.read(&mut *reader)?;
-        if len > reader.remaining() {
-            return Err(DidntRead);
-        }
-        let mut vec: Vec<Locator> = Vec::with_capacity(len);
+        let len: usize = self.read(&mut *reader)?;
+        let mut vec = Vec::new();
         for _ in 0..len {
             vec.push(self.read(&mut *reader)?);
         }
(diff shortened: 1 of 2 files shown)


YuanYuYuan (changes requested):
The guard `len > reader.remaining()` is not sufficient. `Vec::with_capacity(len)` allocates `len × size_of::<Locator>()` bytes, and since `Locator` wraps a `String` (24 bytes on 64-bit), an attacker can pass the check with `len = remaining` and still trigger a 24× memory amplification before any locator is decoded.

The right fix is to replace `Vec::with_capacity(len)` with `Vec::new()`. The loop already terminates safely: when bytes are exhausted, `self.read(&mut *reader)?` returns `Err(DidntRead)` on the first failing attempt, so cost is O(locators actually present), not O(declared `len`). With `Vec::new()` the `len > reader.remaining()` guard becomes redundant and can be dropped too.

evshary:
@YuanYuYuan Thanks for your review. I've updated it accordingly.

PR #2607, opened 2026-05-18
fix(codec): reject noncanonical locator strings

evshary:
## Description

This PR hardens the Locator decoder so it rejects wire strings that parse successfully only because they are first accepted as broader EndPoint syntax and then normalized into a different locator.

### What does this PR do?

The locator codec now verifies that decoding preserves the exact canonical locator string. This prevents decode/re-encode inconsistencies for inputs like locator strings carrying endpoint-only #... config.

```
raw packet --decode--> endpoint with `#` --encode--> result (w/o `#`)
```

Fuzzing found scouting inputs whose locator strings were accepted during decode but normalized into a different locator when re-encoded. That means the codec could accept one wire form and emit another, which breaks decode/encode consistency. Rejecting noncanonical locator strings at decode time closes that gap.

### Related Issues

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/commons/zenoh-codec/src/core/locator.rs
+++ b/commons/zenoh-codec/src/core/locator.rs
@@ -43,7 +43,16 @@ where
     fn read(self, reader: &mut R) -> Result<Locator, Self::Error> {
         let zodec = Zenoh080Bounded::<u8>::new();
         let loc: String = zodec.read(reader)?;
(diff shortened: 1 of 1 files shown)


YuanYuYuan (approved):
LGTM

PR #2612, opened 2026-05-20
fix(protocol): reject malformed endpoint parameter sections

evshary:
## Description

This PR hardens `EndPoint ::from_str` so it rejects malformed metadata and config sections that cannot survive canonicalization.

### What does this PR do?

`EndPoint` parsing now requires `?metadata` and `#config` sections to contain at least one valid parameter entry, and all parsed parameter keys must be non-empty. This prevents malformed endpoint strings from being accepted and then normalized into a different or invalid canonical form.

`EndPoint ::from_str` should only accept parameter sections that remain valid after parsing and canonicalization. Malformed sections like `?;;;` or `#=1` were previously accepted even though canonicalization dropped or changed them, which made parse/stringify/parse inconsistent. Rejecting those inputs at parse time keeps endpoint parsing stable and predictable.

### Related Issues

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/commons/zenoh-protocol/src/core/endpoint.rs
+++ b/commons/zenoh-protocol/src/core/endpoint.rs
@@ -603,6 +603,8 @@ impl TryFrom<String> for EndPoint {
     fn try_from(s: String) -> Result<Self, Self::Error> {
         const ERR: &str =
             "Endpoints must be of the form <protocol>/<address>[?<metadata>][#<config>]";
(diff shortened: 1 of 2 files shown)


YuanYuYuan (approved):
LGTM

Issue #2630, opened 2026-06-02
Aggregating a router's upstream (northbound) declaration forwarding


BOURBONCASK:
### Describe the feature

Hi Zenoh team,

First off — the regions work in 1.9 ([Longwang](https://zenoh.io/blog/2026-04-16-zenoh-longwang/)) is
fantastic. Being able to split a big flat topology into a tree of subregions is exactly what let me push
an edge-to-cloud setup I work on past the old flat-peer limits, and I lean on it heavily.

Running at that larger scale, I bumped into a follow-on declaration-propagation question on the *cloud*
side that I'd love your read on. I've prototyped one approach (described below) but wanted to
sanity-check the direction before proposing it as an actual change.

## The situation I hit

In my setup many downstream routers each forward their local key-expression sub-tree up into a shared
router mesh (cloud), and clients reach resources by their full key-expressions. The shape is pretty general,
though — anything that looks like "many downstream branches feeding into an upstream mesh" runs into the
same thing.

What I see is that every upstream router ends up holding on the order of N×K routing-table `Resource`s:
N downstream branches, K keys each. The footprint matters, but the cost that bites first isn't really
memory (a big cloud box has plenty) — it's the *control-plane work* that scales with N×K: per-declaration
admission, route (re)computation, cold-cache recompute on topology changes, and — in a router mesh —
cross-router propagation, since link-state replicates every declaration to every upstream router. So at
scale the routing table, not the transports, becomes the binding constraint, and it shows up as
convergence time and churn cost rather than just RAM.

Carving the topology into subregions (which I also lean on) helps a related but *different* cost — it
contains churn so one branch's reconnects don't ripple through a shared structure — but it doesn't reduce
the *number* of declarations that propagate upstream and replicate across a router mesh. That count is
what this idea targets, and it's the part isolation alone can't address.

Zenoh already has config-driven aggregation (`aggregation.subscribers` / `publishers`, applied when a
session opens), and it's genuinely useful — but as far as I can tell it only collapses a session's
*own* declarations at that session's boundary. It doesn't change what a router forwards upstream on
behalf of the sessions below it. So a forwarding router's own subscribers collapse to one upstream, but
a downstream session's K subscribers that the router forwards stay K. (I checked this experimentally to
make sure I wasn't missing an existing knob.)

## What I'm thinking

Extend that same config-aggregation idea to a router's northbound forwarding: when a north-bound router
[...]

PR #2631, opened 2026-06-02
feat(routing): upstream (northbound) declaration aggregation

BOURBONCASK:
This follows up on the discussion in #2630.

Up front, so there's no confusion about what this is: it's an experiment I built to deal with a real
routing-table scaling problem at edge-to-cloud scale, and I'm sharing it mostly as a concrete
reference / conversation starter rather than something I expect to land as-is. I understand a broader
Zenoh 2.0 redesign is only just starting to be discussed, with the scope still taking shape — so rather
than assume where this area lands, please read this as "here's one way it could look, and here's what I
learned doing it," offered as possible input to that conversation. I'm very happy to reshape it, cut it
down, rebase it onto whatever makes sense, or just leave it as something others can borrow from.

## What it's for

When many downstream sessions each declare K subscribers/queryables under a shared key-expression prefix
and a router forwards them up into a router mesh, every upstream router ends up holding ~N×K
routing-table `Resource`s (N branches × K keys). The routing table tends to be the first limit you hit.
Zenoh's existing `aggregation.subscribers` / `publishers` only collapses a session's *own* declarations,
not what a router forwards upstream for the sessions below it — so this extends config aggregation to a
router's northbound forwarding, letting an upstream router keep one `Resource` per configured prefix
instead of one per forwarded key.

## What it does

When a north-bound router HAT forwards a downstream subscriber/queryable whose key-expression is included
by a configured `aggregation.upstream.{subscribers,queryables}` prefix, it folds the per-key children
into a single `${prefix}` declaration upstream and suppresses the children there — keeping them
registered in the source region so downward routing is unchanged. There's no new wire type (the aggregate
is an ordinary `DeclareSubscriber` / `DeclareQueryable`), and it's opt-in (an empty `aggregation.upstream`
takes the existing propagation path).

```json5
aggregation: { upstream: { subscribers: ["example/**"], queryables: ["example/**"] } }
```

## A few design notes (in case they're useful for the broader discussion)

The aggregate `Resource`s are created once when the gateway is built and the fold path just looks them
up — creating a `Resource` needs the whole `Tables`, which isn't reachable from inside a single HAT, and
doing it once means each aggregate's match-set is wired exactly once, so reconnect/churn can't pile up
duplicate cross-links.

The aggregate queryable is advertised `complete=false` (presence, not authority), so `BestMatching` falls
through to the real per-key queryable — a genuinely-complete source is never shadowed and no `complete`
state can go stale across owner churn.

For `target=AllComplete`, a route entry that is non-complete, whose matched resource covers the query, and
points at a router, is treated as a transparent forwarder (a small in-process flag on the query route
entry, set only at the router-net fold site — no wire change). `AllComplete` passes through it so the next
router re-applies the filter against its real children; `BestMatching` is untouched.

It's northbound-only, the fold goes through a refcounted ledger, and route caches are invalidated on fold
and teardown. Suspicious prefixes (a bare `**` root, the `@/` admin-space, duplicates, mutual inclusion)
get a startup warning.

## Trade-offs

At the upstream node, per-key ACL / QoS / interceptors and admin-space enumeration see only the
`${prefix}` aggregate (so per-key policy belongs on the forwarding router); the wildcard aggregate can
forward data toward the forwarding router for unsubscribed keys; and liveliness tokens are intentionally
not folded (a liveliness sample's key *is* the token's key, so a folded `${prefix}/**` token couldn't
enumerate per-key presence or signal a per-key loss). These are noted alongside the config in
`DEFAULT_CONFIG.json5` and the schema docs.

## Numbers

From an in-process loopback-TCP benchmark:

- Upstream routing-table cardinality drops from O(N·K) to O(N) — e.g. 5000 → 100 entries at N=100, K=50.
  This is exact by construction (the fold produces one aggregate per branch), independent of RAM.
- Roughly 6–9× less whole-process RSS at the upstream node (K=50) — a process-level measurement and a
  secondary signal; the A/B delta is dominated by the cardinality collapse, which is the load-bearing
  result.

## What changed

It's roughly ~480 lines of routing code plus config and docs, with a similar amount of tests — all behind
the empty-config fast path. The bulk lives in the router HAT (the per-prefix fold/suppress/teardown for
subscribers and queryables). Supporting pieces add the config and its docs, the pre-created aggregates and
the prefix validation in the dispatcher tables, the transparent-forwarder flag on the query route entry
(threaded through the other HATs' query paths), and the tests.

## Tests

- Deterministic, in-process (MockFace, no sleeps): K downstream subscribers collapse to exactly one
  aggregate on the upstream face, and undeclaring all of them withdraws exactly one aggregate.
- Real loopback-TCP integration: subscriber/queryable collapse K→1 plus delivery and teardown; wildcard
  `get` fans out to all children; a missing-key `get` returns empty without hanging; `target=AllComplete`
  reaches a complete child through the aggregate while a non-complete child stays empty (with a negative
  control); two branches under the same prefix don't shadow each other; cross-mesh propagation; plus an
  ignored scale bench.
- The existing unit / `regions` / `queryable` / `matching` / `acl` / `adminspace` / `qos` suites stay
  green, `clippy --deny warnings` is clean, and there's no behaviour change when the config is unset.
  (Also built + run on aarch64 with the same results.)

## Compatibility

No protocol/wire change; additive opt-in config (defaults empty); MSRV (1.75) clean; built on the existing
regions/gateway routing model.

---
## 🏷️ Label-Based Checklist

**Labels:** `new feature`

## 🆕 New Feature Requirements

- [ ] **Feature scope documented** - Clear description of what the feature does and why it's needed
- [ ] **New APIs well-designed** - Public APIs are intuitive, consistent with existing APIs
- [ ] **Comprehensive tests** - All functionality is tested (happy path + edge cases + error cases)
- [ ] **Examples provided** - Usage examples in code comments or separate example files
- [ ] **Documentation added** - New docs explaining the feature, its use cases, and API
- [ ] **Feature flag considered** - Can the feature be enabled/disabled for gradual rollout?
- [ ] **Performance impact assessed** - Memory, CPU, storage implications measured
- [ ] **Integration tested** - Feature works with existing features

**Instructions:**

### DIFF
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -349,6 +349,29 @@
   //   publishers: [
   //     // key_expression
   //   ],
+  //   /// Router upstream (northbound) declaration aggregation. Opt-in; effective only on a
+  //   /// router's north-bound forwarding (distinct from the session-local subscribers/publishers
+  //   /// above). A downstream subscriber/queryable whose key-expression is included by one of
+  //   /// these prefixes is folded into a single `${prefix}` declaration toward the upstream and
+  //   /// suppressed upstream (kept locally for downward routing), so an upstream router holds one
+  //   /// resource per configured prefix instead of one per forwarded child key-expression. The
+  //   /// folded queryable is advertised complete=false (presence, not authority); `target=AllComplete`
+  //   /// queries are still forwarded through the aggregate to the real child queryables.
+  //   /// Trade-offs when enabled — apply per-key policy/introspection at the router that performs the
+  //   /// fold (where child keys are still visible): upstream per-key declaration ACL/QoS/interceptors
+  //   /// and admin-space enumeration see only the `${prefix}` aggregate; the wildcard aggregate may
+  //   /// forward data toward the folding router for unsubscribed keys; liveliness tokens are NOT
+  //   /// folded. Each prefix should cover the key-expression sub-tree a single downstream branch owns;
+  //   /// prefixes should be disjoint, not a bare `**` root, and not under the `@/` admin key-space
+  //   /// (misconfigured prefixes are warned about at startup).
+  //   upstream: {
+  //     subscribers: [
+  //       // key_expression
+  //     ],
+  //     queryables: [
+  //       // key_expression
+  //     ],
+  //   },
   // },
 
   // /// Namespace prefix.
--- a/commons/zenoh-config/src/lib.rs
+++ b/commons/zenoh-config/src/lib.rs
@@ -655,6 +655,44 @@ validated_struct::validator! {
             subscribers: Vec<OwnedKeyExpr>,
             /// A list of key-expressions for which all included publishers will be aggregated into.
             publishers: Vec<OwnedKeyExpr>,
+            /// Router upstream (northbound) declaration aggregation. **Opt-in** and effective only on
+            /// a router's north-bound forwarding — distinct from the session-local `subscribers`/
+            /// `publishers` above. When a north-bound router HAT forwards a downstream subscriber or
+            /// queryable whose key-expression is *included* by one of these prefixes, the per-key
+            /// children are folded into a single `${prefix}` declaration toward the upstream and the
+            /// children are suppressed upstream (they remain registered locally for downward
+            /// routing), so an upstream router holds one routing resource per configured prefix
+            /// instead of one per forwarded child key-expression. The folded queryable is advertised
+            /// `complete=false` (presence, not authority) so it never shadows a distinct complete
+            /// source; `target=AllComplete` queries are still forwarded through the aggregate to the
+            /// real child queryables. Empty by default → behaviour is identical to stock.
+            ///
+            /// **Trade-offs when enabled — apply per-key policy and introspection at the router that
+            /// performs the fold (downstream of the aggregation point), where the child key-expressions
+            /// are still visible:**
+            /// - Per-key declaration ACL / QoS-overwrite / interceptors at the *upstream* node act on
+            ///   the `${prefix}` aggregate, not the individual child keys (granularity degrades to the
+            ///   prefix).
+            /// - Upstream admin-space enumeration (`@/${zid}/router/subscriber|queryable/**`) shows the
+            ///   aggregate, not the per-key children.
+            /// - The aggregate is a wildcard, so data published to *unsubscribed* keys under the prefix
+            ///   is still forwarded toward the folding router (which drops it); keep prefixes as tight
+            ///   as the actually-subscribed sub-tree.
+            /// - Only subscribers and queryables are folded. Liveliness tokens are intentionally NOT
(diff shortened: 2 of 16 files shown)


fuzzypixelz (commented):
A few comments:
- Conceptually, any mode can sit in upstream regions. That this design focuses on router aggregation introduces debatable asymmetries in declaration propagation. The design philosophy of Zenoh doesn't require routers to necessarily be the "backbone" region.
- Propagating aggregated queryables as non-complete and consequently making them indistinguishable on the wire from non-aggregated non-complete queryables feels like a hack and incurs the cost of unnecessary query propagation. Now, it is possible to remedy this using a protocol extension and I would argue that this is the proper way to implement this.
- Given the first two points, this feature introduces significant complexity to the routing subsystem—I'm not convinced that the utility gained here justifies that cost. Especially since entity aggregation can simply be enabled in _all application nodes._ and so I'd like to understand why that route (no pun intended) was not feasible for your use case (e.g. configuration nightmare?)

Overall, the present pull request and linked issue are well-thought out and touch on a real problem. Thank you for working through the problem and exploring the solution space.

BOURBONCASK:
Thanks so much for taking the time to go through this so carefully ! @fuzzypixelz Let me share some deployment context here.

Why not just enable entity aggregation on all application nodes:

My setup is a large robot fleet. Each robot runs many processes isolated by a per-robot namespace (with near-identical internal key-exprs across robots), and between each robot and the cloud sits a single edge-side zenohd that also enforces ACL. The cloud router is what hits the O(N·K) wall.

Enabling aggregation.subscribers/publishers on every application node does collapse each node's own declarations, but in this topology I may ran into two issues (haven't tried yet, I will give a try next week):
1. Over-delivery lands inside the robot. Aggregating a node's own declaration also coarsens the downward routing toward it — a process that only wants robotN/cam/3 ends up advertising robotN/**, so the bridge pushes the whole namespace down over the most constrained (intra-robot) hop. Northbound-only aggregation keeps the downward path per-key and only folds what goes up to the cloud.
2. It may defeats per-key ACL on the bridge. If application nodes pre-aggregate, the bridge's ingress only sees robotN/**, so per-key rules can't be enforced there anymore. Keeping aggregation northbound means the bridge still sees per-key declarations, and only the cloud-facing side is folded.

There's also aggregation not covering queryables today (at least not in configs), which matters for my case. Config-at-scale is a factor too, but the two above are what really blocked me.

On routers being treated specially:
This is a fair point and I'd rather not bake "router == backbone" into it. The property I actually rely on is "a node forwarding declarations on behalf of a downstream region folds them northbound," which isn't router-specific. Would it make more sense to tie this to a region's north/south boundary rather than special-casing router mode? I'd really value your guidance on the right abstraction here.

On complete=false being a hack:
Agreed — that's a fair characterization, and your protocol-extension suggestion sounds like the proper way to do it. I'd be happy to explore that direction.

Lastly, the regions model is a beautiful foundation to build on — it's what made this feel tractable in the first place. I'd be glad to keep exploring directions toward making Zenoh exceptionally scalable.

PR #2633, opened 2026-06-04
fix(util): avoid panic when parsing a bare "s" duration

BOURBONCASK:
### What does this PR do?

`parse_duration` (in `zenoh-util`'s time DSL) panics on the input `"s"`.

It reads the duration's trailing byte to detect the unit, and for `s` it reads a **second** byte to disambiguate `s` (seconds) from `ms` (milliseconds):

```rust
let mut it = s.bytes().enumerate().rev();
match it.next().unwrap() {                    // guarded by the leading is_empty() check
    ...
    (_, b's') => match it.next().unwrap() {   // <-- second next(): NOT guarded
        (i, b'm') => ... // "ms"
        (i, _)    => ... // "s"
    },
    ...
}
```

The leading `is_empty()` guard only covers the *first* `next()`. The input `"s"` (a unit with no numeric part) exhausts the iterator, so the second `unwrap()` panics with `called Option::unwrap() on a None value`. Every other unit-only input (`"u"`, `"m"`, `"h"`, `"d"`, `"w"`) only calls `next()` once and already returns an `Err`; `"s"` is the only unit that double-consumes.

This change handles the exhausted-iterator case like the other unit-only inputs and returns a parse error instead of panicking. It also extends `test_parse_duration` with all unit-only inputs and end-to-end `TimeExpr`/`TimeRange` cases.

Minimal repro on `main`:

```rust
use zenoh_util::time_range::{TimeExpr, TimeRange};

let _ = "now(s)".parse::<TimeExpr>();      // panics
let _ = "[now(s)..]".parse::<TimeRange>(); // panics
```

`parse_duration` is reachable from the public time DSL (`TimeExpr` / `TimeRange`), which backs the standardized `_time` selector parameter (`ZenohParameters::time_range()`). That value comes from the querier's selector, so a malformed `_time` (e.g. `_time=[now(s)..]`) is **untrusted input that can panic the parsing side** — both when sending a `get()` (`ConsolidationMode::Auto` reads `time_range()`) and on a queryable/storage that inspects `_time`.

The intent is clearly to fail gracefully: `time_range()` returns `Option<ZResult<TimeRange>>`, i.e. malformed input is supposed to yield an `Err`, not a panic. This follows the same "reject malformed input instead of panicking" direction as recent parsing fixes.

### Related Issues

N/A — found via review of the time-DSL parsing path.

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/commons/zenoh-util/src/time_range.rs
+++ b/commons/zenoh-util/src/time_range.rs
@@ -427,9 +427,13 @@ fn parse_duration(s: &str) -> Result<f64, ZError> {
     let mut it = s.bytes().enumerate().rev();
     match it.next().unwrap() {
         (i, b'u') => s[..i].parse::<f64>().map(|u| U_TO_SECS * u),
-        (_, b's') => match it.next().unwrap() {
-            (i, b'm') => s[..i].parse::<f64>().map(|ms| MS_TO_SECS * ms),
-            (i, _) => s[..i + 1].parse::<f64>(),
+        (_, b's') => match it.next() {
+            Some((i, b'm')) => s[..i].parse::<f64>().map(|ms| MS_TO_SECS * ms),
+            Some((i, _)) => s[..i + 1].parse::<f64>(),
+            // A bare "s" with no preceding value: behave like the other unit-only
+            // inputs ("u", "m", ...) and return an error instead of panicking on the
+            // exhausted iterator.
+            None => "".parse::<f64>(),
         },
         (i, b'm') => s[..i].parse::<f64>().map(|m| M_TO_SECS * m),
         (i, b'h') => s[..i].parse::<f64>().map(|h| H_TO_SECS * h),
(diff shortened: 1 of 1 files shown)


commons/zenoh-util/src/time_range.rs
```diff
@@ -427,9 +427,13 @@ fn parse_duration(s: &str) -> Result<f64, ZError> {
...
         (i, b'u') => s[..i].parse::<f64>().map(|u| U_TO_SECS * u),
-        (_, b's') => match it.next().unwrap() {
-            (i, b'm') => s[..i].parse::<f64>().map(|ms| MS_TO_SECS * ms),
-            (i, _) => s[..i + 1].parse::<f64>(),
+        (_, b's') => match it.next() {
+            Some((i, b'm')) => s[..i].parse::<f64>().map(|ms| MS_TO_SECS * ms),
+            Some((i, _)) => s[..i + 1].parse::<f64>(),
+            // A bare "s" with no preceding value: behave like the other unit-only
+            // inputs ("u", "m", ...) and return an error instead of panicking on the
+            // exhausted iterator.
```
oteffahi: The comment referencing panicking will be confusing outside of this PR's context. It should only mention that we want to return an error here.

```suggestion
            // A bare "s" with no preceding value: behave like the other unit-only
            // inputs ("u", "m", ...) and return an error.
```

oteffahi (changes requested):
Small nitpick. Otherwise LGTM.

Issue #2634, opened 2026-06-05
Deadlock: dropping a `Querier` with a pending query whose completion callback re-enters the `Session`


ciprianf:
## Summary
Dropping a `Querier` (or calling `Querier::undeclare()`) while one of its queries is still in flight deadlocks if the query's handler has a **completion/drop callback** that calls back into the same `Session`.

The querier destructor removes the pending query **while holding the session `state` write lock**. This drops the query's callback inline, which runs the completion callback, and that callback tries to re-acquire the same lock on the same thread → self-deadlock (`std::sync::RwLock` is non-reentrant).

## Environment
- Zenoh: `main` (crate version `1.9.0`)
- OS: Linux

## Where the problem is
`Session::undeclare_querier_inner` in `zenoh/src/api/session.rs`:

```rust
pub(crate) fn undeclare_querier_inner(&self, querier_id: Id) -> ZResult<()> {
    let mut state = zwrite!(self.0.state);              // (1) WRITE lock held
    ...
    if let Some(querier_state) = state.queriers.remove(&querier_id) {
        // (2) drops each pending query's Callback while the lock is held,
        //     which runs the completion (drop) handler inline
        state.queries.retain(|_, q| q.querier_id != Some(querier_id));
        ...
    }
}
```

The completion callback then calls e.g. `Session::put` → `resolve_put`, which does
`let state = zread!(self.0.state);` — re-acquiring the lock already held at (1).

Relevant locations:
- Lock acquired: `zenoh/src/api/session.rs` — `undeclare_querier_inner`, `zwrite!(self.0.state)`
- Callback dropped under lock: same fn, `state.queries.retain(...)`
- Re-entrant acquire: `zenoh/src/api/session.rs` — `resolve_put`, `zread!(self.0.state)`
- Completion-callback machinery: `zenoh/src/api/handlers/callback.rs` — `CallbackDrop`'s `Drop` impl

This is specific to the **drop/completion** path. The normal **reply** path is safe
because it explicitly `drop(state)` before invoking `callback.call(...)`.

## Why a pending query is required
The deadlock only triggers when the query's callback is dropped *while the lock is held*.
That happens exclusively when the destructor removes a **still-pending** query at (2).
If the query already completed, its callback was dropped earlier (outside the lock), so
there is no re-entrancy. On a standalone session a query with no matching queryables
finalizes immediately, so the reproduction uses a queryable that keeps the query pending.

## Minimal reproduction
[...]

PR #2635, opened 2026-06-05
Fix for the deadlock when dropping a Querier with a pending query whose completion callback re-enters the Session

ciprianf:
See the issue I raised in https://github.com/eclipse-zenoh/zenoh/issues/2634.

## Issue
Dropping a Querier (or calling Querier::undeclare()) while one of its queries is still in flight deadlocks if the query's handler has a completion/drop callback that calls back into the same Session.

The querier destructor removes the pending query while holding the session state write lock. This drops the query's callback inline, which runs the completion callback, and that callback tries to re-acquire the same lock on the same thread → self-deadlock (std::sync::RwLock is non-reentrant).

## Fix

Move the destruction of the queries outside of the lock in `undeclare_querier_inner`.

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/zenoh/src/api/session.rs
+++ b/zenoh/src/api/session.rs
@@ -1681,39 +1681,55 @@ impl Session {
     }
 
     pub(crate) fn undeclare_querier_inner(&self, querier_id: Id) -> ZResult<()> {
-        let mut state = zwrite!(self.0.state);
-        let Ok(primitives) = state.primitives() else {
-            return Ok(());
-        };
-        if let Some(querier_state) = state.queriers.remove(&querier_id) {
+        // Everything that needs the state lock happens in this scope. The
+        // removed `QueryState`s (which own the user callbacks, including
+        // completion/drop callbacks) are returned out of the scope so they are
+        // dropped *after* the lock is released: a completion callback that
+        // re-enters the `Session` would otherwise deadlock re-acquiring the lock.
+        let (primitives, remote_id, removed_queries, send_final) = {
+            let mut state = zwrite!(self.0.state);
+            let Ok(primitives) = state.primitives() else {
+                return Ok(());
+            };
+            let Some(querier_state) = state.queriers.remove(&querier_id) else {
+                return Err(zerror!("Unable to find querier").into());
+            };
             trace!("undeclare_querier({:?})", querier_state);
-            // remove all pending queries from this querier
-            state
+            // Remove all pending queries from this querier.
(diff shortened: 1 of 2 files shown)


zenoh/src/api/session.rs
```diff
@@ -1681,39 +1681,60 @@ impl Session {
...
+                primitives,
+                querier_state.remote_id,
+                removed_queries,
+                send_final,
+            )
+        };
+        // Lock released. Drop the removed queries (and their callbacks) and send
+        // the final interest without holding the state lock.
+        drop(removed_queries);
+        if send_final {
```
DenisBiryukov91: Is there any reason, why send final interest could not stay where it originally was? I believe the issue was only in dropping queries while holding the lock ...

ciprianf: send final interest was done as the last step outside of the lock (note the `drop(state)` just before)

Here we have the same semantics, however the lock is scoped inside the if-statement. I find this to be more readable and safer, but let me know if you prefer otherwise.

DenisBiryukov91: Yes, I would rather prefer to keep sending final interest in the place where it originally was. Since this simply notifies remote peers that we no longer want to receive information about matching queryables - so completely independent from dropping and finalizing queries. This will also somewhat simplify the code removing the need to return multiple variables from the locked scope.

ciprianf: Done. PTAL.

zenoh/src/api/session.rs
DenisBiryukov91, line 1725: It looks like you will also need to explicitly drop state here in the else branch, otherwise for session local querier, _removed_queries will be dropped with state lock being held.

ciprianf: I updated the comment above the `_removed_queries` to make it explicit that the lock is released before the _removed_queries destruction - order of declaration matters.

DenisBiryukov91: Thanks for updating the comment. Maybe it is still better though to write `let _removed_queries = ... ` here, and then explicitly drop state in the else branch to avoid any further confusion and the need for an extra comment ?

ciprianf: I added an explicit drop for the state and for the removed queries to be clear on the order of destruction.

DenisBiryukov91 (approved):

Issue #2409, opened 2026-02-06
Scouting task may deadlock


fuzzypixelz:
### Describe the bug

I have observed a strange deadlock where `TransportManagerStateUnicast::transports: tokio::sync::Mutex` is unlocked, yet the `Net` runtime is blocked waiting for it.

This happens in the `futures::future::select_all` future in `Runtime::scout` where two connections are simultaneously in progress. The first suspended call is:
https://github.com/eclipse-zenoh/zenoh/blob/0682bc526fdd0b19dc9eb1d5ecdbbd344ace8533/zenoh/src/net/runtime/orchestrator.rs#L984-L986
This yields because the mutex is locked somewhere else.

A second future would had been suspended in:
https://github.com/eclipse-zenoh/zenoh/blob/0682bc526fdd0b19dc9eb1d5ecdbbd344ace8533/zenoh/src/api/admin.rs#L243-L246
Only this time, a `Net.block_in_place` is used to acquire the same mutex. I observe that even though the mutex was dropped, Tokio denies this second attempt to acquire the mutex and parks the `block_on` thread because there is a waiter with an assigned permit on the `SelectAll` future.

Using the same code, I am unable to reproduce this deadlock on Linux. I am not yet sure why this is.

In https://github.com/fuzzypixelz/tokio-inaccessible-permit I provide a minimal reproducible example without Zenoh that illustrates the issue observed.

Ultimately, the root issue here is that we-enter the async context through `block_in_place` because `TransportPeerEventHandler` is sync. The `block_in_place` documentation states:

> Be aware that although this function avoids starving other independently spawned tasks, any other code running concurrently in the same task will be suspended during the call to block_in_place. This can happen e.g. when using the [join!](https://docs.rs/tokio/latest/tokio/macro.join.html) macro. To avoid this issue, use [spawn_blocking](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html) instead of block_in_place.

In this case, the failure is much more subtle because of the interaction with `tokio::sync::Mutex`. But the `block_in_place`-`block_on` pattern used extensively in Zenoh to "quickly await this future in a sync callback" **should** be phased out.

### To reproduce

1. Get a macOS machine
2. Clone https://github.com/eclipse-zenoh/zenoh/tree/regions-deadlock
3. Run `reproduce.nu`

### System info

- rev: 17839ed
- os: macOS

fuzzypixelz:
[...]

PR #2637, opened 2026-06-09
fix: don't park the ctrl_lock holder on StartConditions notifications under peer churn

### LINKED ISSUE #2581: Persistent routing lock stall/deadlock under peer churn

godardt:
Closes [#2581](https://github.com/eclipse-zenoh/zenoh/issues/2581)

## What happens

In peer mode with gossip autoconnect, the routing layer can deadlock permanently
when sessions open and close while other peers are connecting. A long-lived
publisher then stalls forever in `put().wait()` (the symptom reported in #2581),
and every other operation that touches routing (declares, accepts, OAM handling,
session close) wedges behind it. In debug builds the same race can instead
poison the routing `ctrl_lock` and turn every rx callback into a `PoisonError`
panic cascade from `demux.rs`.

## Root cause

Two call sites notify `StartConditions` that a peer connector has terminated by
running `terminate_peer_connector_zid` through `ZRuntime::Net.block_in_place(..)`:

- `hat::peer::interests::route_declare_final` (initial-interest finalization),
  reached under the routing `ctrl_lock` via `Face::send_declare`
- the tail of `gossip::link_states`, reached under the routing `ctrl_lock` via
  OAM handling in `DeMux::handle_message`

`block_in_place` parks the calling thread until the future completes, so in both
cases a thread **holding `ctrl_lock` goes to sleep** waiting on the
`StartConditions::peer_connectors` tokio mutex. That closes a cycle, captured
live with `gdb thread apply all bt` on a hung process:

1. An rx thread handles a `DeclareFinal`, takes `ctrl_lock` in
   `Face::send_declare`, and parks in
   `block_in_place(terminate_peer_connector_zid(..))`, waiting on the
   `StartConditions` mutex.
2. The Net runtime's only worker (`worker_threads = 1` by default) runs a gossip
   autoconnect task (`link_states` -> `connect_peer` ->
   `Gateway::new_transport_unicast`) that blocks synchronously on the same
   `ctrl_lock`.
3. With the Net worker wedged, tasks queued on the `StartConditions` mutex are
   never polled again, so the mutex never frees, the rx thread in (1) never
   wakes, and `ctrl_lock` is never released.

## Fix

Spawn the notification on the Net runtime instead of blocking on it.
`terminate_peer_connector_zid` returns nothing and its only effect is to mark
the connector terminated and possibly `notify_one()` a waiting `open()`, so
deferring it by one task-schedule is semantically equivalent. The call already
raced with `add_peer_connector_zid` across threads (terminate-before-add inserts
an already-terminated entry), so the reordering introduces no new interleavings.
With no thread parked while holding `ctrl_lock`, the cycle cannot form.

## Validation

Reproduced with a stress setup that repeatedly opens ~20 short-lived peer
sessions (declaring queryables and issuing queries) against a router while a
publisher streams at 5 kHz, all sessions gossip-linked. Unpatched, it deadlocked
within ~23 iterations on a 24-core Linux machine; patched, 150 iterations across
2/4/8/24-CPU affinities plus repeated full-suite runs completed with zero hangs
and zero panics. The reproducer from #2581 (`z_p2p_declare_final_stress`)
exercises the same churn pattern.

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/zenoh/src/net/protocol/gossip.rs
+++ b/zenoh/src/net/protocol/gossip.rs
@@ -409,11 +409,13 @@ impl Gossip {
             }
         }
         if (!self.wait_declares) || src_whatami != WhatAmI::Peer {
-            zenoh_runtime::ZRuntime::Net.block_in_place(
-                strong_runtime
-                    .start_conditions()
-                    .terminate_peer_connector_zid(src),
-            );
+            // Spawned, not `block_in_place`-ed: `link_states` runs under the
+            // routing `ctrl_lock`, and parking the lock holder deadlocks under
+            // peer churn.
+            let start_conditions = strong_runtime.start_conditions().clone();
+            zenoh_runtime::ZRuntime::Net.spawn(async move {
+                start_conditions.terminate_peer_connector_zid(src).await;
+            });
         }
     }
 
--- a/zenoh/src/net/routing/hat/peer/interests.rs
+++ b/zenoh/src/net/routing/hat/peer/interests.rs
@@ -280,17 +280,22 @@ impl HatInterestTrait for Hat {
                 return Noop;
             }
 
-            zenoh_runtime::ZRuntime::Net.block_in_place(async move {
-                if let Some(runtime) = &ctx.tables.runtime {
-                    if let Some(runtime) = runtime.upgrade() {
-                        tracing::debug!("Terminating peer connector");
-                        runtime
-                            .start_conditions()
-                            .terminate_peer_connector_zid(ctx.src_face.zid)
-                            .await
-                    }
-                }
-            });
+            // Spawned, not `block_in_place`-ed: this runs under the routing
+            // `ctrl_lock`, and parking the lock holder deadlocks under peer
+            // churn.
+            let start_conditions = ctx
+                .tables
+                .runtime
+                .as_ref()
+                .and_then(|runtime| runtime.upgrade())
+                .map(|runtime| runtime.start_conditions().clone());
+            let zid = ctx.src_face.zid;
+            if let Some(start_conditions) = start_conditions {
+                zenoh_runtime::ZRuntime::Net.spawn(async move {
+                    tracing::debug!("Terminating peer connector");
+                    start_conditions.terminate_peer_connector_zid(zid).await;
+                });
+            }
 
             Noop
         } else {


Mensch88:
<h2>Independent Reproduction (C API) &amp; Safety Argument for Ordering Change</h2>
<p>We identified this issue from the other end: our test suite experienced hangs lasting up to 300 seconds, initially reported against <code>zenoh-c</code> as <a href="https://github.com/eclipse-zenoh/zenoh-c/issues/1308">zenoh-c#1308</a> before we localized the root cause. Tracing the source code revealed the same <code>block_in_place</code> under <code>ctrl_lock</code> that this PR addresses, which subsequently led us to #2581 and this PR.</p>
<p>Below are our measurements and a safety analysis regarding the ordering change, which could be the primary concern for merging.</p>
<h3>Different Victim Path, Same Defect</h3>
<p>The reproducer in #2581 stalls a <strong>publisher</strong> within <code>route_data</code> while waiting for the table read lock. Our case blocks an application <strong>declare</strong> operation:</p>
<pre><code class="language-rust">#2  std::sys::sync::mutex::futex::Mutex::lock_contended ()
#3  &lt;...Face as ...Primitives&gt;::send_declare ()
#4  zenoh::api::key_expr::KeyExprWireDeclaration::new ()
#5  zenoh::api::session::Session::declare_nonwild_prefix ()
#6  z_declare_queryable ()
</code></pre>
<p>Here, <code>rx-1</code> parks while holding the lock, consistent with the mechanism described in this PR:</p>
<pre><code class="language-rust">#1  tokio::runtime::park::Inner::park ()
#2  zenoh_runtime::ZRuntime::block_in_place::{{closure}} ()
#3  ...hat::peer::interests...::route_declare_final ()
#4  &lt;...Face as ...Primitives&gt;::send_declare ()
</code></pre>
<p>We observed identical behavior via <code>z_declare_subscriber</code>. Consequently, this fix addresses three distinct victim paths across two separate reproduction cases.</p>
<h3>Measurement Data</h3>
<p>We ran <code>ctest -j 4 --repeat until-fail:10</code> over 95 test cases. Each case spawns a short-lived <code>mode: "peer"</code> process with a per-test timeout of 300 seconds.</p>
<ul><li><strong>Hardware:</strong> 12-core x86-64</li><li><strong>Environment:</strong> <code>zenoh-c</code> 1.9.0 built from source against <code>zenoh</code> <code>release/1.9.0</code></li></ul>

Library Version | Sweeps | Hangs | Total Wall Time
-- | -- | -- | --
Unpatched | 3 | Run 1: 2/95, Run 2: 1/95, Run 3: 3/95 | 348 / 381 / 377 s
PR b898eae Cherry-pick | 3 | 0, 0, 0 (95/95) | 117 / 119 / 119 s
Reverted to Unpatched* | 1 | 2/95 | 385 s


<p><em>*The unpatched library was swapped back in on the same machine immediately after the patched runs. The immediate return of hangs confirms the variable is the library itself, not a transient improvement in system state. The variance in total wall time is driven by whether the 300s timeout triggers.</em></p>
<h3>The Switch to <code>spawn</code> looks safe</h3>
<p>The asynchronous termination introduced by this PR requires a safety argument beyond measurement data. We analyzed the consumers of the notification signal. In <code>zenoh/src/net/runtime/orchestrator.rs</code>, there is exactly one waiter in <code>start_peer()</code>:</p>
<pre><code class="language-rust">if wait_scouting
    &amp;&amp; (scouting || !peers.is_empty())
    &amp;&amp; tokio::time::timeout(delay, self.state.start_conditions.notified())
        .await
        .is_err()
    &amp;&amp; !peers.is_empty()
{
    tracing::warn!("Scouting delay elapsed before start conditions are met.");
}
Ok(())
</code></pre>
<p><strong>Analysis:</strong></p>
<ol><li><strong>Bounded Wait:</strong> The wait is bounded by the scouting <code>delay</code>.</li><li><strong>Graceful Degradation:</strong> The timeout branch only logs a warning and returns <code>Ok(())</code>. A delayed notification degrades to "waited the full scouting delay," not a stall or error.</li><li><strong>No Lost Wakeups:</strong> <code>Notify::notify_one</code> stores a permit if no task is currently waiting. Therefore, a spawned task completing <em>before</em> <code>notified()</code> is called does not result in a lost wakeup.</li></ol><p><strong>Trade-off:</strong><br>The critical failure mode (deadlock) and the new worst-case scenario (warning + bounded wait) both occur under the same condition: a saturated <code>Net</code> runtime. Before this PR, saturation leads to permanent deadlock; after this PR, it results in a bounded wait and a log entry.</p>
<h3>Backport Note</h3>
<p>Commit <code>b898eae</code> cherry-picks cleanly onto <code>release/1.9.0</code> with no conflicts. The resulting <code>zenoh-c</code> 1.9.0 builds and links without changes. There are no technical barriers to backporting this fix to the 1.9.x branch if desired.</p>
<h3>Limitation of Evidence</h3>
<p>Prior to discovering this PR, we independently attempted to fix the issue by modifying only the site in <code>hat/peer/interests.rs</code>. This alone was sufficient to eliminate our deadlocks; we had initially missed the call in <code>gossip.rs</code>. Therefore, our reproducer does not exercise the gossip path and cannot serve as evidence for the validity of the changes in that specific module.</p>

otamachan:
Independent reproduction of this deadlock from a different workload (ROS 2 / `rmw_zenoh`), with a full `thread apply all bt` that shows every participant of the cycle, and an A/B of this PR against a control build.

Standalone reproducer: https://gist.github.com/otamachan/c43030eca21246d56151312fceb91d80

## Setup

ROS 2 Jazzy, `rmw_zenoh_cpp` 0.2.10, zenoh-c `05bd370` (zenoh `2687c513`). All sessions are `mode: "peer"` with gossip autoconnect (`autoconnect: { peer: ["router", "peer"] }`, `to_peer: "greater-zid"`) against a single router, everything on one Linux host.

A router plus 40 resident peer sessions, then repeated waves of 30 new peer sessions joining the existing mesh, all pinned to 4 CPUs. Each "peer" is just a `static_transform_publisher`, i.e. `Session::open()` plus one publisher.

In ROS terms the symptom is that `rmw_init()` never returns for a freshly started node: the process exists but never becomes a node, prints nothing at all, and only `SIGKILL` removes it (it has SIGINT/SIGTERM in `SigCgt` but cannot service them). On a robot this silently disabled a sensor transform for a whole run.

## Backtrace

Five threads, one cycle. `ctrl_lock` is `TablesLock::ctrl_lock`.

| thread | what it is doing |
| --- | --- |
| `rx-0` | **holds `ctrl_lock` + `tables` write**, `DeMux::handle_message` -> `p2p_peer::HatCode::handle_oam` -> `ZRuntime::block_in_place` -> `tokio::runtime::park::Inner::park` |
| `net-0` | `connect_peer` -> `open_transport_unicast` -> `open_link` -> `notify_new_transport_unicast` -> `RuntimeTransportEventHandler::new_unicast` -> `Mutex::lock_contended` (`ctrl_lock`) |
| `acc-0` | `accept_link` -> `notify_new_transport_unicast` -> `RuntimeTransportEventHandler::new_unicast` -> `Mutex::lock_contended` (`ctrl_lock`) |
| `rx-3` | `DeMux::handle_message` -> `Face::send_declare` -> `Mutex::lock_contended` (`ctrl_lock`) |
| main | `rmw_init` -> `z_liveliness_get` -> `LivelinessGetBuilder::wait` -> `Face::send_interest` -> `Mutex::lock_contended` (`ctrl_lock`) |

Full log is `gdb_hang.log` in the gist. This is exactly the cycle described in the PR body, including the `worker_threads = 1` step.

Two things that may be worth adding to the description:

- The `Face::send_interest` frame shows the cycle also traps **session open**, not just long-lived publishers. Any `Session::open()` that issues a liveliness query after the routing tables are wedged blocks forever. That is what makes this fatal for ROS 2 - the node never finishes initialising, so it never shows up in the graph and cannot be recovered in-process.
- #2409 explains why the parked `block_on` never wakes even though nothing visibly holds the `StartConditions` mutex: tokio's `Mutex` can refuse the acquisition when a waiter already has an assigned permit.

## A/B

The gist also documents how to build a patched `libzenohc.so` and swap it into a ROS install with `LD_LIBRARY_PATH`, so the numbers below can be checked independently.

30 rounds each, one round = 30 fresh processes.

| build | rounds with a hang |
| --- | --- |
| stock zenoh-c `05bd370` | 9 / 30 |
| locally built, unpatched (control) | 9 / 30 |
| **locally built + this PR** | **0 / 30** |

The control build rules out toolchain differences. Our production build also carries one unrelated local patch to the transmission pipeline; with that included the numbers are 8 / 30 before and 0 / 30 after this PR.

Raising the Net worker count does **not** avoid it, so the single-worker default is what makes it easy to hit rather than what makes it possible:

| `ZENOH_RUNTIME` | rounds with a hang |
| --- | --- |
| default | 9 / 30 |
| `(net: (worker_threads: 4))` | 9 / 30 |
| `(net: .., acc: .., rx: (worker_threads: 4))` | 7 / 30 |

Happy to run anything else against the reproducer if it helps move this along.

JEnoch:
We just merged an alternative fix to this PR: #2779 
For those of you with reproducers, could you please check if it works for you ?

I'm now preparing a bump of rmw_zenoh with 1.10.1 + #2779

Mensch88:
I backported PR #2779 onto `release/1.9.0` (`81c6c933`, the commit zenoh-c 1.9.0 pins and the one we vendor) and built both libraries from the same tree. The backport mirrors the PR line-for-line across the same three files.

Method unchanged my last report: `ctest -j 4 --timeout 300 --repeat until-fail:10` over ~137 cases, one short-lived `mode: "peer"` process per case, one multicast scouting group. A hang manifests as a 300 s per-test timeout.

| library | workers/pool | sweeps | hung |
|---|---|---|---|
| baseline | 1 | 3 | **3 / 3** (5 timeouts) |
| `#2779` | 1 | 3 | **0 / 3** |
| baseline | 2 | 2 | **0 / 2** |

Runs were executed back-to-back on the same host. All tests passed in the clean sweeps, and the wall-clock difference is purely driven by whether the 300 s timeouts occurred.

The third row serves as a control: our internal wrapper had previously defaulted `ZENOH_RUNTIME` to two workers per pool as a workaround (respecting any explicit caller override). We therefore pinned it back to one worker to reproduce the defect — two workers ran cleanly, one worker consistently hung, with the linked library being the only variable.

In short, #2779 completely eliminates the hang across 3 out of 3 sweeps in our test suite. Thanks for the fix!

As a caveat: our reproducer only goes through entity declaration, so it only provides evidence for the hat/peer/interests.rs, it does not exercise the code path in gossip.rs.

PR #2646, opened 2026-06-16
SHM handoff

yellowhatter:
---
## 🏷️ Label-Based Checklist

**Labels:** `new feature`, `breaking-change`, `release`

## 🆕 New Feature Requirements

## 💥 Breaking Change Requirements

Since this PR contains breaking changes:

- [x] **Breaking changes documented** - Clear list of what breaks
- [x] **Migration guide provided** - Step-by-step instructions for users
- [x] **Deprecation considered** - Could this be done with deprecation warnings first?
- [x] **Version bump planned** - Major or minor version bump needed
- [x] **Alternatives explored** - Why is breaking change necessary?
- [x] **Impact assessed** - How many users/use cases affected?

**Note:** Breaking changes should be discussed with maintainers before implementation.

**Instructions:**

## Summary

This PR introduces a reliable shared-memory buffer handoff mechanism for unicast transports.

The handoff ensures that a shared-memory buffer remains valid while its corresponding network message is waiting in a transport queue or otherwise in flight. The sender retains a hard reference to every transmitted SHM buffer until the receiver has successfully mounted that buffer.

This prevents SHM buffers from being reclaimed or invalidated when message delivery is delayed by:

* Transport backpressure
* Congested or stalled queues
* Slow receivers
* Scheduling delays
* Large reliable-message backlogs

## Problem

SHM payloads are transmitted as descriptors rather than copied into the transport message.

Previously, the sender could release its last strong reference after the message had been submitted to the transport. The transport message could then remain queued for longer than the lifetime protected by the normal SHM watchdog/reference mechanism.

In that situation:

1. The sender serializes an SHM descriptor.
2. The message waits in a transport queue.
3. The original SHM buffer reference is released.
4. The SHM allocation may be invalidated or reused.
5. The receiver eventually processes a descriptor whose underlying buffer is no longer valid.

A successful enqueue therefore cannot be treated as completion of SHM ownership transfer. Ownership must remain with the sender until the receiver has actually opened and mounted the buffer.

## Solution

The PR implements an explicit handoff protocol backed by shared atomic counters.

For every reliable transport direction and priority:

1. The sender leases a counter from its SHM transport metadata segment.
2. The counter identifier is exchanged with the peer during transport establishment.
3. Before enqueueing a message containing SHM buffers, the sender:

   * Creates a hard reference for each buffer.
   * Increments the corresponding shared counter.
   * Stages the references in a handoff transaction.
4. If the message is accepted by the transport pipeline, the transaction is committed.
5. If enqueueing fails or the operation is cancelled, the transaction is rolled back automatically.
6. The receiver mounts each SHM buffer from its descriptor.
7. Only after a successful mount does the receiver decrement the sender's shared counter.
8. A sender-side cleanup task observes the counter and releases acknowledged hard references in FIFO order.

```mermaid
sequenceDiagram
    participant TX as Sender
    participant C as Shared counter
    participant Q as Transport queue
    participant RX as Receiver

    TX->>TX: Create ShmBufHardRef
    TX->>C: Increment pending count
    TX->>Q: Enqueue message and SHM descriptor

    alt Enqueue succeeds
        TX->>TX: Commit handoff transaction
        Q->>RX: Deliver message
        RX->>RX: Open and mount SHM buffer
        RX->>C: Decrement pending count
        TX->>TX: Drop acknowledged hard reference
    else Enqueue fails
        TX->>C: Roll back counter increment
        TX->>TX: Drop staged hard reference
    end
```

The mechanism uses shared-memory atomics after transport establishment, so it does not require an additional acknowledgement message for every transmitted SHM buffer.

## Transport metadata segment

The existing SHM authentication segment is expanded into a transport metadata segment containing:

* Authentication challenge
* SHM metadata version
* Supported SHM protocols
* A pool of atomic handoff counters

Each process creates a local metadata segment and advertises its segment identifier during SHM negotiation.

The peer opens this segment and validates:

* The challenge
* The SHM metadata version
* The supported protocol information

Counter identifiers are then exchanged during the transport open phase.

## Establishment flow

SHM establishment now exchanges the information required for both authentication and handoff.

### `InitSyn`

Alice sends her SHM transport metadata segment identifier.

### `InitAck`

Bob:

* Opens and validates Alice's segment.
* Returns Alice's challenge.
* Sends Bob's metadata segment identifier.

### `OpenSyn`

Alice:

* Validates Bob's challenge.
* Leases local transmission counters.
* Sends the counter identifiers that Bob must decrement after receiving Alice's buffers.

### `OpenAck`

Bob:

* Leases his own transmission counters.
* Returns their identifiers to Alice.

After the exchange, each endpoint has:

* Local counters for buffers it transmits
* Access to the peer's segment for acknowledging buffers it receives

If the handoff state cannot be completely established, SHM handoff is disabled for that link.

## Reliable links and priorities

Handoff tracking is enabled only for reliable communication.

Reliable transports can maintain separate queues for different priorities. A single counter for the entire link would not preserve the ordering relationship between those queues.

The implementation therefore allocates one handoff channel per transport priority. Each channel has:

* Its own shared counter
* Its own FIFO of retained hard references
* Its own acknowledgement and cleanup state

This keeps buffer release ordering aligned with the queue in which the corresponding message was transmitted.

Best-effort communication does not use the handoff mechanism.

## Transactional enqueue semantics

SHM references and counter increments are managed through a handoff transaction.

When a message is mapped for transmission, each SHM slice added to the message is registered with the current transaction.

The transaction remains uncommitted until the transport pipeline accepts the message.

### Successful enqueue

When the message is accepted:

* Staged hard references are moved into the committed FIFO.
* The shared counter increments remain active.
* The references stay alive until acknowledged by the receiver.

### Failed or cancelled enqueue

When the message is rejected, dropped or cancelled:

* Counter increments are reverted.
* Staged references are released.
* No phantom in-flight buffers remain in the handoff state.

The transaction automatically cancels itself when dropped without an explicit commit, which protects early-return and error paths.

## Receive-side semantics

When receiving a message containing an SHM descriptor, the receiver:

1. Deserializes the SHM buffer information.
2. Opens and mounts the corresponding buffer.
3. Replaces the serialized slice with the mounted SHM buffer.
4. Decrements the sender's handoff counter for the message priority.

The counter is decremented only after the buffer has been mounted successfully.

Consequently, a malformed descriptor, unsupported protocol or mount failure cannot incorrectly tell the sender that ownership has been transferred.

## Hard references

The PR adds `ShmBufHardRef`, which retains the confirmed SHM metadata/watchdog ownership required to keep an allocation valid.

Hard references can be created from the supported SHM buffer wrapper types and are stored by the TX handoff queues while buffers are in flight.

The actual payload is not copied. The reference exists only to prevent premature SHM invalidation or reuse.

## Sender-side cleanup

Committed references are stored in FIFO order.

A periodic cleanup task compares:

* The number of locally retained references
* The number of buffers still reported as pending by the shared counter

When the receiver decrements the counter, the cleanup task removes the corresponding number of references from the front of the FIFO.

This means:

* References are retained while the receiver still reports them as pending.
* Acknowledged references are eventually released.
* No per-buffer response packet is required.
* Release may occur shortly after acknowledgement rather than synchronously with it.

If a receiver remains connected but never mounts a buffer, the reference intentionally remains retained. This favors memory safety over prematurely reclaiming an allocation that may still be consumed.

### `zenoh-shm`

* Adds `ShmBufHardRef`.
* Reworks confirmed descriptor ownership around the SHM watchdog.
* Introduces TX and RX handoff counter leases.
* Extends the SHM metadata segment with the counter pool.
* Adds segment and shared-structure equality support.
* Increments the SHM metadata version to reflect the new layout.

### SHM codecs

* Adds codecs for SHM segment references.
* Adds codecs for structures stored in SHM segments.
* Encodes segment identifiers and reopens the corresponding segment while decoding.

### Transport protocol

* Changes the SHM open extension from a scalar value to a buffer capable of carrying the expanded establishment data.
* Exchanges metadata segment identifiers, challenges and counter identifiers.
* Stores established TX and RX handoff state in the SHM negotiation FSM.

### Unicast transport

* Adds per-link TX and RX handoff configuration.
* Integrates handoff transactions into SHM message mapping.
* Commits handoff state only after successful pipeline enqueue.
* Passes receive-side handoff state into SHM buffer mounting.
* Propagates the configuration through the universal and low-latency transport structures.

### Common transport utilities

* Adds a reusable priority-indexed container.
* Adds enabled and disabled handoff configurations.
* Keeps non-SHM and best-effort paths free from handoff bookkeeping.

## Failure and lifecycle behavior

The implementation is designed to preserve the following invariants:

* Every committed TX hard reference has a corresponding increment in a shared counter.
* A counter is decremented only after successful RX mounting.
* References are released in the same FIFO domain in which they were registered.
* Failed enqueue operations do not leave counters incremented.
* Dropped uncommitted transactions automatically roll back.
* Closing a handoff channel stops its cleanup task.
* Destroying the channel releases any references still owned locally.
* Counter leases return their counter identifiers to the segment pool when dropped.

## Compatibility

The SHM metadata version is increased because the shared segment layout and establishment data have changed.

Peers must agree on the SHM metadata version before using the new metadata segment. An incompatible peer must not interpret the new segment layout or use its counters.

The SHM open-extension wire representation also changes to carry the expanded negotiation payload.

These changes affect SHM negotiation rather than ordinary non-SHM payload transport. No application-level API changes are required to benefit from the handoff mechanism.

## Performance and resource impact

For each reliable transport direction, the handoff introduces:

* One shared atomic counter per priority
* One retained hard reference per in-flight SHM slice
* A local FIFO used to track committed references
* Periodic cleanup of acknowledged references

Counter acknowledgements are direct shared-memory atomic operations and do not generate additional network messages.

Best-effort links and messages without SHM buffers do not retain SHM references through this mechanism.

Memory use is proportional to the number of SHM slices that have been committed to the transport but have not yet been acknowledged by the receiver.

## Reviewer focus areas

The most important areas to review are:

* Counter/reference consistency across all enqueue outcomes
* FIFO assumptions within each transport priority
* Multiple-SHM-slice accounting
* Cleanup-task cancellation and link teardown
* Peer disconnect behavior with outstanding references
* Counter lease reuse and exhaustion
* SHM version and wire-format compatibility
* Coverage of every congestion-control path

## Expected result

After this change, an SHM buffer queued on a reliable unicast transport remains valid until the receiver has successfully mounted it, regardless of how long the message spends waiting in the transport queue.

This preserves zero-copy SHM delivery while preventing premature buffer invalidation under backpressure or prolonged queue stalls.


doisyg:
Any progress there ?
Is this worth to test ?

yellowhatter:
> Any progress there ? Is this worth to test ?

Hi @doisyg ! This is still WIP, ETA this week. Not yet testable.

yellowhatter:
Hi @doisyg ! You may try SHM handoff PR since now. I'm still making some tests for it, so not guaranteed to be bug-free, but should work.

YuanYuYuan:
## Concurrency issue in `TxHandoff` staging (likely blocking)

`TxHandoffInner::not_commit` (`io/zenoh-transport/src/unicast/establishment/ext/shm/handoff.rs`) is a single `Mutex<VecDeque<ShmBufHardRef>>` shared per `(link, priority)`. Every `TxHandoffTransaction` for that priority clones the same `Arc`, so concurrent transactions' `push()`/`commit()`/`cancel()` all touch the same queue with no isolation between them.

Nothing serializes this upstream: in `internal_schedule` (`unicast/universal/tx.rs`), the `links` read-lock is dropped before `pipeline.push_network_message` is called, and the pipeline's own per-priority mutex is only acquired *inside* that call — after the handoff `push()` has already staged. So two same-priority sends on the same link can race.

**Failure mode:** transaction A pushes, then transaction B pushes and commits. B's `commit()` drains the whole queue, including A's still-staged entry. If A's own send then fails, A's `Drop::cancel()` finds the queue already empty — a no-op — so A's counter increment is never rolled back and A's hard ref sits in the handoff queue forever, waiting on an RX ack that will never come. That's the same class of leak this PR is meant to fix, reintroduced under concurrent same-priority sends.

I confirmed this with a unit test against the current head (two threads sharing one `TxHandoff`, barrier-synced to force the race), which fails as predicted:

```
assertion `left == right` failed: expected only B's transaction to be committed;
got 2 committed entries — A's uncommitted push leaked into B's commit
  left: 2
 right: 1
```

**Suggested fix:** stage per-transaction (e.g. a local buffer inside `TxHandoffTransaction`, merged into the shared queue only from `commit()`/`cancel()`) — unless same-priority `internal_schedule` calls are meant to be serialized by some guarantee elsewhere in the transport that isn't visible from this diff, in which case it'd be worth documenting.


**Reproducible test:** https://gist.github.com/YuanYuYuan/95382b2e9b1618871cff18416a3008b1

io/zenoh-transport/src/unicast/establishment/ext/shm/segment.rs
```diff
@@ -0,0 +1,315 @@
...
+pub(crate) type ShmCounterID = u16;
+
+#[derive(Debug)]
+#[repr(C)]
+struct ShmTransportMetadata {
+    id_count: u64,
+    challenge: u64,
+    version: u64,
+    protocols: [ProtocolID; 256],
+    shm_counters: [AtomicU32; 762],
```
JEnoch: I understand that within a link 1 SHM counter is used per-channel and hence per-priority. Meaning 8 counters per-link, right ?
762 counters allow to handle 95 links, limiting the number of peers in a p2p system to 96 peers.
That will not be sufficient for some ROS 2 systems where more than 200 peers are deployed.

io/zenoh-transport/src/unicast/establishment/ext/shm/handoff.rs
```diff
@@ -0,0 +1,330 @@
...
+}
+
+impl TxHandoffInner {
+    pub fn new(counter: ShmTXCounterLease) -> Self {
+        let task = Arc::new(TxHandoffTask::new(counter));
+        let token = CancellationToken::new();
+
+        let c_task = task.clone();
+        let c_token = token.clone();
+        ZRuntime::Net.spawn(async move {
```
JEnoch: Does this mean a new task is created for each SHM counter ?
In large p2p systems with many links, might this result in too many tasks that will wake-up every 100ms ?

On the other hand, 100ms sleep seems quite large in case of high frequency of messages. At 100Hz it means 10 messages are retained per-cycle in SHM before to be cleaned-up.

io/zenoh-transport/src/common/shm/interop.rs
```diff
@@ -429,6 +559,9 @@ pub fn map_zslice_to_shmbuf(zslice: &mut ZSlice, shmr: &ShmReader) -> ZResult<()
     // Mount shmbuf
     let smb = shmr.read_shmbuf(shmbinfo)?;
 
+    // Handle RX handoff
+    handoff.on_rx(priority);
```
JEnoch: If the deserialization of shminfo fails, this handoff is never called, causing a leak.

yellowhatter:
> ## Concurrency issue in `TxHandoff` staging (likely blocking)
> 
> `TxHandoffInner::not_commit` (`io/zenoh-transport/src/unicast/establishment/ext/shm/handoff.rs`) is a single `Mutex<VecDeque<ShmBufHardRef>>` shared per `(link, priority)`. Every `TxHandoffTransaction` for that priority clones the same `Arc`, so concurrent transactions' `push()`/`commit()`/`cancel()` all touch the same queue with no isolation between them.
> 
> Nothing serializes this upstream: in `internal_schedule` (`unicast/universal/tx.rs`), the `links` read-lock is dropped before `pipeline.push_network_message` is called, and the pipeline's own per-priority mutex is only acquired _inside_ that call — after the handoff `push()` has already staged. So two same-priority sends on the same link can race.
> 
> **Failure mode:** transaction A pushes, then transaction B pushes and commits. B's `commit()` drains the whole queue, including A's still-staged entry. If A's own send then fails, A's `Drop::cancel()` finds the queue already empty — a no-op — so A's counter increment is never rolled back and A's hard ref sits in the handoff queue forever, waiting on an RX ack that will never come. That's the same class of leak this PR is meant to fix, reintroduced under concurrent same-priority sends.
> 
> I confirmed this with a unit test against the current head (two threads sharing one `TxHandoff`, barrier-synced to force the race), which fails as predicted:
> 
> ```
> assertion `left == right` failed: expected only B's transaction to be committed;
> got 2 committed entries — A's uncommitted push leaked into B's commit
>   left: 2
>  right: 1
> ```
> 
> **Suggested fix:** stage per-transaction (e.g. a local buffer inside `TxHandoffTransaction`, merged into the shared queue only from `commit()`/`cancel()`) — unless same-priority `internal_schedule` calls are meant to be serialized by some guarantee elsewhere in the transport that isn't visible from this diff, in which case it'd be worth documenting.
> 
> **Reproducible test:** https://gist.github.com/YuanYuYuan/95382b2e9b1618871cff18416a3008b1

Fixed! Made priority channel locking

io/zenoh-transport/src/common/shm/interop.rs
yellowhatter: Fixed!

io/zenoh-transport/src/unicast/establishment/ext/shm/handoff.rs
yellowhatter: Made one single global task

io/zenoh-transport/src/unicast/establishment/ext/shm/segment.rs
yellowhatter: I was planning to make growing counter segment based on overcommit mechanics.
Now I made fixed-size array for 351 peer

io/zenoh-transport/src/unicast/establishment/ext/shm/segment.rs
```diff
@@ -0,0 +1,324 @@
...
+            );
+            return false;
+        }
+
+        true
+    }
+
+    fn protocols(&self) -> &[ProtocolID] {
+        &self.protocols[..self.id_count as usize]
+    }
```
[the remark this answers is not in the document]
yellowhatter: SHM subsystem is vulnerable to such kind of attacks at many points. Technically, safety could be achieved by more mature measures, like `mprotect()` etc. I think this is a matter of another work package.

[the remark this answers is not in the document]
yellowhatter: Added bounds check, but segment locking is still needed!

io/zenoh-transport/src/unicast/establishment/ext/shm/handoff.rs
```diff
@@ -0,0 +1,538 @@
...
+    pub fn new_rx(segment: &Arc<RXAuthSegment>, ids: HandoffCounterIds) -> Self {
+        match ids {
+            HandoffCounterIds::Disabled => Self::Disabled,
+            HandoffCounterIds::PerPrio(prio_container) => {
+                let prio_container = prio_container
+                    .map(|counter_id| ShmRXCounterLease::new(segment.clone(), counter_id));
+                Self::PerPrio(prio_container)
+            }
+        }
+    }
```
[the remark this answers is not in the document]
yellowhatter: Added bounds check

io/zenoh-transport/src/unicast/establishment/ext/shm/handoff.rs
```diff
@@ -0,0 +1,538 @@
...
+
+    fn lock(&self) -> LockedTxHandoff {
+        let inner = self.inner.clone();
+        let lock = zlock!(self.inner.not_commit);
+
+        // SAFETY: this is safe because we store Arc to inner together with this reference. Should
+        // track that reference never leaves  LockedTxHandoff
+        let lock: std::sync::MutexGuard<'static, VecDeque<ShmBufHardRef>> =
+            unsafe { std::mem::transmute(lock) };
+        LockedTxHandoff { inner, lock }
```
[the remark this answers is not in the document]
yellowhatter: Here `Note:` note is added and under current constrains this is safe

io/zenoh-transport/src/unicast/establishment/ext/shm/handoff.rs
```diff
@@ -0,0 +1,539 @@
...
+    }
+}
+
+struct LockedTxHandoff {
+    inner: Arc<TxHandoffInner>,
+    lock: std::sync::MutexGuard<'static, VecDeque<ShmBufHardRef>>,
```
YuanYuYuan: Rust drops struct fields in declaration order, so `inner` releases before `lock`. `TxHandoffInner` owns the mutex that `lock` points into. If that `Arc` holds the last strong reference, the drop frees the mutex. `MutexGuard::drop` then unlocks freed memory.

Storing the `Arc` is necessary. It is not sufficient. Field order is what makes the transmute sound. No comment records that.

The window is reachable by construction. `TxHandoffTransaction` carries no lifetime parameter. `internal_schedule` also releases the links read-lock before it pushes. A `del_link` that races an in-flight transaction can leave that transaction holding the last reference.

I have not observed this in a running transport. Minimal reproduction of the two facts, with no zenoh types: https://gist.github.com/YuanYuYuan/e72848fa4c72cec6dc30f9cffc92b444

```suggestion
    // SAFETY: `lock` borrows into the mutex that `inner` owns. Rust drops fields in
    // declaration order. Keep `lock` first, or `inner` frees the mutex too early.
    lock: std::sync::MutexGuard<'static, VecDeque<ShmBufHardRef>>,
    inner: Arc<TxHandoffInner>,
```

io/zenoh-transport/src/unicast/establishment/ext/shm/handoff.rs
```diff
@@ -0,0 +1,539 @@
...
+        let inner = Arc::new(TxHandoffInner::new(counter));
+        Self { inner }
+    }
+
+    fn lock(&self) -> LockedTxHandoff {
+        let inner = self.inner.clone();
+        let lock = zlock!(self.inner.not_commit);
+
+        // SAFETY: this is safe because we store Arc to inner together with this reference. Should
+        // track that reference never leaves  LockedTxHandoff
```
YuanYuYuan: This comment gives the stored `Arc` as the reason the transmute is sound. Field order decides it instead. See the suggestion on `LockedTxHandoff` below.

```suggestion
        // SAFETY: `inner` keeps the mutex alive. LockedTxHandoff drops the guard first.
```

io/zenoh-transport/src/unicast/establishment/ext/shm/segment.rs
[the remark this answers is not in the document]
YuanYuYuan: The check added here is `validate_counter_index`. It guards the counter index on the `ShmRXCounterLease::new` path. That check is correct. `id_count` is a different field, so this report is still open.

`RXAuthSegment::protocols()` returns `&self.protocols[..self.id_count as usize]` over a 256-entry array. The peer writes `id_count` into its own segment. `ShmTransportMetadata::validate` does bound it, but only `AuthUnicast::validate` reaches that code, and it passes the local `TXAuthSegment`. `RXAuthSegment` has no validation. `TransportShmConfig::supports_protocol` performs the read for every SHM-eligible message.

Reproduction: https://gist.github.com/YuanYuYuan/e72848fa4c72cec6dc30f9cffc92b444

```
test create_refuses_an_oversized_protocol_list ... ok
test honest_segment_reads_back_its_own_protocol_count ... ok
test protocols_trusts_the_peer_s_id_count - should panic ...
    panicked at segment.rs:90:24: range end index 300 out of range for slice of length 256
ok
```

The first test is a control. `create` refuses an oversized list, so an honest peer never reaches this state. The panic needs a peer that writes its own segment.

I agree that hardening against a hostile peer is separate work. Two checks still remove this panic: <https://github.com/YuanYuYuan/zenoh/compare/676629b23459241ae49e9121ef0589a551e9d9b6...fix/shm-review-findings>

| | today | with the patch |
|---|---|---|
| `RXAuthSegment::open` | accepts any segment | rejects a bad count or a version mismatch |
| the caller, on rejection | never happens | logs, returns `Ok(())`, keeps the link without SHM |
| `protocols()` | slices `[..300]` of a 256-entry array | clamps to 256 |
| a count rewritten after open | slices `[..300]` | clamps to 256 |
| the sending thread | panics | continues |

The open check alone is not enough. The peer keeps write access to its own segment and can change the count afterwards, so the read clamps as well. The open check is what gives the diagnostic, and it also gives `SHM_VERSION` its first comparison against a peer's segment.

Running the reproduction against that patch turns the third test red: "did not panic as expected". That is the failing baseline for this one.

io/zenoh-transport/src/unicast/establishment/ext/shm/segment.rs
YuanYuYuan: The new array holds 2810 counters. That is 8 per link, so 351 links.

```
counter slots: 2810, priorities per link: 8, links: 351
```

That clears the 200-peer bar at one link per peer. A multilink transport consumes several sets, so it reaches the ceiling sooner.

The number is not what concerns me now. It is what happens at the ceiling. `lease_tx_counter` returns `No available SHM counters`, `new_tx` propagates it with `?`, and `open.rs` maps it to a close reason:

```rust
.send_open_syn(link.link.is_reliable().into())
    .await
    .map_err(|e| (e, Some(close::reason::GENERIC)))?,
```

So the 352nd link closes the transport, and the operator sees only `GENERIC`.

Every other SHM failure in that FSM keeps the link and gives up SHM instead. There are 17 such early returns:

| condition | result |
|---|---|
| extension absent | `Ok(())` |
| decode failure | `Ok(())` |
| challenge mismatch | `Ok(())` |
| `RXAuthSegment::open` fails | `Ok(())` |
| RX segment missing | `Ok(None)` |
| `RxHandoffChannel::new_rx` fails | `Ok(())` |

The two `new_tx` calls are the only places in the FSM that propagate a peer-facing failure out. Every other `?` there reports an encoding or state-machine error. Is that difference intended?

If it is not, here is one patch: <https://github.com/YuanYuYuan/zenoh/compare/676629b23459241ae49e9121ef0589a551e9d9b6...fix/shm-review-findings>

| | today | with the patch |
|---|---|---|
| `new_tx` | returns `Err` | returns `Err`, unchanged |
| `ShmFsm` | propagates with `?` | warns, returns `Ok(None)` |
| `open.rs` | maps to `close::reason::GENERIC` | never reaches it |
| the link | refused | establishes, without SHM |
| the operator sees | a generic close reason | `Continuing without SHM: No available SHM counters` |

`Ok(None)` is the signal the missing-RX-segment path already uses.

Two things to weigh against it. The measurement only covers the ceiling: https://gist.github.com/YuanYuYuan/e72848fa4c72cec6dc30f9cffc92b444 asserts on `new_tx`, and it passes with or without the patch. Nobody has opened 352 transports, so nobody has watched either the refusal or the fallback happen. The case rests on the table above, not on a red test.

Also a correction to my own earlier suggestion in this thread: degrading to `HandoffConfig::Disabled` is not safe. It keeps SHM on and drops the handoff, which restores the use-after-free this PR fixes.

PR #2650, opened 2026-06-20
fix(storage-manager): dispatch on sample_to_store kind for wildcard-delete override

### LINKED ISSUE #2649: Stale Put after a wildcard Delete resurrects the key (storage manager dispatches on original sample kind)
### Describe the bug

In the storage manager, when an incoming `Put` is overridden by a retained wildcard `Delete`, `process_sample` correctly rebuilds `sample_to_store` as a `Delete` (empty payload, at the wildcard delete's timestamp) but dispatches the storage write on the **original** `sample.kind()`. So a stale `Put` arriving after a matching wildcard delete is stored as a `put` of the override's empty payload at the delete's timestamp — resurrecting a key the wildcard-delete tombstone should keep suppressed.

Root cause: in `process_sample`, the storage dispatch is `match sample.kind()` where it should be `match sample_to_store.kind()` (`plugins/zenoh-plugin-storage-manager/src/storages_mgt/service.rs`).

### To reproduce

Single storage on the `memory` volume, timestamping enabled.

1. PUT `a/k` = `V1` at timestamp `T1`.
2. DELETE wildcard `a/**` (matches `a/k`) at `T3`, where `T3 > T1`.
3. PUT `a/k` = `V2` at `T2`, where `T1 < T2 < T3`, delivered **after** the delete.
4. GET `a/k`.

**Expected:** `a/k` is absent — the newer retained wildcard delete overrides the older put.
**Actual:** `a/k` is present, as a `Put` with an empty payload at timestamp `T3`.

A regression test reproducing this is in the linked PR (`plugins/zenoh-plugin-storage-manager/tests/wildcard_delete_override.rs`).

### System info

- Platform: macOS 15 (also reproduced on Linux with the RocksDB backend) — backend-independent, the logic is in the storage manager above the backend
- Zenoh version/commit: 1.9.0 and `main` (`5f338ddb3`)

---
*Disclosure: this bug was diagnosed and the fix developed with the assistance of an AI coding agent (Anthropic Claude); all code and test results were reviewed by me.*

stevenlovegrove:
### What does this PR do?
`process_sample` in the storage manager now dispatches the storage write on `sample_to_store.kind()` instead of the original `sample.kind()`.

When an incoming `Put` is overridden by a retained wildcard `Delete`, `sample_to_store` is rebuilt as a `Delete` (empty payload, at the wildcard delete's timestamp), but the dispatch branched on the original `sample.kind()` (`Put`). A stale `Put` arriving after a matching wildcard delete was therefore stored as a `put` of the override's empty payload at the delete's timestamp, resurrecting a key that the wildcard-delete tombstone should keep suppressed.

In the non-override branch `sample_to_store` is derived from `sample`, so the two kinds already agree — this change only affects the override case.

A regression test is added (`plugins/zenoh-plugin-storage-manager/tests/wildcard_delete_override.rs`, mirroring the existing `wildcard.rs` / `operations.rs` harness). It puts a key, issues a matching wildcard delete at a newer timestamp, then replays a stale put with an older timestamp delivered after the delete, and asserts the key stays absent. It fails before the fix (key resurrected as an empty-payload `Put` at the delete's timestamp) and passes after. Full `cargo test -p zenoh-plugin-storage-manager`: 13 passed, 0 failed.

### Related Issues
Fixes #2649

---
*Disclosure: developed with the assistance of an AI coding agent (Anthropic Claude); all code and test results were reviewed by me.*

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/plugins/zenoh-plugin-storage-manager/src/storages_mgt/service.rs
+++ b/plugins/zenoh-plugin-storage-manager/src/storages_mgt/service.rs
@@ -328,7 +328,14 @@ impl StorageService {
             }
 
             let mut storage = self.storage.lock().await;
-            let storage_result = match sample.kind() {
+            // Dispatch on the kind of `sample_to_store`, not the original `sample`: when an
+            // incoming Put is overridden by a retained wildcard Delete, `sample_to_store` is
+            // rebuilt as a Delete (empty payload, at the wildcard delete's timestamp). Using
+            // `sample.kind()` here would store that override as a `put` of the empty payload,
+            // resurrecting a key the wildcard-delete tombstone should suppress. In the
+            // non-override branch `sample_to_store` is derived from `sample`, so the two kinds
+            // already agree.
+            let storage_result = match sample_to_store.kind() {
                 SampleKind::Put => {
                     storage
                         .put(
--- a/plugins/zenoh-plugin-storage-manager/tests/wildcard_delete_override.rs
+++ b/plugins/zenoh-plugin-storage-manager/tests/wildcard_delete_override.rs
@@ -1,0 +1,157 @@
+//
+// Copyright (c) 2023 ZettaScale Technology
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+// which is available at https://www.apache.org/licenses/LICENSE-2.0.
+//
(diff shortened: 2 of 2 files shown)


stevenlovegrove:
I've signed the ECA, but I don't know how to add the 'bug' label - perhaps that is just a note for project maintainers?

PR #2656, opened 2026-06-26
fix(util/net): make zenoh build on FreeBSD and support bind-to-device there

apbodrov:
## Description

zenoh does not build on FreeBSD at all. `set_bind_to_device_tcp_socket` and `set_bind_to_device_udp_socket` are defined only for `linux | android` and for `macos | ios | windows`, so on FreeBSD `zenoh-link-commons` fails to compile with `cannot find function ... in module 'zenoh_util::net'` — regardless of whether `iface` is ever used.

This PR makes FreeBSD compile, and makes the `iface` option actually work there.

> **Note on the original description of this PR:** it claimed FreeBSD has `IP_BOUND_IF` / `IPV6_BOUND_IF` and that a real implementation was merely out of scope. That was wrong, and the approach has changed accordingly — see below.

### Why FreeBSD can't just be added to the existing `any()` list

There is no `SO_BINDTODEVICE` / `IP_BOUND_IF` / `IPV6_BOUND_IF` on FreeBSD — no such sockopt exists in `netinet/in.h`, `netinet6/in6.h` or `sys/socket.h`. Those are Darwin/Solaris options, not FreeBSD ones.

So `set_bind_to_device_*` cannot be implemented on FreeBSD in any form: by the time either function is called the socket is already bound, and a socket can only be bound once. Adding `freebsd` to the `any()` list would only produce a warning stub, i.e. a function that silently does nothing while the user believes `iface` took effect.

The one mechanism FreeBSD does offer is binding the socket to one of the interface's own addresses. That must happen **before** the single `bind(2)` call each caller already makes — which is why FreeBSD gets an address resolver used at the call sites, rather than another `set_bind_to_device_*` arm.

### What this PR does

- **`zenoh-util`** — adds `resolve_bind_addr_for_interface(iface, addr)`, FreeBSD-only.
- **`zenoh-link-commons`** (`tcp.rs`, `quic/socket.rs`) and **`zenoh-link-udp`** (`unicast.rs`) — on FreeBSD, resolve the bind address before `bind()`. On every other platform the existing `set_bind_to_device_*` call is untouched.

Resolution semantics:

- A **concrete** address is validated against the interface, never rewritten. A mismatched configuration is reported instead of silently binding somewhere the caller never asked for.
- An **unspecified** address (`0.0.0.0` / `::`) selects an address of the same family from the interface, preferring a routable address over a link-local one, skipping loopback.
- A link-local IPv6 address carries the interface index as its scope id. Without it FreeBSD rejects the bind with `EADDRNOTAVAIL`, which makes `iface` + IPv6 unusable.

One caveat worth stating: the interface list is a process-lifetime snapshot, so an address assigned to the interface after startup is not visible. This is inherent to resolving an address instead of naming a device.

### Behaviour on other platforms

Unchanged. Every new code path is behind `#[cfg(target_os = "freebsd")]`.

### Testing

- `cargo check --target x86_64-unknown-freebsd --all-features` for `zenoh-util` and `zenoh-link-commons` — clean. This is the check that fails on `main`.
- `cargo check --all-features` on Linux for those two crates plus `zenoh-link-udp` — clean, no regression.
- Runtime, on FreeBSD 15.1, across two interfaces:
  - a listener with `iface` set binds that interface's address and is reachable on it;
  - an IPv6 listener with `iface` set binds successfully — it fails with `EADDRNOTAVAIL` without the scope-id handling;
  - an explicit address that is not on the interface is rejected with a clear error rather than being silently moved;
  - an explicit address that is on the interface is honoured verbatim;
  - with no `iface`, behaviour is identical to before.

There is no automated coverage for this, because there is no FreeBSD runner in CI — the verification above was done by hand. Happy to follow whatever the project prefers here.

### Known gap, not addressed here

`zenoh-link-udp` still does not compile on FreeBSD, for a reason unrelated to this change: `src/pktinfo/pktinfo_unix.rs` uses the Linux-only `IP_PKTINFO` and `in_pktinfo` (FreeBSD has `IP_RECVDSTADDR` / `IP_RECVIF` instead). This was previously masked, since `zenoh-link-commons` failed to compile first. I've left it alone to keep this PR to one concern — glad to open a separate issue or PR for it.

### Related Issues

None. This is a build fix for FreeBSD support.

---
## 🏷️ Label-Based Checklist

**Labels:** `enhancement`

## ✨ Enhancement Requirements

**Instructions:**

---

**On the one unchecked box:** this PR does add one new public item, `zenoh_util::net::resolve_bind_addr_for_interface`, so "No new APIs added" cannot honestly be ticked. It has to be `pub` because `zenoh-link-commons` and `zenoh-link-udp` call it across crate boundaries, and it is `#[cfg(target_os = "freebsd")]`, so it does not widen the API surface on any platform that builds today.

This is really a build fix rather than an enhancement, so the `enhancement` checklist is a slightly awkward fit. I don't have permission to relabel — if a maintainer retags this as `bug`, the checklist should regenerate accordingly.

### DIFF
--- a/commons/zenoh-util/src/net/mod.rs
+++ b/commons/zenoh-util/src/net/mod.rs
@@ -12,11 +12,14 @@
 //   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
 //
 use std::net::{IpAddr, Ipv6Addr};
+#[cfg(target_os = "freebsd")]
+use std::net::{SocketAddr, SocketAddrV6};
 
 #[cfg(unix)]
 use lazy_static::lazy_static;
 #[cfg(unix)]
 use pnet_datalink::NetworkInterface;
+#[cfg(not(target_os = "freebsd"))]
 use tokio::net::{TcpSocket, UdpSocket};
 use zenoh_core::zconfigurable;
 #[cfg(unix)]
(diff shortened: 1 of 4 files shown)


commons/zenoh-util/src/net/mod.rs
```diff
@@ -466,3 +466,15 @@ pub fn set_bind_to_device_udp_socket(socket: &UdpSocket, iface: &str) -> ZResult
...
+pub fn set_bind_to_device_tcp_socket(socket: &TcpSocket, iface: &str) -> ZResult<()> {
+    tracing::warn!("Binding the socket {socket:?} to the interface {iface} is not supported on FreeBSD");
+    Ok(())
+}
+
+#[cfg(target_os = "freebsd")]
+pub fn set_bind_to_device_udp_socket(socket: &UdpSocket, iface: &str) -> ZResult<()> {
+    tracing::warn!("Binding the socket {socket:?} to the interface {iface} is not supported on FreeBSD");
+    Ok(())
+}
```
diogomatsubara, line 545: Why not add the freebsd target to the any() list of the existing functions instead? Any reason for duplicating them?

diogomatsubara (changes requested):
@apbodrov Thanks for your contribution! You'd need to sign the Eclipse contributor agreement so we're able to merge this change.

commons/zenoh-util/src/net/mod.rs
apbodrov: Good question, and digging into it changed the PR: adding `freebsd` to the `any()` list is not possible, and the stub approach it produced was the wrong answer. I've replaced it.

FreeBSD has no `SO_BINDTODEVICE` / `IP_BOUND_IF` / `IPV6_BOUND_IF` — no such sockopt exists in `netinet/in.h`, `netinet6/in6.h` or `sys/socket.h`. Those are Darwin/Solaris options. (The original description of this PR claimed FreeBSD had `IP_BOUND_IF` since 11.1; that was simply wrong, and I've corrected the description.)

So joining the `any()` list would have meant sharing the macOS/iOS/Windows body — a `tracing::warn` and `Ok(())`. That is not a shared implementation, it is a function that quietly does nothing while the user believes `iface` took effect. The deeper problem is that these functions are called *after* the socket is bound, and a socket can only be bound once, so on FreeBSD there is nothing they could ever do at that point.

What FreeBSD does offer is binding to one of the interface's own addresses, which has to happen *before* the single `bind(2)` each caller already makes. So instead of another `set_bind_to_device_*` arm, FreeBSD now gets an address resolver used at the call sites, and the FreeBSD stubs are gone rather than left as dead code.

Worth flagging: this also means the PR is a build fix, not just a feature — `zenoh-link-commons` does not compile on FreeBSD on `main` today. Fixing that unmasks one more FreeBSD gap in `zenoh-link-udp` (`pktinfo_unix.rs` uses the Linux-only `IP_PKTINFO`/`in_pktinfo`), which I've deliberately left out of this PR and am happy to raise separately.

diogomatsubara:
@apbodrov https://github.com/eclipse-zenoh/zenoh/pull/2087 introduced support to FreeBSD, this PR changed significantly from the initial proposal. Since your initial goal was to build on FreeBSD, please check that fix and confirm if it solves the build issue for you and we can close this PR as superseded.

PR #2657, opened 2026-06-27
fix(rest): parse text/json5 payloads with json5 crate instead of serd…

### LINKED ISSUE #2074: REST API Inconsistencty processing JSON5
When using the REST API, if data is sent with JSON5 encoding (in our application, just so we can parse NaN and Infinity) it attempts to deserialise it as a JSON object and falls back to base64 if it cannot.

This seems reasonable for the other two JSON types, but for JSON5 where we are regularly sending non-compliant JSON, the other end has no way of knowing (I think) whether it needs to do the decode itself or not.

I think JSON5 should instead be treated the same as `TEXT_PLAIN` (which is in the next block of code) where it knows it is text so does not need to base64 encode, but puts the onus on the reader to parse with a JSON5 parser.

Happy to submit a PR if this is acceptable, but I guess it will break anyone who is relying on the current behaviour and doesn't actually use any features of JSON5?

https://github.com/eclipse-zenoh/zenoh/blob/5d4420f8c5356f87bb5c05f8188bb7a863823959/plugins/zenoh-plugin-rest/src/lib.rs#L127-L136

xiao-yang25:
Title: 
 fix(rest): parse TEXT_JSON5 payloads with json5 crate 
 Body: 
 ## What
Use the `json5` crate to parse `TEXT_JSON5` encoded payloads in the REST plugin,                                                                                         
instead of treating them as strict JSON via `serde_json`.                                                                                                                
                                                                                                                                                                         
## Why                                                                                                                                                                   
                                                                                                                                                                         
Fixes #2074                                                                                                                                                              
                                                                                                                                                                         
JSON5 payloads (unquoted keys, single-quoted strings, trailing commas, comments,                                                                                         
NaN/Infinity) are valid TEXT_JSON5 but fail `serde_json` parsing, causing an                                                                                             
unexpected base64 fallback that makes the data unreadable to consumers.                                                                                                  
                                                                                                                                                                         
## Behavior change                                                                                                                                                       
                                                                                                                                                                         
| Scenario | Before | After |                                                                                                                                            
|----------|--------|-------|                                                                                                                                            
| Valid JSON5 (`{a: 1}`) | base64 string | structured JSON (`{"a":1}`) |                                                                                                 
| Strict JSON | structured JSON | structured JSON (unchanged) |                                                                                                          
| Unparseable text | base64 string | raw string |                                                                                                                        
| Non-UTF-8 bytes | base64 string | base64 string (unchanged) |                                                                                                          
                                                                                                                                                                         
> **Note:** This is a breaking change for consumers that relied on the previous                                                                                          
> base64 fallback behavior for JSON5 payloads.                                                                                                                           
                                                                                                                                                                         
## Test plan                                                                                                                                                             
                                                                                                                                                                         
- [x] Unit tests: valid JSON5, strict JSON subset, unparseable fallback                                                                                                  
- [x] `cargo clippy -p zenoh-plugin-rest --all-targets -- --deny warnings`                                                                                               
- [x] `cargo test -p zenoh-plugin-rest`

<!-- 🏷️ Label-Based Checklist START -->

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

<!-- 🏷️ Label-Based Checklist END -->

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5456,6 +5456,7 @@ dependencies = [
  "flume",
  "futures",
  "git-version",
+ "json5",
  "jsonschema",
  "lazy_static",
  "mime",
--- a/plugins/zenoh-plugin-rest/Cargo.toml
+++ b/plugins/zenoh-plugin-rest/Cargo.toml
@@ -38,6 +38,7 @@ base64 = { workspace = true }
 flume = { workspace = true }
 futures = { workspace = true }
 git-version = { workspace = true }
+json5 = { workspace = true }
 lazy_static = { workspace = true }
 mime = { workspace = true }
 schemars = { workspace = true }
--- a/plugins/zenoh-plugin-rest/src/lib.rs
+++ b/plugins/zenoh-plugin-rest/src/lib.rs
@@ -151,8 +151,7 @@ impl JSONSample {
         }
         let base64_encode = |data: &[u8]| base64::engine::general_purpose::STANDARD.encode(data);
         match encoding {
-            // If it is a JSON try to deserialize as json, if it fails fallback to base64
-            &Encoding::APPLICATION_JSON | &Encoding::TEXT_JSON | &Encoding::TEXT_JSON5 => {
+            &Encoding::APPLICATION_JSON | &Encoding::TEXT_JSON => {
                 let bytes = payload.to_bytes();
                 serde_json::from_slice(&bytes).unwrap_or_else(|e| {
                     tracing::warn!(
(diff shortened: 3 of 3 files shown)


xiao-yang25:
Could a maintainer please add the `bug` label to this PR? It fixes #2074, and the PR Checklist workflow currently requires at least one label.

xiao-yang25:
I've tested the changes and added the missing `ZBytes` import. Could a maintainer please approve and run the workflow when convenient? Thank you! @diogomatsubara

PR #2660, opened 2026-07-02
Fix queryable failover

OlivierHecart:
## Description
This is a fix for a regression: queries are not properly rerouted to one of the remaining live complete queryables when one of the complete queryables is undeclared.

Query routes are not disabled on queryable declarations that leave info unchanged. For example, one out of two complete queryables may be undeclared yet the gateway keeps routes that point to the old queryable only because all queryables for resource in question reduce to the same complete queryable.

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**


zenoh/src/net/routing/hat/mod.rs
```diff
@@ -443,6 +443,8 @@ pub(crate) trait HatInterestTrait {
 pub(crate) enum UnregisterEntityResult {
     /// Indicates that the unregisration was a no-op (e.g. the entity has duplicates).
     Noop,
+    /// Indicates that the route should be disabled.
+    DisableRoute { res: Arc<Resource> },
```
fuzzypixelz:
```suggestion
    DisableRoutes { res: Arc<Resource> },
```

A bit of a pedantic remark, but I think it's important to be consistent with the names in `disable_query_routes` and `QueryRoutes`.

zenoh/src/net/routing/hat/mod.rs
```diff
@@ -443,6 +443,8 @@ pub(crate) trait HatInterestTrait {
 pub(crate) enum UnregisterEntityResult {
     /// Indicates that the unregisration was a no-op (e.g. the entity has duplicates).
     Noop,
+    /// Indicates that the route should be disabled.
```
fuzzypixelz:
```suggestion
    /// Indicates that the routes associated with `res` should be disabled.
```

Issue #1432, opened 2024-09-17
Add trust-based authorization (certificate/token)


tjrmarques:
### Describe the feature

# Problem Statement

Currently, Zenoh uses static per-node ACLs as its authorization mechanism, based on either user/passwd or certificate common-name, depending on the authentication method.

Although this is a simple solution and works on non-dynamic small-scale installations, it does not seem to scale well for larger and dynamic environments.

An example use case would be a Zenoh-based secure interface for robot/drone swarms (e.g. Intel's drone light shows), where every vehicle or ground station needs to be authorized to send/receive telemetry and command and control to/from the swarm. As for the previous use case, the problem is that all drones would need to update their user/certificate database (and maybe permissions) for every new drone or ground station added to (or removed from) the swarm.

Making the existing authorization methods work in these scenarios seems to require either:

- updating each node every time a new (relevant) user-password/certificate is added/deleted/changed, which can be really expensive for large deployments and sometimes impossible,
- sharing user/passwd between different nodes with equivalent permissions, which is a clear security breach and prevent auditing,
- sharing certificate CN between different nodes with equivalent permissions or managing certificate CNs to a common root with wildcards (one unique per permission set), which is also a security breach preventing auditing,

An additional problem to consider is that these distributed systems may not be connected to the internet at all, requiring them to be updated in person, or they may have spotty internet connection which may delay remote actions/functionalities or even prevent critical missions from being achieved.

# Solution

The proposed solution is to implement an authorization method based on trust: each node would present a certificate/token with a set of attributes (roles or permissions) that is signed by a (chain of) authorization authority(ies) which, if trusted by the other node (via a trusted certificate/key store or a request to an authorization server), would grant access to the relevant actions/resources.

This "permission-asking" flow is the general idea behind [OAuth2](https://oauth.net/2/), [OpenID Connect](https://openid.net/developers/how-connect-works/) and [Attribute Certificates](https://datatracker.ietf.org/doc/html/rfc5755) (where authorization authorities work on the same trust chain principle has SSL certificate authorities), among others.

The specific mechanism by which nodes acquire such authentication tokens/certificates could be outside Zenoh's scope, but some revocation mechanism may need to be in scope, though it may consist of a simple revocation list that's populated by the application outside Zenoh's scope.

[...]

PR #2662, opened 2026-07-03
Add ACL key expression templates bound to client identity (`${cert_common_name}`, `${username}`)

### LINKED ISSUE #2659: ACL: templated key expressions bound to client identity (`${cert_common_name}`, `${username}`) for dynamic multi-tenant isolation
**Describe the feature**

In multi-tenant deployments (robot fleets, simulation platforms, sandboxed workloads), each tenant should only be able to access its own key space. For example, a client presenting a TLS certificate with CN `t42` should be restricted to `ns_t42/**`.

With the current ACL model this requires one explicit rule/subject/policy triple **per tenant**, because:

- `cert_common_names` only supports exact string matching, and
- `key_exprs` are static — there is no way to reference the authenticated identity of the connection in a key expression.

Since ACL config cannot be updated at runtime (a router restart is required), every dynamically created tenant means editing the config and restarting the router, which disrupts all connected sessions. This does not scale for environments where tenants are created and destroyed on the fly.

**Proposed solution**

Allow binding key expressions to the authenticated identity of the connection, in line with what MQTT brokers offer for the same problem (mosquitto's `pattern ... %u`, EMQX's `${username}`/`${clientid}`):

```json5
access_control: {
  enabled: true,
  default_permission: "deny",
  rules: [
    {
      id: "tenant-own-namespace",
      permission: "allow",
      flows: ["ingress", "egress"],
      messages: ["put", "delete", "declare_subscriber", "declare_queryable", "query", "reply"],
      // NEW: plain-string templates, expanded per connection with the
      // authenticated identity of the remote instance
      key_expr_templates: ["ns_${cert_common_name}/**"],
    },
  ],
  subjects: [
    {
      id: "tenants",
      // NEW: prefix matching for cert common names
      cert_common_name_prefixes: ["t"],
    },
  ],
  policies: [
    { rules: ["tenant-own-namespace"], subjects: ["tenants"] },
  ],
}
```

Supported placeholders:

- `${cert_common_name}` — the TLS/QUIC certificate common name                                                                                                                                                    - `${username}` — the user/password-authenticated username
                                                                                                                                                                                                                  Both can be combined in one template (e.g. `${cert_common_name}/${username}/**`).
                                                                                                                                                                                                                  Subject matching already happens per-transport in `AclEnforcer::new_transport_unicast`, whedentity is available. Templates are expanded there into per-connection policies (reusing the existing `KeBoxTree` machinery). One static config then covers an unlimited number of dynamically created tenants — no config edits, no restarts.                                                                 
**Evaluation semantics**                                                                                                                                                                                          
Template rules are evaluated *as part of their subject*: a subject's effective policy is the union of its static rules and its per-connection expanded template rules, with the usual precedence (explicit deny > explicit allow > default permission). This matters for deny-templates under `default_permis expanded policy as a separate synthetic subject would let the subject's own defaultfallback override its deny template; union semantics keep behavior identical to the equivalent static configuration.

**Validation and failure behavior**

- If a template references an identity attribute the connection does not have (e.g. `${username}` on a TLS-only connection), the rule is skipped for that connection, falling back to static rules / default
permission.
- Substituted values must be usable as (part of) a single key-expression chunk (no `/`, `*`, `$`, `?`, `#`); unsafe values cause the rule to be skipped for that connection.
- Malformed templates — including misspelled placeholders — are rejected at config load timlently at transport establishment (`$` is reserved in key expressions, so any unsubstituted`${...}` fails validation).

**Design notes**

- `key_expr_templates` is a separate plain-string field: `$` is reserved in key expressions (`$*`), so templates cannot be typed as `OwnedKeyExpr`. The result of substitution is validated with
`OwnedKeyExpr::autocanonize`.
- `key_exprs` becomes optional on a rule; at least one of `key_exprs` / `key_expr_templates` must be provided, and both can coexist on the same rule.
- `cert_common_name_prefixes` is a separate subject field (rather than glob syntax inside `silently changing the semantics of existing exact-match configs.
- Fully backward compatible: all new fields are optional, and configurations that don't use them behave exactly as today.

**Alternatives considered**

- Per-tenant static rules + restart: operationally infeasible for dynamic tenants.
- #1432 proposed trust-based authorization (permissions embedded in certificates/tokens) fo; it was closed without implementation. The template approach is a much lighter alternativethat covers the multi-tenant isolation case within the existing ACL model, without introducing a new trust format.

**Status**

We have a working implementation covered by unit tests plus end-to-end tests over TLS, QUIC and user/password transports (tenant isolation in both directions, prefix-unmatched fallback to default deny, deny-template under default allow). All existing ACL/authentication test suites pass unchanged. PR incoming.

cmeng-gao:
### What does this PR do?

Allows ACL rules to define `key_expr_templates`: plain-string key expressions containing placeholders that are expanded **per connection** from the authenticated identity of the remote instance, at transport establishment. Supported placeholders:

- `${cert_common_name}` — the TLS/QUIC certificate common name
- `${username}` — the user/password-authenticated username

Subjects additionally gain `cert_common_name_prefixes` for prefix-based matching.

With a single static rule, every client is confined to its own key space:

```json5
access_control: {
  enabled: true,
  default_permission: "deny",
  rules: [
    {
      id: "tenant-own-namespace",
      permission: "allow",
      flows: ["ingress", "egress"],
      messages: ["put", "delete", "declare_subscriber", "declare_queryable", "query", "reply"],
      key_expr_templates: ["tenant/${cert_common_name}/**"],
    },
  ],
  subjects: [
    { id: "tenants", cert_common_name_prefixes: ["t"] },
  ],
  policies: [
    { rules: ["tenant-own-namespace"], subjects: ["tenants"] },
  ],
}
```

A client presenting a certificate with CN `t42` can then only use `tenant/t42/**`.

In multi-tenant deployments (robot fleets, simulation platforms, sandboxed workloads), tenants are created and destroyed dynamically. Today each tenant requires an explicit rule/subject/policy triple (`cert_common_names` is exact-match only, and `key_exprs` cannot reference the connection identity), and since ACL config cannot be updated at runtime, every new tenant means editing the config and restarting the router — disrupting all connected sessions.

With identity-bound templates, one static config covers an unlimited number of dynamically created tenants: no config edits, no restarts. This is in line with what MQTT brokers offer for the same problem (mosquitto's `pattern ... %u`, EMQX's `${username}`/`${clientid}`).

### Related Issues

Closes #2659. #1432 previously proposed trust-based authorization (permissions embedded in certificates/tokens) for similar scalability reasons and was closed without implementation; this is a much lighter alternative that covers the multi-tenant isolation case within the existing ACL model.

## Implementation details

- **Expansion point**: subject matching already happens per-transport in `AclEnforcer::new_transport_unicast`, where the peer's authenticated identity is available. Templates are expanded there into per-connection policies, reusing the existing `KeBoxTree` machinery.
- **Evaluation semantics**: template rules are evaluated *as part of their subject* — a subject's effective policy is the union of its static rules and its expanded template rules, with the usual precedence (explicit deny > explicit allow > default permission). Evaluating the expanded policy as a separate synthetic subject would let a subject's own default fallback override its deny-templates under `default_permission: allow`; union semantics keep behavior identical to the equivalent static configuration (covered by a dedicated regression test).
- **Validation / failure behavior**:
  - substituted values must be usable as (part of) a single key-expression chunk (no `/`, `*`, `$`, `?`, `#`), otherwise the rule is skipped for that connection;
  - a template referencing an identity the connection does not have (e.g. `${username}` on a TLS-only connection) is skipped for that connection, falling back to static rules / default permission;
  - malformed templates — including misspelled placeholders — are rejected at config load time via probe expansion (`$` is reserved in key expressions, so any unsubstituted `${...}` fails validation).
- **Config shape**: `key_exprs` becomes optional on a rule; at least one of `key_exprs` / `key_expr_templates` must be provided, and both can coexist. `cert_common_name_prefixes` is a separate subject field (rather than glob syntax inside `cert_common_names`) to avoid silently changing the semantics of existing exact-match configs.
- **Backward compatibility / feature gating**: all new config fields are optional and off by default; configurations that don't use them behave exactly as before. `zenoh`'s public API is unchanged (`AclConfigRule` is not re-exported). Happy to gate the new fields behind `unstable` if preferred.
- **Performance impact**: template expansion happens once per transport at establishment (only when template rules are configured and the subject matches); the per-connection policy is a regular `KeBoxTree` consulted through the same cached decision path as static rules. Configurations without template rules take the exact same code path as before.
- **Documentation**: new fields documented in `DEFAULT_CONFIG.json5` (with a worked example) and in the `zenoh-config` doc comments.

## Testing

- Unit tests: template expansion (both placeholders, combined, multiple occurrences), unsafe-substitution rejection, missing-identity rejection, unknown-placeholder rejection at config time, prefix subject matching.
- New end-to-end suite `zenoh/tests/acl_template.rs` over **TLS**, **QUIC** and **user/password** transports: tenant isolation in both directions (cross-tenant subscribe and publish blocked at the router), prefix-unmatched fallback to default deny, deny-template under default allow (regression for the union semantics).
- All existing `acl` and `authentication` test suites pass unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
## 🏷️ Label-Based Checklist

**Labels:** `new feature`

## 🆕 New Feature Requirements

**Instructions:**

### DIFF
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -428,6 +428,27 @@
   //         "**"
   //       ],
   //     },
+  //     {
+  //       /// A rule can alternatively (or additionally) define key expression templates,
+  //       /// expanded per-connection with the authenticated identity of the remote instance.
+  //       /// Supported placeholders: ${cert_common_name} (TLS/QUIC certificate common name)
+  //       /// and ${username} (user/password authentication).
+  //       /// The substituted value must be usable as (part of) a single key expression chunk
+  //       /// (it must not contain '/', '*', '$', '?' or '#'), otherwise the rule is skipped
+  //       /// for that connection; it is also skipped if the template references an identity
+  //       /// the connection does not have. With this rule and "default_permission": "deny",
+  //       /// a client whose certificate common name is "tenant1" can only use "tenant/tenant1/**".
+  //       "id": "rule3",
+  //       "messages": [
+  //         "put", "delete", "declare_subscriber",
+  //         "query", "reply", "declare_queryable",
+  //       ],
+  //       "flows":["egress","ingress"],
+  //       "permission": "allow",
+  //       "key_expr_templates": [
+  //         "tenant/${cert_common_name}/**"
+  //       ],
+  //     },
   //   ],
   //   /// List of combinations of subjects.
   //   ///
@@ -446,6 +467,12 @@
   //       /// Subjects can be cert_common_names when using TLS or Quic
   //       "cert_common_names": [
   //         "example.zenoh.io"
+  //       ],
+  //       /// Subjects can also match cert_common_names by prefix: a remote instance
+  //       /// matches if its certificate common name starts with any listed prefix.
+  //       /// Combined with "cert_common_names" (exact matches) when both are provided.
+  //       "cert_common_name_prefixes": [
+  //         "tenant"
   //       ],
   //       /// Subjects can be usernames when using user/password authentication
   //       "usernames": [
--- a/commons/zenoh-config/src/lib.rs
+++ b/commons/zenoh-config/src/lib.rs
@@ -188,7 +188,15 @@ pub enum LowPassFilterMessage {
 #[serde(deny_unknown_fields)]
 pub struct AclConfigRule {
     pub id: String,
-    pub key_exprs: NEVec<OwnedKeyExpr>,
+    /// Static key expressions this rule applies to.
+    /// At least one of `key_exprs` and `key_expr_templates` must be provided.
+    pub key_exprs: Option<NEVec<OwnedKeyExpr>>,
+    /// Key expression templates, expanded per-connection with the authenticated identity of the
+    /// remote instance. Supported placeholders: `${cert_common_name}` (the TLS/QUIC certificate
+    /// common name) and `${username}` (user/password authentication). Templates are plain
+    /// strings (`$` is reserved in key expressions); the result of the substitution is validated
+    /// as a key expression when the transport is established.
+    pub key_expr_templates: Option<NEVec<String>>,
     pub messages: NEVec<AclMessage>,
     pub flows: Option<NEVec<InterceptorFlow>>,
     pub permission: Permission,
@@ -200,6 +208,10 @@ pub struct AclConfigSubjects {
     pub id: String,
     pub interfaces: Option<NEVec<Interface>>,
     pub cert_common_names: Option<NEVec<CertCommonName>>,
+    /// Prefix-matched certificate common names: a remote instance matches if its certificate
+    /// common name starts with any of the listed prefixes. Combined with `cert_common_names`
+    /// (exact matches) when both are provided.
+    pub cert_common_name_prefixes: Option<NEVec<CertCommonName>>,
     pub usernames: Option<NEVec<Username>>,
     pub link_protocols: Option<NEVec<InterceptorLink>>,
     pub zids: Option<NEVec<ZenohId>>,
--- a/zenoh/src/net/routing/interceptor/access_control.rs
+++ b/zenoh/src/net/routing/interceptor/access_control.rs
@@ -18,7 +18,12 @@
 //!
 //! [Click here for Zenoh's documentation](https://docs.rs/zenoh/latest/zenoh)
 
-use std::{any::Any, collections::HashSet, iter, sync::Arc};
+use std::{
+    any::Any,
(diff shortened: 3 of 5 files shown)


cmeng-gao:
For the label check: this PR is a **new feature** (implements #2659). Could a maintainer add the `new feature` label? Thanks!

langthorsten:
I would love to see this merged. If there was any feedback by the maintainers regarding this PR, I would be willing to contribute if needed.

PR #2670, opened 2026-07-07
Increase `NOFILE` soft limit in SHM subsystem

### LINKED ISSUE #2653: SHM-wrapped messages are dropped when reaching `NOFILE` limits
### Describe the bug

```console
DEBUG zenoh_transport::unicast::universal::rx: Error receiving SHM buffer: Unable to open POSIX shm segment: OS error 24
```

This error event occurs when a Zenoh application's process has reached its OS `NOFILE` (i.e. number of open files) resource limit. In such a scenario, the current SHM implementation fails to `open` the segment and effectively silently drops the received message.

Crates such as [rlimit](https://crates.io/crates/rlimit) make is easy to raise the soft `NOFILE` limit up to the hard limit, something possible on all platforms with no need for special privileges. 

Note that in the case macOS, the soft limit is quite low, at 256 open files.

Now, maxing the soft limit at startup would significantly reduce the change of such errors. The downside is that we would be turning a blind eye to FD leakage bugs that would only surface once the hard limit is reached, i.e. in high-stress use-cases where debugging is significantly more difficult.

For @yellowhatter:

> Yes, 256 is very low limit for SHM.
> I agree that we should raise the soft limit to maximum at startup as a part of some our startup routine.
> 
> For FD leakage I believe we should:
> - rely on safe Rust in managing FDs
> - can override soft limit in tests

### To reproduce

TBD

### System info

- **rev**: 1.9.0

fuzzypixelz:
## Description

Fixes #2653.

### What does this PR do?

- Add rlimit = "0.10.2" as a Unix-only dep of zenoh-shm
- Feature gate increasing the soft limit with `#[cfg(all(unix, not(test), not(feature = "test")))]`
- Raise limit idempotently in `Segment::{create,open,ensure_not_persistent}`, i.e. in codepaths that may open file descriptors.

See #2653.

### Related Issues

Uh, #2653.

---
## 🏷️ Label-Based Checklist

**Labels:** `enhancement`

## ✨ Enhancement Requirements

**Instructions:**


yellowhatter (changes requested):
The limit design is good. However, I think here we should tie rlimit not to SHM, but to `Runtime` initialization. The rationale is - the limit problem is not only SHM-related, it can easily fire on non-shm scenario with many links. So, I suggest moving this functionality into Runtime initialization and change rlimit once we create the first Runtime.

fuzzypixelz:
> The limit design is good. However, I think here we should tie rlimit not to SHM, but to `Runtime` initialization. The rationale is - the limit problem is not only SHM-related, it can easily fire on non-shm scenario with many links. So, I suggest moving this functionality into Runtime initialization and change rlimit once we create the first Runtime.

I am reluctant to go there because I don't have any documented evidence that increasing rlimits is necessary outside SHM. I understand that fd use can grow unbounded in the transport manager as well, but I haven't observed any issues there in the past. The current code is apparently aware of this issue, which explains things such as:

https://github.com/eclipse-zenoh/zenoh/blob/551084b758ce1fa524d368fcb13534c177b2d386/io/zenoh-links/zenoh-link-tcp/src/unicast.rs#L436-L444

I'm not against extending this patch to cover all Zenoh subsystems but I think that such a task would necessitate identifying and re-evaluating all existing rlimit-aware code. 

The snippet I shared and all similar code speaks of the "system open file limit" and seems unaware of (per-process) NOFILE hard/soft limits. So even if we push soft NOFILE limits to hard limits, the code would still be wrong as it speaks of `ulimit` (a per-process rlimit management tool.) I haven't put a lot of effort into studying how the existing subsystems deal with resource limits and this is the only issue I could cite. All this complicates things and thus a second separate patch would be needed, in my opinion.

yellowhatter:
Okay, let's keep it as it is

yellowhatter (approved):

fuzzypixelz:
> Okay, let's keep it as it is

The next step is tracked in #2673.

PR #2678, opened 2026-07-15
fix(liveliness): decouple send_interest replay from calling thread

YuanYuYuan:
## Summary

`Session::liveliness_query` (and any query whose replies go through the default bounded FIFO handler) can self-deadlock the calling thread. `Face::send_interest` walks the routing table and replays every currently-live matching token onto the query's reply channel synchronously, on the same thread that is about to start draining that channel. Once the number of matching results exceeds the handler's default capacity (256 slots), the synchronous send blocks forever waiting for a consumer that structurally cannot run until the function that is blocked returns.

This was originally diagnosed against a downstream ROS 2/zenoh integration hitting it in production on late-joiner nodes with many active peer sessions, and worked around there with a call-site fix (a larger reply-handler capacity) — see [ZettaScaleLabs/hiroz#226](https://github.com/ZettaScaleLabs/hiroz/pull/226). This PR fixes the underlying zenoh-core issue directly.

The fix moves the post-lock replay loop off the calling thread, so the caller can return and start draining its reply channel concurrently with the replay instead of blocking on it.

## Key Changes

- `zenoh/src/net/routing/dispatcher/face.rs`: `Face::send_interest`'s post-lock `send_declare` replay loop now runs via `self.state.task_controller.spawn_with_rt(zenoh_runtime::ZRuntime::Net, ...)` instead of inline on the calling thread. The table walk under `ctrl_lock` is unchanged — it must stay synchronous for table consistency; only the replay fan-out moves. The actual `send_declare` calls run inside `ZRuntime::Net.spawn_blocking(...)`, since they can invoke arbitrary user callbacks that block synchronously for an unbounded time; running that inline on an async worker would starve other work sharing that runtime, including `Session::close_inner`'s own teardown. A panic during the replay is logged (`tracing::error!`) rather than silently discarded.
- `zenoh/tests/liveliness_self_deadlock_{peer,router}.rs`: new regression tests, each self-contained in its own process. Each declares 300 liveliness tokens and issues a liveliness query against its own local table using the default 256-slot handler — the exact vulnerable path. Asserts both prompt completion and that all 300 replies are delivered.
- `zenoh/src/net/tests/regions/mod.rs`: the mock routing-test harness calls `Face::send_interest` directly and asserts on it synchronously; give the now-async replay a brief moment to land before returning.
- `zenoh/tests/routing.rs`: `three_node_combination`'s liveliness recipes now exercise the replay's new async hop through `ZRuntime::Net`, which defaults to a single worker thread shared across the whole process (including `Session::close_inner`'s own teardown). Under CI's contended runners that added scheduling latency occasionally exceeded the file's shared 10s `TIMEOUT`. Rather than loosening that constant for every test in the file, `Recipe::run` gained a `run_with_timeout` variant, and only `three_node_combination`'s recipe calls now use a dedicated, wider `THREE_NODE_COMBINATION_TIMEOUT` (90s); all other tests (`gossip*`, `static_failover_brokering`, `two_node_combination`, `three_node_combination_multicast`, `router_linkstate`) keep the original 10s budget.

## Verification

Direct before/after confirmation: on the parent commit (plain synchronous replay), `liveliness_self_deadlock_peer` hangs and is killed by its wrapper timeout. On this PR's commits, it and `liveliness_self_deadlock_router` both pass in ~0.5s with 300/300 replies drained. Reply ordering is preserved — the collected token declares and the `DeclareFinal` completion marker travel through the same replay task in original order.

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`, `api fix`

## 🐛 Bug Fix Requirements

**Instructions:**


fuzzypixelz (commented):
I was able to reproduce this way:

```nushell
#/usr/bin/env nuze -X

let too_many_tokens = 1..257 | each { |x| z liveliness declare-token $"too-many-tokens/($x)" }
```

```nushell
#/usr/bin/env nuze -X
z liveliness get too-many-tokens/*
# deadlocks!

z liveliness sub --history too-many-tokens/*
# also deadlocks!
```

I am against exceptionally making `Face::send_interest` async for non-FINAL mode. It it an ad-hoc escape hatch in an otherwise consistent sync design within the routing layer.

Let's not forget that the root cause of our troubles here is that (1) Zenoh callbacks can run arbitrary blocking code inside `ZRuntime` and that (2) the stable API itself shoots itself in the foot by implementing blocking handlers such as [`FifoChannel`](https://docs.rs/zenoh/latest/zenoh/handlers/struct.FifoChannel.html).

If you had to take this course—which I don't agree with—I would suggest adding a flag to switch off the async behavior for the sake of keeping routing unit tests fast and deterministic, as they were explicitly designed to be. I've implemented `disable_async_tree_computation` for precisely this reason:

https://github.com/eclipse-zenoh/zenoh/blob/8358cd582eba75f0c3855426391766f8f64a5b21/zenoh/src/net/routing/gateway.rs#L51-L60

If any action is to be taken at all, one idea would be to change the `DefaultHandler` implementation to log an error on timeout:

```diff
@@
impl<T: Send + 'static> IntoHandler<T> for TimeoutFifoChannel {
    type Handler = TimeoutFifoChannelHandler<T>;

    fn into_handler(self) -> (Callback<T>, Self::Handler) {
        let (sender, receiver) = flume::bounded(self.capacity);
        (
            Callback::from(move |t| {
-                if let Err(error) = sender.send(t) {
+                if let Err(error) = sender.send_timeout(t, FIFO_CHANNEL_SEND_TIMEOUT) {
                    tracing::error!(%error)
                }
            }),
            TimeoutFifoChannelHandler(receiver),
        )
    }
}
```

This way, the system would drop the token and log an error instead of dead-locking. I don't know which one is worse!

Another idea would be to change the `DefaultHandler` for liveliness queries to an unbounded FIFO queue. Gateways send all tokens downstream on every liveliness query, so we don't exactly store all tokens received through liveliness queries, but we've considered it in the past, and opted out for ease of implementation instead of concerns about memory usage. Thus is idea of temporarily allocating all known tokens in memory does not frighten me as a default. 

Let's do a back-of-the-envelope calculation of allocated memory for 10k tokens. `Sample` is 224 bytes on my host with `shared-memory` enabled. For a key-expression of length 32 we get a 2.5MB spike in memory usage.

PR #2683, opened 2026-07-16
fix(shm): fail loudly if CARGO_CFG_TARGET_OS is unset

YuanYuYuan:
## Summary

Follow-up to #2680. That PR replaced `commons/zenoh-shm/build.rs`'s `cfg_aliases!` macro with a direct `CARGO_CFG_TARGET_OS` check, using `unwrap_or_default()` on the env var lookup. Review feedback pointed out that `CARGO_CFG_TARGET_OS` is always set by Cargo when it runs a build script normally, so silently falling back to an empty string on a missing/unexpected value would misconfigure `shm_external_lockfile` as `false` with no diagnostic — the wrong failure mode for a build-time invariant that should never actually be violated under normal use.

## Key Changes

- `commons/zenoh-shm/build.rs`: replace `.unwrap_or_default()` with `.expect(...)` on the `CARGO_CFG_TARGET_OS` lookup, so a nonstandard invocation fails loudly at build time instead of silently misconfiguring BSD/Redox platform detection.

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

- [x] **Root cause documented** - see "Summary" above: `unwrap_or_default()` on `CARGO_CFG_TARGET_OS` silently computes `shm_external_lockfile = false` if the var is ever missing, instead of surfacing the problem.
- [x] **Reproduction test added** - no unit test is meaningful for a build-script env-var-presence assumption; the failure mode this guards against (silent misconfiguration on a genuinely-missing `CARGO_CFG_TARGET_OS`) has no reliable way to reproduce under normal `cargo build`, since Cargo always sets it. The fix is a defensive assertion, not a fix for an observed failure.
- [x] **Test passes with fix** - verified via this PR's own CI run (`cargo build`/`cargo test` across the matrix exercise the build script on every job; a passing build confirms `expect()` doesn't fire under normal Cargo invocation).
- [x] **Regression prevention** - any future code path that invokes this build script without Cargo's standard environment will now panic immediately with a clear message, rather than silently producing a wrong `cfg`.
- [x] **Fix is minimal** - one-line change (`.unwrap_or_default()` → `.expect(...)`) in `commons/zenoh-shm/build.rs`; no other files touched.
- [x] **Related bugs checked** - `grep -rn "unwrap_or_default\|unwrap_or(" commons/zenoh-shm/build.rs` confirms this was the only such fallback in the file; no other silent-default pattern exists there.

**Instructions:**

### DIFF
--- a/commons/zenoh-shm/build.rs
+++ b/commons/zenoh-shm/build.rs
@@ -19,7 +19,8 @@ fn main() {
     // family (including all Apple targets, which are BSD-derived) plus
     // Redox. Computed directly instead of pulling in the `cfg_aliases`
     // crate for a single derived flag.
-    let target_os = std::env::var("CARGO_CFG_TARGET_OS").unwrap_or_default();
+    let target_os = std::env::var("CARGO_CFG_TARGET_OS")
+        .expect("CARGO_CFG_TARGET_OS must be set by Cargo when running build scripts");
     let shm_external_lockfile = matches!(
         target_os.as_str(),
         "freebsd"


fuzzypixelz (approved):
Thanks!

PR #2689, opened 2026-07-18
fix: try all resolved addresses when connecting a TLS link

nikitatsym:
## Problem

When connecting a `tls/<hostname>:<port>` endpoint, the TLS link resolves the hostname and dials **only the first** resolved address (`get_tls_addr()` returns `lookup_host().next()`). If that address is unreachable, the retry loop keeps redialing the same dead address until `connect/timeout_ms` expires — even when another resolved address is reachable.

The plain TCP link already handles this correctly: `get_tcp_addrs()` returns all resolved addresses and `LinkManagerUnicastTcp::new_link` iterates over them, collecting errors (`io/zenoh-links/zenoh-link-tcp/src/unicast.rs`).

## Real-world impact

Client on an LTE hotspot whose resolver synthesizes DNS64 AAAA records (`64:ff9b::/96`) without a working NAT64 route — common on mobile networks. `getaddrinfo` returns the synthesized (dead) IPv6 first and the working IPv4 second. A zenoh 1.9.0 client fails with:

```
Can not create a new TLS link bound to DnsName("example.com"):
[64:ff9b::9835:bbdf]:7447: No route to host (os error 65)   (repeated until timeout)
Unable to connect to any of [tls/example.com:7447]. Timeout!
```

`nc` and `openssl s_client` to the same hostname succeed from the same machine because they iterate over resolved addresses.

## Fix

Mirror the TCP link pattern in the TLS link: add `get_tls_addrs()` (all resolved addresses, multicast filtered, same shape as `get_tcp_addrs()`); `LinkManagerUnicastTls::new_link` iterates, uses the first successful TCP connection, collects per-address errors, and reports all attempts on total failure. Listener paths unchanged. QUIC/WS links appear to have the same single-address behavior — happy to extend the same pattern there if desired.

## Testing

- Unit tests for the multi-address connect fallback (`connect_first_reachable` in `zenoh-link-tls`): a dead address falls through to a reachable one; total failure reports every attempted address; an empty address list is an error. Deterministic (loopback only, no DNS).
- `cargo test` / `check` / `fmt` / `clippy` clean on `zenoh-link-tls`.
- Differential test on an affected network (DNS64 without NAT64): a client built from this branch connects (falls through the dead IPv6 to the working IPv4); stock 1.9.0 times out as above.

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/io/zenoh-links/zenoh-link-tls/Cargo.toml
+++ b/io/zenoh-links/zenoh-link-tls/Cargo.toml
@@ -51,5 +51,8 @@ zenoh-protocol = { workspace = true }
 zenoh-result = { workspace = true }
 zenoh-runtime = { workspace = true }
 
+[dev-dependencies]
+tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
+
 [package.metadata.cargo-machete]
 ignored = ["rustls-webpki"]
--- a/io/zenoh-links/zenoh-link-tls/src/unicast.rs
+++ b/io/zenoh-links/zenoh-link-tls/src/unicast.rs
@@ -43,10 +43,13 @@ use zenoh_protocol::{
     core::{EndPoint, Locator, Priority},
     transport::BatchSize,
 };
-use zenoh_result::{zerror, ZResult};
+use zenoh_result::{zerror, Error as ZError, ZResult};
 
 use crate::{
-    utils::{get_tls_addr, get_tls_host, get_tls_server_name, TlsClientConfig, TlsServerConfig},
+    utils::{
+        get_tls_addr, get_tls_addrs, get_tls_host, get_tls_server_name, TlsClientConfig,
+        TlsServerConfig,
+    },
     TLS_ACCEPT_THROTTLE_TIME, TLS_DEFAULT_MTU, TLS_LINGER_TIMEOUT, TLS_LOCATOR_PREFIX,
 };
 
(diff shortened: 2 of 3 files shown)


nikitatsym:
For the label check: this PR is a bug fix (the TLS link only ever dials the first resolved address). Could a maintainer add the `bug` label please?

On the Copilot comment about retrying the TLS handshake across addresses - that's a slightly different scope: this PR is about bringing the TLS link in line with the existing TCP link behaviour, where the iteration covers the connect phase only. Glad to extend it in a follow-up.

nikitatsym:
Thanks for adding the `bug` label. The generated checklist requires an automated reproduction/regression test and verification that no related bugs exist. This PR currently has a real-network differential test, while similar QUIC/WS behavior is intentionally left for follow-ups. Would you prefer an automated test in this PR, or can these checklist items be handled differently?

nikitatsym:
@milyin @diogomatsubara Rebased on latest main and added unit tests covering the multi-address connect fallback. 
The bug-fix checklist items are addressed now. The CI run is pending maintainer approval (fork workflows), could you approve it and take a look when you have a chance?

nikitatsym:
@milyin @diogomatsubara

The only red check is `Unit tests on macos-latest (unstable)`: `zenoh::regions scenario7::test_regions_scenario7_order1_putsub` failed on all 3 retries (388/389 passed otherwise). The same test passes on `ubuntu-latest (unstable)` and `windows-latest (unstable)`, and the other two macOS jobs are green as well. This PR only touches `zenoh-link-tls`, and the runner log shows multicast env issues (`Can't assign requested address, os error 49` on 224.7.1.1), so it looks like a runner issue rather than a regression - recent CI runs on `main` are green too.

Could you re-run that job and take a look when you have a chance?

PR #2691, opened 2026-07-20
fix: parse array-item config key filter from the last `=`

### LINKED ISSUE #2655: `@/<zid>/<mode>/config/plugins/example/a=b/items/id=item1` is mis-parsed
### Describe the bug

The array item update methods run a `split_once('=')` on the PUT key without accounting for the fact that a `=` can occur earlier than the last key-value chunk.

### To reproduce

TBD

### System info

TBD

jpatel3:
### What does this PR do?
`Config::try_insert_json5_array_item` and `Config::try_remove_json5_array_item` accept a key of the form `<array-key>/<field-name>=<field-value>`, where the `<field-name>=<field-value>` field filter is the **last** key segment. They previously used `split_once('=')`, which splits on the **first** `=`.

When the array key itself contains a `=` — for example a plugin subkey addressed via the admin space such as `@/<zid>/<mode>/config/plugins/example/a=b/items/id=item1` — the first `=` is not the filter separator. The key was mis-parsed: `plugins/example/a` was taken as the prefix and `b/items/id=item1` as the field value, so the operation targeted the wrong field (`a`) and failed with a "field filter mismatch" error.

This PR switches both methods to `rsplit_once('=')`, so the filter is always taken from the last `=`.

Array-item config updates/removals over the admin space are broken for any array whose key path contains an earlier `=` (e.g. plugin configuration subkeys). See #2655.

A regression test (`insert_remove_json5_array_item_with_equals_in_array_key`) is added; it reproduces the reported mis-parse (fails before the fix, passes after).

### Verification
- `cargo test -p zenoh-config --all-features` — all tests pass
- `cargo clippy -p zenoh-config --all-features` — clean
- `cargo fmt -p zenoh-config -- --check` — clean

### Related Issues
Fixes #2655

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/commons/zenoh-config/src/lib.rs
+++ b/commons/zenoh-config/src/lib.rs
@@ -1323,7 +1323,7 @@ impl Config {
         key: &str,
         value: &str,
     ) -> Result<bool, validated_struct::InsertionError> {
-        let Some((prefix, field_value)) = key.split_once('=') else {
+        let Some((prefix, field_value)) = key.rsplit_once('=') else {
             return Ok(false);
         };
         let (array_key, field_name) =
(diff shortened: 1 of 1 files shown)


fuzzypixelz (approved):

fuzzypixelz:
I'm converting this to a draft until the syntax ambiguity issue is discussed and resolved.

PR #2693, opened 2026-07-21
fix(zenohd): Exit gracefully instead of panicking on config read failures

### LINKED ISSUE #2421: zeonhd - Exit gracefully on error instead of SIGABRT
### Describe the feature

Currently it seems that when `zeonhd` exits due to an error, such as parsing the config file, it exits with a SIGABRT instead of exiting gracefully (with an exit status).
The default signal handler on linux for SIGABRT is to generate a core dump. This means for something as mundane as a config file error, the system gets littered with core dumps. And because the core dump is written by the kernel, they have root ownership, and the user running `zenohd` (assuming non-root) can't even clean them up.

```
# zenohd -c /dev/null
2026-02-17T17:17:25.409383Z  INFO main ThreadId(01) zenohd: zenohd v1.7.2 built with rustc 1.85.0 (4d91de4e4 2025-02-17)

thread 'main' panicked at zenohd/src/main.rs:113:34:
Failed to load config file: Empty config file at commons/zenoh-config/src/lib.rs:1309.
Aborted (core dumped)
```

This is encountered using the v1.7.2 build from [zenoh-1.7.2-x86_64-unknown-linux-gnu-standalone.zip](https://www.eclipse.org/downloads/download.php?file=/zenoh/zenoh/latest/zenoh-1.7.2-x86_64-unknown-linux-gnu-standalone.zip)

----

The request is to have zenohd exit normally with a non-0 exit status when an error is encountered.

DanBondarenko:
### What does this PR do?
* Avoid panicking on configuration read failures en error to main.
* ~Change signature zenohd's main to `zenoh::Result<()>`.~
* ~Keep error messages identical to as they were before.~ Improve error messages, ensuring output has diagnostic information and a source file path and line number reference.
* Otherwise keep logic as identical to as it was before.
* Add unit tests.
* Add CLI-invocation tests. 
* Add tokio as dev dependency to zenohd (for CLI tests).

* On configuration read failures, a core dump to be generated. Causes [problems](https://github.com/eclipse-zenoh/zenoh/issues/2421).
* In the new failure mode, no `SIGABRT` is emitted – no core dump.

### Related Issues
* Fixes #2421

It's my first contribution. Apologies for mistakes!

---
## 🏷️ Label-Based Checklist

**Labels:** `enhancement`

## ✨ Enhancement Requirements

**Instructions:**

### DIFF
--- a/zenohd/src/main.rs
+++ b/zenohd/src/main.rs
@@ -14,6 +14,7 @@
 use clap::Parser;
 use git_version::git_version;
 use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt, EnvFilter};
+use zenoh::internal::bail;
 use zenoh::{config::WhatAmI, Config, Result, Wait};
 use zenoh_config::{EndPoint, EndPoints, ModeDependentValue, PermissionsConf};
 use zenoh_util::LibSearchDirs;
@@ -76,7 +77,7 @@ struct Args {
     adminspace_permissions: Option<String>,
 }
 
-fn main() {
+fn main() -> Result<()> {
     if let Err(e) = init_logging() {
         eprintln!("{e}. Exiting...");
         std::process::exit(-1);
@@ -85,7 +86,7 @@ fn main() {
     tracing::info!("zenohd {}", *LONG_VERSION);
 
     let args = Args::parse();
-    let config = config_from_args(&args);
+    let config = config_from_args(&args)?;
     tracing::info!("Initial conf: {}", &config);
 
     let _session = match zenoh::open(config).wait() {
@@ -97,9 +98,10 @@ fn main() {
     };
 
     std::thread::park();
+    Ok(())
 }
 
-fn config_from_args(args: &Args) -> Config {
+fn config_from_args(args: &Args) -> Result<Config> {
     let mut inline_config = None;
     for json in &args.cfg {
         if let Some(("", cfg)) = json.split_once(':') {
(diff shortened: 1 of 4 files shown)


zenohd/src/main.rs
```diff
@@ -108,9 +110,9 @@ fn config_from_args(args: &Args) -> Config {
     }
 
     let mut config = if let Some(cfg) = inline_config {
-        Config::from_json5(cfg).expect("Invalid Zenoh config")
+        Config::from_json5(cfg).or_else(|_| bail!("Invalid Zenoh config") as Result<_>)?
```
milyin: Better not to swallow original error message. Maybe we don't need wrapping error message here at all - the `?` is enough. Or if the original message is not informative enough, it's better to fix it on place, in `from_json5`. In any case it's wrong that file name with error is not shown in original code

milyin:
Replacing panic to graceful exit is right thing, but it also causes that less amount of information is shown on failure. It's not enough to mechanically replace panics to `?`, it's necessary to ensure that on each error enough of informaion about the problem is shown

DanBondarenko:
@milyin Thank you for the review! I rewrote it such that error messages are as informative as possible. I also ensured the source file path and line number show up in the error message (but only once, not twice, as each `zerror!()` accumulates these).

To achieve the most readable error messages, I actuallyd change the `main` signature back to `-> ()`. I also added tests which actually invoke the built CLI and assert on exit code and error output. I added tokio as a dev dependency for these tests.

PR #2699, opened 2026-07-23
ci: trigger release workflow on changes to main

diogomatsubara:
## Description
Instead of running the release workflow on a schedule, run it for each push to main. This ties the release workflow to the actual changes landing on main and when development activity is low, we don't waste resources on a daily basis.

I'm not entirely sure on this change though, so I'll leave in draft for now. I'll gather some feedback from the zenoh committers before this can be merged. The reason is that this change has the potential of increase release workflow runs when there's lots of activity consuming even more action minutes, instead of capping them to 20 runs per month. On the other hand, it may give the development team better feedback for each change that lands on main and the concurrency group addition should help on cancelling runs whenever the development activity is intense.

### What does this PR do?
- trigger the release workflow only on a push to the main branch
- remove the scheduled daily run
- add concurrency group to automatically cancel the previous run on the same workflow + branch

Aim to reduce CI consumption and wasteful daily runs independent of changes. When there's intense development activity only the last one is considered.

### Related Issues
Related to https://github.com/eclipse-zenoh/zenoh/pull/2698

---
## 🏷️ Label-Based Checklist

**Labels:** `internal`

## 🏠 Internal Change

**Instructions:**

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -33,6 +33,10 @@ on:
         description: Release branch
         required: false
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   tag:
     name: Branch, bump & tag crates
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -14,8 +14,9 @@
 name: Release
 
 on:
-  schedule:
-    - cron: "0 0 * * 1-5"
+  push:
+    branches:
+      - main
   workflow_dispatch:
     inputs:
       live-run:


diogomatsubara:
I ran some analysis on our history and this is not a good change. I'll close it.

| Month | Expected release workflow triggers |
|---|---:|
| January 2026 | 40 |
| February 2026 | 29 |
| March 2026 | 56 |
| April 2026 | 33 |
| May 2026 | 21 |
| June 2026 | 15 |
| July 1–23, 2026 | 20 |
| **Total** | **214** |

*Based on PR landing commits on `main`, including merge and squash-style commits. The July figure covers only July 1–23, while January is a full month. These results suggest that, given Zenoh’s activity level, the proposed change would likely increase workflow activity rather than reduce it.*

PR #2701, opened 2026-07-23
fix(transport): optionally accept reordered best-effort frames

samconsidine:
### What does this PR do?

Adds an optional receive-side sequence-number window for unfragmented BestEffort Frames arriving over non-streamed links.

When `transport.link.rx.best_effort_reorder_window` is non-zero, Zenoh keeps a bounded bitmap of recently observed sequence numbers. A reordered Frame inside the window is delivered once, while duplicates and Frames older than the window are dropped. Frames are delivered immediately; this does not buffer packets or restore their original delivery order.

The default is `0`, which preserves the existing strict monotonic validation. Reliable Frames, streamed links, multicast, and fragments retain their existing behavior.

Example:

```json5
{
  transport: {
    link: {
      rx: {
        best_effort_reorder_window: 1024,
      },
    },
  },
}
```

Unordered transports such as QUIC DATAGRAM can deliver valid Frames out of sequence. The current receive path advances its sequence-number high-water mark when a later Frame arrives, then treats subsequently arriving earlier Frames as invalid. Forward batching or other network-induced reordering can therefore be amplified into substantial application-visible loss even when the packets reach the receiver.

In a controlled A/B reproduction using the same sender workload and network profile:

| Mode | Sent | Received | Application loss | Invalid SN | RX qdisc drops |
|---|---:|---:|---:|---:|---:|
| Strict/default | 21,871 | 9,509 | 56.5% | 10,024 | 0 |
| Window = 1024 | 21,871 | 21,871 | 0% | 0 | 0 |

With the window enabled, 9,603 late reordered Frames were accepted. The maximum observed reorder distance was 29 sequence numbers (mean 16.05).

The implementation also handles sequence-number rollover and suppresses duplicate delivery. Reordered fragments are deliberately excluded because the current defragmenter tracks a single sequential message.

### Validation

- `cargo test -p zenoh-config test_best_effort_reorder_window_config --lib`
- `cargo test -p zenoh-transport common::seq_num::tests --lib`
- `cargo test -p zenoh-transport --lib` (30 passed, 1 ignored)
- `cargo clippy -p zenoh-config -p zenoh-transport --lib -- -D warnings`

The full local `zenoh-config` test suite has one unrelated existing failure: the TOML-format test runs although that local feature set reports only JSON, JSON5, and YAML support. The new configuration test passes both independently and within that suite.

### Related Issues

No matching existing issue was found.

---
## 🏷️ Label-Based Checklist

**Labels:** `enhancement`

## ✨ Enhancement Requirements

- [ ] **Enhancement scope documented** - Clear description of what is being improved
- [ ] **Backwards compatible** - Existing code/APIs still work unchanged
- [ ] **Tests updated** - Existing tests pass, new test cases added if needed
- [ ] **Performance improvement measured** - If applicable, before/after metrics provided
- [ ] **Documentation updated** - Existing docs updated to reflect improvements
- [ ] **User impact documented** - How users benefit from this enhancement

**Instructions:**

### DIFF
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -684,6 +684,12 @@
         /// more in-flight data. This is particularly relevant when dealing with large messages.
         /// E.g. for 16MiB rx_buffer_size set the value to: 16777216.
         buffer_size: 65535,
+        /// Number of recent sequence numbers remembered for duplicate detection on
+        /// unordered best-effort links. Reordered, unfragmented Frames inside this
+        /// window are accepted once and delivered immediately. No packets are
+        /// buffered or delayed. Set to 0 to use strict monotonic sequence
+        /// validation.
+        best_effort_reorder_window: 0,
         /// Maximum size of the defragmentation buffer at receiver end.
         /// Fragmented messages that are larger than the configured size will be dropped.
         /// The default value is 1GiB. This would work in most scenarios.
--- a/commons/zenoh-config/src/defaults.rs
+++ b/commons/zenoh-config/src/defaults.rs
@@ -313,6 +313,7 @@ impl Default for LinkRxConf {
     fn default() -> Self {
         Self {
             buffer_size: BatchSize::MAX as usize,
+            best_effort_reorder_window: 0,
             max_message_size: 2_usize.pow(30),
         }
     }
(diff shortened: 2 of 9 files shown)


diogomatsubara:
@samconsidine Please, sign the Eclipse Contributor Agreement so we're able to consider this change.

PR #2709, opened 2026-07-27
fix: fail fast when pushing to a transport being closed as unresponsive

### LINKED ISSUE #1876: Blocking push can block the router indefinitely
## Description

When sending "high-throughput" (3.2Mbps in my example) blocking samples through a router, if the network is for congested/cut for some reason, the router logs `Unable to push non droppable network message to XXX. Closing transport!` indefinitely, without actually closing the transport.

## Expected behavior

The router shoud show `Unable to push non droppable network message to XXX. Closing transport!` once or twice, then actually close the transport and stop complaining.

## My setup

My test network is the following:

Image

- Two computers, `a` and `b`, connected through Wi-Fi.
- Each has a router configured to connect to each others in TCP.
- `a` has two z_pub (the modified examples/z_pub), publishing 4096 bytes *blocking* messages at 50Hz.
- `a` and `b` each have one z_sub.
- No shm.

If I cut the connection between `a` and `b`, the `a` router "stop" and shows this error indefinitely.

```
2025-04-03T13:45:19.222370Z ERROR  rx-1 ThreadId(07) zenoh_transport::unicast::universal::tx: Unable to push non droppable network message to b01. Closing transport!
2025-04-03T13:45:24.222530Z ERROR  rx-0 ThreadId(06) zenoh_transport::unicast::universal::tx: Unable to push non droppable network message to b01. Closing transport!
2025-04-03T13:45:29.222667Z ERROR  rx-1 ThreadId(07) zenoh_transport::unicast::universal::tx: Unable to push non droppable network message to b01. Closing transport!
2025-04-03T13:45:34.222975Z ERROR  rx-0 ThreadId(06) zenoh_transport::unicast::universal::tx: Unable to push non droppable network message to b01. Closing transport!
```

Also, the z_sub on the `a` node now receive messages only once every five seconds.

Here are the logs for the `a` router and a tcpdump captured on `a`. 

[capture-a.pcap.gz](https://github.com/user-attachments/files/19589095/capture-a.pcap.gz)
[router-a01.txt](https://github.com/user-attachments/files/19589100/router-a01.txt)

I also did another run with RUST_LOG in trace.

[capture-a.pcap.gz](https://github.com/user-attachments/files/19589122/capture-a.pcap.gz)
[router-a01.txt.gz](https://github.com/user-attachments/files/19589123/router-a01.txt.gz)

In this logs, I do not see ["Closing transport with peer"](https://github.com/eclipse-zenoh/zenoh/blob/b3fd8d86c8f6527fb25808e820039d004e1de980/io/zenoh-transport/src/unicast/universal/transport.rs#L358) which should follow the ["Unable to push non droppable network message to b01. Closing transport!"](https://github.com/eclipse-zenoh/zenoh/blob/b3fd8d86c8f6527fb25808e820039d004e1de980/io/zenoh-transport/src/unicast/universal/tx.rs#L119). It seems the closing task is scheduled but never executed for some reason.

### To reproduce

1. Reproduce a network similar to the one given in example.
2. Run a zenohd router on each computer with the given configuration.
3. Run at least two z_pub on one side, modified to publish at 50Hz. A big payload seems to help trigger the bug more reliably (I tried 4096 bytes). Use the given configuration file, and use `--mode peer`.
4. Run a z_sub on the other computer.
5. Optional: Run a z_sub on the first computer, to showcase it is not receiving its messages either.
6. Cut the network.

The configuration file: [conf.json](https://github.com/user-attachments/files/19588615/conf.json)

I also created a bash file to ease the reproduction if you want. I am however not smart enough to make it work using docker/netns on one computer.

[clogging.sh.gz](https://github.com/user-attachments/files/19588641/clogging.sh.gz)

To use it:
1. On both computers:
    * Set the `interface` env variable to the network interface used to transmit to the other computer.
    * Have `zenohd`, `z_pub` and `z_sub` in your PATH.
2. On computer a, run `./clogging.sh a`.
3. On computer a, run `./clogging.sh b`.
3. After a few seconds, disconnect `b` from the network.

### System info

- Platform: Ubuntu 20.04
- CPU: Intel i7-9750H
- Zenoh: 1.3.1, with default features, in release.

Anishishi:
### Problem

When a peer's TCP connection stays established but the peer stops reading (zero-window; e.g. a frozen or abruptly power-cycled host), pushes of non droppable messages to that transport block for `wait_before_close`. When the deadline expires, `TransportUnicastUniversal::handle_push_result` logs `Unable to push non droppable network message ... Closing transport!` and spawns the close task on the RX runtime.

However, the transmission pipelines of the transport remain enabled until the close task actually runs. If the threads pushing to the dead transport are RX runtime workers (forwarding reliable data or declarations, holding routing locks), they immediately block again on the next message for another full `wait_before_close` each, starving the RX runtime. The close task never runs and the whole session is livelocked, logging the message above forever.

Fixes #1876, related to #2581.

### Fix

Synchronously mark the transmission pipelines of the transport as disabled at the moment the closure is decided. This is a new `TransmissionPipelineProducer::mark_disabled` method that only performs an atomic store; contrary to `disable` it does not acquire the `stage_in` locks (which may be held by a currently blocked pusher).

`push_network_message` now fails fast with `TransportClosed` when the pipeline is disabled, before attempting to acquire the `stage_in` lock. Subsequent pushes to the unresponsive transport therefore return immediately, the RX workers are freed, and the spawned close task can run and perform the regular cleanup.

No behavior change for healthy transports: the check is a relaxed atomic load on the already-existing status flag.

### Reproducer

A standalone reproducer with stock `zenohd` + examples (SIGSTOP-ed subscriber, blocking publishers, liveness probe) is attached to #1876. On current `main` the router wedges forever, alternating the error log between the RX workers every `wait_before_close`; with this PR it recovers right after `wait_before_close`.

### Tests

- New integration test `unicast_unresponsive_peer` simulates the zero-window peer with a stallable TCP proxy and asserts that, after the push failure that triggers the closure, the next push fails fast instead of blocking for another `wait_before_close`. It fails on current `main` and passes with this PR.
- `cargo test -p zenoh-transport --features transport_tcp` passes.
- Validated on a production multi-host `rmw_zenoh` deployment: 100 abrupt reboot cycles with zero occurrences of the wedge (previously reproduced within a few dozen cycles).

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/io/zenoh-transport/src/common/pipeline.rs
+++ b/io/zenoh-transport/src/common/pipeline.rs
@@ -872,6 +872,13 @@ impl TransmissionPipelineProducer {
         &self,
         msg: NetworkMessageRef,
     ) -> Result<bool, TransportClosed> {
+        // Fail fast if the pipeline has already been disabled (e.g. because a previous
+        // non droppable message could not be pushed and the transport is being closed).
+        // This prevents threads from blocking up to `wait_before_close` on a transport
+        // that is already known to be unresponsive.
+        if self.status.is_disabled() {
+            return Err(TransportClosed);
+        }
         // If the queue is not QoS, it means that we only have one priority with index 0.
         let (idx, priority) = if self.stage_in.len() > 1 {
             let priority = msg.priority();
@@ -936,6 +943,17 @@ impl TransmissionPipelineProducer {
         // Lock the channel. We are the only one that will be writing on it.
         let mut queue = zlock!(self.stage_in[priority]);
         queue.push_transport_message(msg)
+    }
+
+    /// Marks the pipeline as disabled without acquiring the `stage_in` locks.
+    ///
+    /// Contrary to [`Self::disable`], this method can safely be called while another
+    /// thread is blocked pushing a message on this pipeline (such a thread holds the
+    /// `stage_in` lock until its deadline expires). New pushes fail fast with
+    /// [`TransportClosed`] and the consumer task terminates on its next wakeup
+    /// (e.g. keep-alive), which in turn unblocks any parked pusher.
+    pub(crate) fn mark_disabled(&self) {
+        self.status.set_disabled(true);
     }
 
     pub(crate) fn disable(&self) {
--- a/io/zenoh-transport/src/unicast/universal/tx.rs
+++ b/io/zenoh-transport/src/unicast/universal/tx.rs
@@ -83,6 +83,26 @@ impl TransportUnicastUniversal {
                 "Unable to push non droppable network message to {}. Closing transport!",
                 self.config.zid
             );
+            // Synchronously mark the transmission pipelines of this transport as disabled
+            // so that any subsequent push to this unresponsive transport fails fast with
+            // `TransportClosed` instead of blocking for another `wait_before_close` period.
+            //
+            // This is critical because the thread executing this code may be an RX runtime
+            // worker that is holding routing locks (e.g. the routing tables) while pushing
+            // declarations. If such a thread kept blocking on pushes to this transport,
+            // the close task spawned below on the same starved runtime would never get to
+            // run, leaving the whole session deadlocked
+            // (see https://github.com/eclipse-zenoh/zenoh/issues/1876 and
+            // https://github.com/eclipse-zenoh/zenoh/issues/2581).
+            {
+                let transport_links = self
+                    .links
+                    .read()
+                    .expect("reading `TransportUnicastUniversal::links` should not fail");
+                for tl in transport_links.get_links().iter() {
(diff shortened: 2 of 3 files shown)


Anishishi:
Could someone with triage/write access please add the `bug` label to this PR?

I don't have permission to add labels myself, and the checklist CI (`Generate / main`) currently fails with:

```
PR must have at least one of the required labels: api-sync, breaking-change, bug, ci, dependencies, documentation, enhancement, new feature, internal
```

`bug` seems appropriate here. Thanks!

PR #2722, opened 2026-07-31
feat: add capnp encoding

ksil:
## Description
This PR adds [Cap'n Proto](https://capnproto.org) as an encoding type.

### What does this PR do?
Constants are added for the cap'n proto encoding type.

Hello, first-time contributor here! We could use `application/octet-stream` for cap'n proto structs over the wire, but my colleagues and I thought it would be nice to add a distinct cap'n proto encoding here directly to differentiate from generic byte streams and also avoid the cost of sending the string "application/capnp" over the wire constantly.

---
## 🏷️ Label-Based Checklist

Current labels: _No labels_

### DIFF
--- a/zenoh/src/api/encoding.rs
+++ b/zenoh/src/api/encoding.rs
@@ -472,6 +472,14 @@ impl Encoding {
         schema: None,
     });
 
+    /// Application-specific cap'n proto-encoded data.
+    ///
+    /// Constant alias for string: `"application/capnp"`.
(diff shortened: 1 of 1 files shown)


milyin:
The encogings set is a predefined list, part of zenoh API. It can't be arbitrarily changed

ksil:
Okay! Thanks for the feedback @milyin. For my own knowledge: is this predefined list defined elsewhere than these Rust files and/or is there a mechanism to change the zenoh API in a more formal way that we should consider moving forward?

PR #2726, opened 2026-08-03
Add configuration in shared_memory/transport_optimization to select the type of messages to optimize

JEnoch:
## Description

This PR replaces the SHM `transport_optimization` on/off granularity with a single
`messages` list under `transport/shared_memory/transport_optimization`, letting users
select exactly which categories of messages are eligible for the SHM transport
optimization. Internally, it also unifies the message-category type shared by downsampling, low-pass
filtering and this new setting.

### What does this PR do?

- **Adds a `messages` field** to the `transport_optimization` section (defaults to `["put", "query", "reply"]`, so behavior is unchanged):
  - A message category present in the list is eligible for SHM optimization; a category
    left out is always sent over the regular network path.
  - e.g. `messages: ["put"]` keeps queries/replies off SHM while preserving it
    for publications.
  - `"delete"` category is accepted but ignored since "delete" messages have no payload and hence are never optimized.

- **Unifies the configuration message-category enum:**
  - Introduces a single `DataMessage` type (`put`, `delete`, `query`, `reply`), replacing
    the previously duplicated `DownsamplingMessage` and `LowPassFilterMessage`, and reuses
    it for `downsampling/messages`, `low_pass_filter/messages` and
    `transport_optimization/messages`.
  - Removes the deprecated `push` message category.

- **Plumbs the selection** from the config down to the transport TX mapping:
  - New `ShmOptimizationPolicy` (per-category flags) carried through `ShmContext` and the
    per-transport SHM contexts (unicast universal, unicast lowlatency, multicast).
  - `map_zmsg_to_partner` takes the policy and skips the whole SHM mapping for a message
    category that is not selected.

- When a category is not in `messages`, only the **implicit** optimization is disabled for it: a large regular payload is no longer auto-copied into SHM and is sent over the regular network path instead. Payloads the application already allocated in shared memory are **always** forwarded as descriptors when the peer supports the protocol, regardless of this setting. The RX path is untouched, so a node keeps decoding SHM messages from peers regardless of its own setting (backward compatible, no wire-protocol change).

- **Adds unit tests** validating that each category independently gates its message type
  through the real implicit-optimization path, plus a check that sub-threshold payloads are
  never promoted.

Today the transport optimization is all-or-nothing: `transport_optimization.enabled` turns
implicit SHM promotion on or off for every message category at once. Selecting categories
individually is useful in its own right, for several reasons:

- **Match the optimization to the traffic profile.** SHM promotion pays off for large,
  bulk, streaming data: the typical pub/sub data plane. Query/reply traffic is usually
  small, one-shot RPC-style exchanges (services, actions, control plane) where the cost of
  promotion (SHM allocation, copy, watchdog bookkeeping, first-use provider init latency)
  often outweighs the benefit. Users can now enable SHM only where it actually helps.

- **Isolate and protect the SHM pool.** The optimization draws from a fixed-size SHM arena
  (`pool_size`). Bursts of small query/reply buffers (e.g. a startup storm of service
  calls) compete with the high-value bulk pub/sub traffic for that arena and can create
  pool pressure. Restricting optimization to publications keeps the pool dedicated to the
  data that benefits most.

- **Mitigation for a known SHM issue.** As one concrete case, the SHM optimization can
  silently drop reliable query traffic under CPU contention when an in-transit SHM buffer
  is invalidated by the watchdog before the receiver mounts it (see #2628). Since queries
  are treated as reliable and not retried (e.g. ROS 2 services over rmw_zenoh), a single
  drop hangs the caller until the query timeout. Dropping `query` and `reply` from
  `messages` keeps queries/replies off the SHM path while preserving SHM benefits for
  publications.

### Breaking changes

- The deprecated `push` value is removed from the message-category lists
  (`downsampling/messages`, `low_pass_filter/messages`); use `put` and/or `delete` instead.

### Related Issues

- Related to #2628
- Related to [ros2/rmw_zenoh#978](https://github.com/ros2/rmw_zenoh/issues/978)

### Did you use Generative AI?

Yes: Claude Opus 5

---
## 🏷️ Label-Based Checklist

**Labels:** `enhancement`

## ✨ Enhancement Requirements

**Instructions:**

### DIFF
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -758,6 +758,14 @@
           pool_size: 16777216,
           /// Allow optimization for messages equal or larger than this threshold in bytes (default `3072`).
           message_size_threshold: 3072,
+          /// Whether SHM optimization is applied to publications (Put) payloads (default `true`).
+          publications: true,
+          /// Whether SHM optimization is applied to query and reply (Request/Response) payloads (default `true`).
+          /// Disabling this keeps queries/replies off the SHM path.
+          /// Set it to `false` for a workaround for the in-transit watchdog invalidation race,
+          /// which could occur under high CPU load.
+          /// See https://github.com/eclipse-zenoh/zenoh/issues/2628).
+          queries_replies: true,
       },
     },
     auth: {
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -758,14 +758,15 @@
           pool_size: 16777216,
           /// Allow optimization for messages equal or larger than this threshold in bytes (default `3072`).
           message_size_threshold: 3072,
-          /// Whether SHM optimization is applied to publications (Put) payloads (default `true`).
-          publications: true,
-          /// Whether SHM optimization is applied to query and reply (Request/Response) payloads (default `true`).
-          /// Disabling this keeps queries/replies off the SHM path.
-          /// Set it to `false` for a workaround for the in-transit watchdog invalidation race,
-          /// which could occur under high CPU load.
-          /// See https://github.com/eclipse-zenoh/zenoh/issues/2628).
-          queries_replies: true,
+          /// The categories of messages the SHM optimization is applied to
+          /// (default: all of "put", "query", "reply").
+          /// Adding "delete" here has no effect since "delete" messages don't have a payload
+          /// and hence are never optimized.
+          /// Removing "query" and "reply" keeps queries/replies off the SHM path. This is
+          /// useful for RPC-style traffic, and a workaround for the in-transit watchdog
+          /// invalidation race that could occur under high CPU load.
+          /// See https://github.com/eclipse-zenoh/zenoh/issues/2628
+          messages: ["put", "query", "reply"],
       },
     },
     auth: {
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -758,13 +758,16 @@
           pool_size: 16777216,
           /// Allow optimization for messages equal or larger than this threshold in bytes (default `3072`).
           message_size_threshold: 3072,
-          /// The categories of messages the SHM optimization is applied to
-          /// (default: all of "put", "query", "reply").
+          /// The categories of messages the *implicit* SHM optimization is applied to,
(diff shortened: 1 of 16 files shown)


OlivierHecart:
It could be good for coherency to use a similar list of messages as in other places in the conf (for example: https://github.com/eclipse-zenoh/zenoh/blob/37d993beebb3371a396b17f5c63cca1a11c3c70d/DEFAULT_CONFIG.json5#L519)

yellowhatter (approved):

JEnoch:
> It could be good for coherency to use a similar list of messages as in other places in the conf

Done. The `trensport_optimization` config is now as such:

```json5
      transport_optimization: {
          /// Enables transport optimization for large messages (default `true`).
          /// Implicitly puts large messages into shared memory for transports with SHM-compatible connection.
          enabled: true,
          /// SHM memory size in bytes used for transport optimization (default `16 * 1024 * 1024`).
          pool_size: 16777216,
          /// Allow optimization for messages equal or larger than this threshold in bytes (default `3072`).
          message_size_threshold: 3072,
          /// The categories of messages the *implicit* SHM optimization is applied to,
          /// i.e. for which a large enough regular payload is automatically copied into
          /// shared memory (default: "put", "query", "reply").
          /// This does not affect payloads the application already allocated in shared
          /// memory: those are always sent over SHM when the peer supports it.
          /// Adding "delete" here has no effect since "delete" messages don't have a payload
          /// and hence are never optimized.
          /// Removing "query" and "reply" keeps auto-promoted queries/replies off the SHM
          /// path. This is useful for RPC-style traffic, and a workaround for the in-transit
          /// watchdog invalidation race that could occur under high CPU load.
          /// See https://github.com/eclipse-zenoh/zenoh/issues/2628
          messages: ["put", "query", "reply"],
      },
```

In `zenoh-config/src/lib.rs` I also introduced a single `DataMessage` enum (`put`, `delete`, `query`, `reply`) which is now used by `transport_optimization/messages`, `downsampling/messages ` and `low_pass_filter/messages` It replaces the `DownsamplingMessage` and `LowPassFilterMessage` types.
Note that `DownsamplingMessage` had a deprecated `push` message category. It no longer exists.

OlivierHecart (approved):

Issue #2675, opened 2026-07-12
session.close() blocks ~10s then raises ZError("close operation timed out") when the peer link has active bursty packet loss


BinHsu:
**Describe the bug**

`session.close()` blocks for ~10 seconds and then raises `ZError("close operation timed out")` when the peer link has active bursty packet loss at close time. This happens under `Reliability.BEST_EFFORT` + `CongestionControl.DROP`, where teardown should not depend on reliable delivery of anything.

Related prior art: #2012 fixed a `Session::close` timeout for a client mid-reconnect (already included in 1.9.0); this report is a second, distinct trigger for the same symptom class — an **established peer-mode TCP link under active packet loss**, not a reconnecting client. #1907 (better close-path logging) would also help diagnose cases like this one.

**Environment**

- eclipse-zenoh (Python) `1.9.0`, official PyPI manylinux wheel (`pip install eclipse-zenoh==1.9.0`)
- Python 3.12, Debian 12 "bookworm" (`python:3.12-slim-bookworm` base image)
- Two peer-mode sessions on a Docker bridge network on one Linux host, default TCP link (no TLS/QUIC/UDP)
- Packet loss emulated with the Linux kernel `netem` qdisc on each container's `eth0`

**Configuration (both peers identical apart from listen/connect roles)**

```python
conf = zenoh.Config()
conf.insert_json5("mode", '"peer"')
conf.insert_json5("scouting/multicast/enabled", "false")
conf.insert_json5("listen/endpoints", json.dumps(["tcp/0.0.0.0:7447"]))
conf.insert_json5("connect/endpoints", json.dumps(["tcp/<peer-host>:7447"]))
```

Publisher: `reliability=zenoh.Reliability.BEST_EFFORT`, `congestion_control=zenoh.CongestionControl.DROP`. No TLS, no SHM.

**Network condition at failure (Gilbert-Elliott bursty loss)**

```
tc qdisc add dev eth0 root netem loss gemodel 0.2222% 2.0000% 100% 0%
```

This targets ~10% steady-state loss with ~1 s mean burst duration at a ~50 pkt/s reference rate (`p`/`r` are the gemodel good→bad / bad→good per-packet transition probabilities — standard netem gemodel math, nothing zenoh-specific). A harsher profile (`gemodel 0.1714% 0.4000% 100% 0%`, ~30% loss, ~5 s mean bursts) triggers the same failure at a similar rate.

**To reproduce**

```python
[...]

PR #2173, opened 2025-09-29
Fix multlink reconnect

DenisBiryukov91:
Fix reconnect logic to account for multiple links on same transport
Resolves #2130

PR #2733, opened 2026-08-08
fix: make the gossip autoconnect task abortable

ryandavid:
## Description
Session close could hang forever under peer churn. `Runtime::close_inner` waits on `terminate_all_async()`, which waits for every tracked task with no timeout, and gossip's autoconnect was spawned with `spawn` -- so the cancellation token never reached it, and it sat in `connect_peer()` against a peer that was itself shutting down.

That does not meet `spawn`'s documented contract, which requires the task to be cancellable or to finish in finite time; connecting to a remote peer is neither. `spawn_abortable` is what the multicast scouting side already uses for the same work (`autoconnect_all`, orchestrator.rs:343 and :360). The task stays tracked, so shutdown still waits for it, but it now observes the token and finishes.

`token.cancel()` is only reached from `terminate_all` and `terminate_all_async`, so nothing changes during normal operation.

Adds a regression test, which fails without this change with "close operation timed out" and passes with it.

### What does this PR do?
Spawns gossip's autoconnect task with `spawn_abortable` instead of `spawn`, so
that session close can no longer wait on it for ever.

```diff
-                    strong_runtime.spawn(async move {
+                    strong_runtime.spawn_abortable(async move {
```

It also adds a regression test, `zenoh/tests/close_under_peer_churn.rs`.

Two peer sessions that discover each other over multicast and then open and close repeatedly deadlock forever inside session close. One process alone always finishes. Two, started together, hang within a handful of cycles and never
recover.

The wait is in `Runtime::close_inner`, on its **first** await - found by instrumenting the two lines, not by reading them. The marker before `terminate_all_async` prints; the one after it never does, so `manager.close()` is never reached and the transports are never closed:

```rust
// zenoh/src/net/runtime/mod.rs
async fn close_inner(&self, _: ()) {
    self.task_controller.terminate_all_async().await;   // <-- never returns
    self.manager.close().await;                         // <-- never reached
    ...
}
```

`terminate_all_async()` is unbounded - it waits for every tracked task, with no timeout:

```rust
// commons/zenoh-task/src/lib.rs
pub async fn terminate_all_async(&self) {
    self.tracker.close();
    self.token.cancel();
    self.tracker.wait().await
}
```

and the task it waits for is gossip's autoconnect, spawned with `spawn`:

```rust
// zenoh/src/net/protocol/gossip.rs:390
strong_runtime.spawn(async move {
    if runtime.manager().get_transport_unicast(&zid).await.is_none() {
        runtime.start_conditions().add_peer_connector_zid(zid).await;
        if runtime.connect_peer(&zid, &locators).await && ... {
            runtime.start_conditions().terminate_peer_connector_zid(zid).await;
        }
    }
});
```

`token.cancel()` therefore does nothing to it, and it is sitting in `connect_peer()` against a peer that is itself shutting down.

`spawn`'s own documentation says it is for a task that

> can be cancelled [by] a token obtained by `TaskController::get_cancellation_token()`, was created via `TaskController::into_abortable()`, **or can run to completion in finite amount of time**

and `terminate_all` puts that obligation explicitly on the caller:

> The caller **must ensure that all tasks spawned with `TaskController::spawn()`… can yield in finite amount of time**

Connecting to a remote peer is not finite, and this task holds no cancellation token, so neither branch of the contract holds. The equivalent task on the multicast scouting path - the same job, connecting to a peer that discovery just turned up - already uses `spawn_abortable`: `autoconnect_all` (`orchestrator.rs:343`, `:360`) and the scouting `responder` (`orchestrator.rs:223`, `:343`, `:355`). Gossip looks like the odd one out.

`spawn_abortable` still tracks the task, so `terminate_all_async` still waits for it; the task simply now observes the token and finishes instead of never. `token.cancel()` is only ever reached from `terminate_all` and `terminate_all_async`, so this changes nothing during normal operation - aborts happen only at shutdown.

## How to reproduce

A ~20-line program, default config, nothing else:

```cpp
#include <zenoh.hxx>
#include <cstdio>
#include <unistd.h>

int main()
{
    for (int cycle = 0; cycle < 50; ++cycle)
    {
        std::fprintf(stderr, "[%d] cycle %d\n", getpid(), cycle);
        zenoh::Session::open(zenoh::Config::create_default());
    }
    std::fprintf(stderr, "[%d] done -- did not deadlock\n", getpid());
    return 0;
}
```

```
./repro & ./repro & wait
```

One process alone always completes all 50 cycles. Two hang, typically by cycle 3-12; the last line each prints is the cycle it died on, and it never recovers (observed >10 minutes). While hung, every zenoh runtime worker is idle in `kevent` with nothing runnable, so this is a wakeup that never arrives rather than contention with a running thread.

### Related Issues
Related to #2409 and tokio-rs/tokio#7892 (the `block_in_place` family).

Not fixed by #2637, which was the obvious candidate - same file, same family (`block_in_place` parking the `ctrl_lock` holder on a tokio mutex under peer churn, which is exactly this workload). It was applied in full and measured first: **still hung 6 runs out of 8**. Its two hunks are in `terminate_peer_connector_zid`'s callers; this is a different task in the same function that is never cancelled. Both look worth having; only this one closes this hang.

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/zenoh/src/net/protocol/gossip.rs
+++ b/zenoh/src/net/protocol/gossip.rs
@@ -387,7 +387,7 @@ impl Gossip {
                 if let Some(locators) = locators {
                     let runtime = strong_runtime.clone();
                     let wait_declares = self.wait_declares;
-                    strong_runtime.spawn(async move {
+                    strong_runtime.spawn_abortable(async move {
                         if runtime
                             .manager()
                             .get_transport_unicast(&zid)
--- a/zenoh/tests/close_under_peer_churn.rs
+++ b/zenoh/tests/close_under_peer_churn.rs
@@ -1,0 +1,139 @@
+//
+// Copyright (c) 2026 ZettaScale Technology
+//
+// This program and the accompanying materials are made available under the
+// terms of the Eclipse Public License 2.0 which is available at
+// http://www.eclipse.org/legal/epl-2.0, or the Apache License, Version 2.0
+// which is available at https://www.apache.org/licenses/LICENSE-2.0.
+//
+// SPDX-License-Identifier: EPL-2.0 OR Apache-2.0
+//
+// Contributors:
+//   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
+//
+use std::{
+    net::{IpAddr, Ipv4Addr, SocketAddr},
+    sync::{
+        atomic::{AtomicBool, Ordering},
+        Arc,
+    },
+    time::Duration,
+};
+
+use tokio::time::timeout;
+use zenoh_config::Config;
+use zenoh_test::get_free_udp_port;
+
+/// Closing a session must not wait forever on a peer connection.
+///
+/// Two peers that have discovered each other over multicast and then open and
+/// close repeatedly used to deadlock in `Session::close`: gossip's autoconnect
+/// task was spawned with `TaskController::spawn`, so the cancellation token
+/// never reached it, and `terminate_all_async()` -- which has no timeout --
+/// waited for a peer connection that would never complete.
+///
+/// A regression here HANGS rather than fails, so every close is bounded by an
+/// explicit timeout and the whole test is bounded by the number of rounds.
+#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+async fn close_does_not_hang_under_peer_churn() {
+    zenoh::init_log_from_env_or("error");
+
+    // A multicast group of this test's own, so it neither discovers nor is
+    // discovered by anything else on the machine -- including other tests
+    // running in parallel.
+    let mcast_addr = SocketAddr::new(
+        IpAddr::V4(Ipv4Addr::new(224, 0, 0, 224)),
+        get_free_udp_port(),
+    );
+
+    // Everything except the multicast group is left at its default, and that
+    // matters twice over: the peers must LISTEN for each other to connect at
+    // all, and autoconnect must stay enabled because the gossip autoconnect
+    // task is the thing under test. Clearing either makes this test pass
+    // against the bug.
+    let config = move || {
+        let mut config = Config::default();
+        config
+            .scouting
+            .multicast
+            .set_enabled(Some(true))
+            .unwrap();
+        config
+            .scouting
+            .multicast
+            .set_address(Some(mcast_addr))
+            .unwrap();
+        config
+            .scouting
+            .multicast
+            .set_interface(Some("127.0.0.1".to_string()))
+            .unwrap();
+        config
+    };
+
+    // The other peers, churning alongside us. Several of them, and all of them
+    // opening AND closing: a long-lived peer plus one side closing does not
+    // reproduce this, and neither does a single neighbour -- there has to be a
+    // connection in flight when a session goes away.
+    let stop = Arc::new(AtomicBool::new(false));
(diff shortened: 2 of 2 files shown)


ixil:
The cause is actually a regression of PR #2012 due to PR #2173 

- 252f6a883 — PR #2012, "Fix Session::close timeout while client attempts to reconnect", 2025-07-24. Changed session.runtime.spawn(...) → spawn_abortable(...) in closed_session. Fixed exactly this, for exactly this reason.
- 225085647 — PR #2173, "Fix multlink reconnect", 2025-10-09. Rewrote closed_session for an unrelated purpose. The WhatAmI::Client arm carrying spawn_abortable was deleted, and the replacement uses two plain spawn calls. The close-timeout fix was silently lost and the site count doubled.

Issue #2675 is also due to this.

We hit this independently and arrived at the same diagnosis, so first: this analysis is correct, and `spawn_abortable` is the right remedy rather than a hand-written `select!` — `run_until_cancelled_owned` drops the task at its next await point, which is what the connect needs.

**The fix looks incomplete, though: there is a second call site with the same defect, and this PR does not cover it.**
`zenoh/src/net/runtime/orchestrator.rs:876` spawns the peer connector with plain `spawn`:
```rust
self.spawn(async move {
    if let Ok(zid) = this.peer_connector_retry(peer).await {
```
That is the **configured `connect.endpoints`** path, whereas `gossip.rs:390` is the gossip-discovery path. Same uncancellable `open_transport_unicast` await underneath, reached two different ways. With only `gossip.rs` fixed, a deployment that lists peers in `connect.endpoints` and does not rely on gossip
still hangs on close.

### Reproducing it without gossip at all
The gossip route needs two peers and multicast churn. The configured route needs neither, which makes for a smaller repro:
- one session, `scouting.multicast.enabled: false`,  scouting.gossip.enabled: false`
- a single `connect.endpoints` entry pointing at a plain `TcpListener` that   accepts the connection and then sends nothing
- open, then close
Close time tracks `transport.unicast.open_timeout` directly:
| `open_timeout` | open      | close     |
| -------------- | --------- | --------- |
| 10000 ms       | 509.95 ms | **9.50 s** |
| 2000 ms        | 504.61 ms | **1.50 s** |

### The trigger condition is narrower than "unreachable"
A **refused** connection returns immediately and the retry sleep that follows is already cancellable via `wait_next_peer_retry` — no hang. The hang needs an endpoint that **completes the TCP handshake but not the zenoh handshake**: a paused or frozen container (`podman pause` reproduces it in one command), a firewall that DROPs, or a wedged process. The distinction is important.

This is also why we think **#2675** ("session.close() blocks ~10s ... when the peer link has active bursty packet loss") is the same root cause reported from the user side — packet loss on an established link stalls the same await. 

### Severity bound
The stall does **not** stack across endpoints — three configured endpoints still cost one `open_timeout`, not three, because the connector tasks overlap. Worth stating so the bug is not read as worse than it is.

### One more site worth a look, not asserted
`orchestrator.rs:578` (`spawn_add_listener` → `add_listener_retry`) is also a plain `spawn` wrapping a retry loop. We have not shown it can hang and are not claiming it does — but it is the same shape, and if the two above are being changed it may be worth auditing at the same time.

AI disclaimer - much of this was found when I was doing work with an agent, and I had it investigate and write things up

PR #2744, opened 2026-08-18
fix(zenoh-ext): stop calling user code while holding subscriber state locks

YuanYuYuan:
## Summary

Two `zenoh-ext` subscribers call user code while they hold a non-reentrant `std::sync::Mutex`. A callback that re-enters the subscriber takes a mutex its own thread already holds. The thread then waits on itself.

This PR records each call under the guard. It replays the call after it releases the guard. `AdvancedSubscriber` and `QueryingSubscriber` both get the fix.

## What changes for a user

| a callback that … | before | after |
|---|---|---|
| undeclares or registers a sample-miss listener | **deadlocks** | works |
| calls `fetch` on a `FetchingSubscriber` | **deadlocks** | works |
| publishes to its own subscriber, conditionally | **deadlocks** | works |
| publishes to its own subscriber on *every* sample | deadlocks | livelocks — still its own recursion (**B5**) |
| panics while another thread stages a call | poisons the mutex, and the subscriber stops working | the subscriber stays usable (**B6**) |

The API does not change. Callbacks still never run concurrently. Samples still arrive in the order the state machine produced them.

## Why it deadlocks

```mermaid
sequenceDiagram
    autonumber
    participant App as Publishing thread
    participant Sub as zenoh-ext subscriber
    participant Core as Session

    App->>Sub: sample arrives
    activate Sub
    Sub->>Sub: zlock!(state)
    Sub->>App: user callback runs, guard held
    App->>Core: publish, fetch, or drop a handle
    Core->>Sub: re-enters on the same thread
    Sub--xSub: zlock!(state) — this thread holds it
    Note over Sub: the thread waits on itself
    deactivate Sub
```

Step 6 takes several shapes. `SampleMissListener::drop` and `SampleMissListenerBuilder::wait` take `statesref`. `FetchingSubscriber::fetch` resolves to `register_handler`, which takes `state`. A callback that publishes re-enters through the session's synchronous local delivery.

The `Drop` shapes reach a user unannounced. A user does not read "drop a value" as "re-enter the middleware".

## The fix

The state walk and the callout interleave, so the callout cannot simply move out of the locked region. Each state gains a FIFO outbox and a `delivering` marker. A thread stages its call under the guard. One thread claims the delivering role and drains the outbox with the guard released.

| | before | after |
|---|---|---|
| state walk | under the guard | under the guard, unchanged |
| the callout | **under the guard** | staged, then called with the guard released |
| delivery order | the guard serialised it | the FIFO preserves it |
| two threads delivering | the guard excluded them | the marker excludes them |
| a callback that re-enters | deadlocks | stages its work and returns |

Each staged call carries the callback handle it is for. An undeclare between the stage and the drain therefore cannot discard the sample.

The two files keep separate copies of this mechanism. `QueryingSubscriber` needs no condition variable. It is also deprecated, so sharing the code would tie a deprecated type's lifetime to the current one's.

## What fails without this

| control | tests | result |
|---|---|---|
| `main`, plus the test file | `AdvancedSubscriber` scenarios | fail: `Got Deadlocked.` |
| commit 1, plus the test file, `querying_subscriber.rs` untouched | 5 reentrancy tests | `3 passed, 2 failed` — the two new ones, both `Deadlocked` |

The second control keeps the first fix and removes only the second. The two new tests therefore pin the second commit.

> [!NOTE]
> Commits 3 to 5 have no failing baseline, and a test cannot honestly give them one. Each closes a race that needs an undeclare, a panic, or a stage to land inside a window bounded by one guard release and the next acquisition. A test that hit such a window would be timing-dependent. A test that missed it would read as proof of absence. The argument for each is the control flow, stated at the code.

## Behaviour changes

| | change | who is affected | action |
|---|---|---|---|
| **B1** | `wait_callbacks()` now blocks. It was a no-op | callers who relied on an immediate return | none, unless you depended on the no-op |
| **B2** | callbacks run with the state guard released | callbacks that observed the implicit exclusion | none: exclusion and ordering are preserved |
| **B3** | another thread may deliver your sample | callbacks that read thread-local state | audit callbacks for thread affinity |
| **B4** | the outbox is not backpressure | a producer faster than the callback | none today: the queue grows where the producer used to block |
| **B5** | an unconditional republish livelocks instead of deadlocking | callbacks that publish on every sample | make the republish conditional |
| **B6** | a panicking callback no longer poisons the state mutex | callbacks that panic | none: calls staged during the panic are discarded |

**B3 and B4 are the two to read.** The subscriber never promised a delivering thread, but one thread delivering another thread's sample is new. And the state mutex used to throttle a fast producer as a side effect; nothing bounds the queue now. A bound means dropping samples or restoring a block. Both are larger decisions than removing the deadlock, so this PR does not take them.

## Performance

| added, per delivered sample | note |
|---|---|
| two more uncontended acquisitions of the state mutex | stage, claim the role, then the empty-check that releases it |
| one `Arc` clone and drop | the callback handle each staged call carries |
| a `VecDeque` push, and a drain into the deliverer's buffer | no allocation in steady state: both buffers keep their capacity |
| one `notify_all` with no waiters | a futex check, not a syscall |

That is tens of nanoseconds on a path that has already crossed the network. Contention improves in the case that used to be worst: a slow callback no longer blocks the threads that arrive during it.

Heap residency is one slot per queued call. A slot holds ~280 bytes, because `Call::Sample` is larger than `Call::Miss`. The queue holds about one call while delivery keeps up. Queue depth under a faster producer is **B4**.

> [!NOTE]
> Both readings come from the code. No benchmark was run.

## Testing

`zenoh-ext/tests/reentrancy.rs` adds 5 scenarios. A timeout alone proves nothing, because "the callback ran and wedged" and "the callback never ran" look the same from outside. So:

- Each scenario asserts that it entered the callback.
- `run_scenario` returns `Completed`, `Deadlocked` or `Panicked`. It therefore never reports an assertion failure as a hang.
- `the_callback_guard_fires_when_the_callback_never_runs` is the positive control. It proves the guard executes.
- The scenarios disable scouting, gossip and listeners. They are single-process and need no network.

Full `zenoh-ext` suite: 30 passed, 0 failed. The checks on this page are the current CI state.

This PR adds one `#[allow]`, on the `Call` enum. `clippy::large_enum_variant` wants `Sample` boxed. Boxing it would put a heap allocation on the hot path to shrink a queue that is empty whenever delivery keeps up.

## Breaking changes

None. No public signature, type or name changes.

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/zenoh-ext/src/advanced_subscriber.rs
+++ b/zenoh-ext/src/advanced_subscriber.rs
@@ -12,7 +12,13 @@
 //   ZettaScale Zenoh Team, <zenoh@zettascale.tech>
 //
 use std::{
-    cmp::min, collections::BTreeMap, fmt, future::IntoFuture, hash::Hash, str::FromStr, sync::Weak,
+    cmp::min,
+    collections::{BTreeMap, VecDeque},
+    fmt,
+    future::IntoFuture,
+    hash::Hash,
+    str::FromStr,
+    sync::{Condvar, Weak},
     time::Instant,
 };
 
@@ -450,6 +456,18 @@ struct State {
     callback: Option<Callback<Sample>>,
     miss_handlers: HashMap<usize, Callback<Miss>>,
     token: Option<LivelinessToken>,
+    /// Calls into user code staged by the state machine, in the order it
+    /// produced them. Drained by `dispatch_outbox` with this mutex released.
+    outbox: VecDeque<Call>,
+    /// Which thread, if any, is currently draining `outbox`. Exactly one at a
+    /// time, which is what keeps delivery ordered and mutually excluded.
+    ///
+    /// The thread id — rather than a bool — lets `wait_for_delivery` avoid
+    /// waiting on itself, which a callback that undeclares from inside delivery
+    /// would otherwise turn into a self-deadlock.
+    delivering: Option<std::thread::ThreadId>,
+    /// Signalled when `delivering` clears. `wait_callbacks` blocks on it.
+    delivery_done: Arc<Condvar>,
     _gc_task: AbortOnDropHandle<()>,
 }
 
@@ -462,9 +480,14 @@ impl State {
         self.miss_handlers.insert(id, callback);
         id
     }
+    /// Removes a miss callback, returning it to be dropped by the caller.
+    ///
+    /// The caller drops it **after** releasing `statesref`: dropping a user
+    /// callback runs user `Drop` code, which is a call-out like any other.
     #[zenoh_macros::unstable]
-    fn unregister_miss_callback(&mut self, id: &usize) {
-        self.miss_handlers.remove(id);
+    #[must_use = "drop the callback outside the state lock"]
+    fn unregister_miss_callback(&mut self, id: &usize) -> Option<Callback<Miss>> {
+        self.miss_handlers.remove(id)
     }
 }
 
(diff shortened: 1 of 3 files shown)


YuanYuYuan:
Opened a follow-up on top of this branch (not a competing fix): https://github.com/YuanYuYuan/zenoh/pull/1 adds an opt-in, debug-time tripwire (`lock-tripwire`) that asserts `dispatch_outbox`'s call-out point never runs with a tracked lock held. Verified it actually catches the pre-fix defect: checked out this fix's parent commit with only the tracking wired in, and the existing reentrancy test panics in under a second naming the exact site instead of hanging. Happy to fold it into this PR instead if that's preferred, or keep it separate for review.

fuzzypixelz (commented):
My 2c: the overall shape of the proposed system in this patch looks like a sample/miss queuing system, queues I would prefer to see on top of the existing stable APIs, through the well-known channel or ring-buffer handlers for example.

So I would rather put some machinery in place to guard against such callback soup and the deadlocks it causes. I have to say I don't yet know exactly what this would look like: documentation? compile-time errors? runtime errors?

Still, I believe that the complexity we bear by abstracting over this is not worth it.

PR #2745, opened 2026-08-18
feat: add close_link method to TransportUnicast API

gabrik:
## Summary

Add `close_link(Link)` to the `TransportUnicastTrait` and expose it through the `TransportUnicast` public wrapper. This allows closing a specific link within a multilink transport without tearing down the entire transport.

## Changes

- **`TransportUnicastTrait`**: new `async fn close_link(&self, link: Link) -> ZResult<()>` on the trait
- **`TransportUnicastUniversal`**: delegates to existing `del_link()`
- **`TransportUnicastLowlatency`**: closes the entire transport (it has at most one link)
- **`MockTransportUnicastInner`**: no-op stub for tests
- **`TransportUnicast`**: public `close_link()` wrapper exposed with doc example

## Motivation

This is a prerequisite for transport link rotation (see #2746), where the rotation engine needs to close individual links within a multilink transport during make-before-break cycles. The existing `del_link` method on `TransportUnicastUniversal` was `pub(crate)` and not exposed through the trait or the public wrapper.

## Tests

- `transport_unicast_close_link_multilink`: closes one of two links, verifies transport survives, then closes the last link and verifies transport closes
- `transport_unicast_close_link_single_link`: closes the only link, verifies transport closes

---
## 🏷️ Label-Based Checklist

Based on the labels applied to this PR, please complete these additional requirements:

**Labels:** `new feature`

## 🆕 New Feature Requirements

Since this PR adds a new feature:

- [x] **Feature scope documented** - Clear description of what the feature does and why it's needed
- [x] **Minimum necessary code** - Implementation is as simple as possible, doesn't overcomplicate the system
- [x] **New APIs well-designed** - Public APIs are intuitive, consistent with existing APIs
- [x] **Comprehensive tests** - All functionality is tested (happy path + edge cases + error cases)
- [x] **Examples provided** - Usage examples in code comments or separate example files
- [x] **Documentation added** - New docs explaining the feature, its use cases, and API
- [x] **Feature flag considered** - Can the feature be enabled/disabled for gradual rollout?
- [x] **Performance impact assessed** - Memory, CPU, storage implications measured
- [x] **Integration tested** - Feature works with existing features

**Consider:** Can this feature be split into smaller, incremental PRs?

**Instructions:**
1. Check off items as you complete them (change `- [ ]` to `- [x]`)
2. The PR checklist CI will verify these are completed

*This checklist updates automatically when labels change, but preserves your checked boxes.*

PR #2746, opened 2026-08-18
feat: transport link rotation engine

gabrik:
## Summary

Implement transport link rotation: a middleware-level capability to automatically close and re-establish transport links on a configurable schedule, enabling cloud-friendly connection cycling.

## Design

- **Make-before-break**: open new link before closing old one — no connectivity gap, no redeclaration storm
- **Fallback to break-before-make**: only if make fails after all retries, as a last resort
- **Interval-only policy**: the RotationPolicy enum has only Interval initially; additional policies can be added later
- **Configurable**: per-endpoint and global, with jitter to avoid synchronized rotation across clients

### Config (commons/zenoh-config)
- New `rotation.rs` module with RotationConf, RotationPolicyConf, RotationModeConf, RotationFallbackConf
- Wired into ConnectConfig
- Documented in DEFAULT_CONFIG.json5

### Rotation Engine (zenoh/src/net/runtime/rotation.rs)
- RotationEngine: periodic timer per endpoint, make-before-break with retry+fallback
- Hooked into spawn_peer_connector() in orchestrator

### Transport Layer
- Uses close_link() API from PR #2745 to close individual links

## Configuration Example

```json5
connect: {
  rotation: {
    enabled: true,
    policy: {
      type: "interval",
      interval_ms: 300000,  // 5 minutes
      jitter_ms: 30000,     // ±30s
    },
    mode: "make_before_break",
    fallback: {
      enabled: true,
      max_retries: 3,
      retry_backoff_ms: 1000,
    },
    rotate_across_locators: true,
  }
}
```

## Depends on
- #2745 (transport: add close_link method to TransportUnicast API)

## Investigation Document
Full investigation and design rationale in `docs/transport_rotation_investigation.md`

---
## 🏷️ Label-Based Checklist

**Labels:** `new feature`

## 🆕 New Feature Requirements

**Instructions:**


fuzzypixelz (commented):
(I haven't reviewed the code; but I have some questions on the design. I will however throw Copilot at the diff.)

> [..] enabling cloud-friendly connection cycling.

Could you please elaborate on this, perhaps with a concrete use-case?

> **Fallback to break-before-make**: only if make fails after all retries, as a last resort

Why should we assume that a given connection attempt would fail because of make-before-break?
 
If we know a priori that make-before-break is incompatible with a certain endpoint, then why go through retry loops? If we cannot know a priori that break-before-make is the solution to connection failure, it seems wrong to speculate and risk destroying the existing while _still_ being unable to re-connect.

> Full investigation and design rationale in docs/transport_rotation_investigation.md

I don't think this document fits within the codebase, perhaps it is better supplied as a file attachment to the pull request description. I understand that it's perhaps useful agentic context, but this codebase is yet to even gain an `AGENTS.md` file :)

gabrik:
> (I haven't reviewed the code; but I have some questions on the design. I will however throw Copilot at the diff.)
> 
> > [..] enabling cloud-friendly connection cycling.
> 
> Could you please elaborate on this, perhaps with a concrete use-case?


Two concrete cloud-related use-cases:
- Ingress controllers in the cloud have limited connection lifetime, because of updates, scale ups/scale downs those ingress controller need to bound the maximum duration of any connection going through them. Thus long-lived TCP (or TLS) connections do not suit could environment. Just to give more context on what it is and how usually and ingress controller works, the ingress controller is usually a proxy, thus all connections go through it, then it takes care of contacting the actual destination. Such destinations are usually reachable only via private IP addresses, and thus the client could not connect directly to them. This approach is used to have a few public IPs, while being able to use a high number of services and replicas internally. 
- Routers in the cloud must be considered ephemeral, they will not live forever, they will be created and destroyed based on the traffic and state of the cluster. It also in general with load-balancing as at each re-connection can end up into a different router.

While UDP and QUIC could be an approach they are still not well supported in Cloud infrastructures.

> > **Fallback to break-before-make**: only if make fails after all retries, as a last resort
> 
> Why should we assume that a given connection attempt would fail because of make-before-break?
> If we know a priori that make-before-break is incompatible with a certain endpoint, then why go through retry loops? If we cannot know a priori that break-before-make is the solution to connection failure, it seems wrong to speculate and risk destroying the existing while _still_ being unable to re-connect.

Because of the `max_links` configuration, if it is set by default at 1 then the make-before-break will fail, if the connection attempt ends up on the same router.
We could skip that rotation if the make-before-break fails, that sounds a reasonable approach.


> > Full investigation and design rationale in docs/transport_rotation_investigation.md
> 
> I don't think this document fits within the codebase, perhaps it is better supplied as a file attachment to the pull request description. I understand that it's perhaps useful agentic context, but this codebase is yet to even gain an `AGENTS.md` file :)
Will clean up!

@fuzzypixelz

fuzzypixelz:
> Two concrete cloud-related use-cases:
> [..]
> While UDP and QUIC could be an approach they are still not well supported in Cloud infrastructures.

All very reasonable.

> Because of the `max_links` configuration, if it is set by default at 1 then the make-before-break will fail, if the connection attempt ends up on the same router.
We could skip that rotation if the make-before-break fails, that sounds a reasonable approach.

Especially since `MAX_LINKS` is a known `Close` reason, so it can be handled as such:

https://github.com/eclipse-zenoh/zenoh/blob/c5d4760e33848be5fc1e4fa76a81fb87799f7379/commons/zenoh-protocol/src/transport/close.rs#L21-L31

gabrik:
@fuzzypixelz Thanks, that makes sense. I traced the connection-establishment path: the peer’s `MAX_LINKS` close reason is recognized when receiving a `Close` during both the [`InitSyn` exchange](https://github.com/eclipse-zenoh/zenoh/blob/bfc6e1b/io/zenoh-transport/src/unicast/establishment/open.rs#L255-L265) and the [`OpenSyn` exchange](https://github.com/eclipse-zenoh/zenoh/blob/bfc6e1b/io/zenoh-transport/src/unicast/establishment/open.rs#L525-L535).

However, the reason is currently only used for logging and then discarded when the generic error is returned. The manager ultimately returns only `Err(e)` from its [`InitTransportError` handling](https://github.com/eclipse-zenoh/zenoh/blob/bfc6e1b/io/zenoh-transport/src/unicast/manager.rs#L825-L836), while `ZResult` carries a boxed generic error rather than a structured transport-close reason ([definition](https://github.com/eclipse-zenoh/zenoh/blob/bfc6e1b/commons/zenoh-result/src/lib.rs#L77-L85)).

PR #2755, opened 2026-08-24
Fix gossip scouting for loopback locators

### LINKED ISSUE #2753: Issue with peer-to-peer discovery in 1.10.0
### Describe the bug

I'm having some issues with peer-to-peer discovery through the router after upgrading to 1.10.0 in ROS2 on localhost. Launching the router, a node, and running a CLI command such as "ros2 node list" discovers nothing. The same domain id is used.

Things I've been able to verify:
- Launching the node as "client" instead of "peer" makes it discoverable.
- Running the router with debug info shows activity when the node launches and ros2 CLI is executed, so they both find the router

Possibly related to #2603 ?

### To reproduce

1. ros2 run rmw_zenoh_cpp rmw_zenohd
2. ros2 launch <node>
3. ros2 node list

### System info

- Platform: Archlinux
- Zenoh 1.10.0
- ROS2 kilted / ROS2 lyrical

fuzzypixelz:
## Description

This patch fixes a regression introduced in #2671: loopback locators were no longer being sent during gossip scouting.

### What does this PR do?

This patch restores the old scouting behavior where loopback locators were filtered out only if they originate in unspecified endpoints.

I've also caught an inadvertant change of behavior in `SessionInfo::locators` introduced in #2671, where said method started returning the aforementioned loopback locators originating in unspecified endpoints.

### Related Issues

Fixes https://github.com/eclipse-zenoh/zenoh/issues/2753.

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/zenoh/src/api/info.rs
+++ b/zenoh/src/api/info.rs
@@ -121,7 +121,7 @@ impl SessionInfo {
     /// ```
     #[zenoh_macros::unstable]
     pub fn locators(&self) -> impl Resolve<Vec<Locator>> + '_ {
-        ResolveClosure::new(|| self.session.runtime().get_locators())
+        ResolveClosure::new(|| self.session.runtime().get_locators_noloopback())
     }
 
     /// Return information about currently opened transport sessions. Transport session is a connection to another zenoh node.
--- a/zenoh/tests/scouting.rs
+++ b/zenoh/tests/scouting.rs
@@ -22,6 +22,8 @@ use zenoh::config::WhatAmI;
 use zenoh::sample::SampleKind;
 use zenoh_config::{Config, ModeDependentValue, WhatAmIMatcher};
 use zenoh_link::EndPoint;
+#[cfg(all(feature = "unstable", feature = "transport_tcp"))]
+use zenoh_test::get_free_tcp_port;
 use zenoh_test::get_free_udp_port;
 
 #[tokio::test(flavor = "multi_thread")]
--- a/zenoh/tests/scouting.rs
+++ b/zenoh/tests/scouting.rs
@@ -22,7 +22,7 @@ use zenoh::config::WhatAmI;
 use zenoh::sample::SampleKind;
 use zenoh_config::{Config, ModeDependentValue, WhatAmIMatcher};
 use zenoh_link::EndPoint;
-use zenoh_test::get_free_udp_port;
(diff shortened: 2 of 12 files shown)


zenoh/tests/scouting.rs
```diff
@@ -22,7 +22,7 @@ use zenoh::config::WhatAmI;
 use zenoh::sample::SampleKind;
 use zenoh_config::{Config, ModeDependentValue, WhatAmIMatcher};
 use zenoh_link::EndPoint;
-use zenoh_test::get_free_udp_port;
+use zenoh_test::{get_free_tcp_port, get_free_udp_port};
```
JEnoch: Clippy (in CI) complains `get_free_tcp_port` is never used. This use should be feature-gated in the same way than the test.

zenoh/src/api/info.rs
```diff
@@ -121,7 +121,7 @@ impl SessionInfo {
     /// ```
     #[zenoh_macros::unstable]
     pub fn locators(&self) -> impl Resolve<Vec<Locator>> + '_ {
-        ResolveClosure::new(|| self.session.runtime().get_locators())
+        ResolveClosure::new(|| self.session.runtime().get_locators_noloopback())
```
JEnoch, line 124: Why removing the loopback from the list returned by `session.info().locators()` ?
If the Session is configured to listen on `tcp/0.0.0.0:7447` I would expect to see it's actually listening on `tcp/127.0.0.1:7447`.

JEnoch (changes requested):

zenoh/src/api/info.rs
fuzzypixelz: This was the behavior before https://github.com/eclipse-zenoh/zenoh/pull/2671, in this patch was focus was on restoring the pre-existing behavior. I agree that it's perhaps for the best to include all locators here, but I felt that such a discussion was beyond the scope of this patch. I am a bit uneasy about the unspecified loopback locator filtering, so I was conservatively aimed to keep things as is until there is consensus on how to deal with this quirk.

JEnoch: I didn't realize it was the former behaviour. Then I agree we shall discuss this aside.

JEnoch (approved):

PR #2757, opened 2026-08-27
feat(link): add CAN FD multicast and ISO-TP unicast links

jerry73204:
### What does this PR do?

Adds two CAN link implementations, both off by default behind new Cargo
features, plus one small extension to the multicast link trait that lets a
medium carry zenoh's QoS onto the wire.

**`transport_can`** -- `zenoh-link-can`, a **multicast** link over a SocketCAN
interface. Endpoint grammar `can/<iface>#bitrate=..;dbitrate=..`. A CAN frame
names a message rather than a destination and every node hears every frame,
which is what a zenoh multicast transport already expects, so a link joins
every other peer on the same bus.

**`transport_isotp`** -- `zenoh-link-isotp`, a **unicast** link over ISO 15765-2.
Endpoint grammar `isotp/<iface>#tx_id=0x..;rx_id=0x..`. The directed identifier
pair is the address. Implemented on the kernel's `CAN_ISOTP` protocol, so
segmentation and flow control happen in-kernel and the link deals in whole
datagrams.

The two share the bus and can run on one peer at once; they are separate link
types because they are separate services.

CAN is the field bus in vehicles and robots. Reaching a zenoh network from
those nodes means either a gateway box or a link, and this is the link.

The multicast link alone is not enough, which is why both are here. zenoh
routes queries and liveliness tokens to unicast faces only -- `mcast_groups`
appears in the pubsub half of each hat and never in `queries.rs` or `token.rs`.
Over a multicast-only link a peer can publish and subscribe and nothing more,
so anything built on a query silently does not work. ISO-TP supplies a real
unicast face on a bus that has no addresses, and lifts the MTU from one frame
to 4095 bytes as a side effect.

### Design notes worth a reviewer's attention

**CAN FD only, deliberately.** Classic CAN is refused at parse rather than
warned about: a 7-byte payload is smaller than zenoh's per-fragment overhead,
so a classic link cannot make progress at all. It is not a slower link, it is a
non-functional one, and carrying the path would have meant shipping a mode
whose symptom is a hang. (The ISO-TP link has no such limit -- segmentation is
below zenoh there, so classic CAN works fine over it.)

**Priority lives in the frame identifier.** CAN arbitrates by identifier,
lowest wins, so putting zenoh's priority in the high bits makes the bus resolve
collisions in zenoh's favour -- a control message beats a bulk transfer on the
wire rather than behind it. That is the one thing this link can do that a
socket cannot, so it needs the priority of the batch being written.

**Flow control is configurable, because the peer may be small.** The ISO-TP
endpoint takes `stmin` and `bs`, the two knobs ISO 15765-2 gives a receiver:
how long the peer must leave between ConsecutiveFrames, and how many it may
send before waiting for another FlowControl. Left unset the kernel says "send
it all as fast as you like", which is right facing another Linux box and wrong
facing a node with one frame of buffer. Each is one octet of the FlowControl
frame, so a wider value is refused at parse rather than truncated.

**The one trait change.** `LinkMulticastTrait` gains
`write_all_with_priority`, which **defaults to `write_all` and discards the
priority**. No existing link implementation changes. The internal
`send_batch` in `io/zenoh-transport/src/multicast/link.rs` grows a `Priority`
argument to feed it; the transmission pipeline already batches per priority, so
a batch belongs to exactly one and this is not an average over mixed traffic.

### Footprint

29 files, +4657/-16. Two new crates account for nearly all of it; existing
code sees ~105 lines, mostly feature plumbing and enum arms.

The PR is **two commits, one per link**, so the CAN link can be taken and
ISO-TP deferred. The first commit adds no ISO-TP files at all.

### Testing

Tests live in `io/zenoh-transport/tests/{multicast_can,unicast_isotp}.rs` and
run on a virtual bus; `ci/vcan-setup.sh` brings one up.

They are `#[ignore]`-gated because they need a `vcan0` interface, so **CI here
will report them as ignored and effectively test nothing**. Run locally with:

```
sudo ./ci/vcan-setup.sh
cargo test --release --features transport_can,transport_isotp \
  -p zenoh-transport --test multicast_can --test unicast_isotp -- --ignored
```

11/11 pass. Coverage: every DLC boundary, transport-driven fragmentation, a
payload well past one batch, a **four-peer** case (each peer tracks three
remotes told apart only by frame identifier -- a bug in peer-address derivation
is invisible with two peers and obvious with four), priority-major identifier
layout, a datagram larger than one ISO-TP frame, a listener accepting more than
one client in sequence, and **two independent identifier pairs transferring
concurrently** on one bus.

Plus `stmin` and `bs` reaching the wire: 1024 bytes at `stmin=1ms` is about 147
ConsecutiveFrames and cannot complete in under 50ms, where unpaced it takes
12ms. These are kernel state-machine behaviour rather than controller
behaviour, so they are measurable on `vcan` and this is a real test rather than
a simulation.

The concurrency test asserts exact message counts per pair, so a socket
accepting the other pair's frames or a shared reassembly buffer fails it. Verified against a
negative control: pointing both pairs at the same identifiers makes it fail
with `pair 0 received 35 messages, expected exactly 20`. The payload is larger
than one frame, so both transfers segment and interleave; `candump` shows 1510
frames on each pair with 12 FirstFrames and 12 FlowControls.

Also verified interoperating with zenoh-pico over the same bus, and end to end
under ROS 2 via rmw_zenoh -- services, actions, action cancellation, parameters
and graph introspection all cross a CAN bus with no router and no TCP.

**Not covered:** anything about timing or arbitration. `vcan` has no bit rate
and queues frames rather than contending for the bus, so `STmin`, `BS` and the
priority-in-identifier mapping cannot be exercised on a virtual bus. The
concurrency test above shows the addressing and per-socket state are
independent; it says nothing about two peers actually colliding on a wire.

### Companion PRs

- eclipse-zenoh/zenoh-pico#1298 -- the same two links in zenoh-pico. Independent
  of this PR; the two interoperate on one bus but neither needs the other to
  build.
- eclipse-zenoh/zenoh-c#1322 -- forwards the two features to zenoh-c. **Depends
  on this PR** and does not build until it lands.


jerry73204:
Two things this PR needs from a maintainer before CI can give a verdict:

- **Workflow approval.** As a first-time contributor the CI workflows need
  "Approve and run".
- **A label.** `Generate / main` requires one of `api-sync`,
  `breaking-change`, `bug`, `ci`, `dependencies`, `documentation`,
  `enhancement`, `new feature`, `internal`. I do not have permission to add
  one -- `new feature` looks right.

Checked locally against what CI runs, on **both** commits independently so the
split stays bisectable:

- `rustfmt +nightly --check --config "unstable_features=true,imports_granularity=Crate,group_imports=StdExternalCrate,skip_children=true" $(git ls-files '*.rs')` -- clean
- `cargo +stable clippy --all-targets --all-features -- --deny warnings` -- clean
- `cargo build --release --features transport_can` on the first commit alone -- builds without the ISO-TP crate present

One thing worth flagging about test coverage: the tests in
`io/zenoh-transport/tests/{multicast_can,unicast_isotp}.rs` are `#[ignore]`-gated
because they need a `vcan0` interface, so **CI will report them as ignored and
effectively test nothing**. `ci/vcan-setup.sh` brings the interface up; with it,
9/9 pass locally. If you would prefer that CI exercise them, adding
`sudo ./ci/vcan-setup.sh` plus `-- --ignored` to a Linux job would do it --
`vcan` is a kernel module with no hardware requirement. I did not want to touch
the CI workflow uninvited.

jerry73204:
CI has run now. Triage of the five failures -- none of them originate here:

| Check | Cause |
|---|---|
| `Lints and doc tests` (ubuntu / macOS / windows) | Pre-existing upstream lint |
| `Coverage (ubuntu-latest, nightly)` | Runner aborted |
| `Generate / main` | Missing label |

**Lints.** The step is `Clippy zenoh no-default-features`, and the error is:

```
error: redundant reference in `format!` argument
   --> zenoh/src/net/runtime/adminspace.rs:772:19
    = note: `-D clippy::useless-borrows-in-formatting` implied by `-D warnings`
```

`adminspace.rs` is not touched by this PR -- the file is byte-identical to the
base commit. #2758 fails at the same file and the same line, so a newer stable
clippy on the runners has started flagging existing code. It will need a
one-line fix on `main` (`version = *LONG_VERSION`); happy to open that
separately if it would help, but it does not belong in this PR.

**Coverage.** `##[error]The runner has received a shutdown signal ... exit code
143` -- the job was terminated by infrastructure, no test failed. A re-run
should clear it.

**Generate / main.** Needs one of the required labels, which I lack permission
to add.

What I *can* verify, run against both commits independently:

- CI's exact `rustfmt +nightly --check --config "..."` invocation -- clean
- `cargo +stable clippy --all-targets --all-features -- --deny warnings` -- clean
- the first commit builds on its own with `--features transport_can`, without
  the ISO-TP crate present

PetervdPerk-NXP:
Hi Jerry,

Thank you for exploring CAN transport ideas for zenoh, and for testing it end to end with zenoh-pico and ROS 2. I fully agree with CAN FD only.

Since ROS 2 over rmw_zenoh is the goal, I looked at the design from that side. Short version: I think one link can do both multicast and unicast, and I would not add ISO-TP at all. Details below.

One CAN fact first, for readers who do not work with CAN every day. Every frame starts with a number, the identifier. When two nodes start sending at the same moment, the lower number goes first and the other one waits. There is no round-robin. So the identifier is three things at once: the priority, the address, and the value a microcontroller programs into its CAN controller hardware filter so that frames it does not need never reach the CPU.

### What ROS 2 puts on the bus

- Every message from rmw_zenoh carries a 35-byte attachment inside the `Put` (sequence number, source timestamp, 16-byte GID, plus the extension header). A `geometry_msgs/Twist` is 52 bytes of CDR, so the `Put` is 89 bytes before any zenoh framing.
- Topic keys are `<domain>/<topic>/<type>/<type_hash>`, more than 100 characters. rmw_zenoh declares them when it creates a publisher, so on the wire they are a 1-byte `ExprId`. Good.
- Services, actions, parameters and graph discovery are queries and liveliness tokens. Transient-local publishers also add a cache queryable and a liveliness token. Zenoh routes queries and liveliness to unicast faces only, never to a multicast group. **So for ROS 2 the multicast link alone gives topics and nothing else. A unicast face is required.**

### Please not ISO-TP

ISO-TP is a diagnostics protocol for pushing one long message through 8-byte frames: a FirstFrame, a FlowControl answer from the receiver, then ConsecutiveFrames. For ROS 2 traffic everything about it works against us:

- **A `Twist` becomes 15 classic frames and about 4 ms.** On your multicast link it is 2 CAN FD frames and about 0.7 ms. An `Imu` (328 bytes of CDR with an 8-character frame id) is 55 frames and about 15 ms, against 7 frames and about 2.8 ms.
- **Every service call is two PDUs, each with its own FirstFrame and FlowControl round trip.** Only one PDU can be in flight per socket, so request and reply wait for each other. One lost frame stops the pair for a full second.
- **The addressing does not scale.** Every pair of peers needs its own static `tx_id`/`rx_id`. With 8 priority classes, 16 nodes already use 94 % of the 11-bit identifier space, and a listener can serve only one client at a time.
- **No topic in the identifier.** A microcontroller cannot filter in hardware. Everything for that peer arrives on one identifier and must be sorted in software after reassembly.
- **The write blocks the whole process.** With `CAN_ISOTP_WAIT_TX_DONE` the kernel makes `write()` sleep until the last frame of the PDU is on the bus, even on a non-blocking socket. Zenoh's TX runtime has one worker thread by default, so a 15 ms `Imu` PDU stops every link in the process, keepalives included.
- Its one real advantage is classic 8-byte CAN hardware, and I see that this is deliberate. Do you have a target that needs classic CAN? If not, I do not think ISO-TP earns its place.

Unicast does not need a second protocol. Your multicast link already puts the sender in the identifier. A frame that also carries a destination node is a unicast frame. The same crate can offer the same socket and wire format through both link traits, the way `udp` already does. A peer's node id is learned from the `Join` it already sends. That is how DroneCAN/Cyphal, J1939 and 6LoCAN do unicast on a bus: source and destination in the identifier, one id per node.

So my suggestion: land the multicast link on its own, drop ISO-TP, and do unicast as below.

### Don't limit the MTU, so we can optmize overhead

**Let the link segment, and report a 4095-byte MTU.** A Cyphal-style tail byte (start and end bits plus a small counter in the last byte of each frame) carries a datagram of any length over several frames, with no flow control and no round trip. The multicast transport takes its batch size from the link MTU. So the moment the link reports 4095 instead of 63:

- the `Join` with the QoS extension (99 bytes, which your own test shows does not fit today) fits, so multicast QoS can be switched on;
- `InitAck` (about 75 bytes with its cookie) fits, which is what the unicast face needs for its handshake;
- your `Fragment` path only runs above 4095 bytes.

Nothing above the link changes. Every peer, zenoh-pico included, has to agree on the batch size, as your README already says.

### Make priority real

`write_all_with_priority` is the right hook, and you already pass the batch priority into it. What makes it inert today is that without multicast QoS the transport always hands the link the default priority. The MTU change above fixes that. Until then the order on the bus is the peer number, not the priority: a background burst from peer 0x100 goes ahead of a control message from peer 0x101, and a 4 KiB message is 71 frames, about 29 ms of that.

### Put the topic in the identifier, not only the sender

`ExprId`s are per face, but every frame on the bus already names its sender. So `(source node, sender's ExprId)` is unique on the bus with no registry and no routing change: the receiving side already keeps, per remote peer, the key ids that peer declared, so a rebuilt `Push` resolves as it is. The link only reads `Push` and `WireExpr`; it never parses `DeclareKeyExpr` and keeps no key table. A subscriber can then filter by topic in hardware, which is what a microcontroller with a few ROS 2 subscriptions on a busy bus wants most.

The part that makes this work for ROS 2: **the `Put` travels byte for byte as zenoh encodes it**, header, attachment, timestamp, encoding and payload. Only the `Frame` header, the SN, the `Push` header and the `WireExpr` are removed, so the rmw attachment simply rides along.

image

image


(CAN FD frames only come in fixed sizes, 8, 12, 16, 20, 24, 32, 48 or 64 bytes, which is why the link already writes a length byte.) To be clear about what this buys for ROS 2: about 11 % less bus time per message, because the rmw attachment is the biggest part either way. The point is not bytes. The point is topic filtering in hardware, priority in the identifier, and unicast on the same frames.

Header by header, this is what moves where. The receiving link rebuilds the batch before it hands it to the transport, so nothing above the link changes:

| your data field today | in the proposal |
|---|---|
| `Frame` header | implied: an 11-bit data frame *is* one `Push` + `Put` on a declared key; one tail-byte bit carries the `R` flag |
| 5-byte SN | the receiving link creates last+1 per sender; the tail-byte counter lets it resync after a lost frame |
| `Push` header + `WireExpr` | key id in the identifier; the link re-emits `WireExpr { scope: id, mapping: Sender }` |
| `Put`, byte for byte | unchanged, attachment included |
| QoS extension (priority) | top 2 bits of the identifier |
| `Fragment` header per frame | tail-byte start and end bits; the link segments to 4095 itself |
| fall-backs | undeclared key or a non-empty `WireExpr` suffix, several messages in one batch, `ExprId` above 31, `Push`-level timestamp or node id: sent as a normal batch with all headers |

An 11-bit layout that fits is `prio(2) | key id(5) | src(4)`: key id above source, so one identifier range covers a topic from every publisher. The source has to be there, because CAN does not allow two nodes to send different data under one identifier, and the receiving transport keys the peer on it. 29-bit identifiers stay for frames that need a destination node or the full 16-bit `ExprId`; they cost about 46 µs more per frame at 500 kbit/s, so plain topic data stays 11-bit. This would be a new format that a peer advertises, with today's format as the fallback, so a zenoh-pico build without it never sees a compact frame.

Peter

PR #2759, opened 2026-08-29
fix: clear pending connection for empty scout locators

stan603-source:
### What does this PR do?

Clears a scouted peer from `pending_connections` when all advertised locators are already covered by configured endpoints. A genuinely empty scouted-locator list now returns before inserting pending state. Regression tests cover both cases.

`Runtime::connect` inserts the peer ZID before filtering locators. The empty-list branch returned without the cleanup performed by the normal path, leaving that ZID permanently pending. Later multicast Hellos with a valid locator were repeatedly rejected as `Already connecting ... Ignore.`

This was reproduced with ten peer-mode nodes under a 256 kbit/s constrained link: D-CRDT state converged but the direct TLS graph remained at 44/45 edges after 75 seconds. With this cleanup, three repeated trials reached 45/45 edges with zero runtime errors.

### Validation

- `cargo fmt --all -- --check`
- `cargo test -p zenoh configured_scouted_locator_does_not_leave_connection_pending --no-default-features`
- `cargo test -p zenoh empty_scouted_locators_do_not_leave_connection_pending --no-default-features`
- `cargo clippy -p zenoh --all-targets --no-default-features -- --deny warnings`
- External 10-node/256 kbit/s formation harness: 3/3 exact 45-edge cliques

### Related Issues

Related to #2753.

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**

### DIFF
--- a/zenoh/src/net/runtime/orchestrator.rs
+++ b/zenoh/src/net/runtime/orchestrator.rs
@@ -1086,6 +1086,10 @@ impl Runtime {
     /// Returns `true` if a new Transport instance is established with `zid` or had already been established.
     #[must_use]
     async fn connect(&self, zid: &ZenohIdProto, scouted_locators: &[Locator]) -> bool {
+        if scouted_locators.is_empty() {
+            return false;
+        }
+
         if !self.insert_pending_connection(*zid).await {
             tracing::debug!("Already connecting to {}. Ignore.", zid);
             return false;
@@ -1484,8 +1488,21 @@ mod tests {
         let zid = ZenohIdProto::rand();
 
         assert!(!runtime.connect(&zid, &[]).await);
-        assert!(runtime.insert_pending_connection(zid).await);
-        runtime.remove_pending_connection(&zid).await;
+        assert!(!runtime.remove_pending_connection(&zid).await);
+    }
+
+    #[tokio::test]
+    async fn configured_scouted_locator_does_not_leave_connection_pending() {
+        let mut config = Config::default();
+        config
+            .insert_json5("connect/endpoints", r#"["tcp/localhost:12345"]"#)
+            .unwrap();
+        let runtime = RuntimeBuilder::new(config).build().await.unwrap();
+        let zid = ZenohIdProto::rand();
+        let locator = "tcp/localhost:12345".parse().unwrap();
+
+        assert!(!runtime.connect(&zid, &[locator]).await);
+        assert!(!runtime.remove_pending_connection(&zid).await);
     }
 
     #[tokio::test(flavor = "multi_thread")]
--- a/zenoh/src/net/runtime/orchestrator.rs
+++ b/zenoh/src/net/runtime/orchestrator.rs
@@ -1116,6 +1116,7 @@ impl Runtime {
                 "Already connecting to locators of {} (connect configuration). Ignore.",
                 zid
             );
+            self.remove_pending_connection(zid).await;
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


stan603-source:
Could a maintainer please apply the `bug` label? The repository checklist workflow requires a classification label, but external contributors do not have label permission.

JEnoch:
Hi @stan603-source,
Thank you for spotting this!
Could you please also sign the Eclipse Contribution Agreement (ECA) here: https://api.eclipse.org/git/eca/status/gh/eclipse-zenoh/zenoh/2759
You'll need to use the same email address than used for your Github account.

zenoh/src/net/runtime/orchestrator.rs
```diff
@@ -1472,6 +1473,20 @@ mod tests {
     use tokio::time::{timeout, Duration};
 
     use super::*;
+    use crate::{net::runtime::RuntimeBuilder, Config};
+
+    #[tokio::test]
+    async fn empty_scouted_locators_do_not_leave_connection_pending() {
```
JEnoch, line 1483: A empty scouted locators list is very unlikely to happen.
What is more likely to happen and should be tested is a configuration with a `connect/endpoints` containing a locator (e.g. `tcp/localhost:12345`) and then the scouted locators list containing this same locator.

Could you please add such a test?

zenoh/src/net/runtime/orchestrator.rs
```diff
@@ -1472,6 +1473,20 @@ mod tests {
...
+        let runtime = RuntimeBuilder::new(Config::default())
+            .build()
+            .await
+            .unwrap();
+        let zid = ZenohIdProto::rand();
+
+        assert!(!runtime.connect(&zid, &[]).await);
+        assert!(runtime.insert_pending_connection(zid).await);
+        runtime.remove_pending_connection(&zid).await;
```
JEnoch: `remove_pending_connection()` is also returning a `bool` that should be asserted as well.

zenoh/src/net/runtime/orchestrator.rs
JEnoch, line 1093: I believe that an empty `scouted_locators` list received as input should be treated differently than the filtered `locators` list ending up as empty: the debug log "Already connecting to locators of ..." makes no sense if the remote session has no scouted locator.

If `scouted_locators` is empty, the function should directly return false without any log (or maybe at trace level, if there are no previous log showing this list is empty)

JEnoch (changes requested):

JEnoch:
Also please include the latest commits from `main` in your branch, to include recent CI fixes regarding Coverage and Lints.

stan603-source:
Thanks @JEnoch — addressed the review requests:

- rebased the branch onto the latest `main`;
- added a regression test where `connect/endpoints` and the scouted locator both contain `tcp/localhost:12345`;
- asserted the `remove_pending_connection()` result to prove no marker remains; and
- separated the genuinely empty `scouted_locators` case so it returns before inserting pending state and emits no misleading configured-endpoint log.

Validation on the rebased branch:

```text
cargo fmt --all -- --check
cargo test -p zenoh configured_scouted_locator_does_not_leave_connection_pending --no-default-features
cargo test -p zenoh empty_scouted_locators_do_not_leave_connection_pending --no-default-features
cargo clippy -p zenoh --all-targets --no-default-features -- --deny warnings
```

All pass locally. The configured/scouted-overlap test was also run once with the cleanup line temporarily removed; it failed on the `remove_pending_connection()` assertion as intended, confirming that it exercises the regression rather than only the return value.

JEnoch (approved):

PR #2761, opened 2026-08-31
fix(uring): collapse an else-if that Clippy flags

YuanYuYuan:
`clippy::collapsible_else_if` in the io_uring reader's completion handler:

```
commons/zenoh-uring/src/linux/api/reader/mod.rs:331:16:
    warning: this `else { if .. }` block can be collapsed
```

The `else { if let .. }` becomes `else if let ..`. The block moves left by four columns and one brace goes away. Nothing else changes.

## What fails without this

Nothing in CI, and that is worth stating plainly rather than leaving a reader to assume otherwise.

| toolchain | reports it |
|---|---|
| Clippy 0.1.93 | yes |
| Clippy 1.98, which CI pins | no |

`Clippy all features` is green either way. This is a portability fix across toolchains, so it has **no failing baseline** and cannot be given one under the toolchain CI uses.

## How it was found

While sweeping the full CI Clippy matrix without `--deny warnings`, so that every lint reported instead of the run stopping at the first error. It appeared under `--features uring`, `--features test`, `--features unstable,test` and `--all-features`, on a commit that contained no local changes — so it belongs to the tree, not to the branch it was found from.

## Breaking changes

None.

---
## 🏷️ Label-Based Checklist

Current labels: _No labels_

### DIFF
--- a/commons/zenoh-uring/src/linux/api/reader/mod.rs
+++ b/commons/zenoh-uring/src/linux/api/reader/mod.rs
@@ -328,28 +328,26 @@ impl Reader {
                     bail!("Task-related uring error: {}, task {:?}", unexpected, index);
                 }
             }
-        } else {
-            if let Some(buf_id) = io_uring::cqueue::buffer_select(e.flags()) {
-                #[cfg(feature = "uring_trace")]
-                tracing::trace!("Read multishot entry: {:?}", e);
+        } else if let Some(buf_id) = io_uring::cqueue::buffer_select(e.flags()) {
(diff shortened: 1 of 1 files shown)


YuanYuYuan:
Closing: no failing baseline. The lint is reported by Clippy 0.1.93 and not by the 1.98 toolchain CI pins, so `Clippy all features` is green either way. It surfaced from a local toolchain mismatch rather than from anything in this tree, and the re-indent is not worth the review cost on its own.

PR #2766, opened 2026-09-03
feat: make loading the WebPKI root trust store optional.

### LINKED ISSUE #2764: When using TLS/QUIC, allow for not trusting the WebPKI Roots.
### Describe the feature

Zenoh unconditionally loads the WebPKI roots into its trust store for TLS/QUIC. https://github.com/eclipse-zenoh/zenoh/blob/main/io/zenoh-links/zenoh-link-tls/src/utils.rs#L349

For my project, *I* am *the* authority on what something is or is not named, and I only want the system to look at my CA. I don't care what Digicert says or does....

Would it be reasonable to make loading the default trust store configurable?

jpace121:
### What does this PR do?
Adds  a new configuration parameter "use_public_pki" to control whether the root trust store for TLS and QUIC connections will load the default WebPKI root certs or not.

The value defaults to true, maintaining the behavior in main today.

Full disclosure: Gemini 3.8 Flash was used to provide an initial review of this PR.

To verify the configuration is being loaded correctly, I set up a peer config with a TLS endpoint as the only valid endpoint, and the currently valid TLS parameters set up such that TLS works. I then launched a simple publisher process with logging set to trace and looked for the messages that are logged when the webpki root store is loaded and when the user cert is added. I verified the right logs showed up at the right time with the new parameter missing, the new parameter set to true, and the new parameter set to false.

I then redid the test for QUIC.

For my project, I would like to use certificates to identify different participants in the network.

Because I control all of the nodes, I don't need or want a public certificate authority to be used. Digicert's opinion on if something is or is not at some some IP is irrelevant. As set up in main, if I turn on TLS, I wind up having to trust the public PKI system, which is just unnecessary. By setting this new parameter to false, my loaded root cert is the only one that is used, which is what I want.

Not loading the public PKI roots also gives me a sort of lazy method of access control. No one without a cert I sign can get access to my system. As set up in main, I would have to also use the other access control methods like passwords on top of just using certs (not that that is a bad idea.)

### Related Issues
Fixes https://github.com/eclipse-zenoh/zenoh/issues/2764

---
## 🏷️ Label-Based Checklist

**Labels:** `new feature`

## 🆕 New Feature Requirements

**Instructions:**

### DIFF
--- a/DEFAULT_CONFIG.json5
+++ b/DEFAULT_CONFIG.json5
@@ -711,6 +711,9 @@
         /// This could be dangerous because your CA can have signed a server cert for foo.com, that's later being used to host a server at baz.com. If you want your
         /// ca to verify that the server at baz.com is actually baz.com, let this be true (default).
         verify_name_on_connect: true,
+        /// Whether or not to load the default WebPKI root certificates.
+        /// If false, only certificates configured in `root_ca_certificate` will be trusted.
+        use_public_pki: true,
         /// Whether or not to close links when remote certificates expires.
         /// If set to true, links that require certificates (tls/quic) will automatically disconnect when the time of expiration of the remote certificate chain is reached
         /// note that mTLS (client authentication) is required for a listener to disconnect a client on expiration
--- a/commons/zenoh-config/src/lib.rs
+++ b/commons/zenoh-config/src/lib.rs
@@ -786,6 +786,7 @@ validated_struct::validator! {
                     connect_certificate: Option<String>,
                     verify_name_on_connect: Option<bool>,
                     close_link_on_expiration: Option<bool>,
+                    use_public_pki: Option<bool>,
                     /// Configure TCP write buffer size
                     pub so_sndbuf: Option<u32>,
                     /// Configure TCP read buffer size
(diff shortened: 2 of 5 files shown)


jpace121:
Looking at the CI:
1. It is looking for one of a list of PR labels, which I'm not sure I can add.
2. It caught that my Eclipse ECA had expired. I've resigned it and relinked my github account. I'm assuming if that job is reran in a little bit after things have a chance to sync, we'll be good.

jpace121:
I updated the checklist. The only thing I am not sure about checking is "comprehensive testing". I didn't add an unit tests, though there aren't any that I can find for the code around what I added either?

oteffahi (changes requested):
Changes look good.

Please add some tests that validate the new config field is taken into consideration. For example:
- A test that expects failure when starting a transport with TLS listener that has no root cert and does not use WebPKI trust store.
- Another test that succeeds in starting a transport with TLS listener with the default config and/or explicit `true`.

These tests should be added under the `zenoh-transport` crate.

Additionally, documentation needs to be updated on [zenoh.io](https://zenoh.io/docs/manual/tls). Repo for documentation: https://github.com/atolab/zenoh-web

jpace121:
Tests added in 0b10f9d274c3142aa7ac5730e53a539c1402350e.

I initially had fewer tests, but then I let Gemini make things complete. If that's too much, let me know. I'm not the one doing long term maintenance here. I will work on updating the documentation next.

jpace121:
Documentation update linked above.

PR #2801, opened 2026-09-18
fix: avoid deadlock when wait_callbacks() runs from its own callback

YuanYuYuan:
## Summary

`SyncGroup::wait()` can hang the calling thread. This happens only in one case: the caller runs inside a callback of the same entity. This is not a race. One thread runs the whole call stack. The bug appears every time this path runs.

## The defect

Look at [`SyncGroup::wait()`](https://github.com/eclipse-zenoh/zenoh/blob/9fcd9cb5d364192c3e8a27e66de76f4bc750d1d5/zenoh/src/api/cancellation.rs#L74-L78) and [`Callback::call()`](https://github.com/eclipse-zenoh/zenoh/blob/9fcd9cb5d364192c3e8a27e66de76f4bc750d1d5/zenoh/src/api/handlers/callback.rs#L105-L109) before this PR.

A `SyncGroup` wraps a `tokio::sync::Semaphore` with `MAX_PERMITS` permits. Each registered callback holds one permit through an on-drop notifier. `SyncGroup::wait()` asks for **all** permits back before it returns. `Callback` is `Clone`, and its dropper is a shared `Arc`. So the dropper runs, and the permit releases, only when the **last** clone dies.

A running callback holds a live clone. That clone holds a permit for the whole call. If code inside that callback drops the last reference to its own entity, the resulting destructor may call `wait_callbacks()` on the same `SyncGroup`. `wait()` then asks for a permit that only this same stack can release. This stack is also the one waiting. The wait can never end.

```mermaid
sequenceDiagram
    participant RX as RX thread
    participant CB as Callback::call
    participant User as user callback
    participant Drop as ~SubscriptionData (drop)
    participant SG as SyncGroup::wait

    RX->>CB: deliver a sample
    activate CB
    Note over CB: a live clone holds<br/>the on-drop permit
    CB->>User: run the callback
    activate User
    User->>Drop: drop the last reference
    Drop->>SG: undeclare().wait_callbacks().wait()
    SG->>SG: acquire_many(all permits)
    Note over SG: needs the permit CB still<br/>holds. Blocks forever.
    deactivate User
    deactivate CB
```

This is not a rare or contrived call pattern. `zenoh-c`'s `ze_undeclare_*` functions call `wait_callbacks()` at every teardown site (subscriber, publisher, queryable, querier, session close, and more). A C or C++ binding built on `zenoh-c` inherits this risk at every entity destruction, including `rmw_zenoh_cpp`. In production, `~SubscriptionData` dropped a subscription. That call reached `ze_undeclare_advanced_subscriber`. The transport RX thread then blocked forever, and the whole session froze. Only a process restart recovered it.

## Why the library should catch this, not just warn about it

The trigger is an ordinary pattern, not a rare or careless call. Dropping the last reference to an entity from inside that entity's own callback is a normal thing for a callback to do. Nothing in the type of `undeclare()` or `wait_callbacks()` warns the caller that this specific call site is dangerous. The risk lives in an implementation detail the caller cannot see: a live `Callback` clone pins a permit for the whole call.

Three independent parties hit this same shape, with no shared code and no contact with each other:

- **Dexory**, in production: [`~SubscriptionData` → `ze_undeclare_advanced_subscriber`](https://github.com/botsandus/zenoh/commit/ff1286f46), about 2 in every 200 robot bringups.
- **[`ros2/rmw_zenoh#993`](https://github.com/ros2/rmw_zenoh/issues/993)**: the same shape on a querier, `~ClientData` → `z_undeclare_querier`, open since June with no comments at the time it was found.
- **[`ZettaScaleLabs/hiroz#284`](https://github.com/ZettaScaleLabs/hiroz/issues/284)**: names this shape directly and states that hiroz's own reentrancy guard cannot see it, because no lock is involved, only a permit.

A defect that three unrelated codebases reach through their own ordinary use is a property of the API, not a misuse of it. That is why every fix proposed so far, Dexory's, hiroz's own dispatch-registry design, and this one, makes the library detect and defer automatically, instead of adding a doc warning and leaving the trap in place.

## The fix

Each `Callback` now records which `SyncGroup`s hold its permits ([`Callback::groups`](https://github.com/YuanYuYuan/zenoh/blob/ec1fb04ce4ba11d7aeb9ca46c4b933852a3658bd/zenoh/src/api/handlers/callback.rs#L116-L124), set through [`set_groups`](https://github.com/YuanYuYuan/zenoh/blob/ec1fb04ce4ba11d7aeb9ca46c4b933852a3658bd/zenoh/src/api/handlers/callback.rs#L176-L180)). A thread-local stack, [`EXECUTING_GROUPS`](https://github.com/YuanYuYuan/zenoh/blob/ec1fb04ce4ba11d7aeb9ca46c4b933852a3658bd/zenoh/src/api/handlers/callback.rs#L21-L35), tracks which groups have a callback running now on this thread. [`SyncGroup::wait()`](https://github.com/YuanYuYuan/zenoh/blob/ec1fb04ce4ba11d7aeb9ca46c4b933852a3658bd/zenoh/src/api/cancellation.rs#L99-L128) and [`wait_async()`](https://github.com/YuanYuYuan/zenoh/blob/ec1fb04ce4ba11d7aeb9ca46c4b933852a3658bd/zenoh/src/api/cancellation.rs#L130-L148) both check this stack before they block. If a callback of the *same* group is already running here, the drain moves to a spawned task, and the call returns right away.

```mermaid
sequenceDiagram
    participant RX as RX thread
    participant CB as Callback::call
    participant User as user callback
    participant Drop as ~SubscriptionData (drop)
    participant SG as SyncGroup::wait
    participant BG as spawned task

    RX->>CB: deliver a sample
    activate CB
    Note over CB: pushes this group onto<br/>EXECUTING_GROUPS
    CB->>User: run the callback
    activate User
    User->>Drop: drop the last reference
    Drop->>SG: undeclare().wait_callbacks().wait()
    SG->>SG: is this group already<br/>on EXECUTING_GROUPS?
    SG->>BG: spawn the drain, return now
    deactivate User
    deactivate CB
    BG->>BG: acquire_many(all permits)<br/>once CB's clone drops
```

The check asks a narrow question on purpose: does a callback of **this** group run now, not does any callback run at all. Entity A's callback may undeclare an unrelated entity B. That call holds none of B's permits, so a wait on B can still block correctly until B's own callback ends. A wider check would drop that guarantee. See "Comparison" below for the concrete cost of the wider check.

[`tests/undeclare_from_callback.rs`](https://github.com/YuanYuYuan/zenoh/blob/ec1fb04ce4ba11d7aeb9ca46c4b933852a3658bd/zenoh/tests/undeclare_from_callback.rs) adds 8 scenarios:

- The two self-joins: a subscriber and a queryable.
- Three controls: no `wait_callbacks()`, a wait from another thread, and a check that the entered-callback guard itself is live.
- The case that separates this fix from a coarser one (the unrelated-entity barrier).
- Two tests that pin the fix's own bookkeeping: nested delivery, and a panicking callback.

## Credit and comparison against Dexory's fix

**Guillaume Doisy ([@doisyg](https://github.com/doisyg), Dexory)** found and fixed this same defect in production. His fix is field-validated at about 2 in every 200 robot bringups: [`botsandus/zenoh@ff1286f46`](https://github.com/botsandus/zenoh/commit/ff1286f46). His check is a thread-local depth counter. It asks "does any callback run on this thread?", not "does a callback of this group run?".

That difference has a measured cost. This PR's own test suite includes a scenario where entity A's callback tears down an unrelated entity B, while B's own callback runs on another thread for 800ms. Nothing here is self-held. B's permit belongs to a different stack. So the barrier is achievable, and current `main` honors it today.

| scenario | Dexory's depth counter | this PR |
|---|---|---|
| subscriber or queryable self-join | fixed | fixed |
| A undeclares unrelated B while B's callback runs elsewhere | barrier dropped, returns in 0ms | barrier honored, waits the full ~800ms |
| diff size | `+50/-1`, 2 files | `+123/-4`, 3 files |

The depth counter is correct for the self-join, and it is smaller. It over-approximates in one case: any callback on the thread is enough to skip the wait, even when that callback holds no permit the wait needs. This PR's per-group tracking costs more code. It keeps the guarantee exact.

## Verified against an external reentrancy checker, both directions

To confirm the defect and the fix outside this PR's own test harness, I wired [`lock-tripwire`](https://github.com/YuanYuYuan/lock-tripwire) — a small, general panic-on-reentrancy checker — into the two real call sites:

- `Callback::call`/`call_with_message`: marks the permit a live clone holds for the call.
- `SyncGroup::wait()`'s `block_in_place(acquire_many(..))`: the guarded call-out.

Two branches on my fork reproduce this, and each builds on its own:

- **[`study/m5-tripwire-demo-red`](https://github.com/YuanYuYuan/zenoh/tree/study/m5-tripwire-demo-red)** — unpatched `main`, reproduces the self-join.
- **[`study/m5-tripwire-demo-green`](https://github.com/YuanYuYuan/zenoh/tree/study/m5-tripwire-demo-green)** — this PR's branch, same instrumentation, completes cleanly.

Reproduce with either branch:

```bash
cargo test -p zenoh --features unstable --test tripwire_m5_demo -- --nocapture
```

On unpatched `main`, the self-join panics in about 1 second. It names the exact acquisition and call-out sites, instead of hanging for the existing harness's 10-second timeout:

```
thread '<unnamed>' panicked at zenoh/src/api/cancellation.rs:76:18:
re-entrancy hazard at zenoh/src/api/cancellation.rs:77: State lock(s) still held,
acquired at zenoh/src/api/handlers/callback.rs:120
```

On this PR's branch, the same scenario completes cleanly. This is not a weaker check. This PR's own early-return branch in `SyncGroup::wait()` returns before the instrumented `block_in_place` call runs at all in the self-join case, so the checker has nothing to catch. A control scenario, ordinary delivery with no self-undeclare, completes cleanly on both branches. This rules out a checker that panics on everything.

## Breaking changes

None. `Callback<T>`'s new field is private. It defaults to empty for any callback never registered with a `SyncGroup`.

---
## 🏷️ Label-Based Checklist

**Labels:** `bug`

## 🐛 Bug Fix Requirements

**Instructions:**


zenoh/src/api/cancellation.rs
```diff
@@ -72,13 +98,58 @@ impl SyncGroup {
...
+        // A's callback may undeclare entity B. That call holds none
+        // of B's permits. So a wait on B can still block until B's
+        // own callback ends. A wider check would break that guarantee.
+        //
+        // This happened in production. rmw_zenoh dropped a
+        // subscription inside `~SubscriptionData`. That call reached
+        // `ze_undeclare_advanced_subscriber`. The transport RX thread
+        // then blocked forever, and the whole session froze.
+        if crate::api::handlers::callback_of_group_running_on_this_thread(self.id()) {
+            tracing::warn!(
```
JEnoch: `warn!` level seems too loud for a condition the caller cannot really act on. If this happen often (e.g. short-lived entity creating/destroying a querier foe each request), it will cause noise in the log.

zenoh/src/api/cancellation.rs
```diff
@@ -72,13 +98,58 @@ impl SyncGroup {
...
 
     #[zenoh_macros::pub_visibility_if_internal]
     pub(crate) async fn wait_async(&self) {
+        // Same self-join check as `wait()`, and for the same reason.
+        // A direct caller of this method — session close, for
+        // example — can also run from inside a callback of this
+        // group. Without this check, the `.await` below would wait
+        // on its own caller's permit and never finish.
+        if crate::api::handlers::callback_of_group_running_on_this_thread(self.id()) {
+            tracing::warn!(
```
JEnoch: Same comment wrt. `warn!` level.

zenoh/src/api/handlers/callback.rs
```diff
@@ -14,9 +14,53 @@
...
+/// Pushes a callback's groups for the length of one call.
+///
+/// `Drop` makes this safe across a panic. A user callback may
+/// panic. If it does, its groups must still pop, or every later
+/// wait on them would take the async path forever.
+struct ExecutingGroups(usize);
+
+impl ExecutingGroups {
+    fn enter(groups: &[GroupId]) -> Self {
+        EXECUTING_GROUPS.with_borrow_mut(|stack| stack.extend_from_slice(groups));
```
JEnoch, line 55: Calling the Thread Local Storage and the RefCell borrow has a cost which can be avoided when the list is empty:
```rust
        if groups.is_empty() {
            return Self(0);
        }
```

zenoh/src/api/handlers/callback.rs
```diff
@@ -14,9 +14,53 @@
...
+        Self(groups.len())
+    }
+}
+
+impl Drop for ExecutingGroups {
+    fn drop(&mut self) {
+        EXECUTING_GROUPS.with_borrow_mut(|stack| {
```
JEnoch, line 65: Calling the Thread Local Storage and the RefCell borrow has a cost which can be avoided if the list is empty:
```rust
        if self.0 == 0 {
            return;
        }
```

JEnoch (changes requested):
In addition of my inline comments, I suggest to clarify the documentation of all the `wait_callbacks()` operations (I counted 9). They currently all have this simple line as documentation:

`/// Block in undeclare operation until all currently running instances of matching listeners' callbacks (if any) return.`

I suggest to add:

```Rust
///
/// # Calling this from inside the entity's own callback
///
/// If this is reached from within a callback of the same entity - for instance by
/// dropping the last reference to the entity inside its callback - the barrier cannot
/// be honored: the running callback holds the very permit the wait needs. In that case
/// the call returns immediately and the remaining callbacks are drained in the
/// background. Callbacks of *other* entities are still waited for as usual.
```

zenoh/src/api/cancellation.rs
OlivierHecart: I agree. 
The goal of this pr is to make this legitimate.

zenoh/src/api/cancellation.rs
OlivierHecart: I agree.

OlivierHecart (changes requested):

YuanYuYuan:
Addressed in `90d46a5d4`:

- Both `warn!` calls downgraded to `trace!`, matching the level this file already uses for routine per-operation logging (`declare_subscriber`, etc.).
- `ExecutingGroups::enter`/`Drop` both short-circuit when the callback holds no `SyncGroup` permits, skipping the thread-local/`RefCell` touch entirely for the common case.
- Added the self-join doc note to all `wait_callbacks()` builders that route through `SyncGroup::wait` — subscriber, queryable, querier, publisher's matching listener, session close, transport events listener, and link events listener (7 in `zenoh`, plus the one on `MatchingListenerUndeclaration` already counted above = 8).

One adjustment: `zenoh-ext`'s `SampleMissHandlerUndeclaration::wait_callbacks()` is the 9th site, and I left it undocumented on purpose — checked its body first, and it's a no-op by design (its own existing comment: miss-listener callbacks already run under the state lock, so there's nothing to drain). Adding the self-join note there would have described behavior that doesn't exist.

OlivierHecart:
It seems to me that this PR introduces an issue for C/C++ bindings: 
The garantie that all callbacks are terminated when `undeclare` terminates was introduced so that the user may safely destroy any context those callbacks may access. 
Now this garantie is violated: after `undeclare` returns, callbacks may still be running. The user may destroy some context that those callbacks are accessing.

